PR #13, opened 2014-10-31
Modernize WebRTC with promises compatibly with hybrid.

jan-ivar:
Let me know if I missed anything. Leaving polish of legacy note to others.


heycam:
Sorry for the delay in replying here.

Initially there were two reasons for disallowing overloading across partial interfaces. The first was the problem of coordinating specs as mentioned above. The second was that overload resolution early on relied on the relative order of each overload; if all the overloads are on a single (partial) interface declaration, it's easy to see the order, but on different partial interface declarations there's no clear order. (By spec publication date perhaps? :smiley:) This second reason doesn't apply any more.

I think @alvestrand makes a good point that you can easily get conflicts with partial interfaces using attributes or other IDL features.  So I think I'm OK with dropping the restriction.

We could require that overloads in partial interfaces are called out with syntax.  For example:

```
interface Base {
  void f();
};
partial interface Base {
  overload void f();
};
```

So we would require `overload` to be used on partial interfaces, and disallow it on the primary interface. That should draw the reviewer's attention to the fact that overloading is happening and that they should be aware of what's in the base/other specs.

Would this work?

heycam:
As for why you might want to call out overloads in syntax rather than anything that might conflict (in which case you might as well just see the `partial` keyword as that hint to the reviewer), function invocation behaviour can change subtly when introducing overloads, whereas for other conflicting features (attributes with the same name, for example) this is just an error.

martinthomson:
@heycam, I'm sure that we could live with more annotations.

martinthomson:
And yes, this might be a valuable construct to have more generally.  (I always liked `override` annotations in subclasses for the same reasons.)

alvestrand:
I'm very much in favour of marking overloads explicitly.
They are a subtle feature, and if used less than carefully, can lead to subtle errors.

Besides, if I intended an overload and spelled the function name wrong, having the "overload" or "override" keyword is the compiler's only chance to figure out that I probably didn't intend what I wrote...

martinthomson:
@alvestrand, I agree.  Even though the marginal utility of marking an overload is slightly less than marking an override.  An override is tightly bound to a very specific function signature, whereas an overload only has a name upon which to validate.  Also, an `overload` marker that references another extension might generate false validation errors in the absence of that other extension.  `override` would be very valuable; `overload` would only be quite valuable.

jan-ivar:
@martinthomson yeah I was going to say, overload and override are not the same thing. An override keyword in a subclass ensures that you're polymorphically matched to a specific base method, so your method gets called when someone invokes the base's. That's different from name-overloading where we're talking about distinct methods with nothing inherently in common other than that they sound the same (if you ignore the call-signature). I'm not seeing what this keyword solves, but I'd do almost anything to unblock this PR at this point, so, sure.

adam-be:
Let's continue the work with this PR and do the cross interface overloading, but leave out the keyword until it's in the WebIDL spec.

adam-be:
Updated examples based @jan-ivar's comments
8aa5d2a, 1c1a0a2

adam-be:
Let's merge this PR and continue working on it in the main branch.

PR #22, opened 2014-11-04
First cut at fixing the whole "queue a task" mess

martinthomson:
This turned out to be a little tricky.  Rather than redefine "queue a task", which would have touched a lot of the spec (see the first commit here, which is basically content-free), I've defined a new label.  "Queue an RTCPeerConnection task" is a new definition that this adds, which is attached to a PC-specific task queue.  This serializes all of the mutation functions: create{Offer|Answer}, set{Local|Remote}Description, {add|remove}Track, createDataChannel and getIdentityAssertion.

This probably needs careful review.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -311,6 +311,35 @@
       [[!RTCWEB-JSEP]].</p>
 
       <section>
+        <h4>RTCPeerConnection Task Queue</h4>
+
+        <p>Each <code>RTCPeerConnection</code> instance maintains a single task
+        queue for the asynchronous operations that it provides.  When the
+        phrase <dfn>queue an RTCPeerConnection task</dfn> is used, the following steps are
+        executed:</p>
+        <ol>
+          <li>
+            If
+            the <a href="#widl-RTCPeerConnection-signalingState"><code>signalingState</code></a>
+            is <code>closed</code>, then either:
+            <ul>
+              <li>For methods that return promises, reject the promise with
+              an <code>InvalidStateError</code>.</li>
+              <li>Otherwise, throw an <code>InvalidStateError</code> exception.</li>
+            </ul>
+            And abort these steps.
+          </li>
+          <li>The user agent then applies the <a>queue a task</a> process
+          described in [[!HTML5]] to enqueue the identified steps.</li>
+        </ol>
+
+        <p>The <dfn>RTCPeerConnection task queue</dfn> is used or all operations
+        on <code>RTCPeerConnection</code> that alter state,
+        including <code>createOffer</code>, <code>setLocalDescription</code>,
+        and <code>addIceCandidate</code>.</p>
+      </section>
+
+      <section>
         <h4>Operation</h4>
 
         <p>Calling <code>new <a>RTCPeerConnection</a>(<var>configuration</var>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -759,6 +797,10 @@
             gathered by the ICE Agent. The options parameter may be supplied to
             provide additional control over the generated answer.</p>
 
+            <p>When <code>createOffer</code> is invoked, <a>queue an
```
adam-be: s/createOffer/createAnswer/

webrtc.html
```diff
@@ -1204,16 +1242,23 @@
...
+            <p>When <code>addIceCandidate</code> is invoked, <a>queue an
+            RTCPeerConnection task</a> to process the incoming candidate:</p>
 
-            <p>If the candidate could not be successfully applied, the user
-            agent MUST queue a task to invoke <var>failureCallback</var> with a
-            <code>DOMError</code> object whose <code>name</code> attribute has
-            the value TBD (TODO InvalidCandidate and InvalidMidIndex).</p>
+            <ol>
+              <li>If the candidate parameter is malformed, throw
+              a <code>SyntaxError</code> exception and abort these steps.</li>
```
adam-be: We can't throw from a queued task. It has to be done as a direct result of a method being called.

webrtc.html
```diff
@@ -1390,26 +1435,16 @@
           <dd>
             <p>Adds a new stream to the RTCPeerConnection.</p>
 
-            <p>When the <dfn id="dom-peerconnection-addstream"><code title=
-            "">addStream()</code></dfn> method is invoked, the user agent MUST
-            run the following steps:</p>
+            <p>When
+            the <dfn id="dom-peerconnection-addstream"><code>addStream()</code></dfn>
+            method is invoked, the user agent MUST
+            <a>queue an RTCPeerConnection task</a> to run the following steps:</p>
```
adam-be: This needs a bit more work. I don't think we should do that before addStream is replaced by addTrack().

webrtc.html
```diff
@@ -1471,21 +1506,13 @@
 
             <p>When the <dfn id="dom-peerconnection-removestream"><code title=
             "">removeStream()</code></dfn> method is invoked, the user agent
-            MUST run the following steps:</p>
+            MUST <a>queue an RTCPeerConnection task</a> to run the following
+            steps:</p>
```
adam-be: See comment about removeStream()

webrtc.html
```diff
@@ -2099,7 +2140,8 @@
 
           <p>When the <dfn id=
           "dom-peerconnection-createdatachannel"><code>createDataChannel()</code></dfn>
-          method is invoked, the user agent MUST run the following steps.</p>
+          method is invoked, the user agent MUST <a>queue an RTCPeerConnection
+          task</a> to run the following steps:</p>
```
adam-be: We can't queue a task here since createDataChannel() returns a DataChannel object.

adam-be:
I'm not sure this is a good idea. The "queue a task" language is used to post the result of a "background" operation, for example something initiated by the network, on the main loop. We do that in a lot of places so I think the links to the "queue a task" concept is good, but I don't think we need it for our special case "queued operations". It's not necessary to do the work on the main loop, and it's not only the presentation of the result (event, callback..) that we want to delay and enforce an order upon, it's doing the work as well. I don't think the existing description of the "queued operations" procedure is perfect, but I captures the functionality. That description is still in the document after this PR btw. It's rather anonymous and located after the RTCPeerConnection costructor description. It should have it's own section and be referenced by each function using it (as this PR does).

alvestrand:
Per discussion f2f in Dallas, closing this pull request. Martin will prepare a new one after thinking about it some more. The tracking bug is #162 .

PR #27, opened 2014-11-04
Documenting practice with candidate events

martinthomson:
Three changes:
* RTCIceCandidate is now less optional and immutable
* sdpMid and sdpMLineIndex are mandatory for the browser to provide
* the candidate attribute on the icecandidate event is now actually nullable and there is documentation that explains what it means

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1943,27 +1943,25 @@
 
         <dl class="idl" data-merge="RTCIceCandidateInit" title=
         "interface RTCIceCandidate">
-          <dt>Constructor (optional RTCIceCandidateInit candidateInitDict)</dt>
+          <dt>Constructor (RTCIceCandidateInit candidateInitDict)</dt>
 
           <dd>The <dfn id=
           "dom-icecandidate"><code>RTCIceCandidate()</code></dfn> constructor
-          takes an optional dictionary argument, <var>candidateInitDict</var>,
+          takes a dictionary argument, <var>candidateInitDict</var>,
           whose content is used to initialize the new
-          <code><a>RTCIceCandidate</a></code> object. If a dictionary key is
-          not present in <var>candidateInitDict</var>, the corresponding
-          attribute will be initialized to null. If the constructor is run
-          without the dictionary argument, all attributes will be initialized
-          to null.</dd>
+          <code><a>RTCIceCandidate</a></code> object. When constructed, values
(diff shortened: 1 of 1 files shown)


martinthomson:
Updated as promised.

webrtc.html
```diff
@@ -1992,27 +1992,25 @@
...
+          takes a dictionary argument, <var>candidateInitDict</var>,
           whose content is used to initialize the new
-          <code><a>RTCIceCandidate</a></code> object. If a dictionary key is
-          not present in <var>candidateInitDict</var>, the corresponding
-          attribute will be initialized to null. If the constructor is run
-          without the dictionary argument, all attributes will be initialized
-          to null.</dd>
+          <code><a>RTCIceCandidate</a></code> object. When constructed, values
+          for <var>candidate</var> and either <var>sdpMid</var>
+          or <var>sdpMLineIndex</var> MUST be provided.</dd>
```
jan-ivar: MUST is for UAs, not users. Or else what? The UA MUST throw... is what we want. Also, if the candidate argument is required then lets use the new WebIDL `require` keyword. Less to write. This defines what happens for at least the candidate case which this prose does not. WebIDL is always better than prose.

Optional and nullable are different concepts. One lets you omit a value (e.g. undefined in JS), whereas the other extends the value space to include null. The prose already requires that a value for candidate be supplied, so omitting it is not allowed, and the `require` keyword is appropriate.

jan-ivar:
Surfacing my file comment here since this got merged already:

MUST is for UAs, not users. Or else what? The UA MUST throw... is what we want. Also, if the candidate argument is required then lets use the new WebIDL `require` keyword. Less to write. This defines what happens for at least the candidate case which this prose does not. WebIDL is always better than prose.

Optional and nullable are different concepts. One lets you omit a value (e.g. undefined in JS), whereas the other extends the value space to include null. The prose already requires that a value for candidate be supplied, so omitting it is not allowed, and the `require` keyword is appropriate.

webrtc.html
martinthomson: I already explained why I didn't use `require`.  I might be OK with nullable AND required on candidate, but this is the `interface` not the `dictionary`.

jan-ivar: I meant required and nullable candidate in the dictionary.

martinthomson: It might be quicker now to generate a PR rather than continue discussing here, given that this has been merged.

PR #29, opened 2014-11-05
Adding mediadiscarded event

martinthomson:
As discussed at TPAC '14 for https://www.w3.org/Bugs/Public/show_bug.cgi?id=21877

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5147,6 +5147,7 @@ sender.insertDTMF("123");
 
     <ol>
       <li>Bug 25724: Allow garbage collection of closed PeerConnections</li>
+      <li>Bug 21877: Adding mediadiscarded event</li>
     </ol>
 
     <h3>Changes since April 10, 2014</h3>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1595,6 +1595,30 @@
             "#dom-peerconnection-ice-connection-state"><code>RTCPeerConnection</code>
             ice connection state</a> changes.
           </dd>
+
+          <dt>attribute EventHandler ondiscardedmedia</dt>
+
+          <dd>
+            <p>This event handler, of event handler event type <code><a href=
+            "#event-discardedmedia">discardedmedia</a></code>,
+            MUST be fired by all objects implementing the
+            <code><a>RTCPeerConnection</a></code> interface. It is fired when
+            inbound media is discarded.  This can be due to packets having an
+            unknown SSRC, payload type, or any other error that makes it
+            impossible to attribute an RTP packet to a
+            specific <code>MediaStreamTrack</code>.  This event is not fired
+            once for each arriving packet; multiple discarded packets for the
+            same SSRC SHOULD result in a single event.</p>
+
+            <p class="note">This event might be delayed while there is an
+            outstanding offer (i.e., in a <a href=
+            "#dom-peerconnection-signaling-state"><code>signalingState</code></a>
+            of <code>have-local-offer</code>
+            or <code>have-remote-pranswer</code>), since packets corresponding
+            to an answer might arrive prior to the answer that describes them.
+            In these states, unknown media might be buffered temporarily to
+            allow time for an answer to arrive.</p>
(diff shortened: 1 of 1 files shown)


fluffy:
What do people think of "one notification per SSRC " ?

martinthomson:
I'd be OK with that, or maybe "at most one notification per SSRC".  There are lots of SSRCs and I don't want to be obligated to report on all of them.

fluffy:
Good point, don't want to be DOS'd by ssrc, but on the other hand I'd be sad  if I got 3 different SSRC and only got notified about one of them. In the early media case, this notification might trigger the JS app to decide which provisions answer to push into setRemote as different media arrives.

vr000m:
RFC7002 defines a metric for packets that are discarded by the jitter/playout buffer, either due to early or late arrival. The early packets in this case occur when the timestamp of the packets are too far in the future compared to current playout and therefore discarded. Consequently, the late packers are packets whose timestamp are in the past compared to the current playout timestamp. 

I suppose these packets should not trigger the `onmediadiscarded` in these cases?

alvestrand:
On 04/06/2015 09:27 PM, Varun Singh wrote:
>
> RFC7002 defines a metric for packets that are discarded by the
> jitter/playout buffer, either due to early or late arrival. The early
> packets in this case occur when the timestamp of the packets are too
> far in the future compared to current playout and therefore discarded.
> Consequently, the late packers are packets whose timestamp are in the
> past compared to the current playout timestamp.
>
> I suppose these packets should not trigger the |onmediadiscarded| in
> these cases?
>
> —
> <https://github.com/w3c/webrtc-pc/pull/29#issuecomment-90230257>.
>
I would say that "mediadiscarded" should happen when you discard an
entire RTP encoded stream (= ssrc), not when you discard a packet from a
stream.

"mediastreamdiscarded" might be a better name.

-- 
Surveillance is pervasive. Go Dark.

vr000m:
Yes, mediastreamdiscarded aligns better with the intention.

I would also like to weigh in on getting at most a notification per SSRC. Additionally, getting the number of packets discarded before playback started via the statsAPI.

alvestrand:
Per discussion in Dallas: We really need to separate the cases of "hey - stuff arrived that we can't associate with any configured track" and "data has been thrown away, we thought you might like to know".

Mozilla has working code that, on the first instance, is able to configure a track that will take the incoming packets and do something sensible with them. Calling the event that triggers that handling "mediadiscarded" just seems wrong.

Martin will write a new proposal.

stefhak:
@alvestrand could you elaborate a bit more about the difference between "hey - stuff arrived that we can't associate with any configured track" and "data has been thrown away, we thought you might like to know"? (I was not in Dallas). The original discussion concluded (if memory serves) that if stuff arrives that can't be associated with a track it would be dropped on the floor - which means the two cases are the same. Apparently this was advanced a bit in Dallas.

alvestrand:
The Dallas conversation included a claim by @martinthomson (I think) that he had examples of applications where data would arrive on a previously unconfigured SSRC, the signal went up to Javascript, and the signal contained enough information that the Javascript could modify the SDP, do a setLocalDescription (or setRemoteDescription or both?), get an addtrack event, and receive the incoming packet on the newly configured track.
Calling the signal that caused this to happen "mediadiscarded" seemed disingenious, since in fact media is not discarded at all.
Of course we can't guarantee that the javascript would be able and willing to do that, and the media subsystem can't hang on to an infinity of incoming packets forever, so there has to be some signal that says that media has *really* been discarded.

That's what I remember of the conversation.

stefhak:
Then I understand better. But it feels like a timing thing then. Say a "new unknown media" event is fired, if the JS does not apply a new SDP that takes care of it a "mediadiscarded" event should be fired after a while (which should be defined). 

Anyway, this PR does not address "unknown media". I'll take this to the list.

alvestrand:
Closing this PR. Solving the problem will require new PRs.

PR #165, opened 2014-12-20
Replace RTCPeerConnection's operations array with a promise-chain.

jan-ivar:
We've successfully replaced Firefox's implementation of the operations array with a promise-chain (http://bugzil.la/1106675 ). I'd like to propose that the spec language benefit from a similar update.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -383,8 +383,8 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to represent a queue of
-            <code>operations</code> with an empty set.</p>
+            <p>Initialize an internal variable to a resolved promise, to
+            represent the start of a chain of <code>operations</code>.</p>
           </li>
 
           <li>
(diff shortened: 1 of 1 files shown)


alvestrand:
Needs to be proposed on the list, and then generated as a new PR, if desirable.

webrtc.html
```diff
@@ -383,8 +383,8 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to represent a queue of
-            <code>operations</code> with an empty set.</p>
+            <p>Initialize an internal variable to a resolved promise, to
+            represent the start of a chain of <code>operations</code>.</p>
```
martinthomson: I believe that it's customary to talk about this as "an internal slot called [[operations]]"

webrtc.html
```diff
@@ -399,23 +399,30 @@
...
+            </ol>
           </li>
 
           <li>
-            <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            <p>Set <var>connection</var>'s <code>operations</code> to the result
+            of transforming <var>p</var> by a rejection handler that returns
+            <var>undefined</var>.</p>
```
martinthomson: Use english too please.  You are saying that you need to ensure that errors from the last operation in the chain need to be handled with `.catch()` so that subsequent operations aren't affected by earlier failures.

webrtc.html
```diff
@@ -399,23 +399,30 @@
...
+            following steps:</p>
+            <ol>
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>closed</code>, abort these steps.</p>
+              </li>
+              <li>
+                <p>Return the result of executing the current call
+                being handled.</p>
```
martinthomson: I find this too vague.  "the current call being handled" isn't well enough defined for me.  Maybe you can assign something to <var>op</var>.

martinthomson:
One thing that I like about this is that it removes the synchronous step.  Having a function run synchronously or asynchronously depending on something that has happened elsewhere is a very good way to ensure that surprises happen.  Allowing the current execution context to run to completion before executing any potentially asynchronous operation is always better than having surprises.

webrtc.html
jan-ivar: sgtm, but orthogonal.

webrtc.html
jan-ivar: This is spec legalese. See http://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting

I suppose we can have both. Want it before or after?
* "In order for errors from the last operation to not affect subsequent operations, set ..."
* "This prevents errors from the last operation from affecting subsequent operations."

webrtc.html
jan-ivar: Agree, but orthogonal.

jan-ivar:
Well put.

webrtc.html
martinthomson: Spec legalese is a great way to produce a very unfriendly spec.  I don't care if the explanation is before or after.

jan-ivar:
See new PR for changes.

Issue #132, opened 2014-12-15
DTMFSender should hang off RTCRTPSender, not MediaStreamTrack


dontcallmedom:
> Originally reported on W3C Bugzilla [ISSUE-27213](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27213) Fri, 31 Oct 2014 18:13:10 GMT
> Reported by Harald Alvestrand
> Assigned to 

Now that RTCRTPSender is available, the abstractions layer better this way.

dontcallmedom:
Closing since #167 was merged

PR #167, opened 2014-12-29
Remove RTCPeerConnection.createDTMFSender and add RTCRtpSender.dtmf.

pthatcherg:
Fixing https://github.com/w3c/webrtc-pc/issues/132

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1363,8 +1363,7 @@
         signalingState</a> is <code>closed</code>, no such event handler can be
         triggered and it is therefore safe to garbage collect the object.</p>
 
-        <p>All <code><a>RTCDTMFSender</a></code>,
-        <code><a>RTCDataChannel</a></code> and
+        <p>All <code><a>RTCDataChannel</a></code> and
         <code><a>MediaStreamTrack</a></code> objects that are connected to a
         <code><a>RTCPeerConnection</a></code> are considered to have a strong
         reference to the <code><a>RTCPeerConnection</a></code> object.</p>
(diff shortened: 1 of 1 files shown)


alvestrand:
Needs to be integrated after #175, since it references RTCRTPSender.

fluffy:
Need to think about how this interacts with SDP - Will all tracks have DTMF negotiated for them ? seems like a waste of PT when it is usually not needed.

pthatcherg:
Cullen:  Would it make sense to have an option in RTCOfferAnwerOptions.dtmf just as we have RTCOfferAnswerOptions.voiceActivityDetection in https://github.com/pthatcherg/webrtc-pc/commit/90af82e02dc141d83048a8497fae5fe4fb365f68?

If that would work, I can propose it on the list.

pthatcherg:
In Chrome at least, we always put DTMF in the offer, so we're already using
the PT.   Meanwhile, we're always negotiating PCMA and PCMU.  In other
words, we're already using lots of PTs.  Meanwhile, is it even possible to
re-negotiate with DTMF later if the DTMF isn't in the original offer?  In
that case, we have to have it in the original offer anyway.

fluffy:
OK - seems reasonable to me

adam-be:
Rebased and integrated.

jesup:
I"ll note that Firefox has not thus far added DTMF to our offers (nor have we integrated support for it yet).  If we're requiring DTMF on all audio channels, that's new - IMHO it should be in OfferAnswerOptions

PR #171, opened 2014-12-30
Add RTCAnswerOptions.

pthatcherg:
Part of https://github.com/w3c/webrtc-pc/issues/128

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -162,7 +162,7 @@
           <dd>
             <p>Sets the <a href="#target-peer-identity">target peer
             identity</a> for the <a>RTCPeerConnection</a>. The
-            <a>RTCPeerConnection</a> will establish a connection to a remote
+            <a>RTCPeerConnection</a> will not establish a connection to a remote
             peer unless it can be successfully authenticated with the provided
             name.</p>
           </dd>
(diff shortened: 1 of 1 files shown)


jesup:
LGTM

stefhak:
This looks reasonable to me.

stefhak:
Isn't this (VAD wanted to be off) something we should do on the RtpSender object instead? It seems to be associated with one track (of kind audio) to be sent over the network, and thus seems fitting to go into the RtpSender object to me.

fluffy:
It's pretty likely that if you want to set it at all, you want it for everything on the PC. I'd prefer to have it an an option for creating offer / answer but I think it makes sense to *also* have it settable at a more fine grain level on the RtpSender.

fluffy:
LGTM

stefhak:
I remain unconvinced about this going inte SDP offer/answer options. This is something that the sender can control - no need to negotiate. And, there are codecs for which there is no SDP attributes to indicate "do not use VAD / DTX / CN" (AMR and AMR-WB). The natural place would be the RTPSender IMO.

stefhak:
I was wrong.... my point was that using VAD or not really belongs to the RTPSender object. But that should not stop this PR to be merged - this PR is about allowing options also for the Answer, and we should do that. 
I'll open an Issue on my point.

PR #175, opened 2015-01-08
Merge doohickey branch to master

juberti:


webrtc.html
```diff
@@ -528,23 +520,60 @@
...
-            <code>MediaStream</code>s may be triggered either by SDP
-            negotiation or by the receipt of media on a given flow. 
+            <code>MediaStreamTrack</code>s may be triggered either by SDP
+            negotiation or by the receipt of media on a given flow.
             <!--  [[OPEN ISSUE: How many <code>MediaStream</code>s are created
                 when you receive multiple conflicting pranswers?]] --></p>
           </li>
 
           <li>
+            Add <var>track</var> to the <code>MediaStream</code>(s) <var>streams</var>
```
adam-be: The variable "streams" turns up out of the blue here and without a proper description.

webrtc.html
```diff
@@ -556,49 +585,51 @@
...
           </li>
         </ol>
 
         <p>When a user agent has negotiated media for a component that belongs
-        to a media stream that is already represented by an existing
-        <code><a>MediaStream</a></code> object, the user agent MUST associate
-        the component with that <code><a>MediaStream</a></code> object.</p>
+        to a media track that is already represented by an existing
+        <code><a>MediaStreamTrack</a></code> object, the user agent MUST associate
+        the component with that <code><a>MediaStreamTrack</a></code> object.</p>
```
adam-be: I think we can't just s/MediaStream/MediaStreamTrack/ here. From the discussion in #3, I think we decided to remove this section.

webrtc.html
```diff
@@ -614,16 +645,12 @@
...
+                <var>connection</var>'s <a href="#receivers-set">set of receivers</a>.</p>
               </li>
 
               <li>
-                <p><a href="#fire-a-stream-event">Fire a stream event</a> named
-                <code title="event-MediaStream-removestream"><a href=
-                "#event-mediastream-removestream">removestream</a></code> with
-                <var title="">stream</var> at the <var>connection</var>
-                object.</p>
+                <p>End the <var>track</var> object.</p>
```
adam-be: The track was ended by the parent step 3 as well. The old step 4.3 fired the "removestream" event, which we have no counterpart for in the new approach, so I think it's fine to just remove step 4.3.

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+
+                <p>Note that this property can change over time.</p>
+              </li>
+
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>stable</code>, then fire a <a href=
+                "#event-negotiation">negotiationneeded</a> event at
+                <var>connection</var>.</p>
```
adam-be: The last two steps should be moved out of the algorithm. But let's do that in a separate PR.

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+          </dd>
+
+          <dt>void removeTrack (RTCRtpSender sender)</dt>
+
+          <dd>
+            <p>Removes <var>sender</var>, and its associated <code>MediaStreamTrack</code>, from the
+            <code><a>RTCPeerConnection</a></code>.</p>
+
+            <!-- onended -->
+            <p>When the other peer stops sending a track in this manner, a
```
adam-be: s/a/an/

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+                <p>Remove <var>sender</var> from <var>connection</var>'s
+                <a href="#senders-set">set of senders</a>.</p>
+              </li>
+
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>stable</code>, then fire a <a href=
+                "#event-negotiation">negotiationneeded</a> event at
+                <var>connection</var>.</p>
```
adam-be: It's not correct to fire this event sync with the function call. But I suspect that it's not introduced in this PR. Lets fix it in a separate PR.

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+        <section>
+      <h3>RTCTrackEvent</h3>
+
+      <p>The <code><a href="#event-track">ontrack</a></code>
+      event uses the
+      <code><a>RTCTrackEvent</a></code> interface.</p>
+
+      <p><dfn id="fire-track-event" title="fire track event">Firing an
+      RTCTrackEvent event named <var>e</var></dfn> with an
+      <code><a>MediaStreamTrack</a></code> <var>track</var> means that an event
```
adam-be: The two other "arguments" to the event should be mentioned here to make it consistent with other events.

webrtc.html
```diff
@@ -3964,21 +4022,35 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+      enable this capability on the <code>MediaStreamTrack</code> object will be
       described in this section. How the media is transmitted to the peer is
       described in [[!RTCWEB-RTP]], [[!RTCWEB-AUDIO]], and
       [[!RTCWEB-TRANSPORT]].</p>
 
-      <p>A <code>MediaStream</code> as defined in [[!GETUSERMEDIA]] may contain
-      zero or more <code>MediaStreamTrack</code> objects. A
-      <code>MediaStreamTrack</code> sent to another peer will appear as one and
+      <p>As described earlier in this document, the object dealt with when
+      sending audio or video to a remote peer is <code>MediaStreamTrack</code>s.
```
adam-be: Scrap this first sentence? It doesn't add anything of value.

webrtc.html
```diff
@@ -4050,15 +4123,15 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+        <code><a>MediaStreamTrack</a></code> is a member of a <code><a>
+        MediaStream</a></code> that has already been created locally
+        by the <code><a>RTCPeerConnection</a></code>, this is observed on the local
         user agent. If this happens for the reason exemplified, or for any
         other reason than the <code><a href=
         "getusermedia.html#dom-mediastream-addtrack">addTrack()</a></code>
-        method being invoked locally on a <code><a>MediaStream</a></code> or
-        tracks being added as the stream is created (i.e. the stream is
-        initialized with tracks), the user agent MUST run the following
-        steps:</p>
```
adam-be: I don't think the part about "tracks being added as a stream is created" should be removed.

new MediaStream(track1, track2); shouldn't fire events.

adam-be:
Let's merge this rather soon and continue to work on in from within the document.

fluffy:
We (someone, does not need to be Justin) need to fix the merge conflicts before we really review on all this. I'm having a hard time sorting out the parts that were removed.

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+      <h2>RTP Media API</h2>
+
+      <p>The RTP media API lets a web application send and receive <code>MediaStreamTrack</code>s
+      over a peer-to-peer connection. Tracks, when added to a RTCPeerConnection, result in
+      signaling; when this signaling is forwarded to a remote peer, it causes
+      corresponding tracks to be created on the remote side.</p>
+
+      <p>The actual encoding and transmission of <code>MediaStreamTrack</code>s is managed through
+      objects called <code>RTCRtpSender</code>s. Similarly, the reception and decoding of
+      <code>MediaStreamTrack</code>s is managed through objects called <code>RTCRtpReceiver</code>s.</p>
```
fluffy: I think we need more here to explain what a RtpSender / Receiver represents. I'm trying to get at the issue of how many of them do you have. Are they 1:1 with tracks, etc ...

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+
+              <li>
+                <p>Remove <var>sender</var> from <var>connection</var>'s
+                <a href="#senders-set">set of senders</a>.</p>
+              </li>
+
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>stable</code>, then fire a <a href=
```
fluffy: This does not seem right in the case things are not in stable. The event is still needed but perhaps it happens later.

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>closed</code>, throw an
+                <code>InvalidStateError</code> exception and abort these
+                steps.</p>
+              </li>
+
+              <li>
+                <p>If an <code>RTCRtpSender</code> for <var>track</var> already exists in
+                <var>connection</var>'s <a href="#senders-set">set of senders</a>,
+                then abort these steps.</p>
```
fluffy: No error in this case ?

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+                frames (video) or equivalently absent content being sent in
+                place of track content.</p>
+
+                <p>Note that this property can change over time.</p>
+              </li>
+
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>stable</code>, then fire a <a href=
```
fluffy: Seems like if it is not stable, we still need this event, just later

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+         <dd>
+           <p>Returns an array of <code><a>MediaStream</a></code> objects
+           representing the <code><a>MediaStream</a></code>s that this event's
+          <code>track</code> is a part of.</p>
+        </dd>
+
+      </dl>
+
+      <dl class="idl" title="dictionary RTCTrackEventInit : EventInit">
+
```
fluffy: How do you find the correct PC? Do we we need that in the Event too?

webrtc.html
```diff
@@ -528,23 +520,60 @@
...
+              <li>
+                <p>Initialize <var>track's</var> <code><a href=
+                "getusermedia.html#dom-mediastreamtrack-id">id</a></code>
+                attribute to the component track id.</p>
+              </li>
+
+              <li>
+                <p>Initialize <var>track's</var> <code><a href=
+                "getusermedia.html#dom-mediastreamtrack-label">label</a></code>
+                attribute to "<code>remote audio</code>" or "<code>remote
```
fluffy: is this right? it seems like the label should come from SDP not be fixed.

webrtc.html
```diff
@@ -528,23 +520,60 @@
...
+                <p>Initialize <var>track's</var> <code><a href=
+                "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
+                attribute to "<code>audio</code>" or "<code>video</code>"
+                depending on the media type of the incoming component.</p>
+              </li>
+
+              <li>
+                <p>Initialize <var>track's</var> <code><a href=
+                "getusermedia.html#dom-mediastreamtrack-id">id</a></code>
+                attribute to the component track id.</p>
```
fluffy: where would the ID come from from ? particularly when multiple early media streams are received before Answer.

webrtc.html
```diff
@@ -528,23 +520,60 @@
...
+                "getusermedia.html#dom-mediastreamtrack-label">label</a></code>
+                attribute to "<code>remote audio</code>" or "<code>remote
+                video</code>" depending on the media type of the incoming
+                component.</p>
+              </li>
+
+              <li>
+                <p>Initialize <var>track's</var> <code><a href=
+                "getusermedia.html#dom-mediastreamtrack-readystate">readyState</a></code>
+                attribute to <code>muted</code>.</p>
```
fluffy: why muted ?

webrtc.html
```diff
@@ -528,23 +520,60 @@
...
+              </li>
+
+            </ol>
+          </li></p>
 
             <p class="note">The creation of new incoming
-            <code>MediaStream</code>s may be triggered either by SDP
-            negotiation or by the receipt of media on a given flow. 
+            <code>MediaStreamTrack</code>s may be triggered either by SDP
+            negotiation or by the receipt of media on a given flow.
```
fluffy: In the case where media created it, how does one know which Stream it is in?

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+          </dd>
+
+          <dt>attribute EventHandler ontrack</dt>
+
+          <dd>This event handler, of event handler event type <code><a href=
+          "#event-track">track</a></code>, MUST be fired
+          by all objects implementing the <code><a>RTCPeerConnection</a></code>
+          interface. It is called any time a <code>MediaStreamTrack</code> is added
+          by the remote peer. This will be fired only as a result of
+          <code>setRemoteDescription</code>. ontrack happens as early as
```
fluffy: not sure this is right. Early media creates a track before the setRemote. In the case of a PRanswer one can get multiple tracks before after a PRanswer but before the Answer.

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+          possible after the <code>setRemoteDescription</code>. This callback
+          does not wait for a given media track to be accepted or rejected via
+          SDP negotiation.</dd>
```
fluffy: This might need a bit more detail explaining the difference for PRans and ans. We also need to describe what happens  to a track that arrived as early media but was rejected in the Ans. (it should be terminated)

webrtc.html
```diff
@@ -2075,6 +1826,314 @@
...
+        <dt>readonly attribute RTCRtpReceiver receiver</dt>
+
+        <dd>
+          <p>The <dfn id=
+          "dom-trackevent-receiver"><code>receiver</code></dfn> attribute
+          represents the <code><a>RTCRtpReceiver</a></code> object associated with
+          the event.</p>
+        </dd>
+
+        <dt>readonly attribute MediaStreamTrack track</dt>
```
fluffy: I'm worried that when the event was created by media before the SDP, one will not know the stream.

adam-be:
I've updated the doohickey branch with fixes for my comments

webrtc.html
fluffy: I think what we have here is wrong, but I'm happy if we resolve it just by putting an open issue note and not changing it

webrtc.html
juberti: These events are on the PC, so this shouldn't be an issue (same situation exists for other events, and it's easy to deal with this by binding your pc to your event handler in JS if desired)

webrtc.html
juberti: resolved

webrtc.html
juberti: Tracks initially start as muted, until media is received. This isn't new to this PR.

webrtc.html
juberti: For PRANSWER, there is no issue - PRANSWER is a remote description. The tricky part is media before PRANSWER. We need to decide if we are going to handle this or not.

webrtc.html
juberti: This is true. I don't think this is solvable without wire protocol changes.

If we decide to handle this scenario (media before *any* remotedescription), we'll need to somehow generate new MediaStreams for each track, and then migrate tracks as needed when the remote description arrives.

webrtc.html
juberti: done

webrtc.html
juberti: done

webrtc.html
juberti: done

webrtc.html
juberti: done

webrtc.html
juberti: This whole block needs to move down to section 5.

juberti: This text has been moved as discussed. I am fine with including an open issue note for the track label, but since this is ancient we could just do it later. (S 9.2.2 in the current spec)

webrtc.html
juberti: If this is MID, we can get it from media. If it is MSID, it needs to come from the remote description. I think there is a decent case for it being MID, but since this is also not new, I think we should deal with it later (e.g. when we get the .mid pull request).

juberti:
Just uploaded a new version. I believe all comments have been addressed.

I will take one more pass over this today to see if there are any remaining nits, but I think this is ready to be merged now.

.travis.yml
```diff
@@ -1,13 +0,0 @@
...
- - git clone https://github.com/dontcallmedom/webidl-checker.git
- - git clone https://github.com/dontcallmedom/widlproc
- - git clone https://github.com/halindrome/linkchecker.git
- - cd widlproc && make obj/widlproc && cd ..
- - pip install html5lib lxml
-script:
- - phantomjs --ignore-ssl-errors=true --ssl-protocol=tlsv1 respec/tools/respec2html.js -e -w webrtc.html output.html
- - python webidl-checker/webidl-check output.html
- - "! (perl -T linkchecker/bin/checklink -S 0  -q -b output.html |grep  \"^\")"
-env: WIDLPROC_PATH=$TRAVIS_BUILD_DIR/widlproc/obj/widlproc
```
fluffy: Need to not delete this

webrtc.html
```diff
@@ -2033,6 +1684,537 @@
...
+                  "getusermedia.html#dom-mediastreamtrack-id">id</a></code>
+                  attribute to the component track id.</p>
+                </li>
+
+                <li>
+                  <p>Initialize <var>track's</var> <code><a href=
+                  "getusermedia.html#dom-mediastreamtrack-label">label</a></code>
+                  attribute to "<code>remote audio</code>" or "<code>remote
+                  video</code>" depending on the media type of the incoming
+                  component.</p>
```
fluffy: Still seem that setting the label like this is wrong. It should be transported over the SDP

webrtc.html
```diff
@@ -2033,6 +1684,537 @@
...
+          </ol>
+
+          <div class="note">
+            <p>Since the beginning of this specification, remote MediaStreamTracks have been
+            created by the setRemoteDescription call, one track for each non-rejected
+            m-line in the remote description. This meant that at the caller,
+            MediaStreamTracks were not created until the answer was received, and any media
+            received prior to a remote description (AKA "early media") would be discarded.
+            If any form of remote description is provided (either an answer or a pranswer),
+            this issue does not occur.</p>
```
fluffy: This was one of the key issues that doohickey hope to fix

webrtc.html
```diff
@@ -2033,6 +1684,537 @@
...
+                setRemoteDescription for any offered tracks that were rejected.
+                For the answerer, ontrack will continue to fire during setRemoteDescription,
+                as it does today (this is necessary to allow the answerer to reject
+                offered tracks by stopping them).
+                </li>
+              </ol>
+            </p>
+
+            <p>For now, we simply make note of this issue, until it can be
+            considered fully by the WG.</p>
```
fluffy: I want to get a very concrete plan of how to resolve this in here before we merge this. If we are in the "going to talk to the WG" about if we fix this, then I don't think this represents what we agreed too and we should go back to drawing board on what doohickeys are. The ability to deal with early media and an offer to receive more tracks than you are receiving was a key part of this. I'm also very uncomfortable merging something that leaves with dealing with all the on negotiation needed events in any state other than stable.

.travis.yml
juberti: OK, I think this was just a merge thing. Will fix.

webrtc.html
juberti: This has been in the spec (S 9.2.2) for a while, not new to this PR.

webrtc.html
juberti: Ability to receive more tracks than offered is already present, with OfferToReceiveX.

The onnegotiationneeded stuff is not new (same thing exists for AddStream), so I don't think it needs to be resolved here, but I think it is easily fixable. I am happy to do a followup PR for it.

But to the core issue - what to do about media before a remote description - I think does need to be discussed. I can remove the "WG" from the note and we can discuss as a small group - but it clearly needs discussion, given the limitations. I think this is entirely consistent with what we talked about.

In addition to the limitations above, there exists a critical issue in that if the caller is NATed, it can't receive media until it has a remote description and can start doing its own STUN binding checks to open pinholes.

alvestrand: Cullen, you have repeatedly been asked to make suggestions on how to deal with early media.
My preference is to just bit-bucket packets before first answer; if an application knows it's going to interwork with legacy only, it can generate a pr-answer with no candidates and no msid-semantic, and tracks will pop up per the -msid document.
I've yet to see a case that is compelling to me where that workaround is not sufficient.

juberti: Agree. Happy to hop on a call Thursday to understand how we can get this landed.

elagerway: Agree with Harold. I see no cogent reason at this point to continue to hold this up, weeks have drifted by and still we see no sign of this PR landing, very disappointing.

PR #178, opened 2015-01-30
Identity provider interface redesign

### LINKED ISSUE #172: Identity provider selection invisible <iframe>: what's the motivation?
Hey guys,

The TAG is interested in helping get the [Identity Provider Selection](http://w3c.github.io/webrtc-pc/#identity-provider-selection) stuff on a bit firmer ground. It seems like the use of the invisible `<iframe>`-with-extra-restrictions is trying to grope at some deeper concept, and we'd like to be able to understand what the goal is and maybe make this concept reusable across specs or by developers directly, instead of having WebRTC kind of take an HTML element, grab some interesting behavior from it, modify that behavior in a few ways, and then say "effectively, do this" :)

So, just to get the conversation started, could you help us understand the motivation for this? For example, why would an invisible `<iframe>` be better than a simple HTTP request? (I can guess several reasons, but would be interested in which of them are applicable.)

Also, what is the group's current thoughts on this mechanism in general? (Especially any implementers who have looked in to it.) I noticed the "TODO: This is not sufficient unless...", which might indicate that you guys were planning to re-think it and replace it with something better?

martinthomson:
As I described in my email of 2014-12-30, this is a complete rework of the identity provider infrastructure.

I've implemented this in Firefox and hope to enable this in Firefox 38.

This should address @domenic's concerns and therefore (I hope) closes #172.

There are changes to the rtcweb security architecture, but those are mostly deleting text.  I'm starting on those now.


martinthomson:
Hmm, it looks like the WebIDL validation engine doesn't recognize the `required` keyword.  @dontcallmedom, what do you recommend here?

dontcallmedom:
@martinthomson I've fixed the WebIDL validation engine and it now properly recognizes the `required` keyword, and the pull request now passes Travis validation.

domenic:
Looked at the realm section and it looks very nice to me. Thanks!

martinthomson:
Note: Jan-Ivar noted that the idpLoginUrl doesn't need to be a promise, so it now isn't.

webrtc.html
```diff
@@ -3894,25 +3869,17 @@ <h2 id="sec.identity-proxy">Identity</h2>
       <p>The interaction with the IdP is designed to decouple the browser from
       any particular identity provider; the browser need only know how to load
       the IdP's JavaScript, which is deterministic from the IdP's identity, and
-      the generic protocol for requesting and verifying assertions. The IdP
+      the generic interface to generating and validating assertions. The IdP
```
ekr: s/deterministic/the location of which is determined by/

webrtc.html
```diff
@@ -3920,8 +3887,10 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
           the IdP provided shall be used.</li>
 
           <li>If the <code>setIdentityProvider()</code> method has not been
-          called, then the browser can use an IdP configured into the
-          browser.</li>
+          called, then the user agent MAY use an IdP configured into the
+          browser.  However, a user agent-provided IdP MUST NOT cause the
+          <code>createOffer</code> or <code>createAnswer</code> invocation to
```
ekr: I don't understand this text. What does MUST NOT cause it to fail mean?

webrtc.html
```diff
@@ -3932,52 +3901,178 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
...
+        <p>The user agent instantiates an isolated interpreted context, a
+        JavaScript <a
+        href="http://people.mozilla.org/~jorendorff/es6-draft.html#sec-code-realms"><dfn>realm</dfn></a>
+        that operates in the origin of the loaded JavaScript. Note that a
+        redirect will change the origin of the loaded script.</p>
+
+        <p>The <a>realm</a> is populated with a global that implements
+        <code>WorkerGlobalScope</code> [[!WEBWORKERS]].</p>
+
+        <p>The user agent provides instance of
```
ekr: an instance of

webrtc.html
```diff
@@ -3932,52 +3901,178 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
...
+        <code><a>RTCIdentityProviderRegistrar</a></code> named
+        <var>rtcIdentityProvider</var> in the global scope of the <a>realm</a>.
+        This object is used by the IdP to interact with the user agent.</p>
+
+        <p>Since it is an essential security property of the web sandbox that a
+        page is unable to insert objects into content from another origin, using
+        the global <var>rtcIdentityProvider</var> instance ensures that the IdP
```
ekr: This doesn't seem maximally clear. Can you rewrite it to make it clearer what the property is?

webrtc.html
```diff
@@ -3932,52 +3901,178 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
...
+      <p>An IdP proxy implements the <code><a>RTCIdentityProvider</a></code>
+      callback interface, which is the means by which the user agent is able to
+      request that an identity assertion be generated or validated.</p>
+
+      <p>Once instantiated, the IdP script is executed.  The IdP MUST call the
+      <code>register()</code> function on the
+      <code><var>RTCIdentityProviderRegistrar</a></code> instance during script
+      execution.  If an IdP is not registered during this script execution, the
+      user agent cannot use the IdP proxy and MUST fail any attempt to interact
+      with the IdP.</p>
```
ekr: any future attempt?

webrtc.html
```diff
@@ -3932,52 +3901,178 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
...
+          executed.  This registers an instance of
+          <code><a>RTCIdentityProvider</a></code> with the user agent.</p>
+        </dd>
+      </dl>
+
+      <section>
+        <h2>Interface Exposed by Identity Providers</h2>
+
+        <p>The <code>RTCIdentityProvider</code> interface is exposed by identity
+        providers.</p>
```
ekr: And is called by the PeerConnection object, right?

webrtc.html
```diff
@@ -3932,52 +3901,178 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
...
+            generation of an identity assertion.</p>
+
+            <p>The <var>contents</var> parameter includes the information that the
+            user agent wants covered by the identity assertion.  A successful
+            validation of the provided assertion MUST produce this string.</p>
+
+            <p>The <var>origin</var> parameter identifies the origin of the
+            <code><a>RTCPeerConnection</a></code> that triggered this request.  An
+            IdP can use this information as input to policy decisions about
+            use.</p>
```
ekr: You need to make clear that this can be trusted by the IdP.

webrtc.html
```diff
@@ -3932,52 +3901,178 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
...
+
+            <p>The IdP selects the identity to assert.  The optional
+            <var>usernameHint</var> parameter is the same value that was pass to
+            <code>setIdentityProvider</code>.</p>
```
ekr: passed to

webrtc.html
```diff
@@ -3932,52 +3901,178 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
...
+            contain all information necessary to assert identity.  This value is
+            consumed by the validating IdP.</p>
+          </dd>
+        </dl>
+
+        <dl class="idl" data-merge="RTCIdentityProviderRegistrar"
+            title="dictionary RTCIdentityProviderDetails">
+          <dt>required DOMString domain</dt>
+          <dd>
+            <p>The domain name of the IdP that validates the associated identity
```
ekr: validated

webrtc.html
```diff
@@ -3932,52 +3901,178 @@ <h4 id="sec.identity-proxy-communications">Identity Provider
...
+            assertion.</p>
+          </dd>
+
+          <dt>DOMString protocol = "default"</dt>
```
ekr: Why does this have a default value?

webrtc.html
```diff
@@ -3985,46 +4080,66 @@ <h4 id="sec.create-identity-proxy">Instantiating an IdP Proxy</h4>
...
       <ol>
-        <li>The <code>RTCPeerConnection</code> instantiates an IdP proxy as
-        described in <a href="#sec.identity-proxy-communications">Identity
-        Provider Selection section</a> and waits for the IdP to signal that it
-        is ready.
+        <li>
+          <p>The <code>RTCPeerConnection</code> instantiates an IdP as described
+          in <a href="#sec.identity-proxy-communications"></a> and <a
+          href="#sec.register-idp"></a>.  If the IdP cannot be loaded,
+          instantiated or the IdP proxy is not registered, this process
```
ekr: instantiated,

webrtc.html
```diff
@@ -3985,46 +4080,66 @@ <h4 id="sec.create-identity-proxy">Instantiating an IdP Proxy</h4>
...
+        <li>
+          <p>The IdP returns a Promise to the <code>RTCPeerConnection</code>.
+          If the user has been authenticated by the IdP, and the IdP is willing
+          to generate an identity assertion, the IdP resolves the promise with
+          an identity assertion in the form of an
+          <code><a>RTCIdentityAssertionResult</a></code>.</p>
+
+          <p>This step depends entirely on the IdP. The methods by which an IdP
+          authenticates users or generates assertions is not specified, though
+          this could involve interacting with the IdP server or other
```
ekr: this -> they

webrtc.html
```diff
@@ -3985,46 +4080,66 @@ <h4 id="sec.create-identity-proxy">Instantiating an IdP Proxy</h4>
...
+          servers.</p>
+        </li>
 
-        <li>The <code>RTCPeerConnection</code> MAY store the identity assertion
-        for use with future offers or answers.</li>
+        <li>
+          <p>The <code>RTCPeerConnection</code> MAY store the identity assertion
+          for use with future offers or answers.</p>
+        </li>
```
ekr: Hmm... it's not clear how to know if it can be reused. Minimally you need the same key, but what if the IdP has anti-replay stuff? Seems like more is needed here.

webrtc.html
```diff
@@ -4037,23 +4152,28 @@ <h3 id="sec.identity-proxy-assertion-request">Requesting Identity
...
-        "#event-idpassertionerror">idpassertionerror</a> event. This URL might
-        be to a page where a user is able to enter their (IdP) username and
-        password, or otherwise provide any information the IdP needs to
-        authorize a assertion request.</p>
-
-        <p>An application can load the login URL in an IFRAME or popup; the
-        resulting page then provides the user with an opportunity to provide
-        information necessary to complete the authorization process.</p>
+        <p>An IdP MAY reject an attempt to generate an identity assertion if
+        they are unable to verify that a user is authenticated. This might be
```
ekr: they are -> it is

webrtc.html
```diff
@@ -4074,90 +4194,123 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+          identity validation unless there is a <a
+          href="#target-peer-identity">target peer identity</a>.
+        </li>
+        <li>
+          <p>The <code>RTCPeerConnection</code> loads the identity assertion
+          from the session description and decodes the base64 value, then parses
+          the resulting JSON.  The <var>idp</var> paramter of the resulting
+          dictionary contains a <var>domain</var> and an optional
+          <var>protocol</var> value that identifies the IdP, as described in
+          [[!RTCWEB-SECURITY-ARCH]].</p>
```
ekr: You should probably try to match tense here. Either be descriptive or imperative but not both (as in the first and second items on this list)

webrtc.html
```diff
@@ -4074,90 +4194,123 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
-            by the IdP.</li>
+        <li>
+          <p>The <code>RTCPeerConnection</code> decodes the contents and
+          validates that it contains at least one fingerprint value for each
+          <code>a=fingerprint</code> attribute in the session description.  This
+          ensures that the certificate that is, or will be, used by the remote
+          peer for communications is covered by the identity assertion.</p>
+
+          <p>A user agent MUST NOT permit communication with a peer, if that
+          peer offers a certificate that doesn't match an
```
ekr: Suggest "If a peer offers a certificate... a user agent MUST NOT..

webrtc.html
```diff
@@ -4074,90 +4194,123 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+        </li>
 
-          <ul>
-            <li>Ensuring that there is only a single a=fingerprint value across
-            all media sections in the SDP that matches the fingerprint provided
-            by the IdP.</li>
+        <li>
+          <p>The <code>RTCPeerConnection</code> decodes the contents and
+          validates that it contains at least one fingerprint value for each
+          <code>a=fingerprint</code> attribute in the session description.  This
```
ekr: I'm having trouble reading this. "At least one fingerprint value for each a=fingerprint"? Why would it have >1 ?

webrtc.html
```diff
@@ -4074,90 +4194,123 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+        </li>
       </ol>
 
-      <p>The IdP might fail to validate the identity assertion by providing an
-      "ERROR" response to the validation request. Validation can also fail due
-      to the additional checks performed by the browser. In both cases, the
-      process terminates and no identity information is exposed to the
-      application or the user.</p>
+      <p>This process can fail at almost every step of the above process.  If
+      the process fails, the <code><a href=
```
ekr: process...process...process

ekr:
This mostly looks good, but I see some issues as noted above.

fluffy:
Still waiting for update from EKR review ...

webrtc.html
martinthomson: Removed this.  I was under some mistaken impression that this would break something.

What does break is if the user agent-provided IdP asks for login and the site doesn't handle the IdpLoginError correctly.  That's just a consideration for the user agent though.

webrtc.html
martinthomson: I've had another go at this.

webrtc.html
martinthomson: The default corresponds to the default on `setIdentityProvider`.  It allows the default identity provider to have a slightly more compact encoding for their assertions.

webrtc.html
martinthomson: Well, the scope is to the RTCPeerConnection instance.  If the assertion is no good for a second PC, that's OK.  The intent is to avoid contacting the IdP for renegotiation exchanges, since we don't plan on tearing down the connection at that point.  I'll note that if a fresh identity assertion is needed, you need to make a new PC.

webrtc.html
martinthomson: Yeah, that's overly precise.  Our code permits multiples (and ignores them) by virtue of doing a lookup.

martinthomson:
I've updated the PR based on ekr's review.  @fluffy, @burnburn, PTAL

stefhak:
I've not done a detailed review, but given that this PR is based on the updated implementation and that it has been reviewed by many (and have had updates after the reviews) I think it is an improvement over what is in the draft now. I think we should merge this PR.

webrtc.html
```diff
@@ -4138,90 +4260,123 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
-      <code>RTCPeerConnection</code>, any error MUST cause <code><a href=
-      "#dom-peerconnection-setremotedescription">setRemoteDescription</a></code>
-      to fail.</p>
+      <p>If identity validation fails and there is no a <a
+      href="#target-peer-identity">target peer identity</a>, the value of the
+      <code><a href=
+      "#widl-RTCPeerConnection-peerIdentity">peerIdentity</a></code> MUST be set
+      to a new, unresolved promise instance.  This permits the use of
+      renegotiation (or a subsequent answer, if the session description was a
+      provisional answer) to resolve or reject the identity.</p>
```
burnburn: I don't quite understand how/where in the algorithm this new unresolved promise eventually gets resolved one way or another.  Can you briefly point me to what I missed in the section?

burnburn:
I finally had a chance to compare the old and new versions side by side.  There are a few typos and editorial changes worth fixing/making eventually, and I'm sure others will propose changes down the road, but it is definitely complete enough to go in without any further delay.

webrtc.html
martinthomson: @burnburn, just read a few paragraphs up.

Resolved (L4341), rejected (L4354).

martinthomson:
It looks like this needs another rebase...  I'll do that tomorrow; @burnburn, if you do notice anything specific, I'll fix those.  It's hard to proofread your own stuff.

webrtc.html
burnburn: I'm still not completely convinced, but I can always send a new question or PR later if needed.  We can go ahead and put this in after the rebase.

PR #184, opened 2015-02-16
Will NOT connect unless identity is verified

### LINKED ISSUE #183: Illogical specification of peerIdentity in PeerConnection constructor
The following fragment looks a bit weird. Why would a connection fail if the authentication succeeds? I think that "unless" has to be replaced with "if".

> **peerIdentity** of type DOMString
> Sets the target peer identity for the RTCPeerConnection. The RTCPeerConnection will establish a connection to a remote peer unless it can be successfully authenticated with the provided name.
> ~ https://w3c.github.io/webrtc-pc/#widl-RTCConfiguration-peerIdentity

/cc @martinthomson because you authored this fragment in bd9cee2c6ca089ea9a9e33892e52d8001ddd80e6

martinthomson:
Closes #183

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -161,7 +161,7 @@
           <dd>
             <p>Sets the <a href="#target-peer-identity">target peer
             identity</a> for the <a>RTCPeerConnection</a>. The
-            <a>RTCPeerConnection</a> will establish a connection to a remote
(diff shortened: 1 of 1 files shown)


juberti:
lgtm

burnburn:
Applied manually in commit 4c6d983a88dab6026bc82f98b337466eb063cc17

PR #185, opened 2015-02-19
updateIce->setConfiguration

juberti:
Renames updateIce to setConfiguration and moves it next to
getConfiguration.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -372,7 +372,7 @@
           <li>
             <p>Validate the <code><a>RTCConfiguration</a></code> argument by
             running the steps defined by the <a href=
-            "#dom-peerconnection-updateice">updateIce()</a> method.</p>
+            "#dom-peerconnection-setconfiguration">setConfiguration()</a> method.</p>
           </li>
 
           <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -987,6 +987,64 @@
...
+                with <code>SyntaxError</code> and jump to the step labeled
+                <em>Return</em>.</p>
+              </li>
+
+              <li>
+                <p>If the candidate could not be successfully applied, reject
+                <var>p</var> with a <code>DOMError</code> object whose
+                <code>name</code> attribute has the value TBD (TODO
+                InvalidCandidate and InvalidMidIndex) and jump to the step
+                labeled <em>Return</em>.</p>
```
martinthomson: It seems like this could be a little more extensively defined than "successfully applied", even if it is just a reference to JSEP.

webrtc.html
```diff
@@ -1010,7 +1114,7 @@
             connectivity being established.</p>
 
             <p>When the <dfn id=
-            "dom-peerconnection-updateice"><code>updateIce()</code></dfn>
+            "dom-peerconnection-setconfiguration"><code>setConfiguration()</code></dfn>
             method is invoked, the user MUST run the following steps to process
             the <code><a>RTCConfiguration</a></code> dictionary:</p>
```
martinthomson: Where are the steps?

Also, I don't see anything about the peerIdentity stuff.  I'd need to think about it, but I think that the basic idea would be that it couldn't be unset once set, so removing or changing a pre-existing value wouldn't be allowed, even an absent value.  

Setting peerIdentity when there was previously none is tricky.  If there is an existing connection with a matching validated identity, that might be OK.  However, we can't permit one to be added, since that might allow the media isolation protections to be circumvented.

webrtc.html
juberti: Agree, although this is not new to this PR. (This is just a text move)

webrtc.html
juberti: The steps are in the text below, lines 1121-1207 in the new file.

I think it would be fine to prevent any changes to peerIdentity. We can relax this in the future if needed.

adam-be:
The diff of this PR is a bit unfortunate. It's hard to see what's actually been changed due to a lot of moved code. I'd prefer to do this change in two separate steps.

burnburn:
@adam-be does that mean you will be doing the verify that the only changes are:
1. changing updateIce to setConfiguration, and 
2. moving text without additional changes

or are you waiting for Justin to do it?  Or would you like me to review it?

adam-be:
My intention was for Justin to update it, but since I was a bit unclear with my intention I can do the split in a new PR.

adam-be:
PR #211 replaces this one.

PR #186, opened 2015-02-26
v1 of onnegotiationneeded

### LINKED ISSUE #182: The rules for when to fire Onnegotiationneeded need to be clear
We need clear specifications for when to fire the "negotiationneeded" event, as per mailing list discussion  in February 2015 (subject "Specify the negotiationneeded event").

juberti:
v1 of text for onnegotationneeded, without any of the formatting stuff yet. PTAL and see if this looks like it's on the right track.

Fixes #182

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -853,6 +853,12 @@
                   </li>
 
                   <li>
+                    <p>If the supplied description matches the state of all
+                    tracks, clear the negotiation-needed flag.
+                    </p>
+                  </li>
+
+                  <li>
                     <p>Set <var>connection</var>'s description attribute (
                     <code><a>localDescription</a></code> or
                     <code><a>remoteDescription</a></code> depending on the
@@ -1862,6 +1868,39 @@
         </dl>
       </section>
     </section>
+    <section>
+      <h3>Session Negotiation Model</h3>
+      <p>Many changes to the PeerConnection state will require communication with
+        the remote side via the signaling channel. The app can be kept informed
+        as to when it needs to do signaling by listening to the
+        negotiationneeded event.</p>
+      <p>If an operation is performed on a RTCPeerConnection that requires signaling,
+        the connection will be marked as needing negotiation. Examples of such
+        operations include adding, stopping, or removing a track.</p>
+      <p>The negotiation-needed state is cleared when setLocalDescription is
+       called (either for an offer or answer), and the supplied description
+       matches the state of the tracks/datachannels that currenly exist on the
+       RTCPeerConnection. Specifically, this means that all live tracks have
+       an associated section in the local description with their MSID,
+       all stopped tracks have been removed from the local description,
+       and, if any data channels have been created, a data section exists in
+       the local description.</p>
+      <p>Note that setLocalDescription(answer) will clear the
+      negotiation-needed state only if the offer had a corresponding section for
+      all the tracks/datachannels on the answerer side. Otherwise, a new offer by
+      the answerer is still needed, and so the state is not cleared.
+      </p>
+      <p>For simplicity, the negotiation-needed state is only cleared by
+       setLocalDescription. It is far better to fire too many negotiationneeded
+       events than too few; spurious negotiationneededs are essentially harmless,
+       whereas missing events may prevent proper operation of the application.</p>
+      <p>When the RTCPeerConnection is marked as needing negotiation, and it was
+        not already marked as such:</p>
+      <p>If the signaling state is stable, schedule a task to fire an
+        <a href="#event-negotiation">negotiationneeded</a> event on <var>connection</var>.</p>
+      <p>Otherwise, wait for the signaling state to return to stable, and then
+         fire this event.</p>
+    </section>
 
     <section>
       <h3>Interfaces for Connectivity Establishment</h3>
@@ -2087,19 +2126,7 @@
               </li>
 
               <li>
-                <p>If <var>connection</var>'s <a href=
-                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
-                signalingState</a> is <code>stable</code>, then fire an <a href=
-                "#event-negotiation">negotiationneeded</a> event at
-                <var>connection</var>.</p>
-
-                <div class="note">
-                  <p>Throughout this document, we need to indicate what should
-                  happen if a change happens and the state is not stable, most
-                  likely to fire the negotiationneeded event when the state
-                  eventually returns to stable.
-                  </p>
-                </div>
+                <p>Mark <var>connection</var> as needing negotiation.</p>
               </li>
             </ol>
           </dd>
@@ -2147,11 +2174,7 @@
               </li>
 
               <li>
-                <p>If <var>connection</var>'s <a href=
-                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
-                signalingState</a> is <code>stable</code>, then fire a <a href=
-                "#event-negotiation">negotiationneeded</a> event at
-                <var>connection</var>.</p>
+                <p>Mark <var>connection</var> as needing negotiation.</p>
               </li>
             </ol>
           </dd>
@@ -2589,6 +2612,11 @@
               transport</a> and configure it according to the relevant
               properties of <var>channel</var>.</p>
             </li>
+
+            <li>
+              <p>If <var>channel</var> was the first RTCDataChannel created on
+              this connection, mark the connection as needing negotiation.
+              </p>
           </ol>
         </dd>
 
@@ -5249,7 +5277,8 @@ sender.insertDTMF("123");
           <td><code><a>Event</a></code></td>
 
           <td>The browser wishes to inform the application that session
-          negotiation needs to be done at some point in the near future.</td>
+          negotiation should now be done (i.e. a createOffer call followed by
+          setLocalDescription).</td>
         </tr>
 
         <tr>
@@ -5447,7 +5476,7 @@ sender.insertDTMF("123");
       <li>Update misleading local/RemoteDescription attribute text.</li>
 
       <li>Add RTCBundlePolicy.</li>
-      
+
       <li>All callback-based methods have been moved to a legacy section, and
       replaced by same-named overloads using Promises instead.</li>
 
--- a/webrtc.html
+++ b/webrtc.html
@@ -1914,15 +1914,15 @@
       <section>
         <h4>Firing An Event</h4>
         <p>When the <code><a>RTCPeerConnection</a></code> <var>connection</var>
-            is marked as needing negotiation, and it was not already marked as such:</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1862,6 +1868,39 @@
         </dl>
       </section>
     </section>
+    <section>
+      <h3>Session Negotiation Model</h3>
+      <p>Many changes to the PeerConnection state will require communication with
+        the remote side via the signaling channel. The app can be kept informed
```
alvestrand: Suggest adding "in order to have the desired effect" at the end of the sentence. IE you can add tracks as much as you want, but the likely desired effect is to have them sent to the other side, and that won't happen until negotiation is done.

webrtc.html
```diff
@@ -2087,19 +2126,7 @@
...
-                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
-                signalingState</a> is <code>stable</code>, then fire an <a href=
-                "#event-negotiation">negotiationneeded</a> event at
-                <var>connection</var>.</p>
-
-                <div class="note">
-                  <p>Throughout this document, we need to indicate what should
-                  happen if a change happens and the state is not stable, most
-                  likely to fire the negotiationneeded event when the state
-                  eventually returns to stable.
```
alvestrand: I think this needs to be added in two places only:
- SetLocalDescription's "set signalingState accordingly" (step 7 of the last list in setLocalDescription): If resulting state is "stable", and "needs_to_fire_negotiation_needed" flag is set, fire NN.
- SetRemoteDescription's "set signalingState accordingly" (this text is missing in the current draft).

Question: Should the needs_to_fire_negotiation_needed flag be an observable property?

(I've invented the term "needs_to_fire_negotiation_needed" as "what gets set when "Mark connection as needing negotiation" is done).

webrtc.html
```diff
@@ -1862,6 +1868,39 @@
...
+       the local description.</p>
+      <p>Note that setLocalDescription(answer) will clear the
+      negotiation-needed state only if the offer had a corresponding section for
+      all the tracks/datachannels on the answerer side. Otherwise, a new offer by
+      the answerer is still needed, and so the state is not cleared.
+      </p>
+      <p>For simplicity, the negotiation-needed state is only cleared by
+       setLocalDescription. It is far better to fire too many negotiationneeded
+       events than too few; spurious negotiationneededs are essentially harmless,
+       whereas missing events may prevent proper operation of the application.</p>
```
stefhak: I don't think this kind of motivation for the design belongs in the normative text.

webrtc.html
```diff
@@ -1862,6 +1868,39 @@
...
       </section>
     </section>
+    <section>
+      <h3>Session Negotiation Model</h3>
+      <p>Many changes to the PeerConnection state will require communication with
+        the remote side via the signaling channel. The app can be kept informed
+        as to when it needs to do signaling by listening to the
+        negotiationneeded event.</p>
+      <p>If an operation is performed on a RTCPeerConnection that requires signaling,
+        the connection will be marked as needing negotiation. Examples of such
```
stefhak: In my thinking this was not limited be fired only as a result of an operation the application does, it could be things happening under the hood happening (such as a resource not being available for use any longer, cpu/power issues forcing a re-negotiation etc). This was the main motivator for the event to me - there is no way the app would know about this otherwise.
Of course this would give reason for the follow up question: what if a re-negotiation would be good, but not needed (for example if a HW codec has become available)?

stefhak:
Thanks for getting this out! Ideally I would have preferred an operation along what Harald proposes in http://lists.w3.org/Archives/Public/public-webrtc/2015Feb/0037.html, but that is perhaps hard to implement in some cases.

webrtc.html
juberti: These things seem pretty hand-wavy to me right now - it's not clear there's anything that the runtime would do without app knowledge (at least presently) that would *require* a renegotiation.

Similarly for 'optional' renegotiations - I don't think we have any well-defined scenarios where we would want to drive a renegotiation without the app understanding why.

Generally, I think onNN is useful for simplifying the app's own logic about negotiation; it doesn't need to think about when to call createOffer. But I don't think we should imbue it with any more additional responsibilities than that.

webrtc.html
```diff
@@ -853,6 +853,12 @@
                   </li>
 
                   <li>
+                    <p>If the supplied description matches the state of all
+                    tracks, clear the negotiation-needed flag.
```
martinthomson: Probably needs a more rigorous definition here.  I.e., If the description includes media sections for all output tracks, and an application media section if data channels have been opened.

Question regarding renego: if a track is negotiated off in one round, but not subsequently removed, is there any expectation that a new offer will attempt to add it each time?

webrtc.html
juberti: Done.

webrtc.html
martinthomson: I tend to agree with Justin here.  If there is a browser-driven scenario that is well-enough defined, then I'm not opposed to this, but I think that the situation is complex enough to warrant this sort of help.  After all, there are some operations on local tracks that don't result in a need for negotiation and having application developers understand the minute detail of what needs what would be doing them a disservice.

webrtc.html
juberti: Updated setLocal. setRemote already refers to setLocal's processing algorithm, so no action needed there.

I don't see much value in the app inspecting n_t_f_n_n, so I didn't go down that road in this PR.

webrtc.html
juberti: Removed.

juberti:
Updated based on feedback, PTAL.

webrtc.html
martinthomson: I agree with Stefan, removing the paragraph would avoid confusion.  This is only design detail (and arguably incorrect, as this design ensures fewer events than might otherwise be generated).

webrtc.html
martinthomson: I think that an internal property is OK.  Applications that require more complex behaviour will likely understand the complexities well enough to maintain their own understanding of what the variable's value is; the value is largely there for applications that don't care to track state, so they are unlikely to do anything other than observe the event.

martinthomson:
Needs a change log entry.

webrtc.html
juberti: Making sure I understand:

A adds audio and video tracks to its PC. B stops video via mst.stop(), resulting in SDP with m=video 0 to its setLocalDescription, and to A's setRemoteDescription. A's video track obviously does not stop at this point, so how do we prevent a subsequent createOffer by A from including the video track again?

This seems like it needs to cause some state change on A's RtpSender.

juberti: Added reference to the definition.

martinthomson: You got it.

Changing the RtpSender might work.  Perhaps the best thing to do is to open an issue and address it separately.

webrtc.html
```diff
@@ -1862,6 +1868,39 @@
...
+      the answerer is still needed, and so the state is not cleared.
+      </p>
+      <p>For simplicity, the negotiation-needed state is only cleared by
+       setLocalDescription. It is far better to fire too many negotiationneeded
+       events than too few; spurious negotiationneededs are essentially harmless,
+       whereas missing events may prevent proper operation of the application.</p>
+      <p>When the RTCPeerConnection is marked as needing negotiation, and it was
+        not already marked as such:</p>
+      <p>If the signaling state is stable, schedule a task to fire an
+        <a href="#event-negotiation">negotiationneeded</a> event on <var>connection</var>.</p>
```
ekr: Don't we want the task to not fire the event if in the meantime sLD has been called.

ekr:
This looks like it's generally on the right track.

webrtc.html
juberti: yes, good catch. Basically we want to check again when the task runs to ensure negotiation-needed is still set.

juberti: PTAL

webrtc.html
```diff
@@ -1862,6 +1880,52 @@
...
+          that requires signaling,
+          the connection will be marked as needing negotiation. Examples of such
+          operations include adding or stopping a track, or adding the first
+          data channel.</p>
+      </section>
+      <section>
+        <h4>Clearing Negotiation-Needed</h4>
+        <p>The negotiation-needed state is cleared when
+          <code>setLocalDescription</code> is
+          called (either for an offer or answer), and the supplied description
```
fluffy: I'm sort of wondering if  the match is needed. I'm thinking it can be cleared whenever setLocal is called. If the setLocal does not have some of the tracks, then those tracks would be removed ? Thoughts on this?

webrtc.html
```diff
@@ -1862,6 +1880,52 @@
...
+        the remote side via the signaling channel, in order to have the desired
+        effect. The app can be kept informed as to when it needs to do signaling,
+        by listening to the negotiationneeded event.</p>
+      <section>
+        <h4>Setting Negotiation-Needed</h4>
+        <p>If an operation is performed on an <code><a>RTCPeerConnection</a></code>
+          that requires signaling,
+          the connection will be marked as needing negotiation. Examples of such
+          operations include adding or stopping a track, or adding the first
+          data channel.</p>
```
stefhak: I'll argue once more: I think we should not limit Negotiation-Needed to happen _only_ as a result of an operation performed on the PC. If the UA for some reason needs the session to be renegotiated, it should be allowed to fire this event.

I propose changing
"If an operation is performed on an RTCPeerConnection that requires signaling, the connection will be marked as needing negotiation. Examples of such operations include adding or stopping a track, or adding the first data channel."
to
"Whenever an RTCPeerConnection requires signaling, the connection will be marked as needing negotiation. This usually happens as a result of performing an operation such adding or stopping a track, or adding the first data channel on the RTCPeerConnection, but can also happen as a result of internal changes in the implementation (such as resources becoming unavailable in runtime)."

webrtc.html
juberti: It can't always be from setLocal. As mentioned below, if the offerer offers a single track (m= line), and the answerer adds two tracks in response, setLocal can't clear the state. However, it can clear the state if a single track is added in response.

Similar situations exist for data channels.

webrtc.html
juberti: How about I just add a new paragraph:

"Internal changes within the implementation can also result in the connection being marked as needing negotiation. For example, if a MediaStreamTrack enters the ended state due to its source device becoming unavailable."

stefhak: SGTM!

juberti: Done, PTAL

stefhak: I can't see it in the diff. But if you have committed this, I'm sure it LGTM.

juberti: oops, hit the wrong button yesterday. Now PTAL.

stefhak: LGTM!

juberti:
Any final comments? Hoping this can get landed during the editor's call Thursday.

PR #192, opened 2015-02-26
Added support for auth tokens. Fixes #190

### LINKED ISSUE #190: Add support for TURN authentication via OAuth2 tokens
As described in https://tools.ietf.org/html/draft-ietf-tram-turn-third-party-authz-13.

The implications for RTCPeerConnection are pretty minor, I think. Suggestion:

enum RTCIceCredentialType {
  "password", // normal long-term auth
  "token" // OAuth2 token
};

Dictionary RTCIceServer {
  DOMString credential;
  DOMString credentialType;  // if unset, treat as "password"
  DOMString username;
  DOMString or sequence<DOMString> urls;
};

juberti:


juberti:
Can anyone explain how I add a new I-D reference? (i.e. TRAM-TURN-THIRD-PARTY-AUTHZ)

juberti:
@dontcallmedom do you know how I can add a new I-D reference?

dontcallmedom:
to add new references, the preferred way is to make a pull request against specref (as I've just done in my [pull request](https://github.com/tobie/specref/pull/172) for this ref )

An alternative is to use the [localBiblio object in ReSpec](http://www.w3.org/respec/ref.html#localbiblio).

juberti:
ok - thanks!

webrtc.html
```diff
@@ -203,9 +233,13 @@
 
         <p>An example array of RTCIceServer objects is:</p>
 
-        <p><code>[ { "urls": "stun:stun1.example.net" }, { "urls":
-        "turn:turn.example.org", "username": "user", "credential": "myPassword"
-        } ]</code></p>
+        <p><code>[
+          { "urls": "stun:stun1.example.net" },
+          { "urls": "turn:turn.example.org",
+            "username": "user",
```
stefhak: (this is a little beside the point of this PR, but anyway): I think we should have examples using "stuns" and "turns" as well.

stefhak:
Apart from the note that we should have stuns and turns examples as well this LGTM.

webrtc.html
jan-ivar: and maybe an example of urls plural while we're at it (preferably with turn where username and password are reused)?

burnburn:
I'm fine with this.  @stefhak or @jan-ivar Can one of you either provide a PR with some examples like you mention or open an issue requesting those examples so we can remember until someone volunteers (unless you've already done the issue and I haven't come across it yet)?

PR #193, opened 2015-02-26
Add .mid property to sender/receiver. Fixes #191

### LINKED ISSUE #190: Add support for TURN authentication via OAuth2 tokens

### LINKED ISSUE #191: Add RTCRtpSender.mid (same for RTCRtpReceiver)
.mid contains the value of the a=mid:XXXX attribute for the m= section that is associated with the RTCRtpSender/Receiver.

juberti:


webrtc.html
```diff
@@ -2400,14 +2409,25 @@
...
+          </dd>
+
+          <dt>readonly attribute DOMString mid</dt>
+          <dd>
+            The <dfn id="dom-rtpreceiver-mid"><code>RTCRtpReceiver.mid</code></dfn>
+            attribute is the value of the a=mid SDP attribute that is immutably
+            associated, via <code>setRemoteDescription</code>, with this
+            <code><a>RTCRtpReceiver</a></code> object.
+            In the case where no a=mid attribute is present in the remote
+            description, a random value will be generated.
```
burnburn: Is this the right behavior (random value for receiver when no a=mid in remote description)?  I'm not challenging, just asking.

burnburn:
Except for the annoying white space changes (which I can manually adjust if necessary), I'm fine with this.  Would like an SDP expert to comment on my random value question above.

alvestrand:
We have pushback on generating a random value - it seems strange.
Should we represent a missing value as either missing property or empty string?
Changing the MID in RFC 5888 is a SHOULD NOT, not a MUST NOT. Will it ever change?

webrtc.html
fluffy: I lean towards null not random but could live with random too. However, I think it needs to be clear that it's a stable random value and that any future read of that same media section would return the same random value. @juberti, can you confirm that is what you had in mind for the random?

fluffy:
Other than sorting out this random issue or marking it an open issue, all LGTM

alvestrand:
If the .mid property is supposed to reflect the a=mid value in the SDP, then let's let it reflect that and leave it as None for the case where the SDP doesn't have one. (Will only happen on setRemoteDescription, since setLocalDescription is generated by us, and a=mid is required there. I don't think we should allow SDP mangling to remove it.)

webrtc.html
juberti: yes - a stable value, so that the application could still use .mid to tell its receivers apart. 

Probably should say "unique value" or "unique random value"

juberti:
The reason for the random value is that 
a) I didn't want applications that were using .mid as a key for a RtpReceiver to suddenly malfunction when they get a null value.
b) we probably still want to generate an a=mid in any SDP we send, regardless of what we get in an offer. 5888 doesn't seem to prohibit this.

This may also have implications for RTCIceCandidate.sdpMid.

PR #15, opened 2014-10-31
security considerations drafted by Harald

dontcallmedom:
see also https://www.w3.org/Bugs/Public/show_bug.cgi?id=20806

PR #194, opened 2015-02-27
Security considerations section - second PR

alvestrand:
Adding security considerations. This replaces PR #15 .

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5408,7 +5408,7 @@ sender.insertDTMF("123");
 
     <p>This means that data and media can be shared between
     applications running in different browsers, or between an
-    application running in a browser and something that is not a
+    application running in the same browser and something that is not a
     browser, something that is an extension to the usual barriers in
     the Web model against sending data between entities with different
     origins.</p>
@@ -5419,16 +5419,17 @@ sender.insertDTMF("123");
     with other entities as it chooses.</p>
 
     <p>A mechanism, <code>peerIdentity</code>, is provided that gives
-    Javascript the option of requesting media that the Javascript
+    Javascript the option of requesting media that the same javascript
     cannot access, but can only be sent to certain other entities.</p>
 
-    <p>While a web application already before WebRTC will know the IP
-    address to which the application is delivered, setting up
+    <p>Even without WebRTC, a  Web application will know the public IP
+    address to which the application is delivered. Setting up
     communications exposes additional information about the browser’s
-    network context to the web application, including the set of IP
-    addresses available to the browser for WebRTC use. Some of this
-    information has to be passed to the corresponding party to enable
-    the establishment of a communication session.</p>
+    network context to the web application, including the set of
+    (possibly private) IP addresses available to the browser for WebRTC
+    use. Some of this information has to be passed to the
+    corresponding party to enable the establishment of a communication
+    session.</p>
 
     <p>Revealing IP addresses can leak location and means of
     connection; this can be sensitive. Some mitigation is possible by
@@ -5437,7 +5438,7 @@ sender.insertDTMF("123");
 
     <p>Since the browser is an active platform executing in a trusted
     network environment (inside the firewall), it is important to
-    limit the damage that the browser can do to other units on the
+    limit the damage that the browser can do to other elements on the
     local network, and it is important to protect data from
     interception, manipulation and modification by untrusted
     participants.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5418,11 +5418,12 @@ sender.insertDTMF("123");
     has been allowed to access media, it is free to share that media
     with other entities as it chooses.</p>
 
-    <p>A mechanism, <code>peerIdentity</code>, is provided that gives
+    <p>A mechanism, <code><a>peerIdentity</a></code>, is provided that gives
     Javascript the option of requesting media that the same javascript
     cannot access, but can only be sent to certain other entities.</p>
 
-    <p>Even without WebRTC, a  Web application will know the public IP
+    <p>Even without WebRTC, the Web server providing a Web application
+    will know the public IP
     address to which the application is delivered. Setting up
     communications exposes additional information about the browser’s
     network context to the web application, including the set of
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM!

juberti:
lgtm

webrtc.html
```diff
@@ -5396,10 +5396,87 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+    limit the damage that the browser can do to other elements on the
+    local network, and it is important to protect data from
+    interception, manipulation and modification by untrusted
+    participants.</p>
+
+    <p>Mitigations include:</p>
+    <ul>
+      <li>Always requesting permission to communicate using ICE. This
+      ensures that the browser can only send to partners who you have
+      shared credentials with.</li>
```
fluffy: I don't think that prompting the user is one of the mitigations we are OK with - needs more discution with WG

webrtc.html
```diff
@@ -5396,10 +5396,87 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+      <li>Requesting ongoing permission using ICE continued
+      consent. This enables a receiver to withdraw consent to
+      receive.</li>
```
fluffy: So probably need to explain the consent is the media consent stuff and not a prompt to user.

webrtc.html
```diff
@@ -5396,10 +5396,87 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+      <li>Requiring encryption of data, with strong per-session keying
+      (DTLS-SRTP).</li>
```
fluffy: so this is not a mitigation to revealing IP and is required not something one might do

webrtc.html
```diff
@@ -5396,10 +5396,87 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+      <li>Requiring the use of congestion control. This ensures that
+      WebRTC cannot be used to flood the network.</li>
+    </ul>
+
+    <p>These measures are specified in the relevant IETF documents.</p>
```
fluffy: I'd be happier to see us just totally defer to the IETF specs (and add refs) and could make this easier to get done

webrtc.html
```diff
@@ -5396,10 +5396,87 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+    applications running in different browsers, or between an
+    application running in the same browser and something that is not a
+    browser, something that is an extension to the usual barriers in
+    the Web model against sending data between entities with different
+    origins.</p>
+
+    <p>The WebRTC specification provides no user prompts or chrome
+    indicators for communication; it assumes that once the Web page
+    has been allowed to access media, it is free to share that media
+    with other entities as it chooses.</p>
```
fluffy: Might be worth mentioning there are modes where the web page does not get access to the data but it is still sets up an audio / video stream

webrtc.html
```diff
@@ -5396,10 +5396,87 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+
+    <p>Revealing IP addresses can leak location and means of
+    connection; this can be sensitive.</p>
+
+    <p>Mitigating the exposure of IP addresses to the application requires
+    limiting the IP addresses that can be used, which will impact the
+    ability to communicate on the most direct path between endpoints.
+    Browsers are encouraged to provide appropriate controls for deciding
+    which IP addresses are made available to applications, based on the
+    security posture desired by the user.
```
fluffy: I think we need WG discussion around what mitigations would be appropriate.

webrtc.html
```diff
@@ -5396,10 +5396,87 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+    address to which the application is delivered. Setting up
+    communications exposes additional information about the browser’s
+    network context to the web application, including the set of
+    (possibly private) IP addresses available to the browser for WebRTC
+    use. Some of this information has to be passed to the
+    corresponding party to enable the establishment of a communication
+    session.</p>
+
+    <p>Revealing IP addresses can leak location and means of
+    connection; this can be sensitive.</p>
```
fluffy: Also worth noting that the connection to the webserver already leaked some of this

fluffy: We should also look at the Network API and see what it reveals and what happens.

fluffy: I think it might be worth discussing the difference of split VPN and non split VPN on this.

webrtc.html
juberti: This confused me the first time as well, but I think the intent here is that "ICE is always used, and so you can only communicate with endpoints that are expecting inbound traffic", not that an infobar shows up.

I agree it should be reworded to make it clear it's the remote endpoint's permission that we are talking about.

webrtc.html
martinthomson: Yes, we should do that to the greatest extent possible.

martinthomson:
I think that I would have liked to have seen @fluffy's comments addressed before merging this.  @burnburn, is there a plan to address those?

webrtc.html
alvestrand: Replying to 3 points in bullets:
- The fact that the connection to the webserver leaked an IP address is already in there.
- The Network API https://dvcs.w3.org/hg/dap/raw-file/tip/network-api/Overview.html has been discontinued. Last I looked at it, it didn't do anything with addresses; it was just "3G up, Ethernet down".
- In the Chrome investigation, we found not only that VPNs were troublesome, they were more diverse than a division into "split VPN" and "non split VPN" would account for. In particular, one particular type of leakage showed up only with one specific VPN product, not with other products designed to achieve the same purpose. This is too complex for a security considerations section.

webrtc.html
alvestrand: We've had that. This is the result.

webrtc.html
alvestrand: This text has been hanging around for many months (since my original drafting, which was discussed on the list). It's time to integrate it.

webrtc.html
alvestrand: The word "consent" was used here because "consent (to communicate)" is the term used in the IETF base spec. The fact that people act as if "consent" means "user prompt" no matter what the context is is unfortunate, but probably will pass. The text is, after all, rather explicit that this section does not do user prompting.
Still, edits can be suggested.

PR #195, opened 2015-02-27
RTCRtpSender.replaceTrack first cut.

jan-ivar:


webrtc.html
```diff
@@ -2389,6 +2389,58 @@
...
+              <li>
+                <p>Let <var>withTrack</var> be the argument to this method.</p>
+              </li>
+
+              <li>
+                <p>Run the following steps asynchronously:</p>
+              </li>
+              <ol>
+              <li>
+                <p>If <var>withTrack</var> is compatible with the sender's existing
```
juberti: This seems like a run-on sentence. Consider making the steps into an explicit list.

webrtc.html
```diff
@@ -2389,6 +2389,58 @@
...
+                renegotiation, then the sender should seamlessly switch to transmitting
+                <var>withTrack</var> in place of what it was sending, set the
+                <code><a href="#dom-rtpsender-track">RTCRtpSender.track</a></code>
+                attribute to <var>withTrack</var>, and resolve <var>p</var> with
+                <var>undefined</var>. Track ids are to be ignored in this
+                comparison, and the remote receiver continues to see the same
+                Track ids as before.</p>
+              </li>
+              <li>
+                <p>If <var>withTrack</var> is not compatible in this manner, reject
```
juberti: I think you should give examples of what constitutes a 'compatible' track.

webrtc.html
```diff
@@ -2389,6 +2389,58 @@
           attribute is the track that is associated with this
           <code><a>RTCRtpSender</a></code> object.</p>
         </dd>
+
+        <dt>Promise&lt;void&gt; replaceTrack (MediaStreamTrack withTrack)</dt>
```
juberti: Is there any precedent for the withTrack syntax in this document (as opposed to |track|)

webrtc.html
```diff
@@ -2389,6 +2389,58 @@
...
+                <var>p</var> with <var>IncompatibleMediaStreamTrackError</var>.</p>
+              </li>
+              </ol>
+
+              <li>
+                <p><em>Return:</em> Return <var>p</var>.</p>
+              </li>
+            </ol>
+
+            <div class="note">
```
juberti: I think onnegotiationneeded could be invoked if negotiation was in fact needed. An example would be useful to explore this option.

I think the track could come from any stream. On the receive side, the track will maintain its existing grouping, since there is no negotiation. (This is probably worth spelling out explicitly)

webrtc.html
```diff
@@ -2389,6 +2389,58 @@
...
+              <li>
+                <p>If <var>withTrack</var> is compatible with the sender's existing
+                track, in the sense that it can be switched to without needing
+                renegotiation, then the sender should seamlessly switch to transmitting
+                <var>withTrack</var> in place of what it was sending, set the
+                <code><a href="#dom-rtpsender-track">RTCRtpSender.track</a></code>
+                attribute to <var>withTrack</var>, and resolve <var>p</var> with
+                <var>undefined</var>. Track ids are to be ignored in this
+                comparison, and the remote receiver continues to see the same
+                Track ids as before.</p>
```
juberti: Track -> track

webrtc.html
jan-ivar: No precedent, just an attempt to clarify that it is the replacer, not the replacee. E.g. instead of, say, replaceTrackWith(track).

jan-ivar:
Incorporate feedback. No example yet. Seeing how negotiation is fitting in, and after some discussions with API people today, I may be coming around on perhaps using a simpler readwrite attribute instead like I believe was it you or @pthatcherg mentioned at TPAC? Should I try that here or set up another PR for comparison?

jan-ivar:
Pinging @martinthomson as well.

jan-ivar:
In short, there is apparently nothing wrong with:

    try { sender.track = myAudioTrack; }
    catch (e) { console.log(“Use ” + sender.track.kind); }

juberti:
I would do it as separate PR. IIRC, the Promise-based version allowed the application to know authoritatively when the old track had been removed and the new track inserted in its place, which seems useful.

webrtc.html
juberti: OK, that makes sense. Otherwise, I can imagine people thinking that |track| is the track to be replaced.

webrtc.html
```diff
@@ -2389,6 +2389,83 @@
...
+                  and abort these steps.</p>
+                </li>
+                <li>
+                  Set the
+                  <code><a href="#dom-rtpsender-track">RTCRtpSender.track</a></code>
+                  attribute to <var>withTrack</var>.
+                <li>
+                  <p>Determine if negotiation is needed to transmit
+                  <var>withTrack</var> in place of the sender's existing track.
+                  Ignore what <code>MediaStream</code> a track resides in and the
```
juberti: a->the

webrtc.html
```diff
@@ -2389,6 +2389,83 @@
...
+                  <p>resolve <var>p</var> with <var>undefined</var>.</p>
+                </li>
+              </ol>
+              <li>
+                <p>Return <var>p</var>.</p>
+              </li>
+            </ol>
+
+            <div class="note">
+              <p>Things like dimensions and frameRate do not require negotiation.
```
juberti: frame rate

jan-ivar:
Will do. Yeah, that's the one advantage. FWIW our implementation is currently synchronous.

juberti:
It's more complex obviously when negotiation is needed, and unclear what the state will be when the promise resolves. Is the existing track still present? Or the new track present but in a paused state?

jan-ivar:
Yes, though this seems true for negotiation in general, no?

juberti:
I think it is clearer when a new track is added or one removed; obviously the new track doesn't start streaming until the remote side accepts it. Here it's much more ambiguous when the old one ends and the new one starts.

jan-ivar:
I suppose ideally one would want the old stream to keep playing until the new one can be seamlessly swapped in on the remote end? In the last line of the other PR I tried to accommodate something like that, but we may want to be explicit about this if this is the desired behavior (I think it is, but not sure how it fits in with negotiation of other stuff). By other stuff I mean that the text on RtpSender in general - which I like - slides to that there'll be other properties that can be updated and require negotiation, besides perhaps adding and removing? Btw, what's to happen for removing a track? Mute-immediately-then-remove, or keep-playing-then-remove?

jan-ivar:
I've added text to this PR to cover track-renaming on the remote receiving end on the first subsequent renegotiation. We ran into this problem in our implementation, and we think the best way to solve replaceTrack vs. renegotiation is to "come clean" when we renegotiate, and have the remote receiver use SSRCs to detect when a new track id is really an existing track that has been renamed, and not treat it as remove/add track (an optimization of sorts).

jan-ivar:
@juberti - a question on pc.addTrack: Since MediaStreamTrack and RTCPeerConnection are both browser objects, peerConnection could simply look up what streams a track belongs to. So why the multiple stream args to addTrack? Is it a control surface for users to withhold some associations (or even compose different ones)? If so, does replaceTrack need the same? It's not clear to me how addTrack affects stream associations, or how replaceTrack should affect stream associations.

juberti:
I don't think it needs the arguments because there is no way to communicate them or for the remote side to use them - the remote side doesn't see any change to its existing track or associated groupings.

The stream args for addTrack are there to make sure the placing of a track into a stream is an explicit operation, and therefore operations like stream.removeTrack don't affect the PC state.

jan-ivar:
Ah, so peerConnection has a separate picture of stream associations to be maintained separately to control what the remote sees? This makes replaceTrack simple I suppose as the new track takes on all associations of the old, regardless of what streams it actually resides in. Thanks for clarifying.

So to update associations remotely, tracks have to be removed and re-added?

juberti:
exactly

alvestrand:
Now that RTPSender has landed, can someone present this proposal to the list?

webrtc.html
```diff
@@ -2389,6 +2389,88 @@
...
+                    the connection ends or a renegotiation happens, whichever is
+                    sooner. If a renegotiation happens, then the remote receiver
+                    SHOULD use SSRCs to detect tracks that have been renamed in
+                    this manner, and update their ids and groupings, rather than
+                    treat them as new tracks.</p>
+                  </li>
+                  <li>
+                    <p>If negotiation is needed, then fire a <a href=
+                      "#event-negotiation">negotiationneeded</a> event at the
+                    <a>RTCPeerConnection</a> associated with this sender.</p>
```
burnburn: At this point you have already set RTCRtpSender.track to the new track, before calling negotiation needed.  Is that the right behavior?
Sorry if this was covered in the prior discussions -- it wasn't clear to me.

jan-ivar: As I understand the [session negotiation model](http://w3c.github.io/webrtc-pc/#session-negotiation-model), local settings still reflect what has been set locally, even when a PeerConnection becomes "dirty" with the `negotiationneeded` flag.

E.g. if I do `var newSender = pc.addTrack(newTrack);` then `newSender` will be in pc.getSenders() even if negotiation is pending. Since senders are extensions of PeerConnection, I would expect them to work the same, e.g. sender.track reflects what was last set.

If someone could validate my understanding then that would be great.

burnburn: Oops, duh, I somehow forgot this was initiated by the JS (i.e., user code) as sender.replaceTrack().  All SDP-affecting controls in WebRTC are essentially instructions to the browser on what should be done at the next negotiation.  So yes, I would expect this behavior (thus answering my own question).

alvestrand:
I have sent out a note on the webrtc list today (April 13) with the subject line "ReplaceTrack - need to evaluate alternatives" to solicit further comments.

webrtc.html
```diff
@@ -2389,6 +2389,88 @@
...
+                  <p>resolve <var>p</var> with <var>undefined</var>.</p>
+                </li>
+              </ol>
+              <li>
+                <p>Return <var>p</var>.</p>
+              </li>
+            </ol>
+
+            <div class="note">
+              <p>Things like dimensions and frame rate do not require negotiation.
```
fluffy: Just one clarification here - changing frame rate, resolution, or bitrate to something that it still inside the negotiated envelope does not require renegotiating but changing it to something outside the envelope does.

PR #197, opened 2015-03-04
Add ORTC attribution

juberti:
@stefhak

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5958,7 +5958,7 @@ sender.insertDTMF("123");
     specification was provided by many people including Martin Thomson, Harald
     Alvestrand, Justin Uberti, and Eric Rescorla.</p>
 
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5957,6 +5957,10 @@ <h3 id="sec-constraints">Constraints</h3>
     Haza&euml;l-Massieux, for their support. Substantial text in this
     specification was provided by many people including Martin Thomson, Harald
     Alvestrand, Justin Uberti, and Eric Rescorla.</p>
+
+    <p>The RTCRtpSender and RTCRtpReceiver objects were initially conceived in the
```
alvestrand: Can we use the word "described" here?
The conception of these objects seems to have a lot more murky history, but the description is clearly based on the one in ORTC.

juberti:
Updated accordingly.

PR #200, opened 2015-03-09
onsignalingstatechange: readyState -> signalingState

bruun:
This seems to be outdated, please correct me if I'm wrong. 
As for the anchor tag, I followed the syntax from `onicegatheringstatechange`
for consistency.

I have no idea if this is the right approach for changing the spec, please let me know if this needs to go through the WG mailing list instead.


juberti:
lgtm

adam-be:
Thanks for fixing

PR #203, opened 2015-03-11
Rewrote mitigations text for security considerations section.

alvestrand:
Leaving "under consideration" marker on the IP adresss revealing text only.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5494,9 +5494,9 @@ if (sender.dtmf) {
 
     <p>Mitigations include:</p>
     <ul>
-      <li>An UA will always requesting permission from the correspondent
+      <li>An UA will always request permission from the correspondent
       UA to communicate using ICE. This
-      ensures that the browser can only send to partners who you have
+      ensures that the UA can only send to partners who you have
       shared credentials with.</li>
       <li>An UA will always request ongoing permission to continue sending
       using ICE continued
--- a/webrtc.html
+++ b/webrtc.html
@@ -5455,7 +5455,7 @@ if (sender.dtmf) {
     will know the public IP
     address to which the application is delivered. Setting up
     communications exposes additional information about the browser’s
-    network context to the web application, including the set of
+    network context to the web application, and may include the set of
     (possibly private) IP addresses available to the browser for WebRTC
     use. Some of this information has to be passed to the
     corresponding party to enable the establishment of a communication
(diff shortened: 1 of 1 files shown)


burnburn:
There are some minor typos and grammar changes I would make after applying, but I am fine with the text otherwise.  Let's hear from others.

stefhak:
LGTM.

alvestrand:
Adam, do you want to push the button? Dan can fix his language issues afterwards...

webrtc.html
```diff
@@ -5462,6 +5462,19 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
     connections between participants. One will normally assume that
     the IP address of TURN servers is not sensitive information.</p>
 
+    <p>Mitigating the exposure of IP addresses to the application requires
+    limiting the IP addresses that can be used, which will impact the
+    ability to communicate on the most direct path between endpoints.
+    Browsers are encouraged to provide appropriate controls for deciding
+    which IP addresses are made available to applications, based on the
+    security posture desired by the user.</p>
+
```
fluffy: Has the WG discussed and come to consensus around this because my understanding was we were still at the point that WebRTC would expose internal IP addresses by default. Perhaps I am just missing it but this is such a hot thing, I'd prefer if someone could point the editors at the discussion around where we decided this before we merge this. I'm not claiming I definitely know what the consensus is, I just think we need to go make check this text roughly matches the discussion before we put it in.

martinthomson: I believe that this is still largely in line with what we discussed at the IETF meeting.  That said, I think that Justin wanted some more specific guidance on what the controls might do.  I'd be OK with text to this effect being landed, especially since the following note makes it clear that we intend to consider the matter further.

alvestrand: That's why the note is on the next line saying that the WG has NOT come consensus on this.
I believe we have consensus on saying at least this, and may need to have stronger language.

webrtc.html
```diff
@@ -5469,13 +5482,23 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
-    <div class="note">
-      <p>The working group is actively discussing what additional text
-      is appropriate for this section.  In particular, the group
-      expects to list ways to mitigate the exposure of IP addresses
-      but has not yet reached consensus on the details of this
-      list.</p>
-    </div>
+    <p>Mitigations include:</p>
+    <ul>
+      <li>An UA will always requesting permission from the correspondent
```
alvestrand: Nit: requesting -> request

alvestrand:
Fixed my grammar. Push history is confusing because I got confused about "git rebase"; sorry.

burnburn:
Yes, will adjust other grammar stuff after we push the button.  Otherwise LGTM.

Issue #179, opened 2015-02-03
Potentially private IP addresses revealed


stefhak:
Host IP addresses are made available to an application that creates a PeerConnection and lets it start gathering ICE candidates, and this can be done without the user knowing.
This can be sensitive in certain scenarios (e.g. VPN) as recently discussed, OTOH host IP addresses (if reachable from a peer) are the "best" for establishing connectivity.
This issue is opened to keep track of any decisions made regarding this problem.

martinthomson:
it's not sufficient to disable host addresses.  If you care about fingerprinting, then reflexive addresses can be problematic as well.

juberti:
Disabling host addresses will require use of a STUN server, as well as a NAT that supports hairpinning, to allow connections between peers on the same LAN.

martinthomson:
If we consider this to have the same requirements as drive the [ICE transport policy](http://w3c.github.io/webrtc-pc/#idl-def-RTCIceTransportPolicy), then relay candidates are the only ones that should be permitted.

That said, I think that addressing this concern is not the business of the specification.  The specification need only note what information is exposed and describe the consequences of that exposure.

stefhak:
(@martinthomson: about the reflexive addresses and fingerprinting: are these not made available to the site already when the page is fetched - this is where the server sends the document?)

martinthomson:
@stefhak, the concerns that I've seen raised are with respect to users who are behind a proxy or VPN.  Presumably, the reflexive address that is revealed as a result is different to one that might be revealed by a STUN request made from a host candidate.

alvestrand:
Two different concerns:
- Local IP addresses are revealed to the Web page (we can hide them from correspondents) - but these are vital for efficient local connections. Fingerprint/benefit tradeoff. I suggest we document these and move on.
- Public IP addresses other than the one used as a source for browsing are revealed. This appears to depend on specific VPN configuration properties. More investigation needed.

dontcallmedom:
To add another angle, these IP addresses are potentially revealed to the server, but also to the peer.

[WebRTC Security Architecture](https://tools.ietf.org/html/draft-ietf-rtcweb-security-arch-10#section-5.4) sets requirements on the API to help avoid leaking IP addresses to the peer — we probably should review whether these requirements are currently addressed.

It also says protecting from the server is not a requirement; we probably would want to import some of the language there in the security considerations.

[...]

PR #206, opened 2015-04-13
Harmonizing issue references

dontcallmedom:
This pull request brings more uniformity in what is marked up as "issue" in the document.

I've opted to use `class="issue"` for things that were marked up as notes, but were really todos or at least meant to be temporary until something happens.

I have also configured respec to refer to github issues when using the `data-number` attribute on issues (used for #179 and #141 for now); to avoid surprises (such as discovering "issues" hidden in the media capture streams doc as we were calling for last call approval), I would like to set up a Travis CI job that alerts us for any issue that is not linked to github, or for anything marked as an issue in the doc that refers to a closed issue.

But before I can look into that, we need to import all the things marked up as issues in the doc as issues in github (if they still apply).

(using the `data-number` attribute means that other issues are no longer numbered; that's probably a feature since these numbers were not very useful in the first place)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -184,7 +184,7 @@
           [[!TRAM-TURN-THIRD-PARTY-AUTHZ]], Section 6.2.</dd>
         </dl>
 
-        <div class="note">
+        <div class="issue">
           <p>Should we have a "none" type, for cases where no authentication is
           needed? (e.g. STUN)
           </p>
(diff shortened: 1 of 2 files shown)


juberti:
Good idea. lgtm.

burnburn:
Yeah, LGTM too.

adam-be:
Rebased and pushed.

PR #209, opened 2015-04-13
makes peerIdentity nullable and default to null in RTCConfiguration

dontcallmedom:
no point in requiring it all the time
aligns with firefox's webidl http://mxr.mozilla.org/mozilla-central/source/dom/webidl/RTCConfiguration.webidl

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -157,7 +157,7 @@
             <p>Indicates which <a href="#target-bundle-policy">BundlePolicy</a> to use. </p>
           </dd>
 
-          <dt>DOMString peerIdentity</dt>
+          <dt>DOMString? peerIdentity = null</dt>
 
           <dd>
             <p>Sets the <a href="#target-peer-identity">target peer


martinthomson:
This is superceded by #178, and rots it unnecessarily.

dontcallmedom:
really? I checked #178 before submitting this, and it didn't seem to touch on RTCConfiguration at all. 

Feel free to close this PR if it is indeed misdirected.

martinthomson:
Sorry, I checked the wrong attribute.  Different comments then:

Given that `peerIdentity` is a dictionary member, what value is there in permitting an explicitly null value?  Is there some sort of general design practice that requires that optional dictionary members be made nullable?

jan-ivar:
Not that I know of. IMHO there's rarely a need to extend the value space of strings to include `null`, unless `null` has some special meaning distinct from `""`. As Martin points, out dictionary members may be omitted entirely as well.

I think `DOMString peerIdentity = "";` is best, and matches the others here.

martinthomson:
@jan-ivar and I have a disagreement about `""` and its role here, because that's an invalid string.

That said, I'm not enthusiastic about being more tolerant of varying unputs.  We have `undefined` anyway (without `required`), and I see no need to accept `null`, `0`, or `""`.  All of which indicate someone tried to do something with the parameter, but failed to do so correctly.

dontcallmedom:
I got confused on making `peerIdentity` nullable, it is indeed unclear that this brings anything useful here.

As an after-thought, `null` could be used to unset a peerIdentity in setConfiguration; but I see now that's the current preference is to [not allow change to `peerIdentity` from setConfiguration](https://github.com/w3c/webrtc-pc/pull/185/files#r25026865). And in any case, it shouldn't be the default.

(I wonder if that's a sign that `peerIdentity` should be handled in a separate parameter.)

martinthomson:
I don't think that a small difference in handling is a strong enough reason to have a separate configuration path.  Presumably, `setConfiguration` will have to prohibit other sorts of changes.

Issue #34, opened 2014-11-07
[clarification] may iceServers in RTCConfiguration be null?


AVGP:
In 4.2.1.1 the spec reads:

    iceServers of type sequence<RTCIceServer>
    An array containing URIs of servers available to be used by ICE, such as STUN and TURN server.

which doesn't specify that it may be "null" as well. At least that's what Chrome seems to allow me to do just fine. So I'm wondering if it's just me being too easily surprised here, the spec being a tad fuzzy or the implementation being off...

Sorry if this is a silly question :/

dontcallmedom:
as far as I can tell, Chrome now barfs if iceServers is null:
```javascript
new webkitRTCPeerConnection({iceServers:null});
//VM124:2 Uncaught TypeError: Failed to construct 'RTCPeerConnection': Malformed RTCConfiguration```

That being said, I'm not sure why the spec requires iceServers, or RTCConfiguration at all.

jan-ivar:
I think RTCConfiguration is required because without iceServers, connections will always fail (except to peers inside the same LAN), a common pilot error.

Firefox does let you omit RTCConfiguration because Mozilla has a default stun server [but we're in the process of removing it](https://bugzilla.mozilla.org/show_bug.cgi?id=1143827).

Enforcing invariants: No reason to support `{ iceServers:null }` over `{ iceServers:[] }`

All that said, I believe Firefox currently allows allow passing in `{ iceServers:[] }` for LAN-only use, and that seems like a good idea to me. Not sure why we're disallowing it.

juberti:
For demo pages, e.g. two PCs in the same page, there is no need for a STUN server.

I think iceServers:[] should be tolerated, along with iceServers being omitted completely, or even a null RTCConfiguration, since use of null is common in many of the existing demo pages. 

I don't see a need to allow iceServers:null, since there are simpler ways of expressing the same thing.

jan-ivar:
Firefox does not allow a `null` RTCConfiguration, never has, but agree with everything else.

juberti:
http://webrtc.github.io/samples/src/content/peerconnection/pc1/ passes null and seems to work OK in FF.

jan-ivar:
Indeed it does. I asked our resident guru and just learned that dictionary arguments are always nullable, and that passing `null` is the same as passing `{}`. So problem solved.

alvestrand:
[...]

PR #210, opened 2015-04-13
Optional RTCConfiguration in RTCPC constructor

dontcallmedom:
[WebIDL requires it](http://heycam.github.io/webidl/#dfn-optional-argument):
"If the type of an argument is a dictionary type or a union type that has a dictionary type as one of its flattened member types, and this argument is either the final argument or is followed only by optional arguments, then the argument MUST be specified as optional. "
Firefox has it optional too: http://mxr.mozilla.org/mozilla-central/source/dom/webidl/RTCPeerConnection.webidl

see also #34

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -593,7 +593,7 @@
         <h3>Interface Definition</h3>
 
         <dl class="idl" title="interface RTCPeerConnection : EventTarget ">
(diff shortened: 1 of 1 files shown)


juberti:
Agree. I can try to hit this in my updateIce PR.

alvestrand:
The remaining issue with updateIce is in text that is updated in #211

Issue #168, opened 2014-12-29
Should updateIce() update the returned value of getConfiguration()?


ibc:
It is not clear in the latest spec. If I call `updateIce(newPcConfig)` on a `RTCPeerConnection`, should `getConfiguration()` return such a `newPcConfig`?

juberti:
updateIce is now setConfiguration, so... yes.

dontcallmedom:
I think this can be closed @stefhak @alvestrand

stefhak:
I'd like to see the PR land first on this one. When reading the current version it's unclear to me if the configuration is set or not.

alvestrand:
#211 is the current PR.

alvestrand:
Closing now that #211 is merged.

PR #211, opened 2015-04-15
Rename updateIce() to setConfiguration() + Refactoring (move around some definitions)

adam-be:
This is a split PR #185 into three separate commits to make it easier to review.

It's based on the same rev as the original PR to make it easier to diff.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -999,109 +999,6 @@
             signaling state</a>.</p>
           </dd>
 
-          <dt>void setConfiguration (RTCConfiguration configuration)</dt>
-
-          <dd>
-            <p>The setConfiguration method updates the ICE Agent process of gathering
-            local candidates and pinging remote candidates.</p>
-
-            <p>This call may result in a change to the state of the ICE Agent,
-            and may result in a change to media state if it results in
-            connectivity being established.</p>
-
-            <p>When the <dfn id=
-            "dom-peerconnection-setconfiguration"><code>setConfiguration()</code></dfn>
-            method is invoked, the user MUST run the following steps to process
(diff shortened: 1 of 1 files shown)


adam-be:
Is this new name is fully compatible with how this method is defined to work? The name setConfiguration() sort of hints that it's this new configuration that must be used; nothing else. The old method, updateIce(), was defined to accept deltas and leaving members out would leave the corresponding setting unchanged.

If we want to keep the new name, it might be better to redefine its behavior.

Ping @juberti

jan-ivar:
Agree it's unintuitive with the new name. It's probably better behavior as well - with precedent in track.applyConstraints (no partial apply) - and a good time to change it if we're going to change it.

martinthomson:
Why can't this just say that this does a complete replacement of the configuration, or throws if that is not possible.  Doesn't this have to be asynchronous?  Can't this affect the operation of an outstanding `createOffer()`?

jan-ivar:
I think that's the "better behavior" yes. It doesn't need to be asynchronous that I can see, and updateIce wasn't (it's implicitly "called" by the RTCPeerConnection constructor too).

If I change a wheel while I'm driving then I'm sure it has an effect. But since we're dealing with timing, I don't see how it needs to be deterministic.

adam-be:
I'm leaning towards the replacement behavior as well. It' still easy to tweak the configuration by reading the current, changing it and doing a set.

alvestrand:
@juberti do you agree that setConfiguration() overwrites the configuration no matter what, as suggested in bug #168 ?

adam-be:
PR #213 replaces this one.

PR #213, opened 2015-04-21
PR #211 + Rewrite setConfiguration() to replace the current configuration.

adam-be:
Uses a required dictionary member RTCIceServer.urls instead of "manually" throwing if not present.

PR #214, opened 2015-04-21
Certificate management API

### LINKED ISSUE #147: Certificate management is underspecified
> Originally reported on W3C Bugzilla [ISSUE-21880](https://www.w3.org/Bugs/Public/show_bug.cgi?id=21880) Tue, 30 Apr 2013 18:14:58 GMT
> Reported by Martin Thomson
> Assigned to 

The current specification is silent on how peer certificates are managed.

Current implementations appear to mint a new certificate for every new connection.  This is good from a privacy perspective because it ensures that no two sessions can be linked based solely on certificate contents.  Of course that assumes that the certificate doesn't contain a DN that contains linkable information.

Volatile certificates create a minor performance issue (certificate generation is fast, but it's not that fast) and a major traceability issue.  If there is no stability in certificates, then you lose the ability to audit sessions.  It's not possible to determine if the peer you were talking to is really the person you expected - they are probably unable to produce the certificate that they used during the session because the certificate has since been destroyed.

This leads to a need for some degree of persistence, with certain constraints:
 1. Certificates cannot be the same across different origins, i.e., your identity must be unlinkable across origins.
 2. It must be possible to use a different certificate when talking to different people on the same site, i.e., your identity must be unlinkable across peers.
 3. Certificates need to be destroyed when cookies are destroyed, i.e., you must be able to control linkability over time.

Requirement 2 is the most interesting case because this implies that there needs to be some interaction with the application to achieve this, not just the browser.  Only the application has any hope of having prior knowledge of identity such that this would be possible.

Note: A DTLS peer acting in the server role presents their certificate without any prior knowledge of the identity of the peer, and a client presents their certificate without actually verifying the server identity.

This implies that there needs to be some control surface to allow applications to control certificate options.  This needs to be separate to the existing peer identity constraints because these privacy protections are needed in the absence of identity assertions.  In fact, anonymity cases are where these are most needed.

martinthomson:
I've discussed this with Ryan and think that this accomplishes the goals we set out for identity/key management.

This is necessarily larger than the previous proposal; it reuses less of what WebCrypto defines.  The positive side of this is that because it stores certificates as well as keying material, it ensures that fingerprints are consistent over time.

Closes #147.


webrtc.html
```diff
@@ -1254,8 +1301,14 @@
             <p>When this method is call, the user agent MUST construct new
             <code><a>RTCConfiguration</a></code> object to be returned, and
             initialize it using the ICE Agent's <a href=
-            "#ice-transports-setting">ICE transports setting</a> and <a href=
-            "#ice-servers-list">ICE servers list</a>.</p>
```
dontcallmedom: did you mean to remove this? there is dangling open &lt;a href=

webrtc.html
```diff
@@ -2582,6 +2635,111 @@
       </dl>
     </section>
 
+    <section>
+      <h2 id="sec.cert-mgmt">Certificate Management</h2>
+      <p>
+        The certificates that <code>RTCPeerConnection</code> instances use to
+        authenticate with peers are use
```
dontcallmedom: extraneous "are"

webrtc.html
martinthomson: It just reflowed by accident.  I'll try to correct it somehow.

webrtc.html
```diff
@@ -1241,7 +1288,7 @@
             for trickle ICE, as defined in Section 4.1.9 of [[!RTCWEB-JSEP]].  Prior to
             the completion
             of <a href="#dom-peerconnection-setremotedescription"><code>setRemoteDescription</code></a>,
-            this value is <code>null</code>.
+              this value is <code>null</code>.
```
alvestrand: oops?

webrtc.html
```diff
@@ -2582,6 +2636,110 @@
...
+        ([[<dfn id="RTCCertificate-certificate">certificate</dfn>]])
+        that <code>RTCPeerConnection</code> uses to authenticate with a
+        peer.</p>
+
+        <p>For the purposes of this API, the
+        [[<a href="RTCCertificate-certificate">certificate</a>]] slot contains
+        unstructured binary data.</p>
+
+        <p>Note that a <code>RTCCertificate</code> might not directly hold
+        private keying material, this might be stored in a secure module.</p>
```
alvestrand: Does this mean that if the certificate is persisted, the persisted object may have links to something that is not part of the persistent storage?
Do we have some kind of assurance that if persistent storage is cleared, the linked-to objects in the secure module get cleared out too? This makes me nervous about relative lifetimes and well-defined failure modes.

webrtc.html
```diff
@@ -4062,7 +4220,9 @@ <h3 id="sec.identity-proxy-assertion-request">Requesting Identity
         <li>The <code>RTCPeerConnection</code> sends a "SIGN" message to the IdP
         proxy. This message includes the material
         the <code>RTCPeerConnection</code> desires to be bound to the user's
-        identity.</li>
+        identity.  This includes the fingerprint of the certificate that was
+        selected or generated during the construction of
```
alvestrand: Given that "certificates" above is plural and "certificate" here is singular:
Do we have an interface for seeling which of the certificates from the "certificates" option is actually selected during the construction of the RTCPeerConnection?
Is this certificate selected during construction or during connection?

alvestrand:
Some commentary....

alvestrand:
@ekr @rsleevi have you reviewed this?

dontcallmedom:
I think @alvestrand meant to ping @sleevi

webrtc.html
```diff
@@ -2582,6 +2636,110 @@
...
+      </p>
+
+      <dl class="idl" title="partial interface RTCPeerConnection">
+        <dt>static Promise&lt;RTCCertificate&gt; generateCertificate
+          (AlgorithmIdentifier keygenAlgorithm)</dt>
+        <dd>
+          <p>This function takes an argument that describes how the private key associated with the certificate is generated.  The <var>keygenAlgorithm</var> argument uses the WebCrypto
+          [[!WebCryptoAPI]] <a href="https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#dfn-AlgorithmIdentifier">AlgorithmIdentifier</a>
+          type.  The <var>keygenAlgorithm</var> value MUST be a valid argument
+            to <a href="https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#SubtleCrypto-method-generateKey"><code>Crypto.subtle.generateKey</code></a>.</p>
```
sleevi: I'd be surprised if the TAG doesn't flag it, so it may be worth pre-emptively engaging them (and WebCrypto) for how we can facilitate this.

This sentence establishes a normative requirement, but doesn't actually provide a description of what that normative requirement means. For example, an implementation COULD interpret "a valid argument" to mean "conforms to the AlgorithmIdentifier" WebIDL specification.

At a minimum, you should reference the algorithm from WebCrypto for normalizing an algorithm ( https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#algorithm-normalization-normalize-an-algorithm ), wherein the input is an AlgorithmIdentifier alg and an operation "op"

A rough stab at how that might be expressed algorithmically

1. Let *foo* be the result of performing the (normalize an algorithm)[https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#algorithm-normalization-normalize-an-algorithm] algorithm, with `alg` set to *keygenAlgorithm* and `op` set to **generateKey**.
2. Let *bar* be the result of performing the (normalize an algorithm)[https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#algorithm-normalization-normalize-an-algorithm] algorithm, with `alg` set to *keygenAlgorithm* and `op` set to **sign**.
3. (something something about rejecting the promise if an error is thrown, or maybe that's maybe that's implicit in Promise IDL handling because I can never remember)
4. (something something about performing a deep equality check)

Of course, the proposed API **does not work** for valid use cases. For example, for RSASSA-PKCS1-v1_5, the input is a https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#dfn-RsaHashedKeyGenParams (a dictionary with keys for *name*, *modulusLength*, *publicExponent*, *hash*), but the sign parameters are null. For RSA-PSS, you have https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#dfn-RsaHashedKeyGenParams vs https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#dfn-RsaPssParams

If you meant for the AlgorithmIdentifier to just be a proxy for "name", then you probably simply want to be looking up the `name` key of *keygenAlgorithm*, and then dealing with the **[supportedAlgorithms](https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#dfn-supportedAlgorithms)** internal slot in some way.

To put it differently, I have *no* idea how such a combination of specs would be implemented in Chrome, nor am I sure how best to recommend text that could be implemented.

webrtc.html
```diff
@@ -2582,6 +2636,110 @@
...
+
+          <p>The resulting certificate MUST NOT include information that can be
+          linked to a user or <a>user agent</a>.  Randomized values for
+          distinguished name and serial number SHOULD be used.</p>
+
+          <p>A <a>user agent</a> MUST reject a call
+          to <code>generateCertificate()</code> that uses an algorithm that is
+          not supported: either because the algorithms are not supported or it
+          cannot use a certificate produced by those algorithms
+          with <code>RTCPeerConnection</code>.</p>
```
sleevi: This is another area of ambiguity; there is NO requirement that the TLS implementation, the TLS client cert implementation, the WebRTC implementation, or the Web Crypto implementation use the same cryptographic stacks.

(And indeed, as we're going through in Chrome, we actually **do** have a case where things are different depending on what you're doing). The ambiguity here is "an algorithm that is not supported", because it could be referencing Web Crypto, WebRTC (at an API layer), or the DTLS implementation used for WebRTC.

webrtc.html
sleevi: Yes, it does. WebCrypto was designed to "accommodate" this, in as much as API calls start failing, but I agree that overall this is a bad user experience, and I think we're unlikely to ever actually implement in Chrome, and I think anyone who did try to implement it is in for a bad time. But at least Web Crypto provides for defined behaviours if they did (it'd just be what developers hated)

webrtc.html
martinthomson: I think that you are incorrectly assuming that this needs to *use* the WebCrypto infrastructure in any way.  `AlgorithmIdentifier` is just a dictionary.  This is just using the already-defined syntax in WebCrypto to describe what it means.

Also, we aren't allowing applications to say anything about the signing process.  You're assuming a lot when you invent step 2 there.

I find the W3C habit of overspecificity in specs stifling.  What is wrong with describing inputs and outputs?  The question you need to ask here is this: Are you capable of taking information that identifies a key generation algorithm and either turning that into a certificate+private key or failing?  Why does it matter so much how that happens?  

The question of whether the algorithm identification is sufficiently well specified is definitely something I'm willing to engage on here.  The bit where the browser chooses the signature algorithm is here largely in response to your feedback, but I'm happy to iterate on that.  To that end, I will definitely reference the normalization algorithm, though it's really only a hint for implementations.

webrtc.html
martinthomson: Do you think that it is impossible to work out how to thread that needle?  It might mean knowing what you support in the various involved components, but that's hardly special.  All this takes is identifying what options work and rejecting all others.

Why do you think we have to identify what parts of the browser internals need to support anything?  The browser either supports X or it doesn't.  (Again, WebCrypto has no play here, the spec is only being used because it provides a reusable method for identifying key generation algorithms and parameters.)

webrtc.html
martinthomson: I think that Harald's concern is that you might store the keys in a removable HSM, remove the HSM, clear cookies for the site, and now there there is some private keying material on the HSM.  The concern is that this is now a sharp object lying around, awaiting some future compromise.

That's what PFS is for :)

The allowance in WebCrypto only really covers the case where the keying material is disconnected, causing all operations on keys to reject.  I agree that might be annoying, but I don't see any point losing sleep over something you can't change.  The only alternative is to have CryptoKeys share fate with the private keying material, but that means that you don't get to have nice things like HSMs.

sleevi: I disagree. There's equally the concern where you have the certificate, clear the HSM, and now you have... what?

webrtc.html
sleevi: I think it's underspecified and not reliably implementable. I know you mean well, but I think I laid out why it's problematic.

Again, "the browser either supports X or it doesn't" is a statement with many interpretations, few of them interoperable.

webrtc.html
sleevi: > I think that you are incorrectly assuming that this needs to use the WebCrypto infrastructure in any way.
> `AlgorithmIdentifier` is just a dictionary. This is just using the already-defined syntax in WebCrypto to describe what it
> means.

Then I think this shares the same fundamental flaw of your previous PR, in which it looks like a duck, but it neither walks, talks, acts like a duck, but you're still calling it a duck.

> Also, we aren't allowing applications to say anything about the signing process. You're assuming a lot when you 
> invent step 2 there.
>
> I find the W3C habit of overspecificity in specs stifling. What is wrong with describing inputs and outputs?

I'm surprised you didn't realize how closely related these two statements are. The fact that reasonable people (or at least, I hope I am), the people most likely to be implementing or involved it in, the people responsible for this, can reach different conclusions from what _you_ meant is **exactly** why the "stifling" process of specing exists. To ensure that we can get compatibility.

You said the operation must support signing. I described to you how, to Web Crypto, one describes whether or not the algorithm supports signing. That's "clearly" not what you meant, but that's not at all clear in what you propose.

The habit is not saying "You must do exactly X" - it's that you must, from the observer's point of view, have taken these inputs and produced these outputs. Any side-effects of the outputs that can be observed by script **MUST** be consistent between implementations, as decades of experience have shown us people _will_ come to rely on this.

> To that end, I will definitely reference the normalization algorithm, though it's really only a hint for implementations.

I don't really care what you do, and it was more an example of how you _might_ do it, one that was admittedly flawed. What I want to ensure with this PR is that it's unambiguous what needs to happen, and as it stands, it's quite ambiguous and (as a result), not _consistently_ implementable.

webrtc.html
martinthomson: I didn't mean to suggest that this proposal was in any way different to WebCrypto in that regard.  I intended to agree with you.

webrtc.html
martinthomson: Good catch, this needs to be plural here.

martinthomson: Oh, but this is likely to change significantly with the identity rework.

webrtc.html
martinthomson: I don't think that you did explain.  But you used a lot of words and that can cause a message to be obscured sometimes.

sleevi: You said in your reply that "Web Crypto" has no play. But that's not what the spec says. The spec doesn't say _anything_. That's the problem.

webrtc.html
martinthomson: OK, let try to be clear: you are concerned about using duck-typing into WebCrypto concepts here.  I'm sorry that you feel that way, but I'm not inclined to back down on this. 

This is only duck-typing, not reuse.  I'm fully prepared to break any dependencies on WebCrypto by using copy and paste.  I can produce a duck made with all-new meat if necessary.  That's just more work for me.

Regarding specification style, I accept that you are explaining a method for getting to the same outcome. I'm happy to talk about how the approach you advocate produces bad work that fails to properly respect the Priority of Constituencies, but this isn't the time and place.

webrtc.html
```diff
@@ -2582,6 +2636,116 @@
...
+          [[!WebCryptoAPI]] <a href="https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#dfn-AlgorithmIdentifier">AlgorithmIdentifier</a>
+          type.  The <var>keygenAlgorithm</var> value MUST be a valid argument
+            to <a href="https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#SubtleCrypto-method-generateKey"><code>Crypto.subtle.generateKey</code></a>.</p>
+
+          <p>The algorithm name used in the <code>AlgorithmIdentifier</code>
+          MUST be valid for both key generation and signing.
+          The <var>keygenAlgorithm</var> value MUST produce a non-error result
+          when normalized according to the
+          WebCrypto <a href="https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#algorithm-normalization">algorithm
+          normalization process</a> [[!WebCryptoAPI]].  The following algorithm
```
sleevi: As noted in the link you provided, and what I specified earlier, algorithm normalization requires both an input and a desired operation.

Line 2669 says "MUST be valid for both key generation and signing".

Possible interpretations:
1) Normalize for generateKey
2) Normalize for sign
3) Normalize for generateKey and sign
4) Normalize for some other operation
5) Just convert DOMString inputs to { name: "foo" } form

Only due to following this PR, not due to the proposed language itself, I suspect you want 1.
However, from following this PR, it's unclear to me whether or not it is *possible* to request a 512-bit key. Are you just using the names, or are you also using the generateKey parameters in full?

If using the generateKey parameters, then it's unclear if it's valid to request a 512-bit RSA key. Or an even exponent. Or a non-prime exponent. or e=3 when the browser really, really, really hates non-F4.

martinthomson: It's 1, yes. I'll fix that.

The browser has a responsibility to decide whether it likes the arguments.  That's down below: the bit you thought was ambiguous.

fluffy:
LGTM (other than might need rebase)

martinthomson:
Rebased.

stefhak:
Some comments:
a) "sequence<RTCCertificate> certificates;" is added to the RTCConfiguration. Is that the right place (I'm thinking about that all other parts of the configuration can be updated - but this one not and it is just said "can't be updated")?
b) Same area: it is said that certs in the RTCConfig can't be updated, but surely that only applies if the configuration has been used in a PeerConnection? Before that everything could be changed?
c) further, there is no error check whether or not the certificates supplied to the configuration are valid or not. What should happen if the app supplies something, partly or completely, bogus? (RTCConfiguration is said to be validated by running the updateIce steps, but there is no check regarding the certs there)
d) On line 2664 is said "That set MUST include an ECDSA certificate with a private key on the P-256 curve and a signature with a SHA-256 hash." This is the first time we introduce this kind of reqs in this document. Does it belong here, or is it part of something we reference? And if it belongs here, we should at least reference.
e) The term DTLS is used a bit, but never explained or any ref done (this was actually the case already from the IsoelatedTrack stuff). Compare to ICE, this term is used too, but there is a reference and some explanation as well.
f) On line 2708 "browser SHOULD select SHA-256"; same comment as d) above (but here is a reference actually!)
g) Line 2721: see comment d) above
h) In section 5.5.1 "X.509" is used. But no explanation and no reference.

martinthomson:
> a) "sequence certificates;" is added to the RTCConfiguration. Is that the right place (I'm thinking about that all other parts of the configuration can be updated - but this one not and it is just said "can't be updated")?

I want this set before createOffer or setRemoteDescription are called.

> b) Same area: it is said that certs in the RTCConfig can't be updated, but surely that only applies if the configuration has been used in a PeerConnection? Before that everything could be changed?

I didn't see any point in making that distinction.  Learn to commit.

> c) further, there is no error check whether or not the certificates supplied to the configuration are valid or not. What should happen if the app supplies something, partly or completely, bogus? (RTCConfiguration is said to be validated by running the updateIce steps, but there is no check regarding the certs there)

I didn't deem validation to be necessary: if the certificate was generated by the browser, it needs to be good.  That said, I think that there is probably something to do here in light of the potential addition of expiry.

My problem here is that `updateIce()` says nothing about what to do if validation fails.  I'm going to open a separate issue on this; I'd rather see a "validate RTCConfiguration" algorithm specified that can be used prior to setting the values as currently described.

> d) On line 2664 is said "That set MUST include an ECDSA certificate with a private key on the P-256 curve and a signature with a SHA-256 hash." This is the first time we introduce this kind of reqs in this document. Does it belong here, or is it part of something we reference? And if it belongs here, we should at least reference.

Yeah, I'm working on updating the security architecture, but I have no direct control over any of this.

> e) The term DTLS is used a bit, but never explained or any ref done (this was actually the case already from the IsoelatedTrack stuff). Compare to ICE, this term is used too, but there is a reference and some explanation as well.

I figured that this was already present.  If it is not, I don't think that is a fault of this change.

> f) On line 2708 "browser SHOULD select SHA-256"; same comment as d) above (but here is a reference actually!)
g) Line 2721: see comment d) above
h) In section 5.5.1 "X.509" is used. But no explanation and no reference.

> Update: Regarding d) f) and g) above: should we not reference http://rtcweb-wg.github.io/security-arch/#rfc.section.5.5 here?

Maybe.

alvestrand:
Since this PR references updateIce, which we are renaming to setConfiguration, I suggest we submit that one after #213 has been committed, and this one is updated to use that interface.

Having the "certificates" be immutable creates an issue with setConfiguration; do we require that an update of the configuration use getConfiguration first to get the value it can pass in with setConfiguration, or do we exceptionally say that "passing in Null will not blank the value of this item"?

jan-ivar:
Do you mean `undefined` e.g. dictionary member absence? RTCConfiguration.certificates is not nullable (and besides I would interpret a `null` as a desire to erase rather than leave alone).

jan-ivar:
That still doesn't solve that `pc.setConfiguration(pc.getConfiguration());` must work.

Have we considered attributes (with RTCConfiguration still serving as the constructor InitDict)?

alvestrand:
This needs a rebase, but I think we're reasonably OK otherwise. @martinthomson can you rebase?

fluffy:
I think it would be better to define the MTI and preferd crypto algo in the IETF security draft.

martinthomson:
@fluffy, I agree, but...  

That means significantly more work for me, and that's not going to happen for some time (I'm going to prioritize implementation of this over document shuffling).  And it means pulling in a whole lot more context into those documents.  If @ekr is willing to sign on to adding that context, it might be good to move those parts.  Right now, they have no home.

ekr:
Yes, I'm happy to put this in the IETF doc.

martinthomson:
How do you propose that we proceed then?  I was going to propose that we take this and move MTI parts later.

ekr:
Yes, that sounds good to me.

PR #215, opened 2015-04-21
Adding expires attribute to certificate management

martinthomson:
Using private keying material for ever is not a good idea.  This adds an expiration to the certificates parameters.

This adds to #214.  If you don't take that PR, this isn't really valid.


fluffy:
I suspect this looks fine but lets get #214 resolved first then look at this one

alvestrand:
Looking at the diff, the PR includes #214 too. Can we get a diff that only contains the expires attribute text?

martinthomson:
@alvestrand, I was simply waiting for #214 to be completed before I rebased this.  It should be OK now.  Dependencies are hard.

PR #218, opened 2015-04-22
Enqueue addIceCandidate

ekr:
ISTM that addIceCandidate needs to be queued because it has to be ordered wrt setRD.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -458,7 +458,8 @@
 
         <p>Once the RTCPeerConnection object has been initialized, for every
         call to <code>createOffer</code>, <code>setLocalDescription</code>,
-        <code>createAnswer</code> and <code>setRemoteDescription</code>;
+        <code>createAnswer</code>, <code>setRemoteDescription</code>,
+        and <code>addIceCandidate</code>;
         execute the following steps:</p>
 
         <ol>


alvestrand:
This has Travis errors, but they seem unrelated to the PR.

stefhak:
I think Ekr is right, it should be queued.

adam-be:
This is needed. addIceCandidate() needs the remote description set so (possibly) waiting behind setRemoteDescription() in the queue makes sense.

The change is suspiciously tiny, but I believe adding the method name to the list of queued operation is sufficient as it's currently defined.

jan-ivar:
agree

fluffy:
LGTM

juberti:
Convinced by Adam's argument

webrtc.html
```diff
@@ -458,7 +458,8 @@
 
         <p>Once the RTCPeerConnection object has been initialized, for every
         call to <code>createOffer</code>, <code>setLocalDescription</code>,
-        <code>createAnswer</code> and <code>setRemoteDescription</code>;
+        <code>createAnswer</code>, <code>setRemoteDescription</code>,
+        and <code>addIceCandidate</code>;
```
jan-ivar: Do we also need to add it below where it says:
> The general idea is to have only one among `createOffer`, `setLocalDescription`, `createAnswer` and `setRemoteDescription` executing at any given time.

?

PR #219, opened 2015-04-23
Fix broken links

dontcallmedom:
Required to make Travis CI happy again, now that the bug that prevented the linkchecker to run has been fixed.


adam-be:
The link checker succeeds locally on @dontcallmedom's machine, so this is probably at issue on the Travis server. Merging.

Issue #139, opened 2014-12-15
WebRTC spec should explicitly specify all causes of a PeerConnection-sourced track being muted


dontcallmedom:
> Originally reported on W3C Bugzilla [ISSUE-25513](https://www.w3.org/Bugs/Public/show_bug.cgi?id=25513) Wed, 30 Apr 2014 07:12:28 GMT
> Reported by Kiran
> Assigned to 

As discussed in Bug 25361 [1], WebRTC spec should explicitly specify all the muted events of a MediaStramTrack related to RTCPeerConnection

1. Tracks sourced from m-mline and marked as RecvOnly in SDP should be considered muted tracks.
2. Muted state resulted due to doohickey actions.

// May be some more other events related to PeerConnection.

[1] https://www.w3.org/Bugs/Public/show_bug.cgi?id=25361

stefhak:
Closing since #220 has been merged.

PR #220, opened 2015-04-23
Clarify muted state (proposed fix for issue #139)

adam-be:


adam-be:
Perhaps the algorithm for updating muted state should go into the Media Stream and Capture spec and only be referred to from webrtc-pc.

adam-be:
This actually only deals with one issue brought up in #139.

adam-be:
Updated PR to refer to the Media Capture and Stream spec for the specifics of (un)muting a track.

webrtc.html
```diff
@@ -4625,16 +4629,21 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
-      <p>In addition, a <code>MediaStreamTrack</code> has its
-      <code>muted</code> attribute set to <code>true</code> on the
-      remote peer if the local user agent makes inactive the corresponding
-      <code><a>MediaStreamTrack</a></code> that is being sent.</p>
+      <p>When an <code><a>RTCPeerConnection</a></code> receives data on an RTP
+      source for the first time, it MUST <a href="#update-track-muted">update
+      the muted state</a> of the corresponding <code><a>MediaStreamTrack</a>
+      </code> with the value <code>false</code>.</p>
+
+      <p>When an <code><a>RTCPeerConnection</a></code>'s RTP source is
```
juberti: This seems underspecified. How long is "temporarily unable"?

alvestrand: This is from getusermedia - the muted state means that no data is coming, but the track is not promising that it's permanent - so it's not ended.
It could be for a very long time. Does it matter to any of the rest of the spec how long it is?

juberti: I think some guidance to implementors is advisable to make clear in which situations this should occur (e.g. no packets for N seconds, or some specific state change like an ICE disconnection)

adam-be:
Rebased and updated to clarify "temporarily unable to receive media".

Issue #198, opened 2015-03-05
ontrack events need to fire before setRemoteDescription resolves


martinthomson:
The spec says:
> This will be fired only as a result of setRemoteDescription.   Onnaddstream [sic] happens as early as possible after the setRemoteDescription. 

The latter sentence is not right.  It makes it impossible to build some forms of negotiation based on onaddstream because there is no definitive event that marks the end of the set of events.  setRemoteDescription resolving is that event.  For instance, if an application can only display one video track, it has to choose, and if it doesn't know that the set of options presented by a peer is complete, making an optimal choice is impossible.

It should say something like:
> The onaddstream event fires only during the processing of setRemoteDescription.  All streams and tracks that are signaled in a remote description MUST be exposed in an onaddstream event prior to the time that the promise returned by setRemoteDescription resolves.

ibc:
100% agreed.

jan-ivar:
Works for me.

juberti:
Agreed, although onaddstream is now ontrack.

It should also say that getRemoteTracks must return the complete set of tracks if called immediately after the promise is resolved.

jan-ivar:
Yes, I gather the same rule applies. We'll wanna fix up old getRemoteStreams() as well, as it'll probably need to stick around for quite some time.

martinthomson:
@juberti 
> It should also say that getRemoteTracks must return the complete set of tracks if called immediately after the promise is resolved.

I'd rather say that getRemoteTracks() MUST produce a track when `ontrack` fires for that track.  I think that's stronger.

i.e., the algorithm is roughly:
```js
newtracks.forEach(track => {
  remoteTracks.push(track);
  dispatchEvent(new TrackAddedEvent('track', track));
});
resolveSetRemoteDescriptionPromise();
```

adam-be:
This should not be documented in the track/stream even handler attribute description. The latest spec addresses this to some extent.

We need to do some work in section: 5.1.3 Processing Remote MediaStreamTracks and in the description of setRemoteDescription().

adam-be:
Proposed fix: #221

stefhak:
Closing as #221 has been merged.

maitrungduc1410:
This save my day

My problem is that `ontrack` doesn't fire, then I change the order or my code, move `ontrack` on top of `setRemoteDescription`, then it works.

Thanks

fippo:
[...]

PR #221, opened 2015-04-24
Define when RTCRtpReceivers are created and dispatced (issue #198)

adam-be:
* Remove all text about the event dispatching procedure from the event handler attribute description.
* Add text in setRemoteDescription() that describes the extra steps needed before resolving the promise.
* Use "media description" instead of "component" to describe what populates the fields of the new MediaStreamTrack. I would like to reference the JSEP spec here, but I couldn't find a suitable definition to refer to.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1028,7 +1028,7 @@
             <p>If the <code><a>RTCSessionDescription</a></code> argument is
             applied successfully, the user agent MUST <a href=
             "#dispatch-receiver">dispatch a receiver</a> for all new
-            media descriptions [[!SDP]] before queuing the <a href=
+            media descriptions [[!RTCWEB-JSEP]] before queuing the <a href=
             "#setlocal-resolve">setLocalDescription() resolve task</a>.</p>            
 
             <p>If an <code>a=identity</code> attribute is present in the session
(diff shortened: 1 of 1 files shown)


fluffy:
@adam-be what are you like to have in JSEP and we can add something there

adam-be:
@fluffy, I created an issue for this in the JSEP repo
https://github.com/rtcweb-wg/jsep/issues/132

alvestrand:
Tagging @juberti for comment

adam-be:
Updated to reference JSEP instead.

webrtc.html
```diff
@@ -2293,39 +2299,46 @@
...
+                  attribute to the media description track id.</p>
                 </li>
 
                 <li>
                   <p>Initialize <var>track's</var> <code>label</code>
                   attribute to "<code>remote audio</code>" or "<code>remote
-                  video</code>" depending on the media type of the incoming
-                  component.</p>
+                  video</code>" depending on the media type of the media
+                  media description.</p>
```
stefhak: "hte media media"?

adam-be:
Rebased. Hitting the button when Travis says OK.

PR #222, opened 2015-04-24
Replace operations array with a chain.

jan-ivar:
#165 Take two (wouldn't let me push changes to a closed PR).

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -416,7 +416,8 @@
           </li>
 
           <li>
-            <p>Set <var>connection</var>'s <code>operations</code> to the result
+            <p>To prevent having errors from one operation cancel subsequent
+            operations, set <var>connection</var>'s <code>operations</code> to the result
             of transforming <var>p</var> by a rejection handler that returns
             <var>undefined</var>.</p>
           </li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -383,8 +383,9 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to a resolved promise, to
-            represent the start of a chain of <code>operations</code>.</p>
+            <p>On the <var>connection</var>, initialize an internal slot
+            <var>operations</var> to a resolved promise, to represent the start
+            of a promise chain.</p>
           </li>
 
           <li>
@@ -399,27 +400,38 @@
 
         <ol>
           <li>
-            <p>Let <var>p</var> be the result of transforming
-            <code>operations</code> by a fulfillment handler that runs the
+            <p>Let <var>operation</var> be a function executing the steps of the
+            call in question.</p>
+          </li>
+          <li>
+            <p>If <var>connection</var>'s <a href=
+              "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+              signalingState</a> is <code>closed</code>, throw an
+              <code>InvalidStateError</code> exception and abort these steps.</p>
+          </li>
+          <li>
+            <p>Let <var>p</var> be the result of transforming <var>connection</var>'s
+            <var>operations</var> by a fulfillment handler that runs the
             following steps:</p>
             <ol>
               <li>
                 <p>If <var>connection</var>'s <a href=
                 "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
-                signalingState</a> is <code>closed</code>, abort these steps.</p>
+                signalingState</a> is <code>closed</code>, return
+                <var>undefined</var> and abort these steps.</p>
               </li>
               <li>
-                <p>Return the result of executing the current call
-                being handled.</p>
+                <p>Return the result of executing <var>operation</var>.</p>
               </li>
             </ol>
           </li>
 
           <li>
-            <p>To prevent having errors from one operation cancel subsequent
-            operations, set <var>connection</var>'s <code>operations</code> to the result
+            <p>To point to the new tail of the chain, and to prevent having
+            errors from one operation cancel subsequent operations, set
+            <var>connection</var>'s <var>operations</var> to the result
             of transforming <var>p</var> by a rejection handler that returns
-            <var>undefined</var>.</p>
+            <var>undefined</var> (this has no effect on <var>p</var>).</p>
           </li>
 
           <li>
@@ -431,9 +443,8 @@
         <code>setLocalDescription</code>, <code>createAnswer</code> and
         <code>setRemoteDescription</code> executing at any given time. If
         subsequent calls are made while one of them is still executing, they
-        are added to a queue and processed when the previous operation is fully
-        completed. It is valid, and expected, for normal error handling
-        procedures to be applied.</p>
+        are added to a chain and processed upon fulfillment or rejection of the
+        previous operation's promise.</p>
 
         <p>Additionally, during the lifetime of the RTCPeerConnection object,
         the following procedures are followed when an ICE event occurs:</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -383,8 +383,8 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to represent a queue of
-            <code>operations</code> with an empty set.</p>
+            <p>Initialize an internal variable to a resolved promise, to
+            represent the start of a chain of <code>operations</code>.</p>
           </li>
 
           <li>
@@ -399,23 +399,30 @@
 
         <ol>
           <li>
-            <p>Append an object representing the current call being handled
-            (i.e. function name and corresponding arguments) to the
-            <code>operations</code> array.</p>
+            <p>Let <var>p</var> be the result of transforming
+            <code>operations</code> by a fulfillment handler that runs the
+            following steps:</p>
+            <ol>
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>closed</code>, abort these steps.</p>
+              </li>
+              <li>
+                <p>Return the result of executing the current call
+                being handled.</p>
+              </li>
+            </ol>
           </li>
 
           <li>
-            <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            <p>Set <var>connection</var>'s <code>operations</code> to the result
+            of transforming <var>p</var> by a rejection handler that returns
+            <var>undefined</var>.</p>
           </li>
 
           <li>
-            <p>When the asynchronous operation completes (either successfully
-            or with an error), remove the corresponding object from the
-            <code>operations</code> array. After removal, if the array is
-            non-empty, execute the first object queued asynchronously and
-            repeat this step on completion.</p>
+            <p>Return <var>p</var>.</p>
           </li>
         </ol>
 


jan-ivar:
Discuss here: https://lists.w3.org/Archives/Public/public-webrtc/2015Apr/0137.html

webrtc.html
```diff
@@ -383,8 +383,8 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to represent a queue of
-            <code>operations</code> with an empty set.</p>
+            <p>Initialize an internal variable to a resolved promise, to
```
alvestrand: What internal variable? Is it a temporary, or is it referred to somewhere else?

webrtc.html
```diff
@@ -399,23 +399,31 @@
 
         <ol>
           <li>
-            <p>Append an object representing the current call being handled
-            (i.e. function name and corresponding arguments) to the
-            <code>operations</code> array.</p>
+            <p>Let <var>p</var> be the result of transforming
```
alvestrand: What is the reference for "transforming X by a fulfilment handler"? Please add pointer.

webrtc.html
```diff
@@ -399,23 +399,31 @@
...
+            <code>operations</code> by a fulfillment handler that runs the
```
alvestrand: What does "operations" refer to? If it is the internal variable referred to above, then that sentence needs rephrasing.

webrtc.html
```diff
@@ -383,8 +383,8 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to represent a queue of
-            <code>operations</code> with an empty set.</p>
+            <p>Initialize an internal variable to a resolved promise, to
+            represent the start of a chain of <code>operations</code>.</p>
```
alvestrand: What does "operations" refer to?

webrtc.html
```diff
@@ -399,23 +399,31 @@
...
+            <code>operations</code> by a fulfillment handler that runs the
+            following steps:</p>
+            <ol>
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>closed</code>, abort these steps.</p>
+              </li>
+              <li>
+                <p>Return the result of executing the current call
```
alvestrand: What does "return" mean in this context? Which code path gets a returned value?

webrtc.html
```diff
@@ -399,23 +399,31 @@
...
-            <code>operations</code> array.</p>
+            <p>Let <var>p</var> be the result of transforming
+            <code>operations</code> by a fulfillment handler that runs the
+            following steps:</p>
+            <ol>
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>closed</code>, abort these steps.</p>
+              </li>
```
alvestrand: And what happens once these steps are aborted?

alvestrand:
I'm sorry, I don't understand the proposed text.
I can't even figure out if new operations are supposed to be prepended or appended to the promise-chain.

webrtc.html
jan-ivar: It's from http://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting
See also other comments in https://github.com/w3c/webrtc-pc/pull/165

webrtc.html
jan-ivar: That's what I meant yes. How does it need to be phrased? I tried as much as possible to rely on existing language in order to keep general cleanup of this section orthogonal to this PR.

alvestrand: The problem is that "operations" is undefined. I left a suggestion at the place where it's likely you intended it to be defined.

webrtc.html
alvestrand: If you write the sentence as "Initialize an internal variable <code>operations</code> to a resolved promise, to represent...", the grammar makes sense (and "operations" is defined for later use).

webrtc.html
jan-ivar: Nothing happens, just like before. Here are some [code comments on this](http://mxr.mozilla.org/mozilla-central/source/dom/media/PeerConnection.js?rev=b11576671375#383) that may help.

webrtc.html
jan-ivar: Thanks! Will do.

jan-ivar:
New operations are appended to the promise chain (there's no way to prepend to a promise chain, if that helps). If you have any other specific questions, maybe I can help.

Would looking at code rather than spec legalese help? From [Firefox's implementation](http://mxr.mozilla.org/mozilla-central/source/dom/media/PeerConnection.js?rev=b11576671375#383):

    _chain: function(operation) {
      this._checkClosed(); // out here DOMException line-numbers work.
      let p = this._operationsChain.then(() => {
        if (!this._closed) {
          return operation();
        }
      });
      // don't propagate errors in the operations chain (this is a fork of p).
      this._operationsChain = p.catch(() => {});
      return p;
    },

Basically, tiny functions that (check we're not closed and) execute and return the result from `opereation` are .then()'ed onto the promise held by `this._operationsChain` (initialized to `Promise.resolve()`), and the resulting promise from that (which is the promise that can fail and needs to be returned to the caller) is assigned to `p`.

Then it fixes up the `this._operationsChain` pointer to point to the new tail of the chain: which is the tail promise returned from doing a .catch() on `p`. We do a catch because the chain should always continue regardless of errors (it is an internal conveyor-belt of sorts).

This relies on a feature that multiple .then()/.catch() functions can be added to a single promise, and such functions are executed in FIFO order when the promise is resolved. Importantly, these are not dependent on each-other, rather each one represents a fork in the chain with its own failure path (i.e. `p`' still returns a failure to the caller, in spite of the catch() in the internal fork. Makes sense?

webrtc.html
alvestrand: The head of that document is "a collection of interesting ideas". How stable is it going to be?

Personally, I find "call p.then(fulfilment handler, undefined)" *MUCH* more readable than "transforming x by a fulfilment handler". Is there *any* reason why we should prefer the more convoluted langauge?

alvestrand:
I'd be very happy if you were to add into the PR a reference to an actual spec document saying what you say above (with chapter and verse references).
I'd also be very much happier if you were to propose text that I could actually map into the operations above.
The operations seem reasonable. I find the description incomprehensible.

webrtc.html
jan-ivar: Heading aside, that's the tag document describing guidelines for using promises in specs, and is actively updated by @domenic - If we're going to follow a guideline here I believe this is the authoritative one. While this particular phrase may not be the clearest, isn't fixing that best done at the tag level, so everyone benefits and we're consistent?

jan-ivar:
@domenic, any tips on better phrasing here? I followed your guideline so it must be your fault ;-)

webrtc.html
domenic: p.then(fulfillment handler, undefined) is not correct because what happens if some script on the page does `Promise.prototype.then = function () { throw new Error("No, I am your father."); }`.

I can change the heading of the document to "TAG Finding" if that would help.

jan-ivar: Couldn't hurt.

webrtc.html
jan-ivar: FWIW this language predates this patch. Do we want a separate issue?

alvestrand:
To be clear, before accepting this language, I want this PR to include:

1) A normative dependency on Domenic's guidance document
2) Domenic's assurance that he'll fight for such a reference being accepted by the approving bodies when the time comes to push this document to CR and people freak out over the normative references.

@domenic, the fact that we have to use language that's extremely hard to parse for anyone who hasn't been part of the language development process instead of snippets of the code that we're all having to read anyway is an architectural issue, and the TAG should probably take it up as such.

The way we solve it when using Javascript to implement browsers is to declare special contexts that users can't get at and say "our feature-implementing Javascript runs here" - if it works for implementations, perhaps it works for specifications too.

jan-ivar:
I think treating new things different from old things dates a spec. Promises are part of WebIDL; Why would we need normative references to talk about them?

jan-ivar:
Incorporated definitions and more text.

alvestrand:
@jan-ivar the definitions I'm complaining about ("transforming by a fulfilment handler") is not part of WebIDL. If it was part of WebIDL, I'd follow the reference and read it there. They're part of Dominic's guidelines for usage of that version of WebIDL, and it's *that* definition that needs to be either referenced or copied.
(The fact that the WebIDL we're using is an editor's draft, which defines ECMAScript by a reference to a private draft in a private repository is just how the world is at the moment. Sooner or later it'll all shake out.)

jan-ivar:
I've opened an issue. Lets see if we can make the language clearer without resorting to normative references just to describe descriptions.

jan-ivar:
#### Q: Any precedent for async methods or attributes that queue a task to start?

Hi @annevk, @domenic, @bzbarsky, we’re having trouble mixing synchronous and asynchronous APIs in WebRTC, and I would love your JS expertise vetting two ideas!

The first problem is a synchronous `sender` control interface to ongoing live transmission.

The idea is to treat all attributes set synchronously as bulk:

    sender.foo = x;
    sender.bar = y;
    Promise.resolve().then(() => {
      // both x and y apply to the transmission
    });
    // neither x nor y apply to the transmission yet
    console.log(sender.foo); // x
    console.log(sender.bar); // y

It must guarantee that x and y are always applied together on the live transmission, never one without the other. It might need a `sender.onerror` to react to things that can't be caught synchronously.

Is this sound? Intuitive? Too limiting? Any precedence for something like this?

The alternative seems to be the set/get-methods-with-dictionary pattern:

    let settings = sender.getSettings();
    settings.foo = x;
    settings.bar = y;
    sender.setSettings(settings).then(() => {
      // both x and y apply to the transmission
    });
    // neither x nor y apply to the transmission yet
    settings = sender.getSettings();
    console.log(settings.foo); // not x
    console.log(settings.bar); // not y

Thoughts? Preference?

The second problem is an interface `pc` with four async methods that internally chain when called so only one runs at a time. It’s a state-machine. One of the methods is called createOffer.

We thought the internal chaining was a good idea to protect the state-machine from users calling methods at the wrong time (this was pre-promises), but there are some unintended surprises:

    pc.addTrack(x);
    pc.createOffer().then(offer => {
      // if chain was empty, then offer includes x only
      // if chain was non-empty, then offer includes x and y
    });
    pc.addTrack(y);

To solve this, one idea is to queue a task before invoking the method even when the queue is empty, resulting in the offer always including x and y. Intuitive? Any precedence to rely on here?

Another idea is to rip out the internal queue and just throw if people call one of these mutually-exclusive methods when another one is outstanding. That would mean that `pc.createOffer()` above would reject its promise in the "chain was non-empty" scenario.

Does that make any sense?

All thoughts welcome!

bzbarsky:
I'm pretty sure there is precedent for things where the setter of an attribute queues a task to do some processing.  That processing can examine the then-current state of the objects involved.

If you just do that, then in your first example the `// both x and y apply to the transmission` bit would not hold, since the task would not have run by then.  You can maybe get around that by using a microtask instead, but I wouldn't worry about doing that unless there's a _really_ good reason to do it.

I don't have a strong opinion about this vs the "pass in a dictionary" approach, except that one more naturally gives you a Promise to use for waiting until the settings are applied.

> To solve this, one idea is to queue a task before invoking the method even when the queue is empty

If the queue is nonempty, is it guaranteed that it will get flushed within a single event loop turn?  As in, if at a later time someone does `pc.addTrack(z)`, is there some guarantee that `z` won't be included in the offer?

Throwing when things are in the wrong state is certainly an option... if we think it would be web-compatible at this point.

jan-ivar:
Thanks! Is a "microtask" the same as `.then()` ?

> > To solve this, one idea is to queue a task before invoking the method even when the queue is empty

Ugh, sorry I meant "when the *chain* is empty" here (I've edited my comment). What we have internally in `pc` is a chain of asynchronous operations, where the next one starts when the previous one ends. We used to erroneously refer to this as a queue, and I slipped up.

> If the queue is nonempty, is it guaranteed that it will get flushed within a single event loop turn? As in, if at a later time someone does pc.addTrack(z), is there some guarantee that z won't be included in the offer?

The chain can remain non-empty for a long time, and the cutoff for `pc.addTrack(z)` being included would be when it's the internal chained createOffer implementation's turn to actually execute. The rationale for chaining behind another async call to begin with is that the new call was done at the "wrong time", and the new call is effectively delayed until the "right time", so it makes sense (we think) for it to run with whatever exists at that (right) time, which may include `z`. If one knows what's on the queue then one can use `pc.onsignalingstatechange` to detect succession in the chain and predict the outcome (in theory).

Since `pc.addTrack(z)` is sync and the others are async, it is easy to get ahead of oneself. I hope users use explicit promises now that we support them, and avoid this confusing implicit behavior.

annevk:
(To @alvestrand's issue above, the promise language should become part of IDL. IDL should make it easy to work with promises and it should just define what "a new promise" etc. actually means.)

To @jan-ivar's issue, the setters on the `sender` object set internal slots. I would have the setters do something like this:

1. Set the corresponding internal slot to the given value.
1. If the *already queued flag* is set, return.
1. Set the *already queued flag*.
1. Queue a task to run these substeps:
  1. Do whatever you were planning on doing on the internal slots.
  1. Unset the *already queued flag*.

The getters can just return the corresponding internal slot.

(You could also make this a microtask I suppose if synchronization needs to happen faster.)

bzbarsky:
> Is a "microtask" the same as .then()

Let's say yes.  (In reality, nothing actually defines the behavior here so far, but the intent is that those are the same concept.)

> The chain can remain non-empty for a long time

Then I don't see how the "queue a task" approach really helps.  It'll cover up some races against the chain emptying, but not others.

alvestrand:
If "microtask" isn't defined, let's not use the term.
As I understand it, a .then is run by the Javascript event loop, just like regular queued tasks.
More undefined terms makes life harder.

bzbarsky:
"microtask" is perfectly well defined.  What's not defined is whether `then()` is a microtask or not.

`then()` is run by the JavaScript event loop, but regular queued tasks are not: those are run by the browser event loop. The interaction of those two event loops is not defined.

alvestrand:
@bzbarsky not defining the relationship between stuff executed by then() and stuff executed by the browser event loop sounds like a bug. Who should we call on to fix that?

domenic:
This looks like a lot of contortions, resulting in a fairly unintuitive API. Or at least I'm pretty confused, although I admit I don't have the full background.

> The alternative seems to be the set/get-methods-with-dictionary pattern:

This looks a lot nicer. Although I'd quibble about the naming, e.g. it should probably be `sender.settings` for the getter, as otherwise the asymmetry between sync `getSettings` and async `setSettings` is weird.

If that code is representative then I'd consider a dedicated utility for overriding existing settings, e.g. instead of that you could do

```js
sender.setSettings({ foo: x, bar: y }).then(() => {
  // both x and y apply to the transmission
});
// etc.
```

bzbarsky:
> Who should we call on to fix that?

Some combination of hixie and TC39.  But they know about the problem, and have for a while now.  It's just that there is no agreement on how exactly to solve the problem.

jan-ivar:
@domenic - Thanks for your comments. The idea would be that users would not worry about synchronization, and just do:

    sender.foo = x;
    sender.bar = y;

If that's not intuitive then this approach doesn't have much going for it, as KISS was supposed to be its main benefit.

>     sender.setSettings({ foo: x, bar: y }).then(() => {
>       // both x and y apply to the transmission
>     });

Will this overwrite `{ fum: z }`? Interestingly, we're in the process of renaming [`updateIce()`](http://w3c.github.io/webrtc-pc/#dom-peerconnection-updateice) to `setConfiguration()` (its scope got broadened) and change it to overwrite absent members in the process, reasoning that the word "set" (in combination with a same-named getter) implied overwriting with the entire dictionary rather than with just members that were specified. Default values also have unintended consequences in the partial overwrite pattern.

domenic:
> If that's not intuitive then this approach doesn't have much going for it

Yeah I think

```js
// neither x nor y apply to the transmission yet
console.log(sender.foo); // x
console.log(sender.bar); // y
```

is very unintuitive.

> Will this overwrite { fum: z }?

I was thinking no, but I agree the naming is a bit confusing. Indeed maybe `updateSettings` is clearer in that regard. Maybe you even need both `setSettings` and `updateSettings`.

jan-ivar:
@domenic - On the chance that I've explained it poorly, with "apply to the transmission" I mean the underlying implementation rather than anything observable from JS. The API to the settings is synchronous, while the implementation acts on the settings a bit later. How could a synchronous API to transmission or another time-based operation work any other way?

domenic:
Yes, the expectation is generally that when you do `implementationObj.prop = value`, that will apply to the underlying implementation immediately. Anything else is pretty bizarre.

> How could a synchronous API to transmission or another time-based operation work any other way?

If you're operating on something out-of-process (which I imagine an ongoing transmission is), then the API for interfacing with it should definitely not be synchronous.

jan-ivar:
It is applied immediately as far as JS is concerned. Compare to sync APIs like HTMLMediaElement.srcObject and .onloadedmetadata. That's sort of where I was going. Some (fairly key) actions (like loading or transmission) are on other threads (not other processes), but when they start exactly seem almost immaterial. I want your opinion, just making sure we're not miscommunicating. RTCPeerConnection has a mix of sync and async today, so I'd be surprised if there was a clean line here,  but I'd be happy to hear if there is.

stefhak:
We should probably close this one given that #236 seems like a replacement.

alvestrand:
Replaced by #236.

Issue #160, opened 2014-12-15
add new acces for the active remote/local SDP


dontcallmedom:
> Originally reported on W3C Bugzilla [ISSUE-25808](https://www.w3.org/Bugs/Public/show_bug.cgi?id=25808) Mon, 19 May 2014 15:58:19 GMT
> Reported by Cullen Jennings
> Assigned to 

add two new ones and decide if current ones which are propsoed sdp need a name change.

PR #225, opened 2015-04-30
Support for pending and current SDP

fluffy:
The will close https://github.com/rtcweb-wg/jsep/issues/16
and fixes https://github.com/w3c/webrtc-pc/issues/160
and fixes https://github.com/w3c/webrtc-pc/issues/141 

I think Justin and someone else from the editor team should review this before it is merged.


webrtc.html
```diff
@@ -754,12 +754,12 @@
...
             from one media format to a different, incompatible format, the
             <code><a>RTCPeerConnection</a></code> must be able to
-            simultaneously support use of both the old and new local
+            simultaneously support use of both the current and pending local
             descriptions (e.g. support codecs that exist in both descriptions)
             until a final answer is received, at which point the
-            <code><a>RTCPeerConnection</a></code> can fully adopt the new local
-            description, or rollback to the old description if the remote side
-            denied the change.</p>
+            <code><a>RTCPeerConnection</a></code> can fully adopt the prending local
```
juberti: pending

webrtc.html
```diff
@@ -903,13 +903,89 @@
...
+                    type of "offer", and it has a version that is later than the
+                    currentLocalDesc, then the pendingLocalDesc will be set to
+                    the argument and the state will transition to
+                    have-local-offer.  </p> </li>
+
+                    <!-- B transition rollback to haveLocalOffer to stable -->
+                    <li> <p>If the local description was set, and the
+                    <code><a>RTCSessionDescription</a></code> argument is an
+                    Offer, and it has a version that is the same as the
+                    currentLocalDesc, then this is a rollback and the
```
juberti: I thought this would happen via type:rollback.

fluffy: type:rollback is much more direct as an indication so sounds good to me. I'll update the PR

fluffy:
I think that fixed all Justin's comments

webrtc.html
```diff
@@ -903,13 +901,87 @@
...
+                    <code><a>RTCSessionDescription</a></code> argument has a
+                    type of "offer", and it has a version that is later than the
+                    currentLocalDesc, then the pendingLocalDesc will be set to
+                    the argument and the state will transition to
+                    have-local-offer.  </p> </li>
+
+                    <!-- C transition haveRemoteOffer or haveLocalProvAnswer to
+                    stable --> <li><p>If the local description was set, and the
+                    <code><a>RTCSessionDescription</a></code> argument has a
+                    type of "answer", then this complets an offer answer
```
stefhak: complets -> completes

stefhak:
Bikeshed: I'm not sure about the renaming of localDescription into currentLocalDesc (and the same for the remote variant). I think localDescription and pendingLocalDescription works better.
But I think the PR looks good, and that we can merge.

alvestrand:
I'm certain I don't like localDescription -> currentLocalDesc. It breaks running code, for no particularly good reason that I can discern.

webrtc.html
```diff
@@ -903,13 +901,87 @@
...
+                  
                   <li>
-                    <p>Set <var>connection</var>'s description attribute (
-                    <code><a>localDescription</a></code> or
-                    <code><a>remoteDescription</a></code> depending on the
-                    setting operation) to the
-                    <code><a>RTCSessionDescription</a></code> argument.</p>
-                  </li>
+
+                  <p>Set the <var>connection</var>'s varios description attributes
```
alvestrand: Nit: varios -> various. Word can be deleted, too.

webrtc.html
```diff
@@ -903,13 +901,87 @@
...
+                  by executing one of the following. </p>
+                  
+                  <p class="Issue"> NOTE: The principles of pending and current SDP
```
alvestrand: Class needs to be "issue" (lower case) in order to get it correctly formatted.

webrtc.html
alvestrand: Should there be a test for state here?

webrtc.html
```diff
@@ -903,13 +901,87 @@
...
+                    stable --> <li><p>If the local description was set, and the
+                    <code><a>RTCSessionDescription</a></code> argument has a
+                    type of "answer", then this complets an offer answer
+                    negotiation and currentLocalDesc is set to the argument,
+                    currentRemoteDesc is set to the value of pendingRemoteDesc,
+                    then pendingRemoteDesc and pendingLocalDesc are set to null,
+                    and the state will transitions to stable.  </p> </li>
+
+                    <!-- D transition stable to haveRemoteOffer --> <li> <p>If
+                    the remote description was set, and the
```
alvestrand: This section is under setLocalDescription.

I think the idea of having a processing model described under a method call is a bad one. We have a section (4.3.1 Operation) that has multiple processing models already (and should be clear that this is what they are). If we want a single list, I think it should be moved up there, and pointed to from both methods.

webrtc.html
```diff
@@ -903,13 +901,87 @@
...
+                    InvalidStateError error will be generated.  </p> </li>
+
+                     <!-- H transition haveRemoteOffer to haveLocalProvAnswer
+                    --> <li><p>If the remote description was set, and the
+                    <code><a>RTCSessionDescription</a></code> argument has a
+                    type of "pranswer", then pendingRemoteDesc will be set to
+                    the argument and the state will transitions to
+                    have-remote-pranswer unless the state was stable,
+                    have-remote-offer, or have-local-pranswer in which case an
+                    InvalidStateError error will be generated.  </p> </li>
```
alvestrand: The previous step ("or the type is wrong for the current signaling state") already checks state (but - bug - doesn't say what's legal in each state). There should be no need to repeat the check here.

webrtc.html
```diff
@@ -1035,18 +1132,47 @@
...
+          <dd>
+            <p>The <dfn id=
+            "dom-peerconnection-pendingRemoteDesc"><code>pendingRemoteDesc</code></dfn>
+            attribute MUST return null if the PeerConnection is in the stable
+            state and otherwise must return the remote
+            <code><a>RTCSessionDescription</a></code> that is in the process of
+            being negotiated plus any remote candidates that have been supplied
+            via <code><a href=
+            "#dom-peerconnection-addicecandidate">addIceCandidate()</a></code>
+            since then.</p>
```
alvestrand: Referent for "then"?
Assuming that this is "since the setLocalDescription call was started"?

webrtc.html
```diff
@@ -1873,6 +1999,17 @@
             may be applied as a response to an SDP offer or as an update to a
             previously sent SDP "pranswer".</p>
           </dd>
+
+          <dt>rollback</dt>
+
+          <dd>
+            <p>An RTCSdpType of "rollback" indicates that a description should be
+            treated as an canceling the current SDP negotiation and moving back
+            to the SDP [[!SDP]] offer and answer from the previos stable
```
alvestrand: Nit: previos -> previous

webrtc.html
```diff
@@ -1873,6 +1999,17 @@
...
+            state. Note the previous stable state could be null if there has
+            not yet been a successful offer-answer negotiation.</p>
```
alvestrand: Nit: note -> note that
The previous state can't be null - the localDescription can be null. According the definition, "stable" is the name of the state before negotiation takes place too (RTCPeerState enum definition).

alvestrand:
The sense of the PR looks good to me, but there are lots of nits and where-to-put-things that need fixing.

webrtc.html
fluffy: The only state this is not expected in is "stable", but if you get it in stable it still might be the best thing to do to just use it vs cause an error. I can go either way on this. Thoughts?

webrtc.html
fluffy: I was thinking that we might want to do this later  but the way we have now minimizes the changes from existing doc so I was thinking of making the technical change in this PR, then having WG look at it and later make the organization PR to move it around as you suggest. That said, what does the editor team think. I can go either way

webrtc.html
fluffy: I think I would rather rephrase the earlier error check as having this here helps keep the flow in the right place but I'm sort of missing where the earlier one happens. Perhaps it can be clarified a bit. Don't really care deeply about where the text is as long as it is clear this causes an error.

fluffy:
next steps on this one ?

alvestrand:
Adam wanted a read through it - if he's OK with it, and can fix the needed rebase, he can merge it.
If he needs help, he'll ask.
(I think that text block needs moving still, but we can do that later.)

jan-ivar:
> The localDescription attribute MUST return pendingLocalDesc if it is not null and otherwise it MUST return currentLocalDesc. 

and

> The pendingLocalDesc attribute MUST return null if the PeerConnection is in the stable state and otherwise must return the local RTCSessionDescription that is in the process of being negotiated plus any local candidates that have been generated by the ICE Agent since the offer or answer was created.

This PR adds four attributes where only two seem needed to me: `stableLocalDescription` and `stableRemoteDescription` (ice candidates added everywhere).

ekr:
We had consensus in the DC interim to add pending.

jan-ivar:
I see an API optimization.

`localDescription` seems indistinguishable from `pendingLocalDesc` in this PR, other than that the latter is `null` in stable state (when `localDescription` and `stableLocalDescription` would be identical).

Seems we have more attributes than information to me, and I find reasoning about 3 attributes (some of which toggle to and from `null` in the state machine!) harder to reason about than I think it needs to be here.

jan-ivar:
`let isPending = pc.localDescription != pc.stableLocalDescription;`

ekr:
Please see Justin's comment in https://github.com/rtcweb-wg/jsep/issues/16 form March 13, where he lays out the options.

jan-ivar:
That's where I posted initially and moved my comment here. @juberti would you mind commenting here?

alvestrand:
@jan-ivar let isPending = pc.localDescription != pc.stableLocalDescription; only works because the v= number is supposed to increment in each negotiation iteration, so even if you renegotiate the same proposal, you get descriptions that don't compare equal.

jan-ivar:
Thanks for the info. I just meant this to be illustrative, not saying anyone should actually rely on such a test over checking the signaling state directly.

adam-be:
Regarding moving text into the Operations section, I think it's a good idea to not move too much text around as we're adding new functionality. That makes it hard to see what's actually new text. However, I support doing the change in a follow-up PR.

The new text in the setRemote/LocalDescription processing model specifically dictates how the new attributes (currentLocalDesc, pendingLocalDesc, ...) should be set. So, the descriptions of these attributes should say that the attribute MUSUT return the value that the UA last set it to (plus candidates). We can keep the text that describes what the attribute represents, but without the MUSTs.

The text in the processing model should link to types (e.g. RTCSdpType) and attributes.

Regardless if we go for a 3 or 2 attribute approach; the "Desc" abbreviations hurts my eyes.

fluffy:
Splitting these up to separate threads.... On the "Desc" abbreviation ... We could change all of these to Description or we could make it as an open issue asking WG what they want. I think other people had complained Description was too long but is is more consistent with rest of spec. Iether way this is pretty editorial so if Adam just wants to pick one, good with me and people can comment on the list to change it.

fluffy:
Adam says 
    "The new text in the setRemote/LocalDescription processing model specifically dictates how the new   attributes (currentLocalDesc, pendingLocalDesc, ...) should be set. So, the descriptions of these attributes should say that the attribute MUST return the value that the UA last set it to (plus candidates). We can keep the text that describes what the attribute represents, but without the MUSTs."

I'll update the PR to address this. Note that remoteDescription and localDescription I will leave as they are because the attributes is the only place that explains how they are synthesized.

fluffy:
On the topic of the text in spec processing should link to stuff ... totally agree but there are lot of places where this needs some fixing so rather do a pass over the whole spec and fix it that worry about it for stuff that I don't know if we are merging. Right now it just makes it harder to review the diffs. Lets get the actual text in then we can worry about the stuff that sets the color of the text.

adam-be:
Thanks for fixing this @fluffy. I noted that there's text saying that the remote descriptions should be updated with local candidates generated by the Ice Agent instead of remote candidates from addIceCandidate(), but I think those are just copy-paste errors.

I was thinking about initializing all descriptions to null in the constructor algorithm so we don't have to have exceptions for unset values everywhere. I can fix the candidate issue mentioned above at the same time.

PR #228, opened 2015-05-09
Initial pass at adding QoS API

fluffy:
Can the editor team take a quick review of this and see if it seems like it is on the right track.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3036,7 +3036,7 @@
 	<dt>attribute QOSValue qos</dt>
 	  <dd>
 	    The <dfn
-	    id="dom-rtpreceiver-qos"><code>RTCDataChannel.qos</code></dfn>
+	    id="dom-rtcdatachannel-qos"><code>RTCDataChannel.qos</code></dfn>
 	    attribute is the value of the QoS setting that is used when sending
 	    packets for this RTCDataChannel.
 	  </dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -2056,31 +2056,39 @@
     </section>
 
 	<section>
-	<h2>QoS Model</h2>
+	<h2>Priority and QoS Model</h2>
 
 	<p> Many applications have multiple media flows of the same
-	data type and often some are more important than others.
-	WebRTC uses the Quality of Service (QoS) framwork described in
-	[[!TSVWG-RTCWEB-QOS]] to provide DSCP markettings for packets
-	that will help provide QoS in some networking envriontments.
-	The QOS seting can be used to indicate the relative priority
-	of variso flows. Likewise, in a video conference where the
-	flows in the conference is of the same data type but contains
-	different media types, the flow for audio may be more
-	important than the video flow. JavaScript applications can
-	tell the browser whether a particular media flow is high,
-	medium, low or of very low importance to the application.
-	</p>
+	data type and often some of the flows are more important than
+	others.  WebRTC uses the priority and Quality of Service (QoS) framework
+	described in [[!RTCWEB-TRANSPORTS]] and [[!TSVWG-RTCWEB-QOS]]
+	to provide priority and DSCP marketing for packets that will
+	help provide QoS in some networking environments. The
+	priority setting can be used to indicate the relative priority
+	of various flows. The priority API allows the JavaScript
+	applications to tell the browser whether a particular media
+	flow is high, medium, low or of very low importance to the
+	application by setting the <code>RTCRtpSender.priority</code>
+	or <code>RTCDataChannel.priority</code> attribtues to one of
+	the follwing values.  </p>
 
 	<section>
-	<h4>QOSValue Enum</h4>
-	<dl class="idl" title="enum QOSValue">
-	  <dt>very-low</dt> <dd>See [[!TSVWG-RTCWEB-QOS]]</dd>
-	  <dt>low</dt> <dd>See [[!TSVWG-RTCWEB-QOS]]</dd>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2054,6 +2054,36 @@
         </dl>
       </section>
     </section>
+
+	<section>
+	<h2>QoS Model</h2>
+
+	<p> Many applications have multiple media flows of the same
```
alvestrand: Should this also contain a reference to rtcweb-transport, for the intra-browser prioritization aspect?

webrtc.html
```diff
@@ -2481,7 +2511,16 @@
             associated, via <code>setLocalDescription</code>,
             with this <code><a>RTCRtpSender</a></code> object.
           </dd>
-        </dl>
+
+	  <dt>attribute QOSValue qos</dt>
+	  <dd>
+	    The <dfn
+	    id="dom-rtpreceiver-qos"><code>RTCRtpSender.qos</code></dfn>
```
alvestrand: copy/paste problem? receiver in hidden tag, sender in text?

alvestrand:
I think this looks like a reasonable direction. Note: The data channel also needs a qos attribute, as specified in -transport-.

alvestrand:
Note this output in the Travis build:

"file:/home/travis/build/w3c/webrtc-pc/build/output.html":3194.39-3194.39: error: Duplicate ID “dom-rtpreceiver-qos”.

stefhak:
QoS can mean so much, e.g.
- Packet marking to enable the network to prioritize (which this PR is about)
- In a situation when there is congestion, back off a little less on one flow and a little more on another
- If shortage of computational resources, try to maintain the framerate for one video, but allow another to slip
- etc.

Three questions:
1. Is this QoS API supposed to tell the UA only about the packet marking, or give a guidance for all processing that affects QoS and where there can be a limited amount of resources?
2. (If going beyond the packet marking) how would we spec what the UA should do based on this? (I seem to remember something in an I-D about different queues being scheduled depending on QoS - is that part of the answer?)
3. Should there be some text explaining why an app developer should not put "highest" prio on all tracks?

webrtc.html
fluffy: thanks - yep - typo

fluffy:
@alvestrand  - I think I do have the qos attribute for the - search for RTCDataChannel.qos (still trying to figure out best way to reply to github review comments)

fluffy:
@alvestrand @stefhak - I'll rework this to point at rtcweb-transport which I think will take care of most the comments. Do you think the attribute name should be changed from qos to priority ?

alvestrand:
You and me both wrt github comments - I think all the other review systems I've used are better than Github's default system.
I like priority better than qos for an attribute name. Priority low/medium/high is just one input into the qos algorithm (media type being the other for RTP), so the qos name is too expansive.
Now I see the datachannel.qos attribute!

webrtc.html
```diff
@@ -2481,7 +2511,16 @@
...
             with this <code><a>RTCRtpSender</a></code> object.
           </dd>
-        </dl>
+
+	  <dt>attribute QOSValue qos</dt>
+	  <dd>
+	    The <dfn
+	    id="dom-rtpsender-qos"><code>RTCRtpSender.qos</code></dfn>
+	    attribute is the value of the QoS setting that is used when sending
+	    packets for this RTCRtpReceiver.
```
alvestrand: I think this is a place where it says Receiver instead of Sender.

stefhak:
I also like "priority", and I even think "TransportPriority" or similar could be even better because it scopes what this API does. OTOH, when the API is on the RTPSender object you get a pretty good idea of that it applies to transport (the only other thing I could think of being in scope of "Priority" there would be codec resources).

alvestrand:
@dontcallmedom Travis build fails with

ERROR: Bad reference: [RTCWEB-TRANSPORTS] (appears 1 times)
ERROR: Bad reference: [TSVWG-RTCWEB-QOS] (appears 1 times)

Can you make these references legal?

fluffy:
After the reference thing gets fixed, what's the next step on this on?

alvestrand:
Adam will merge it.

dontcallmedom:
I've filed a [request to add a reference for TSVWG-RTCWEB-QOS](https://github.com/tobie/specref/pull/190).

The RTCWEB-TRANSPORT reference already exists in the db, but the trailing "S" should be removed.

fluffy:
OK - I think this is ready to be merged. (Thanks Dom for ref update)

webrtc.html
```diff
@@ -2481,7 +2519,17 @@
             associated, via <code>setLocalDescription</code>,
             with this <code><a>RTCRtpSender</a></code> object.
           </dd>
-        </dl>
+
+	  <dt>attribute PriorityValue priority</dt>
```
juberti: I think this is not the right place for this attribute. As a per-ssrc attribute, it should be on RtpEncodingParams (which haven't landed in the spec yet).

alvestrand: I don't think I agree.
a) RtpEncodingParams don't have a proposal yet. That means it's at risk.
b) -transport's prioritization algorithm talks about prioritization in terms of various tracks being sent at various priorities. Burying them one level down seems like the wrong idea.

Suggest we leave it on RTP sender, but say (when RtpEncodingParams gets proposed) that it can be overridden (to a lower value only, I assume) for individual SSRCs.

alvestrand:
Superseded by #234

PR #229, opened 2015-05-22
Removing optionality from constructors

martinthomson:
Removing the weird optionality from RTCSessionDescription and its constructor.

Making RTCIceCandidateInit.candidate required.  The dictionary argument to the constructor is not optional anyway (which is a WebIDL violation).

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1919,24 +1919,22 @@
 
         <dl class="idl" data-merge="RTCSessionDescriptionInit" title=
         "interface RTCSessionDescription">
-          <dt>Constructor (optional RTCSessionDescriptionInit
-          descriptionInitDict)</dt>
+          <dt>Constructor (RTCSessionDescriptionInit descriptionInitDict)</dt>
 
           <dd>The <dfn id=
           "dom-sessiondescription"><code>RTCSessionDescription()</code></dfn>
-          constructor takes an optional dictionary argument,
+          constructor takes an dictionary argument,
           <var>descriptionInitDict</var>, whose content is used to initialize
-          the new <code><a>RTCSessionDescription</a></code> object. If a
-          dictionary key is not present in <var>descriptionInitDict</var>, the
-          corresponding attribute will be initialized to null. If the
-          constructor is run without the dictionary argument, all attributes
-          will be initialized to null. This class is a future extensible
-          carrier for the data contained in it and does not perform any
-          substantive processing.</dd>
+          the new <code><a>RTCSessionDescription</a></code> object. If
+          the <code>sdp</code> dictionary key is not present
+          in <var>descriptionInitDict</var>, the corresponding attribute will be
+          initialized to null. This class is a future extensible carrier for the
+          data contained in it and does not perform any substantive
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1909,24 +1909,22 @@
...
         <dl class="idl" data-merge="RTCSessionDescriptionInit" title=
         "interface RTCSessionDescription">
-          <dt>Constructor (optional RTCSessionDescriptionInit
-          descriptionInitDict)</dt>
+          <dt>Constructor (RTCSessionDescriptionInit descriptionInitDict)</dt>
 
           <dd>The <dfn id=
           "dom-sessiondescription"><code>RTCSessionDescription()</code></dfn>
-          constructor takes an optional dictionary argument,
+          constructor takes an dictionary argument,
```
jan-ivar: a dictionary

webrtc.html
```diff
@@ -1909,24 +1909,22 @@
...
           <var>descriptionInitDict</var>, whose content is used to initialize
-          the new <code><a>RTCSessionDescription</a></code> object. If a
-          dictionary key is not present in <var>descriptionInitDict</var>, the
-          corresponding attribute will be initialized to null. If the
-          constructor is run without the dictionary argument, all attributes
-          will be initialized to null. This class is a future extensible
-          carrier for the data contained in it and does not perform any
-          substantive processing.</dd>
+          the new <code><a>RTCSessionDescription</a></code> object. If
+          the <code>sdp</code> dictionary key is not present
```
jan-ivar: Dictionaries have members, not keys. I know this mistake predates this patch, but I checked and this looks like the sole place in this spec where we refer to a dictionary 'key' in this way, so maybe a good time to nip this?

jan-ivar:
> The dictionary argument to the constructor is not optional anyway (which is a WebIDL violation).

Interestingly, the [RTCPeerConnection constructor](http://w3c.github.io/webrtc-pc/#interface-definition), violates WebIDL in the same way.

I think making iceServers required here would fit existing intent (and alleviate some prose around iceServers).

Wanna do this here or in a new issue?

martinthomson:
Let's talk about that on the list.  I think that we might want to make it optional.

adam-be:
If the sdp property on RTCSessionDescription is nullable, I see no reason why the corresponding member in the init dictionary shouldn't be nullable as well. That way, we can give it a default value of null and remove some prose.

martinthomson:
Default nullable WFM.

adam-be:
@martinthomson, can you please update the PR? If you don't have the time, let me know and I can do it.

martinthomson:
Note that this most recent change is going to change the serializer behaviour slightly.

`{ "type": "rollback", "sdp": null }` rather than `{ "type": "rollback" }`

Happy to make the change, but I think that this is less good now that I look at it.

martinthomson:
Oh, belay that, we already have that problem.  I'm not sure what prose we can remove though.

martinthomson:
Found it.

adam-be:
Thanks. LGTM.

Issue #232, opened 2015-05-28
No way to limit bufferedAmount without polling


bemasc:
Discussed in this thread:
https://lists.w3.org/Archives/Public/public-webrtc/2015May/0073.html

I will propose a change based on that discussion.

PR #233, opened 2015-05-28
Add a "bufferedamountlow" event.

bemasc:
Fixes https://github.com/w3c/webrtc-pc/issues/232

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3240,9 +3240,9 @@
           <code title="event-RTCDataChannel-bufferlow"><a href=
           "#event-datachannel-bufferlow">bufferlow</a></code> event fires.
           The <code><a href=
-          "#dom-datachannel-lowbufferlimit">lowBufferLimit</a></code>
-          defaults to zero, but the application may change its value at any
-          time.</p>
+          "#dom-datachannel-lowbufferlimit">lowBufferLimit</a></code> is
+          initially zero on each new <code><a>RTCDataChannel</a></code>, but
+          the application may change its value at any time.</p>
         </dd>
 
         <dt>attribute EventHandler onopen</dt>
--- a/webrtc.html
+++ b/webrtc.html
@@ -3253,6 +3253,13 @@
         objects implementing the <code><a>RTCDataChannel</a></code>
         interface.</dd>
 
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3227,6 +3227,24 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+          "#dom-datachannel-bufferedamount">bufferedAmount</a></code> is
+          considered to be low.  When the <code><a href=
+          "#dom-datachannel-bufferedamount">bufferedAmount</a></code>
+          decreases from above this threshold to equal or below it, the
+          <code title="event-RTCDataChannel-bufferlow"><a href=
+          "#event-datachannel-bufferlow">bufferlow</a></code> event fires.
+          The <code><a href=
+          "#dom-datachannel-lowbufferlimit">lowBufferLimit</a></code>
+          defaults to zero, but the application may change its value at any
+          time.</p>
```
martinthomson: I think that you want to say that this value starts at zero rather than defaults.

bemasc: OK, I've improved the wording here.

bemasc:
I didn't add an `onbufferlow` because I think `onwhatever` is rightly deprecated (and there's no compatibility concern for new events), but if you want to keep parallelism I'm happy to add it.

OK, I've changed the names to `bufferedAmountLowThreshold` and `bufferedamountlow`.

jan-ivar:
@bemasc - Do you have a link about `onwhatever` being deprecated?

bemasc:
[Mozilla](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Event_handlers) says "If you're working with a modern browser, use addEventListener() instead."

`onwhatever` has been removed entirely from the DOM Level 2 Events and is never mentioned in the [living standard](https://dom.spec.whatwg.org/#events).  Admittedly, I haven't found any explicit guidance on whether to include it for new events or not.

jan-ivar:
Thanks for the link. PeerConnection has `onxxx` for all its events I think, and AFAIK they were all added after this advice.

bemasc:
OK, I've added `onbufferedamountlow`.

PR #234, opened 2015-05-28
Add RTCRtpParameters, RTCRtpSender.getParameters, and RTCRtpSender.se…

pthatcherg:
…tParameters.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2513,10 +2513,41 @@
         <h3>RTCRtpSender Interface</h3>
 
         <p>The <code>RTCRtpSender</code> interface allows an application to control how a given
-        <code>MediaStreamTrack</code> is encoded and transmitted to a remote peer.
-        When attributes on an <code>RTCRtpSender</code> are modified, the encoding is either
-        changed appropriately, or a negotiation is triggered to signal the new encoding
-        parameters to the other side.</p>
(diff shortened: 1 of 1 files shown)


fluffy:
I think we need to have a bunch of list discussion about this first. I would much rather do this as part of 1.1 not 1.0. The ORTC work has shown that we need to think carefully about this. We need to figure out how we are going to represent the encodings before we start adding API for it.

pthatcherg:
In this PR, I have only included the things that have had a lot of list
discussion, and which WG chairs specifically asked me to make a PR.

For example, there has been lots of discussion, and I believe consensus
(except for one WG member) on RtpParameters, getParameters, and
setParameters. That goes all the way back to TPAC 2014. The only reason I
didn't write a PR months ago for this if that RtpSenders only landed
recently.

Similarly, there has been lots of discussion and I believe rough consensus
on having a .active and a .priority, so I have included both.

It sounds to me like the only big issue remaining, and the one you
explicitly mention, is whether to have a sequence of encoding parameters or
not. In 1.0, it would only allow a single value anyway (no simulcast), so
it's not much different, except that it would allow future expansion to
support multiple encodings.

We've discussed that on the list before also. But I'd be glad to continue
that discussion. That's probably part of why the chairs asked me to make a
PR: to clarify what details we have left to discuss.

alvestrand:
The chairs conclude that this approach has the consensus.

PR #235, opened 2015-05-29
Modernize getStats.

jan-ivar:
We didn't do getStats when we added promises to the spec, which seems like an oversight.

I've heard rumors since last year that getStats is moving - https://github.com/w3c/webrtc-stats/issues/3 - but it's still here and an eye-sore, so here's a PR to update it to promises like the others. If it is moving, consider this a poke. ;)

Disclaimer: This section appears not to have been updated in a while, so I had to take some liberties with the language (e.g. I believe `TypeError` is the current thinking on invalid parameters rather than the non-existent `InvalidParameters`?)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1576,6 +1576,44 @@
             invoke <var>failureCallback</var> with <var>r</var> as the
             argument.</p>
           </dd>
+
+          <dt>void getStats(MediaStreamTrack? selector, RTCStatsCallback
+          successCallback, RTCPeerConnectionErrorCallback failureCallback)</dt>
+
+          <dd>
+            <p>When the <code>getStats</code> method is called, the
+            user agent MUST run the following steps:</p>
+
+            <ol>
+              <li>
+                <p>Let <var>selector</var> be the method's first argument.</p>
+              </li>
+              <li>
+                <p>Let <var>successCallback</var> be the callback indicated by
+                the method's second argument.</p>
+              </li>
+              <li>
+                <p>Let <var>failureCallback</var> be the callback indicated by
+                the method's third argument.</p>
+              </li>
+              <li>
+                <p>Invoke <a
+                  href="#widl-RTCPeerConnection-getStats-Promise-RTCStatsReport--MediaStreamTrack-selector">
+                  RTCPeerConnection.getStats()</a>
+                with <var>selector</var> as the sole argument, and let
+                <var>p</var> be the resulting promise.</p>
+              </li>
+            </ol>
+
+            <p>Upon fulfillment of <var>p</var> with value <var>report</var>,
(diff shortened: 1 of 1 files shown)


alvestrand:
Do we need an entry in the backwards compatibility section for this?

jan-ivar:
Do you mean [4.3.3 Legacy Interface Extensions](
http://htmlpreview.github.io/?https://raw.githubusercontent.com/jan-ivar/webrtc-pc/getstats/webrtc.html#legacy-interface-extensions)? If so it's in the PR.

webrtc.html
```diff
@@ -3849,29 +3888,21 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+            <code>getStats()</code></dfn> method is
+          invoked, the user agent MUST run the following steps:</p>
 
           <ol>
             <li>
-              <p>If the <code><a>RTCPeerConnection</a></code> object's <a href=
-              "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
-              signalingState</a> is <code>closed</code>, throw an
-              <code>InvalidStateError</code> exception.</p>
-            </li>
```
burnburn: I can't see where you have moved the signalingState check to.

burnburn:
I've reviewed this and, in general, it looks good.  Two comments:  1) I commented above on the signalingState check being dropped, and 2) I'm not sure we need a legacy version of this since the signatures (to me) don't appear to collide.  If so, then browser polyfill can easily support the original syntax.  Then again, I may have missed a reason why they can't.  Thoughts?

webrtc.html
burnburn: Or did you just drop it because we plan to keep some form of stats info alive even for closed channels?

alvestrand:
Re backwards compatibility: Removing an old API breaks all code that doesn't use a polyfill. So it's something browsers do after carefully counting the number of users.
I don't think the spec is harmed much by having the old method around too (as you say, the signatures don't collide).
FWIW - poyfills DO have a problem with overloads, at least as I've seen them - they have to "cover" both the old and the new signature, and do JS-level interpretation to figure out which signature they're supposed to trigger on.
And yes, it's been very useful in practice to be able to access stats after closing, so the state check SHOULD go - it's not appropriate here.

burnburn:
Sorry, when I said browser polyfill above, I meant browser support, not app authors having to load a polyfill -- of course that breaks running code!

Next time it would be better to separately remove the state check (even as a separate commit within the same PR) rather than just combining it all together.

But I'm okay with this one now.

webrtc.html
jan-ivar: Yes, see https://github.com/w3c/webrtc-stats/issues/3

PR #236, opened 2015-06-12
Replace operations array with a chain, preserving existing behavior.

jan-ivar:
Replacing #222, this PR preserves the existing desired behavior of `pc.addTrack(X); pc.createOffer(); pc.addTrack(Y);` not including `Y` in the offer, except in this corner-case:

    pc.setRemoteDescription(desc); // not checking for failure!
    pc.addTrack(X); pc.createOffer(); pc.addTrack(Y);

where `Y` *is* part of the offer, by nature of the createOffer being delayed. This is considered a natural outcome of a deprecated pattern. Use promises.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -465,8 +465,9 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to represent a queue of
-            <code>operations</code> with an empty set.</p>
+            <p>On the <var>connection</var>, initialize an internal slot
+            <var>operations</var> to a resolved promise, to represent the start
+            of a promise chain.</p>
           </li>
 
           <li>
@@ -493,23 +494,42 @@
 
         <ol>
           <li>
-            <p>Append an object representing the current call being handled
-            (i.e. function name and corresponding arguments) to the
-            <code>operations</code> array.</p>
+            <p>Let <var>operation</var> be a function executing the steps of the
+            call in question.</p>
+          </li>
+          <li>
+            <p>If <var>connection</var>'s <a href=
+              "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+              signalingState</a> is <code>closed</code>, throw an
+              <code>InvalidStateError</code> exception and abort these steps.</p>
+          </li>
+          <li>
+            <p>Let <var>p</var> be the result of transforming <var>connection</var>'s
+            <var>operations</var> by a fulfillment handler that runs the
+            following steps:</p>
+            <ol>
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>closed</code>, return
+                <var>undefined</var> and abort these steps.</p>
+              </li>
+              <li>
+                <p>Return the result of executing <var>operation</var>.</p>
+              </li>
+            </ol>
           </li>
 
           <li>
-            <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            <p>To point to the new tail of the chain, and to prevent having
+            errors from one operation cancel subsequent operations, set
+            <var>connection</var>'s <var>operations</var> to the result
+            of transforming <var>p</var> by a rejection handler that returns
+            <var>undefined</var> (this has no effect on <var>p</var>).</p>
           </li>
 
           <li>
-            <p>When the asynchronous operation completes (either successfully
-            or with an error), remove the corresponding object from the
-            <code>operations</code> array. After removal, if the array is
-            non-empty, execute the first object queued asynchronously and
-            repeat this step on completion.</p>
+            <p>Return <var>p</var>.</p>
           </li>
         </ol>
 
@@ -517,9 +537,8 @@
         <code>setLocalDescription</code>, <code>createAnswer</code> and
         <code>setRemoteDescription</code> executing at any given time. If
         subsequent calls are made while one of them is still executing, they
-        are added to a queue and processed when the previous operation is fully
-        completed. It is valid, and expected, for normal error handling
-        procedures to be applied.</p>
+        are added to a chain and processed upon fulfillment or rejection of the
+        previous operation's promise.</p>
 
         <p>Additionally, during the lifetime of the RTCPeerConnection object,
         the following procedures are followed when an ICE event occurs:</p>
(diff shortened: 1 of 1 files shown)


jan-ivar:
For reference, here is a sample implementation, from https://bugzil.la/1169386

    _runOperation: function(func) {
      this._throwIfClosed(); // out here DOMException line-numbers work.
      let p;
      if (!this._operationsLength) {
        p = func();
      } else {
        p = this._operations.then(() => {
          // Don't _throwIfClosed() inside the chain, because spec behavior as of
          // this writing is to NOT reject outstanding promises on close. This is
          // what happens most of the time anyways, as the in-parallel steps stop
          // calling us once closed, hanging the chain. However, a task may already
          // have been queued on us, so if we're one of those then sit back.
          if (!this._closed) {
            return func();
          }
        });
      }
      this._operationsLength++;
      // don't propagate errors in the operations chain (this is a fork of p).
      // Note: Decrement must run before other p.then's, so don't extend fork!!
      this._operations = p.then(() => this._operationsLength--,
                                () => this._operationsLength--);
      return p;
    },

stefhak:
Apart from the comment I made, this LGTM.

webrtc.html
```diff
@@ -517,9 +545,8 @@
         <code>setLocalDescription</code>, <code>createAnswer</code> and
         <code>setRemoteDescription</code> executing at any given time. If
```
stefhak: addIceCandidate should be added here

adam-be:
Some questions/comments:

This patch uses a terminology that is quite different from the rest of the spec. Some additions to the terminology section might be helpful.

This algorithm checks for the closed state. Does that mean that we can remove the same check from the queued operations individual algorithms?

With this approach, is the intention that type conversions (that, e.g., throw TypeError on bad arguments) should be done when an operation is taken of the queue or directly when called?

jan-ivar:
The terminology follows the [TAG promise-guide](http://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting).

Nice catch about the closed state. I worked backwards from an implementation where it made sense to me to centralize this check rather than duplicate it in the individual methods, but I see that it perhaps belongs in the individual methods. I should probably move the initial closed check to avoid double-checking when we call the method directly in that case. 

I'm happy either way, as long as we're consistent. Some methods in the spec, like `createOffer` and `createAnswer` do not (yet?) seem to have processing models with steps that check for closed state, so we should remember to fix those if we're not centralizing the checks.

Let me know what you prefer, and I can update the PR.

On implementations that use WebIDL compilers, TypeErrors from binding code happen immediately when the method is called from content JS, so I think that is the right answer.

I don't see any TypeErrors or validation errors in prose of our individual methods, though that would have been an interesting question.

adam-be:
Sorry taking so long to respond to this one.

Regarding the closed state check, I don't have a strong opinion. Perhaps it's good to have the individual operations self-contained with their own checks as you suggest.

It would be great to have some input from some implementors in the Chrome team (ping @pthatcherg @juberti) about this approach to specify the behaviour.

pthatcherg:
To my eyes, the PR text is longer and harder to read and understand than the existing text, which is shorter and easier to understand.

What deficiency is there in the current text that we're trying to overcome? 

Defining a queue of operations as a chain of promises looks awkward and unnecessary.  And if I were to read a spec defined as a chain of promises, I would probably eventually say to myself, "oh, this is a complicated way of defining a queue of operations;  I'll just implement that".   I'd prefer to just stick with defining a queue of operations as a .... queue of operations.

adam-be:
I tend to agree with Peter here. If this is the behaviour we want I think it's great if it can be implemented with a Promise chain, but I don't think we necessarily need to specify it that way since it becomes a bit complex to read.

ekr:
I agree with Peter and Adam.

Also, I wonder if this is actually equivalent because of the subtle distinctions in scheduling between events and promised. Note: I have not tried to evaluate this.

alvestrand:
There seems to be no consensus to merge this. Closing.

jan-ivar:
Sorry for taking so long to check in on this (my mistake for thinking it a done deal). This PR solves several problems with the original text, primarily that it hand-waves, uses ill-defined terms, doesn't do any error handling, and translates to incorrect code. I would like to see it reopened to discuss and fix this.

> Append an object representing the current call being handled (i.e. function name and corresponding arguments

`operations.push({ foo: "createOffer", bar: [options] });` ? I think at least the implementation function is needed here, not just its name.

> execute the function from the front of the queue asynchronously

There is no function at the front of the queue, rather an object which isn't executable. But these are just nits.

More importantly, what does "execute the function asynchronously" mean? I don't see this defined anywhere. Surely, since the methods in question contain processing steps that, among other things, create and return a promise `p`, these processing steps have to be executed *synchronously* for this to make any sense.

There is no mention of return values whatsoever in these steps, so how are promises to make it back to the content script?

> When the asynchronous operation completes (either successfully or with an error), remove the corresponding object from the operations array.

This is step 3 in a sequence of steps, yet the impact of "when" is unclear. Does the sequence block until step 3 has happened? As written, it sounds like a call to `setLocalDescription` will not return until its "asynchronous operation" has succeeded or failed, which would be wrong. `setLocalDescription` should return immediately with the promise from its processing steps.

It is also not well defined what constitutes "completion" of an "asynchronous step". There is also no mention of how errors are to be handled, since the above is the only mention of "error".

> After removal, if the array is non-empty, execute the first object queued asynchronously and repeat this step on completion.

Nit: objects cannot be executed, and again these functions should be executed synchronously. But more importantly, this algorithm offers no separation of calls, and it's unclear how return values (promises) from subsequently queued steps are to make their way back to the calling code. This might have worked when we still had methods that took callbacks and returned nothing, but it is totally broken now.

All this is fixed by this PR.

There was a complaint about length and complexity. As noted in comments, this PR also adds checking of the `closed` state which the old algorithm did not have. This could be removed for parity as mentioned. I think you'll find that with the necessary fixes for all the problems with the old text, this PR will be far shorter and sweeter, never-mind that it relies on standard established patterns rather than reinventing the wheel.

alvestrand:
Jan-Ivar, you have 3 people who have read your text and think it makes the spec worse. So far, I haven't seen a single person agreeing with you that it makes the text better.

If you want to continue this approach, I suggest you approach a single person and discuss the problem you are trying to solve, and see if you can come up with something that looks compelling.
Just repeating your arguments doesn't seem to be working.

jan-ivar:
@alvestrand I think everyone who commented initially did so on the assumption that the existing naive prose was fine, and that the merits to be judged were solely ones of readability. I have seen no objections since I pointed that out that the existing text is wrong. 

I've resubmitted this PR for consideration as a solution to https://github.com/w3c/webrtc-pc/issues/299 as https://github.com/w3c/webrtc-pc/pull/307, so lets move the discussion there.

PR #237, opened 2015-06-15
RTCRtpSender.replaceTrack second cut

jan-ivar:
@stefhak - As you requested, a rebased attempt at capturing or at least getting closer to list consensus, to further discussion. Please comment.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2557,21 +2557,21 @@
               </li>
 
               <li>
-                <p>Run the following steps asynchronously:</p>
+                <p>If <var>withTrack's</var> <code><a href=
+                  "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
+                attribute differs from the </var><code><a href=
+                  "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
+                of
+                <code><a href="#dom-rtpsender-track">RTCRtpSender.track</a></code>,
+                then reject
+                <var>p</var> with <code>IncompatibleMediaStreamTrackError</code>,
+                return <var>p</var> and abort these steps.</p>
+              </li>
+
+              <li>
+                <p>Run the following steps in parallel:</p>
               </li>
               <ol>
-                <li>
-                  <p>If <var>withTrack's</var> <code><a href=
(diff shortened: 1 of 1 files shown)


jan-ivar:
Replaces #195

webrtc.html
```diff
@@ -2537,6 +2537,91 @@
...
+                    treat them as new tracks. The id of the remote track MUST
+                    NOT change in this case.</p>
+                  </li>
+                  <li>
+                    <p>If negotiation is needed, then fire a <a href=
+                      "#event-negotiation">negotiationneeded</a> event at the
+                    <a>RTCPeerConnection</a> associated with this sender.
+                    In this case, <var>p</var> MUST be resolved with
+                    <var>undefined</var> the next time renegotiation is
+                    successfully completed.</p>
```
stefhak: Should there not be text saying that if renegotiation fails, the promise should be rejected? And then the old track should continue flowing (and be returned from sender.track)?
Would we need something similar to the pending SDPs here, a "pending track" that will (or will not) replace the track when negotiation happens?

webrtc.html
```diff
@@ -2537,6 +2537,91 @@
...
+              </li>
+
+              <li>
+                <p>Run the following steps in parallel:</p>
+              </li>
+              <ol>
+                <li>
+                  Set the
+                  <code><a href="#dom-rtpsender-track">RTCRtpSender.track</a></code>
+                  attribute to <var>withTrack</var>.
```
stefhak: Should that not happen once the promise is resolved? Especially if renegotiation is needed, there may be some time before the withTrack is used, and renegotiation may fail.

martinthomson:
From Seattle: we are going to fail if the track can't be seamlessly replaced.

martinthomson:
From Seattle: With the decision to consider Peter's `addMediaSection` proposal, we will have nullable track values on `RTCRtpSender` (and `RTCRtpReceiver`).  I don't think that this needs to change to accommodate that now, but I wanted to record the fact that the argument might need to become nullable too.  This would allow us to model `removeTrack` in terms of `replaceTrack(null)`.

jan-ivar:
Replaced by https://github.com/w3c/webrtc-pc/pull/303.

PR #241, opened 2015-06-19
Add RtpSender.transport, RtpReceiver.transport, RTCDtlsTransport, RTCIceTransport, etc

pthatcherg:


alvestrand:
Please fix the errors that phantomjs reports from Travis. The html5valdiator errors are a known bug in the HTML5 validator, and Dom has proposed a fix.

webrtc.html
```diff
@@ -2625,6 +2687,194 @@
...
+            and receive packets.  The underlying transport may not be
+            shared between multiple <code>RTCDtlsTransport</code>
+            objects.
+          </dd>
+
+          <dt>readonly attribute RTCDtlsTransportState state</dt>
+
+          <dd>
+            <p>The <dfn id=
+            "dom-dtlstransport-state"><code>state</code></dfn>
```
dontcallmedom: I suggest using &lt;code id="dom-dtlstransport-state">state&lt;/code> (no &lt;dfn>) to avoid the dup id error of ReSpec.

pthatcherg: OK, I have done that for all of the places where ReSpec was complaining.  Actually, it found several real mistakes, which I fixed.

elagerway:
I will circle back with @pthatcherg to see where we are on this one.

elagerway:
Peter is off on mini-vacation until Monday, he said he will endeavour to look at it when he is back at his desk.

pthatcherg:
I have fixed all the errors and merged with master.

burnburn:
Hopefully Peter, Harald, and I can look at this together in Prague to see if it's ready to go in.

elagerway:
Ok, with that in mind I am assigning this to you Dan.

webrtc.html
```diff
@@ -2776,6 +2776,35 @@
...
 
+          <dt>readonly attribute RTCDtlsTransport transport</dt>
+
+          <dd>
+            <p>The <code id="dom-rtpsender-transport">RTCRtpSender.transport</code>
+            attribute is the transport over which media from
+            <code>RTCRtpSender.track</code> is sent in the form of RTP
+            packets.  When BUNDLE is used,
+            many <code>RTCRtpSender</code> objects will share
+            one <code>RTCRtpSender.track</code> and all will all send
```
alvestrand: I think you mean RTCRtpSender.transport here.

pthatcherg: Fixed

elagerway:
There is ongoing discussion on the public-webrtc mail list regarding this PR: https://lists.w3.org/Archives/Public/public-webrtc/2015Jul/0080.html

@pthatcherg any comment from you on where we are on this today?

webrtc.html
```diff
@@ -2776,6 +2776,35 @@
...
 
+          <dt>readonly attribute RTCDtlsTransport transport</dt>
+
+          <dd>
+            <p>The <code id="dom-rtpsender-transport">RTCRtpSender.transport</code>
+            attribute is the transport over which media from
+            <code>RTCRtpSender.track</code> is sent in the form of RTP
+            packets.  When BUNDLE is used,
+            many <code>RTCRtpSender</code> objects will share
+            one <code>RTCRtpSender.transport</code> and all will all send
```
adam-be: One "all" at much I think

webrtc.html
```diff
@@ -2840,6 +2898,137 @@
...
+          <dd>
+            This event handler, of event handler event type <code><a href=
+            "#event-dtlstransportstatechange">dtlstransportstatechange</a></code>,
+            MUST be fired any
+            time the <a href=
+            "#dom-dtlstransport-state"><code>RTCDtlsTransport</code> state</a> changes.
+          </dd>
+        </dl>
+
+        <h3>RTDtlsConnectionState Enum</h3>
```
adam-be: There's a mismatch between the header and the name of the enum. I guess it should be:
s/RTDtlsConnectionState/RTCDtlsTransportState/

webrtc.html
```diff
@@ -2840,6 +2898,137 @@
...
+          <dd>
+            <p>The <dfn id=
+            "dom-dtlstransport-transport"><code>RTCDtlsTransport.transport</code></dfn>
+            attribute is the underlying transport that is used to send
+            and receive packets.  The underlying transport may not be
+            shared between multiple <code>RTCDtlsTransport</code>
+            objects.
+          </dd>
+
+          <dt>readonly attribute RTCDtlsTransportState state</dt>
```
adam-be: Should this one be called readyState to align with other APIs (WebSocket, EventSource, XHR, DataChannel, ...)?

webrtc.html
```diff
@@ -2840,6 +2898,137 @@
...
+          <dd>
+            <p>An array containing the remote certificates in use by
+            the remote side.</p>
+          </dd>
+
+          <dt>attribute EventHandler onstatechange</dt>
+
+          <dd>
+            This event handler, of event handler event type <code><a href=
+            "#event-dtlstransportstatechange">dtlstransportstatechange</a></code>,
```
adam-be: It's common to align the EventHandler attribute name (onstatechange) with the Event type (dtlstransportstatechange). See comment above about the state attribute.

adam-be:
Didn't we have a shared base class for RTCRtpSender and Receiver at some point? Feels like a lot of stuff are repeated here.

Future work:
* Add the RTCDtlsTransport state event to the Event Summary section

webrtc.html
pthatcherg: Fixed

webrtc.html
pthatcherg: Fixed.

webrtc.html
pthatcherg: Sure.  Changed to "readyState".

pthatcherg: Actually, I changed it back to be consistent with the IceTransport.state, which should be consistent with the PeerConnection.iceconnectionstate (not icereadystate).

webrtc.html
pthatcherg: So do you want to change the attribute name to DtlsTransport.ondtlstransportstatechange or the Event type to statechange?  Neither seems good to me.

pthatcherg:
I don't think we ever had a common base for RtpSender and RtpReceiver.  I don't think it's worth it to have one (since it's nice to have comments that say things like "the transport on which media is sent" and not "the transport on which media is sent or received".  But even if we decided to do that, I'd prefer to merge the PR and address that separately.

pthatcherg:
Do you want me to add a TODO for the future work (Event summary section)?

webrtc.html
adam-be: I don't find the options here ideal either. That's why I'm leaning more towards aligning with other APIs and go with readyState (state attribute), readystatechange (event type) and onreadystatechange (event handler attribute). Those names aren't super great either, but developers are familiar with them and will know what to do.

adam-be:
The comment about a common base class was more of a side-note.

Perhaps it's enough to create an issue for updating the Event Summary section. It's a quick fix when we get the event naming sorted out. :)

webrtc.html
pthatcherg: I sent an email to the list asking what name the WG would prefer (state, readystate, orconnectionstate).  I don't feel too strongly about it, but I thought it would be better for it to be consensus before changing it in the PR.

webrtc.html
```diff
@@ -2807,6 +2807,35 @@
...
           </dd>
 
+          <dt>readonly attribute RTCDtlsTransport transport</dt>
+
+          <dd>
+            <p>The <code id="dom-rtpsender-transport">RTCRtpSender.transport</code>
+            attribute is the transport over which media from
+            <code>RTCRtpSender.track</code> is sent in the form of RTP
+            packets.  When BUNDLE is used,
+            many <code>RTCRtpSender</code> objects will share
```
adam-be: Please use links when referring to types like <code>RTCRtpSender</code> to make them clickable. So it should be <code><a>RTCRtpSender</a></code>. This isn't crucial for the PR, but every time we say that it's easy to fix in an follow-up PR, we never do it. :)

pthatcherg: Types are now clickable.

adam-be:
Thanks for fixing the links and adding the Event Summary section as well. The last commit broke Travis, but at least one error was some strange Respec.js bug (it doesn't seem possible to have the same text between two dfn-tags even though their ids are different). I've made a PR against your PR branch that fixes those.

Otherwise this LGTM.

pthatcherg:
Thanks!  I merged your PR of my PR into my PR.

PRs of PRs.  We have to go deeper.

PR #242, opened 2015-06-22
Remove SyntaxError on malformed ICE candidate

dontcallmedom:
* it is not defined what a "malformed" ICE candidate is
* no browser does this today
* from a flow perspective, an error will be triggered in the next step (can't be applied)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1107,12 +1107,6 @@
               </li>
 
               <li>
-                <p>If the candidate parameter is malformed, reject <var>p</var>
-                with <code>SyntaxError</code> and jump to the step labeled
-                <em>Return</em>.</p>
-              </li>
-
-              <li>
                 <p>If the candidate could not be successfully applied, reject
                 <var>p</var> with a <code>DOMError</code> object whose
                 <code>name</code> attribute has the value TBD


stefhak:
@dontcallmedom this should go to the list, can you do that?

jan-ivar:
[It's](http://w3c.github.io/webrtc-pc/#dom-peerconnection-addicecandidate) the candidate *parameter* that is checked for malformity.  That's a dictionary, which integrity is somewhat guaranteed by WebIDL. Any chance this language predates that and can be removed without discussion?

jan-ivar:
This prose also hops between *candidate parameter* and *candidate* without defining either very well. Just an observation.

dontcallmedom:
[List discussion](https://lists.w3.org/Archives/Public/public-webrtc/2015Jul/thread.html#msg7)

stefhak:
PR #302 would make this OBE.

dontcallmedom:
@stefhak I don't think it would — as far as I can tell, #302 doesn't remove the syntax error on addIceCandidate()

stefhak:
@dontcallmedom I agree. I think your PR makes sense, if for no other reason than that it conforms to setLocal/Remote (there is no "malformed" check there).

PR #248, opened 2015-06-30
Fix error handling for certificate management

martinthomson:
The certificate management description inadvertently requires the use of InvalidAccessError exclusively.  That's not good.

Otherwise, if the algorithm isn't supported, we should report NotSupportedError.


ekr:
LGTM

PR #256, opened 2015-07-23
Fix prose on getStats() wo/selector + move type check to sync section

jan-ivar:
Fix WebIDL and prose to handle the case where no selector is passed correctly, and move type check to the synchronous section, like we do over in https://github.com/w3c/mediacapture-main/pull/182.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4179,18 +4179,17 @@ sender.setParameters(params)
 
           <ol>
             <li>
-              <p>Let <var>p</var> be a new promise.</p>
-            </li>
-
-            <li>
               <p>Let <var>selectorArg</var> be the methods first argument.</p>
             </li>
 
             <li>
               <p>If <var>selectorArg</var> is neither <code>null</code> nor a
               valid <a href="#stats-selector">selector</a>,
-              reject <var>p</var> with a
-              new <code>DOMError</code> with name <code>TypeError</code>.</p>
+              return a promise rejected with a <code>TypeError</code>.</p>
+            </li>
+
+            <li>
+              <p>Let <var>p</var> be a new promise.</p>
             </li>
 
             <li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -4166,7 +4166,7 @@ sender.setParameters(params)
 
       <dl class="idl" title="partial interface RTCPeerConnection">
         <dt>Promise&lt;RTCStatsReport&gt; getStats(
-          optional MediaStreamTrack? selector)</dt>
+          optional MediaStreamTrack? selector = null)</dt>
 
         <dd>
           <p>Gathers stats for the given <a href="#stats-selector">selector</a>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4166,7 +4166,7 @@ <h2 id="sec.stats-model">Statistics Model</h2>
 
       <dl class="idl" title="partial interface RTCPeerConnection">
         <dt>Promise&lt;RTCStatsReport&gt; getStats(
-          optional MediaStreamTrack? selector)</dt>
+          optional MediaStreamTrack? selector = null)</dt>
```
alvestrand: I'm not sure what the correct syntax is here, but it seems to me that "optional MediaStreamTrack? selector = null" makes the selector optional in 3 different ways. If it has a default value, it doesn't need to be optional, does it?
I suppose either

optional MediaStreamTrack selector

or 

MediaStreamTrack? selector = null

would have the same effect, but stacking them seems strange.

jan-ivar: > MediaStreamTrack? selector = null

`WebIDL.WebIDLError: error: Mandatory arguments can't have a default value.`

Unlike in C++, a WebIDL default does not implicitly make a parameter optional apparently.

Having an explicit `null` overload for "select everything" makes some sense I suppose (if one buys into that `null` vs. `undefined` is useful for anything at all in the JS language).

martinthomson: In this case then it is either optional, nullable and a default of null; or it is merely optional. The difference being that the former permits an explicit null value whereas the latter would require omission to convey the same meaning. I prefer the option that permits only one way. Just make it optional.  ...or use overloads so that you can highlight the fact that these are very very different APIs depending on what is passed in.

jan-ivar: My only hesitation is that it seems to hinge on there being no more arguments. E.g. `getStats(undefined, onSuccess, onFailure);` just seems wrong for some reason.

martinthomson: Oh yeah, the legacy junk.  I wouldn't bother fixing that one.  Leave it nullable.

You could "fix" it by adding overloads, but that's just going to cause more work for us to dress up an API point we ultimately want to remove.

jan-ivar: Sure, but I'm not making my point well: What if the new getStats did take more arguments? Would we go with `null` or `undefined`? I think `null`.

Furthermore, I think `null` makes sense semantically here. E.g. in this case:

    var selector = findSelectorPreference();
    return getStats(selector);

Here `selector` can be `null`, which explicitly means "select all", which is different from `undefined` which might mean "no preference". Lets not force people to do this:

    return getStats((selector === null? undefined : selector);

martinthomson: The reason I suggested that this be an overload is that I don't think people will have a convergent code path that covers both "I want everything" and "I want this specific thing".  The data that is returned and the use cases for it are quite different.  In that case you don't have a weird selector thing, you just have:

```js
function getAllStats(pc) { return pc.getStats(); }
function getTrackStats(pc, track) { return pc.getStats(track); }
```

BTW, just checked what we (gecko) do with extra arguments to functions when there are overloads or optional arguments.  Using a constructor that optionally initializes a value that defaults to `null`:

```idl
[Constructor(optional DOMString x)]
interface Foo { readonly attribute DOMString? x; };
```

Then you get this:
```
>>> new Foo()
Foo { x: null }
>>> new Foo('123')
Foo { x: "123" }
>>> new Foo(null)
Foo { x: null }
>>> new Foo(123)
TypeError: Argument 1 of Foo.constructor can't be converted to a string.
```

So there is no need to play games with types, pass the `null` if that is what you have.

jan-ivar: But you're testing `DOMString`. When I remove `nullable` from `getStats` in WebIDL like this:

```webidl
Promise<RTCStatsReport> getStats (optional MediaStreamTrack selector)
```

and call `pc.getStats(null)` then I get:

    TypeError: Argument 1 of mozRTCPeerConnection.getStats is not an object., line 56

adam-be:
This change looks good to me.

adam-be:
We probably need to fix one thing. When the promise is rejected we don't jump to the end or prevent the async section to be run. I would actually return a rejected promise directly when something goes wrong and then define p just before the async section (like they do in the promise writing guide).

jan-ivar:
I like that.

adam-be:
:+1:

PR #258, opened 2015-07-24
Add RTCRtpParameters.codec, RTCRtpCodecParameters, and RTCEncodingPar…

pthatcherg:
List discussion needed.


webrtc.html
```diff
@@ -2530,23 +2539,55 @@
...
+            <p>The RTP payload type.  This value can be set in the
+            RTCRtpParameters.encodings[0].payloadType to control which
+            codec should be used to send a given encoding.
+            </p>
+          </dd>
+
+          <dt>DOMString name</dt>
+
+          <dd>
+            <p>The name of the codec, such as "opus", "vp8", and "h264".</p>
```
pthatcherg: FYI, we can change this to match the results from the discussion on getCapabilities around mime type vs name.

martinthomson: Codecs don't have names, they have media types.

webrtc.html
```diff
@@ -2530,23 +2539,55 @@
...
           <dd>
             <p>
-              Indicates the relative priority of this encoding, across
-              all RtpSenders of a given PeerConnection.  When there is
+              The relative priority of this encoding across all
+              RtpSenders of a given PeerConnection.  When there is
               limited bandwidth available to a PeerConnection, higher
               prioirty encodings will be sent with more bandwidth, and
               lower priority encodings will be sent with less
               bandwidth.
```
martinthomson: This isn't enough information to implement consistent behaviour.

pthatcherg: This is being updated in https://github.com/w3c/webrtc-pc/pull/259.  You may like that version better.

And that really has nothing to do with this PR.  It looks like I accidentally changed some grammar/formatting.

webrtc.html
pthatcherg: Like I said, I'll just change this to match what is decided in getCapabilities.

pthatcherg: OK, it's been changed to mimeType to match https://github.com/w3c/webrtc-pc/pull/269.

At Harald's request, I also added clockRate and sdpFtmpLine.

martinthomson:
This is an odd control surface.  If you take the view that these are domain objects, then you could represent them with interfaces.  I think that at this point, it would make a lot of sense to take a step back and try to model this out with interfaces.  Then you could use actual objects as the basis of the selection, rather than this odd payload type indirection.

alvestrand:
As noted in Seattle:
The PR needs to include info about what happens at renegotiation - whether payload type goes back to "browser selects", PT gets kept no matter if it works or not, or whether it's "browser, do something sensible".

alvestrand:
From Seattle:
Another approach in this PR is to make the "codecs" sequence updateable - allow reorder and deletion, but not adding new ones or change existing ones.

pthatcherg:
Here is that approach: https://github.com/w3c/webrtc-pc/pull/298/f

On Wed, Sep 9, 2015 at 10:45 AM, Harald Alvestrand <notifications@github.com

alvestrand:
Closing per Seattle discussions, obsoleted by #298.

Issue #21, opened 2014-10-31
Use of MUST and must is completely inconsistent throughout


martinthomson:
Sometimes it's capitalized, other times not.

PR #262, opened 2015-07-31
fix for #21

fluffy:


stefhak:
LGTM.

stefhak:
@martinthomson can you review this PR? (Should be quickly done)

martinthomson:
LGTM

webrtc.html
```diff
@@ -2095,14 +2095,14 @@
...
-            <p>An RTCSdpType of "offer" indicates that a description should be
+            <p>An RTCSdpType of "offer" indicates that a description MUST be
             treated as an [[!SDP]] offer.</p>
           </dd>
 
           <dt>pranswer</dt>
 
           <dd>
-            <p>An RTCSdpType of "pranswer" indicates that a description should
+            <p>An RTCSdpType of "pranswer" indicates that a description MUST
```
alvestrand: Not happy with "MUST" here; what about "is intended to be treated as"?

stpeter: That does seem better in this instance.

PR #264, opened 2015-08-11
Adapt to new ReSpec rules on title in dfn tags

dontcallmedom:
removes warnings generated with newest version of respec (which makes Travis CI fails)

cf https://lists.w3.org/Archives/Public/spec-prod/2015JulSep/0020.html and https://lists.w3.org/Archives/Public/spec-prod/2015JulSep/0022.html


dontcallmedom:
sorry, just saw this was fixed separately

PR #266, opened 2015-08-18
Tidy up setLocal/RemoteDescription processing model

adam-be:


alvestrand:
@fluffy can you check that this looks right to you?
There should be few/no functional changes, but text should be clearer.

fluffy:
Added to my list ... this stuff has some subtle weirdness so it is going to take me some time to review this.

fluffy:
LGTM - and if anyone else wants to review, I have a diff that removes the moves from the changes. I actually wish we had done this as two PR, one that did the moves around in the document and one built on top of that which made the changes. But nice work from Adam. Thanks

adam-be:
@fluffy, have you looked at the separate commits in the PR? It should be possible to inspect the move and the functionality update separately since these have been done in separate commits. It wouldn't be a problem to split this into two PRs either.

stefhak:
@pthatcherg @martinthomson  any comments on this?

webrtc.html
```diff
@@ -625,6 +625,313 @@
           </li>
         </ol>
 
+        <p>To <dfn id=
+        "update-ice-gathering-state">update the ice gathering state</dfn> of an
```
martinthomson: You can use `data-lt` attribute for `dfn` elements and then you don't have to use `href` attributes as long as the text matches.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+        <var>newState</var>, the User Agent MUST queue a task that sets
+        <var>connection</var>'s <a href=
+        "#dom-peerconnection-ice-gathering-state">iceGatheringState</a>
+        attribute to <var>newState</var> and fires a simple event named
+        <code><a href=
+        "#event-icegatheringstatechange">icegatheringstatechange</a></code> at
+        <var>connection</var>.</p>
+
+        <p>To <dfn id=
+        "update-ice-connection-state">update the ice connection state</dfn> of an
```
martinthomson: We're removing ice connection state, but this is good.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+              <li>
+                <p>If <var>connection</var>'s
+                <a href="#dom-peerconnection-signaling-state">signaling
+                state</a> is <code>closed</code>, the user agent MUST return a
+                promise rejected with an <code>InvalidStateError</code>.</p>
+              </li>
+
+              <li>
+                <p>If <var>description</var> is set as a local description and
+                contains a different set of ICE credentials, then the ICE Agent
```
martinthomson: ICE credentials needs a definition, or you could explicitly note that it is a different ufrag on ANY media section.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+                MUST trigger an ICE restart, and run the following steps:</p>
```
martinthomson: Why is the ufrag check and conditional ICE restart not just a step in the algorithm?

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+
+                <ol>
+                  <li>
+                    <p>If <var>connection</var> is not already gathering,
+                    <a href="#update-ice-gathering-state">update
+                    <var>connection</var>'s ice gathering state</a> to
+                    <code>gathering</code>.</p>
```
martinthomson: This doesn't match with the changes to add `RTCIceTransport`, since the gathering state will be deprecated (?) or turned into a composite state.  I think that Peter's rules for composite states need to be followed.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+                  </li>
+
+                  <li>
+                    <p>If <var>connection</var>'s <a href=
+                    "#dom-peerconnection-ice-connection-state">ice connection state</a>
+                    is <code>completed</code>, <a href=
+                    "#update-ice-connection-state">update
+                    <var>connection</var>'s ice connection state</a> to
+                    <code>connected</code>.</p>
```
martinthomson: As with gathering state.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+                  </li>
+                </ol>
+              </li>
+
+              <li>
+                <p>The user agent MUST start the process to apply
+                <var>description</var>.</p>
```
martinthomson: Needs a reference to JSEP here.

martinthomson: Note that when you run in parallel, you have two things to do:
1. await the completion of any other tasks that have been enqueued.
2. check that the pc is not closed
3. then apply the description

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+              <li>
+                <p>If the content of <var>description</var> is invalid or if
+                <var>description</var>'s <code><a href=
+                "#widl-RTCSessionDescription-type">type</a></code> is wrong for
+                the current <a href=
+                "#dom-peerconnection-signaling-state">signaling state</a> of
+                <var>connection</var>, then reject <var>p</var> with an
+                <code>InvalidSessionDescriptionError</code> and abort these
+                steps.</p>
+              </li>
```
martinthomson: Move this to the algorithm section.  This structure is confusing.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+              <li>
+                <p>Let <var>p</var> be a new promise and return <var>p</var>.</p>
+              </li>
+            </ol>
+          </li>
+
+          <li>
+            <p>If the process to apply <var>description</var> fails for
+            any reason, then user agent must queue a task runs the
+            following steps:</p>
```
martinthomson: Make this part of the algorithm.  It will read better.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+              <li>
+                <p>If <var>description</var> is set as a local description, and
+                its content matches the state of all tracks and data channels,
+                <a href="#clearing-negotiation-needed">as defined below</a>,
+                clear the negotiation-needed flag.
+                </p>
+              </li>
+
+              <li>
+                <p>Set <var>connection</var>'s description attributes
```
martinthomson: What description attributes?  Be precise.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+
+                <p class="issue">NOTE: The principles of pending and current SDP
+                were agreed by the WG but the details the this step has not
+                yet been fully reviewed. TODO - review this.</p>
+
+                <ul>
+                  <!-- A) transition stable to haveLocalOffer -->
+                  <li>
+                    <p>If <var>description</var> is set as a local description
+                    of type "offer", and it has a version that is later than
```
martinthomson: The version isn't relevant here.  Checking the version is a precondition on applying the description, not a post-condition.

martinthomson:
Feedback provided.

webrtc.html
jan-ivar: It's hard to do anything after one returns. FWIW I like the pattern we follow in media-capture: 1) `let p = ...` 2) `Run steps in parallel: (indent all steps here one level in)`, 3) `return p;` - But I know there are some alternative views on this.

> await the completion of any other tasks that have been enqueued.

If you mean the operations chain here, then I feel strongly that we can avoid having its workings spill over into the individual methods it is operating on. That is, once https://github.com/w3c/webrtc-pc/issues/299 is addressed.

webrtc.html
```diff
@@ -625,6 +625,313 @@
...
+        <p>To <dfn id=
+        "update-ice-connection-state">update the ice connection state</dfn> of an
+        <code><a>RTCPeerConnection</a></code> instance <var>connection</var> to
+        <var>newState</var>, the User Agent MUST queue a task that sets
+        <var>connection</var>'s <a href=
+        "#dom-peerconnection-ice-connection-state">iceConnectionState</a>
+        attribute to <var>newState</var> and fires a simple event named
+        <code><a href=
+        "#event-iceconnectionstatechange">icegatheringstatechange</a></code> at
+        <var>connection</var>.</p>
```
pthatcherg: Since we're replacing iceConnectionState with "connectionState" which is an aggregate state, I think we need to define these rules for changing state to be in terms of IceTranpsorts.

webrtc.html
adam-be: Yes. This PR is not up to date on that.

I note that #291 adds connectionState, but it doesn't remove iceConnectionState. Are we keeping it in the spec?

webrtc.html
adam-be: Yeah, this needs some work. I was planning to move this part to the "resolve section" because here we're talking about properties of the description before we consider it parsed and verified.

I'll update with your second proposed solution, and then we can create a separate issue if it needs more work.

webrtc.html
adam-be: I'm not sure what you mean by "the algorithm", but I think we're seeing the same problems here. See previous comment.

webrtc.html
adam-be: That's very likely since this PR is over a month old. I'll look at it.

webrtc.html
alvestrand: I interpreted the Seattle discussions as a consensus to add connectionState, but leave iceConnectionState in for now - we already have people with code that depends on it, and the information it shows isn't *quite* what we want for connectionState.

adam-be:
Thanks for the comments. I get the feeling that people are fairly OK with the changes to the state updating and corresponding event firing which constituted the main functional changes introduced by this PR. Then we have the new connectionState, but it's a bit tricky since it's not landed in the spec yet.

I'll do a first update that merges the three old separate algorithms into one which should address a lot of @martinthomson's comments about the overall structure. I'm planning to base it on one of the examples in the Promise Writing Guide.

webrtc.html
adam-be: Good point.

webrtc.html
adam-be: Thanks for the heads up, but I'm a bit hesitant to add stuff about connectionState into this PR before the connectionState PR has landed.

alvestrand:
@martinthomson @pthatcherg can you review again?

adam-be:
Merged latest master. Should be mergeable again.

PR #271, opened 2015-08-22
Add PC.createRtpSender.

pthatcherg:


alvestrand:
This is an alternate approach to the one mentioned in https://github.com/w3c/mediacapture-main/issues/221 - @pthatcherg could you file an issue describing what we need that functionality for?

pthatcherg:
Do you mean file an issue explaining why we want an RtpSender with a
null/fake/empty track?  In other words, explaining that we want to warm up
ICE/DTLS?

alvestrand:
Den 24. aug. 2015 18:50, skrev pthatcherg:
> Do you mean file an issue explaining why we want an RtpSender with a
> null/fake/empty track? In other words, explaining that we want to warm up
> ICE/DTLS?

Yes.

In particular, that you want this at the answerer side without touching
any real device - because you don't want to turn on the camera light or
trigger any permission prompt.

pthatcherg:
Done: https://github.com/w3c/webrtc-pc/issues/274

webrtc.html
```diff
@@ -2504,6 +2504,30 @@
...
+
+          <dd>
+            <p>Equivalent to
+            calling <a href="#dom-peerconnection-addtrack"><code>addTrack(null)</code></a>,
+            but the returned RtpSender is configured to send a track
+            of the given kind.<p>
+
+            <p>The value
+              of <a href="#dom-rtpsender-track"><code>RtpSender.track</code></a>
+              of the returned RtpSender will be null.  When the track
```
martinthomson: `<code>null</code>`

webrtc.html
```diff
@@ -2504,6 +2504,30 @@
...
+              is set, it must have the same kind.</p>
+
+            <p>Session parameters will be negotiated exactly as
+              though <a href="#dom-peerconnection-addtrack"><code>addTrack</code></a>
+              were called with a track of the given kind, except that
+              when track-specific information is needed, suitable
+              values will be provided in their place such that the
+              RtpReceiver on the remote endpoint will be able to
+              receive the media once the track is set.  For examle, if
```
martinthomson: s/examle/example

webrtc.html
```diff
@@ -2504,6 +2504,30 @@
...
+              a track ID is needed, a unique value will be generated
+              in its place.</p>
+          </dd>
```
martinthomson: Rather than examples, I think that we need to be more precise here.  I understand that that is hard, and I think that I'd be OK with deferring that to a follow-up instead of gating this PR on having the extra detail.

For instance, we should describe *precisely* how the track ID is generated, if indeed it is needed.

Codec negotiation is trickier.  I would think that a largely software-based implementation would choose to advertise the software codecs it supports.  But we need to address the issue of sources that have encoders. Personally, I have concluded that generic codecs are all we need to concern ourselves with, since we no longer have any good way of predicting where media will come from. That means that source-specific encoders are properly exceptional and that they won't get good support for these sorts of features.

webrtc.html

webrtc.html
pthatcherg: This is removed now

webrtc.html
pthatcherg: I added two TODOs, one for track ID and one for hardware codecs.

webrtc.html
```diff
@@ -2504,6 +2504,31 @@
             </ol>
           </dd>
 
+          <dt>RTCRtpSender createRtpSender (DOMString kind)</dt>
+
+          <dd>
+            <p>Equivalent to
+            calling <a href="#dom-peerconnection-addtrack"><code>addTrack(null)</code></a>,
```
adam-be: Perhaps we should describe this differently since calling addTrack() with null results in a TypeError.

Should RTCRtpSender have a readonly type-attribute that is initialized when created (either with a MediaStreamTrack or a DOMString as described here)? That way you can read out the type if no track is set, and a new track can be verified against the predefined type.

martinthomson: I think that a type attribute would be a good idea given the possibility that we now make the only way of identifying the type (the track) optional.

pthatcherg: RtpSender.kind is a good idea for an attribute.  I have added it.

pthatcherg: I have updated the text to not be worded in terms of addTrack(null).  See if you like it better.

webrtc.html
```diff
@@ -2504,6 +2504,26 @@
...
+          <dd>
+            <p>Creates an <a><code>RTCRtpSender</code></a> and goes
+            through the same session neogitation
+            as <a href="#dom-peerconnection-addtrack"><code>addTrack</code></a>,
+            except that there is no source of media attached to
+            the <a><code>RTCRtpSender</code></a> (The value
+            of <a href="#dom-rtpsender-track"><code>RTCRtpSender.track</code></a>
+            of the returned RTCRtpSender will be <code>null</code>)
+
+            <p>When track-specific information is needed by session
```
martinthomson: Check alignment of text.

martinthomson:
```
    readonly    attribute MediaStreamTrack track;
```

This needs to be made nullable.

webrtc.html
pthatcherg: Done

pthatcherg:
Done

alvestrand:
Closing per Seattle discussion, obsoleted by #293

PR #272, opened 2015-08-22
Add RtpSenderInit (alternative to PC.createRtpSender)

pthatcherg:


alvestrand:
Per Seattle decision.

PR #283, opened 2015-08-27
Adding unassignedmedia event

martinthomson:
@stefhak, @fluffy, I think that this is a good replacement for #29.

This is mostly a restoration of the previous one, but I removed provision for delaying the event (that hides information).  I also provided a little better justification.  We can talk about justification in more detail at the interim.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1561,6 +1561,53 @@
             the <a href="#dom-peerconnection-ice-gathering-state"><code>RTCPeerConnection</code>
             ice gathering state</a> changes.
           </dd>
+
+          <dt>attribute EventHandler onunassignedmedia</dt>
+
+          <dd>
+            <p>
+              This event handler, of type <code><a>unassignedmedia</a></code>,
+              MUST be fired by objects implementing
+              the <code><a>RTCPeerConnection</a></code> interface.  It is fired
+              when authenticated media packets arrive from the remote peer that
+              are not assigned to a <code><a>MediaStreamTrack</a></code>.
+              Unauthenticated packets are always discarded without generating an
+              event.
+            </p>
+            <p>
+              This event allows an application to learn of new inbound media
+              that was not negotiated correctly.  An application can use this
+              information for diagnostic purposes, or it might use this event to
+              trigger local negotiation.
+            </p>
+            <p>
+              Media packets can be discarded for several reasons.  In some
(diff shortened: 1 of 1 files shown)


stefhak:
Thanks @martinthomson

stefhak:
LGTM

webrtc.html
```diff
@@ -1561,6 +1561,53 @@
...
+
+          <dd>
+            <p>
+              This event handler, of type <code><a>unassignedmedia</a></code>,
+              MUST be fired by objects implementing
+              the <code><a>RTCPeerConnection</a></code> interface.  It is fired
+              when authenticated media packets arrive from the remote peer that
+              are not assigned to a <code><a>MediaStreamTrack</a></code>.
+              Unauthenticated packets are always discarded without generating an
+              event.
```
fluffy: What do you mean by unauthenticated. There are two cases here ... one is a ranodm packet with where SRTP integrity fails, and one where a valid DTLS session is set up and the packet is valid SRTP encrypted with keys set up in DTLS but there is not yet a answer with a fingerprint that for that DTLS session. We don't want to drop those - the whole point is to get notified about them.

fluffy:
one comment but looking reasonable good. I still would like the option of playing media where I did not yet know the fingerprint of the far end and this PR seems to keep avoiding addressing that problem.

webrtc.html
```diff
@@ -2089,6 +2136,52 @@
...
+          <dt>readonly attribute DOMString? mid</dt>
+         <dd>The value of any mid header extension on the discarded media
+          packet, if that header extension is present.</dd>
+        </dl>
+
+        <dl class="idl" title="dictionary RTCUnassignedMediaEventInit : EventInit">
+          <dt>unsigned long ssrc</dt>
+          <dd>See <a><code>RTCUnassignedMediaEvent.ssrc</code></a></dd>
+          <dt>unsigned byte payloadType</dt>
+          <dd>See <a><code>RTCUnassignedMediaEvent.payloadType</code></a></dd>
```
alvestrand: Wasn't this syntax (type.field) what was objected to because it looks like an accessor to a static? @adam-be has been fixing some of these - he may have comments.

alvestrand:
Closing per Seattle discussion, to be replaced by new text (responsible @pthatcherg)

Issue #278, opened 2015-08-26
How does the JS know if ICE is connected but DTLS isn't?


pthatcherg:
The discussion around what happens when DTLS-SRTP fails and how we should signal that up to the JS isn't resolved.  Most people like a generic "PeerConnection.onerror" or "PeerConnection.onconnectionbusted".  I went to write a PR for PeerConnection.onerror, but found it very hard to define it in way that's useful for the JS.  

The more I thought about it, the more I realized that the real fundamental issue is that JS apps really want to know if the PeerConnection is "connected", "connecting", "disconnected, but could reconnect if you kicked it", or "completely busted".  The state we currently have, and which is currently used by JS apps is PeerConnection.iceconnectionstate.  But it's ICE-specific and the information the JS really wants is if the whole PeerConnection is working, including both ICE and DTLS.

This is particularly a problem is something goes wrong with DTLS, even if only temporarily.  For example, if ICE succeeds, but DTLS fails, how does the JS know that?  Or if DTLS takes 10 seconds longer to connect than ICE, how does the JS know that?  Or, in my original mailing list post, what if something goes wrong with DTLS-SRTP, how does the JS know that?  

I thought maybe PeerConnection.onerror could solve this, but as I went to write the PR, I found it didn't really fit as a solution.  

I'd like to see us address this f2f at the interim in September.

alvestrand:
The other use case I've seen argued for "onerror" is stuff that doesn't map neatly to a state, such as when the credentials used towards the TURN server don't validate.
The effect on ICE connection state is difficult to figure out (it depends on whether any of the other candidates succeed), but it's definitely something that's bogus and that the browser can't fix.

ekr:
Peter, I think we should address this. Generally, I think we want to have two orthogonal categories:

- Severity (OK, connecting, busted but recovering, busted, ...)
- Where the problem happened (ICE, DTLS, internal...)

juberti:
I was going to make a specific PR for TURN errors, so that might help inform this discussion.

alvestrand:
Fixed by #291

PR #291, opened 2015-09-05
Add PeerConnection.connectionState.

pthatcherg:
As a possible solution to https://github.com/w3c/webrtc-pc/issues/278


juberti:
My initial sense is that this is probably a mistake. I think we should just make the existing top-level state reflect the overall 'connection state' (e.g. ICE + DTLS, perhaps with a rename), and if you want more data, poll the individual IceTransport/DtlsTransport objects.

pthatcherg:
FYI, I've updated the PR to be ICE+DTLS instead of just DTLS (It's not
pc.connectionState rather than pc.dtlsState).

juberti:
As such, I think this should replace .iceState rather than adding a new state.

pthatcherg:
I think that makes sense.  It shouldn't be hard to polyfill either the old one (.iceConnectionState) or the new one (.connectionState) once DtlsTransport and IceTransport are implemented.

martinthomson:
From Seattle: a "new" state will be added, but this is otherwise OK.

webrtc.html
```diff
@@ -1561,6 +1574,18 @@
...
+
+          <dt>attribute EventHandler onconnectionstatechange</dt>
+
+          <dd>
+            This event handler, of event handler event type <code><a href=
+            "#event-connectionstatechange">connectionstatechange</a></code>,
+            MUST be fired by all objects implementing the
+            <code><a>RTCPeerConnection</a></code> interface. It is
+            called any time the <a href=
+            "#dom-peerconnection-connection-state"><code>RTCPeerConnection</code>.connectionstate</a>
```
martinthomson: Check case.

alvestrand: the attribute name should be connectionState, not connnectionstate, I think.

alvestrand:
Bernard will update.

webrtc.html
pthatcherg: Fixed the case.

webrtc.html
```diff
@@ -1558,6 +1571,18 @@
...
+
+          <dt>attribute EventHandler onconnectionstatechange</dt>
+
+          <dd>
+            This event handler, of event handler event type <code><a href=
+            "#event-connectionstatechange">connectionstatechange</a></code>,
+            MUST be fired by all objects implementing the
+            <code><a>RTCPeerConnection</a></code> interface. It is
+            called any time the <a href=
+            "#dom-peerconnection-connection-state"><code>RTCPeerConnection</code>.connectionState</a>
```
adam-be: I will do a PR that removes all the "it is called when" text from EventHandler attribute descriptions. But that shouldn't stop this PR from landing.

PR #284, opened 2015-08-28
Add .onicecandidateerror, for indicating ICE gathering errors

juberti:
A PR based on the TPAC 2014 proposal: 
https://docs.google.com/presentation/d/1nnKS9GHcOB7M1ZCHMGHRtGk9TCXBoGkISb1qIsDq2Qk/edit#

Please ignore the trailing whitespace removal.

PR #292, opened 2015-09-05
Add PeerConnection.onfatalerror and .onwarning.

pthatcherg:
As a possible solution to https://github.com/w3c/webrtc-pc/issues/278

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1561,6 +1561,34 @@
             the <a href="#dom-peerconnection-ice-gathering-state"><code>RTCPeerConnection</code>
             ice gathering state</a> changes.
           </dd>
+
+          <dt>attribute EventHandler onfatalerror</dt>
+
+          <dd>
+            This event handler, of event handler event type <code><a href=
+            "#event-fatalerror">fatalerror</a></code>,
+            MAY be fired by objects implementing the
+            <code><a>RTCPeerConnection</a></code> interface. It may be
+            called when
+            the <a href="#dom-peerconnection"><code>RTCPeerConnection</code></a>
+            experiences a fatal error and cannot continue.  It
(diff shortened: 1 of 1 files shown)


juberti:
Can you give some examples of when either would be fired?

What constitutes 'fatal'? Isn't that really the app's decision (e.g. if audio broke but video did not)?

pthatcherg:
I think fatal means "you might as well close down the PeerConnection".  The
initial scenario that get me writing this PR is
https://github.com/w3c/webrtc-pc/issues/278, where SRTP fails to
initialize.  If we have an aggregate DTLS or DTLS+ICE state, then we might
not need it to be fatal, since setting the state to "failed" would
effectively do the same thing.

On maybe that's an argument against having an a fatal event: if it's fatal,
set aggregate state to "failed" and just provide a diagnostic message.

martinthomson:
This needs to be a lot more precise.  I think that by now we should know exactly what conditions are "fatal" and what requires warnings.

pthatcherg:
I didn't call it a strawman for nothing :).

So here's a question:  if we ditch the "onfatalerror" and just had
"onwarning", would that valuable by itself just to have a way of sending
diagnostic messages to the JS?

juberti:
Unless we have examples, I'm not sure.

I don't think it's reasonable to try to report browser bugs (e.g. SRTP fail) to the application. What should it do with them? 
Better to use UMA or similar mechanisms.

pthatcherg:
Would your TURN error you were mentioning before be an example?  Or did you
have something else in mind for handling it?

juberti:
No, TURN errors are handled by #284

alvestrand:
Closing per Seattle discussions - not needed.

Issue #187, opened 2015-02-26
How to prevent tracks from being renegotiated after rejection


martinthomson:
Situation:

1. A adds a track and sends an offer.
2. B applies the offer, but rejects the track (calls .stop() on it).
3. A receives the answer and doesn't send the track.
4. A creates a second offer.

As currently specified, the second offer will include the rejected track.

@juberti suggests that we mark the corresponding RtpSender (maybe it can have a `rejected` property).

Also to consider: what if A *wants* to include the track in the second offer?  Should it be forced to remove the dead track and re-add it?  Maybe the `rejected` property can be made mutable, so that it can be set back to false.

stefhak:
Could not that sender move to a closed state (with a close event being fired)? Subsequent offers will not contain the track, and if the application wants to try to send it again it can add it directly (no need to remove it because it is closed and not part of "this RTCPeerConnection object's set of [active] senders.").

juberti:
Senders don't have any states at the moment, but this sounds similar to what Martin/I mentioned above.
I prefer 'closed'/'ended' to 'rejected'. I also think readding makes more sense than making the closed state mutable.

juberti:
@pthatcherg wdyt?

juberti:
@pthatcherg ping

pthatcherg:
I think having an RtpSender.state == "closed" makes a lot of sense (I don't like "rejected").  I think having other states such as "active" or "inactive" depending on the state of the m-line direction also makes sense (sendrecv/sendonly/recvonly/inactive).

However, if an m-line has a port 0, doesn't that "close" the m-line for both the RtpSender and the RtpReceiver with that MID?  If the answerer called addTrack() to create an RtpSender, and then called receiver.track.stop() to close the m-line, wouldn't the closing of the m-line also close the RtpSender just created?  If that's the case, I think we should have an API point that makes this more clear (something like closeSenderAndReceiver).  Or am I wrong and it's possible to close an m-line just for the receiver, but not for the sender?

PR #279, opened 2015-08-26
Add PeerConnection.createRtpReceiver.

pthatcherg:
This has not been proposed on this list yet.  I will do so now, with this as a reference.

PR #293, opened 2015-09-09
Add RTCRtpTransceiver and PeerConnection.addMedia.

pthatcherg:
This is intended to replace createRtpSender (https://github.com/w3c/webrtc-pc/pull/271/) and createRtpReceiver (https://github.com/w3c/webrtc-pc/pull/279/) with something more powerful and complete.  It also fixes https://github.com/w3c/webrtc-pc/issues/187.

Needs list discussion, and I'd like to present on it at the f2f.


webrtc.html
```diff
@@ -2559,6 +2557,17 @@
             </ol>
           </dd>
 
+          <dt>RTCRtpSender addMedia (DOMString kind, RTCRtpTransceiverInit init)</dt>
```
adam-be: RTCRtpTransceiver as return value?
First argument: (MediaStreamTrack or DOMString) media?

webrtc.html
```diff
@@ -2953,6 +2962,99 @@
...
+            will offer to receive RTP and receive RTP if the remote
+            peer accepts.  If false, indicates that
+            the <a><code>RTCRtpReceiveer</code></a> will not offer to
+            receive RTP and will not receive RTP.</p>
+          </dd>
+        </dl>
+
+        <dl class="idl" title=
+          "interface RTCRtpTransceiver">
+          <dt>readonly attribute mid</dt>
```
adam-be: DOMString type?

webrtc.html
```diff
@@ -2953,6 +2962,99 @@
...
+        <!-- TODO: I'm sure we can think of a more creative name. -->
+        <h3>RTCRtpTransceiver Interface</h3>
+
+        <p>The <a><code>RTCRtpTransceiver</code></a> interface
+        represents a combination of an <code>RTCRtpSender</code> and
+        an <a><code>RTCRtpReceiver</code></a> that share a
+        common <a><code>mid</code></a>.
+
+        <dl class="idl" title="dictionary RTCRtpTransceiverInit">
+          <dt>bool send = true</dt>
```
adam-be: s/bool/boolean/

adam-be:
I couldn't get my microphone working during this session on the f2f meeting.

As I understand this PR, there's API for the A-side to add media and inspect existing transceivers. Have you thought anything about how the B-side API looks?

webrtc.html
```diff
@@ -2446,6 +2424,26 @@
             defined and the order does not have to be stable between calls.</p>
           </dd>
 
+          <dt>sequence&lt;RTCRtpTranseceivers&gt; getMedia()</dt>
```
dontcallmedom: This should probably be Transceiver (here and in the rest of the text)

webrtc.html
```diff
@@ -2446,6 +2424,26 @@
...
+            representing the RTP transeceivers that are currently
+            attached to this
+            <code><a>RTCPeerConnection</a></code> object.</p>
+
+            <p>The <dfn id=
+            "dom-peerconnection-gettranseceivers"><code>getTranseceivers()</code></dfn>
+            method MUST return a new sequence that represents a snapshot of all
+            the <code><a>RTCRtpTranseceiver</a></code> objects in this
+            <code><a>RTCPeerConnection</a></code> object's <a href=
+            "#transeceivers-set">set of transeceivers</a>. The
```
dontcallmedom: not sure where that link is supposed to link to

martinthomson:
Consensus from the meeting was that the name was bad.  I don't think that there was wide agreement on a name, but I recall `addMediaSection` being mentioned.  If that is the name then `RTCTransceiver` could - and maybe should - be named `RTCMediaSection`.

jan-ivar:
`RTCMediaLine`?

pthatcherg:
Yes, addMedia is a bad name.  But how is RTCRtpTransceiver?  If it's
palatable, then we could simply call the method to create one
createRtpTransceiver.  It's a bit advanced, but it exists mostly for
advanced use cases anyway.

pthatcherg:
What was mentioned on one my slides was the full "bear hug":
addSdpMediaSection with RTCSdpMediaSection.  But it wasn't met with much
approval.

alvestrand:
Met with approval in Seattle (modulo name bikeshed). Should be reviewed and merged.

alvestrand:
Waiting for naming discussions to conclude. Also, Travis errors.

webrtc.html
pthatcherg: Yep, fixed it.

webrtc.html
pthatcherg: Added place to link to.

webrtc.html
pthatcherg: Yes and yes.

webrtc.html

webrtc.html

pthatcherg:
FYI, I have change the name in the PR to createRtpTransceiver until we come
up with a better one.

webrtc.html
```diff
@@ -2559,6 +2560,25 @@
             </ol>
           </dd>
 
+          <dt>RTCRtpTransceiver createRtpTransceiver(MediaStreamTrack or
+          DOMString track_or_kind, RTCRtpTransceiverInit init)</dt>
+
```
dontcallmedom: The union type `MediaStreamTrack or DOMString` needs to be wrapped in parenthesis, i.e. `(MediaStreamTrack or DOMString) track_or_kind`

webrtc.html
```diff
@@ -2953,6 +2973,99 @@
...
+            <p>If true, indicates that
+            the <a><code>RTCRtpTransceiver</code></a>'s <a><code>RTCRtpSender</code</a>
+            will offer to send RTP and send RTP if the remote peer
+            accepts.  If false, indicates that
+            the <a><code>RTCRtpSender</code></a> will not offer to
+            send RTP and will not send RTP.</p>
+          </dd>
+
+          <dt>bool receive = true</dt>
+
```
dontcallmedom: `boolean` :)

webrtc.html
```diff
@@ -2953,6 +2973,99 @@
...
+          <dd>
+            <p>The <code id="dom-rtptransceiver-sender">RTCRtpTransceiver.sender</code>
+            attribute is
+            the <a><code id="dom-rtptransceiver-sender">RTCRtpSender</code></a>
+            corresponding to the RTP media that may be sent
+            with mid = <a><code>RTCRtpTransceiver.mid</code></a>.</p>
+          </dd>
+
+          <dt>readonly attribute RTCRtpReceiveer receiveer</dt>
+
```
dontcallmedom: extraneous 'e' (should be receiver)

webrtc.html
```diff
@@ -2953,6 +2973,99 @@
       </section>
 
       <section>
+        <!-- TODO: I'm sure we can think of a more creative name. -->
+        <h3>RTCRtpTransceiver Interface</h3>
+
+        <p>The <a><code>RTCRtpTransceiver</code></a> interface
+        represents a combination of an <code>RTCRtpSender</code> and
+        an <a><code>RTCRtpReceiver</code></a> that share a
+        common <a><code>mid</code></a>.
```
dontcallmedom: I suggest removing the `<a>` around `mid` since it doesn't point to a target.

webrtc.html
```diff
@@ -2953,6 +2973,99 @@
...
+
+        <dl class="idl" title="dictionary RTCRtpTransceiverInit">
+          <dt>boolean send = true</dt>
+
+          <dd>
+            <p>If true, indicates that
+            the <a><code>RTCRtpTransceiver</code></a>'s <a><code>RTCRtpSender</code</a>
```
dontcallmedom: missing `>` for the `</code` tag

webrtc.html
```diff
@@ -2953,6 +2973,99 @@
...
+            with mid = <a><code>RTCRtpTransceiver.mid</code></a>.</p>
+          </dd>
+
+          <dt>readonly attribute RTCRtpReceiveer receiveer</dt>
+
+          <dd>
+            <p>The <code id="dom-rtptransceiver-receiveer">RTCRtpTransceiver.receiver</code>
+            attribute is
+            the <a><code id="dom-rtptransceiver-receiveer">RTCRtpReceiver</code></a>
+            corresponding to the RTP media that may be received
```
dontcallmedom: Remove ` id="dom-rtptransceiver-receiveer"` from the 2nd `<code>`

webrtc.html
```diff
@@ -2953,6 +2973,99 @@
...
+            value that will be added to the next value returned by
+            <code>PeerConnection.createOffer</code>.</p>
+          </dd>
+
+          <dt>readonly attribute RTCRtpSender sender</dt>
+
+          <dd>
+            <p>The <code id="dom-rtptransceiver-sender">RTCRtpTransceiver.sender</code>
+            attribute is
+            the <a><code id="dom-rtptransceiver-sender">RTCRtpSender</code></a>
```
dontcallmedom: remove ` id="dom-rtptransceiver-sender"` from the 2nd `<code>`

webrtc.html
```diff
@@ -2953,6 +2973,99 @@
...
+            attribute is
+            the <a><code id="dom-rtptransceiver-receiveer">RTCRtpReceiver</code></a>
+            corresponding to the RTP media that may be received
+            with mid = <a><code>RTCRtpTransceiver.mid</code></a>.</p>
+          </dd>
+
+          <dt>readonly attribute boolean stopped</dt>
+
+          <dd>
+            <p>The <code id="dom-rtptransceiver-receiveer">RTCRtpTransceiver.stopped</code>
```
dontcallmedom: remove ` id="dom-rtptransceiver-receiveer"`

webrtc.html
pthatcherg: fixed

webrtc.html
pthatcherg: fixed

webrtc.html
pthatcherg: Wow, I did some sloppy search and replace.

Fixed.

webrtc.html
pthatcherg: fixed

webrtc.html
pthatcherg: fixed

webrtc.html
pthatcherg: fixed

webrtc.html
pthatcherg: fixed

webrtc.html
pthatcherg: fixed

pthatcherg:
Sorry about all the typos.  I guess I was a little too rushed when writing this.  Hopefully you found all of them.

webrtc.html
```diff
@@ -2559,6 +2560,25 @@
             </ol>
           </dd>
 
+          <dt>RTCRtpTransceiver createRtpTransceiver((MediaStreamTrack or
+          DOMString track_or_kind), RTCRtpTransceiverInit init)</dt>
```
dontcallmedom: the closing parenthesis needs to be after DOMString, not track_or_kind

adam-be:
Do we have a plan to add the receiving side APIs in this PR as well to make it a complete solution?

Now we have addStream(), addTrack() and createRtpTransceiver() that have overlapping functionality. Are we planning to keep addTrack() as well. It doesn't seem to be widely implemented.

webrtc.html
```diff
@@ -2446,6 +2427,26 @@
             defined and the order does not have to be stable between calls.</p>
           </dd>
 
+          <dt>sequence&lt;RTCRtpTransceivers&gt; getMedia()</dt>
```
adam-be: s/getMedia/getTransceivers/

webrtc.html
adam-be: s/track_or_kind/trackOrKind/ ? We don't use underscore as a separator in any other identifiers.

alvestrand:
My reading is that simple applications would use addTrack, but that what addTrack does can be defined in terms of the operations defined for createRtpTransceiver() and replaceTrack() (+ some query calls), and that apps that require more precise control over the SDP would call createRtpTransceiver() directly.

When we added addTrack(), I think we agreed to deprecate and/or remove addStream(), but we haven't gotten around to that yet.

webrtc.html
pthatcherg: Both fixed.

webrtc.html
pthatcherg: fixed

pthatcherg:
addStream is gone and replaced with addTrack, so on overlap there.

There is some overlap of addTrack and createRtpTransceiver, but Harald explains it well.  I agree with his explanation: addTrack for the simple case, createRtpTransceiver for the advanced one.

pthatcherg:
What do you mean by "add the receiving side APIs"?  What's missing?  The RtpReceivers are already there.

adam-be:
So it's only the side that initiates adding media to a session that gets an RTCRtpTransceiver? The other side simply gets a track event with an RTCRtpReceiver?

We might want to add an example using these new features.

alvestrand:
For completeness of this concept, this PR needs to state in the description of addTrack, setLocalDescription and setRemoteDescription that these functions can create transceivers.

An example or two would be a Good Thing.

pthatcherg:
Adam: There are three ways to add a transceiver:
1.  Call addTransceiver: Will always add a transceiver, which will appear as a new m-line in the next call to createOffer.
2.  Call addTrack: May either add a transceiver or use an existing one, which will appear as a new or modified m-line in the next call to createOffer.
3.  Call setRemoteDescription: If the remote side offered a new m-line, it will cause a new Transceiver to be created and .ontrack to be fired.  

I think the question you are really asking is:  In case 3, does ontrack give a Transceiver, or just a Receiver?  I think it should give a .transceiver as well as a .receiver.

pthatcherg:
Harald is right.  I need to update update addTrack, removeTrack, setLocalDescription, and setRemoteDescription, and give some examples.  According to the f2f, I also need to specify the MSID used and put .streams in the TransceiverInit dictionary.

adam-be:
> I think the question you are really asking is: In case 3, does ontrack give a Transceiver, or just a Receiver? I think it should give a .transceiver as well as a .receiver.

@pthatcherg, yes.

alvestrand:
If ontrack gives a transceiver, it can show the receiver as transceiver.receiver, so no extra receiver argument is needed.

alvestrand:
WRT streams, I think they belong on RTPSender, not transceiver - since the streams argument affects what is sent, but not what is received.
Unless we want Transceiver to contain all the baggage we think we might want to not carry forward...

pthatcherg:
OK, I updated addTrack, removeTrack, addTransceiver, track-event.  I also added a warm-up example.  

Harald, I put the streams in RtpTransceiverInit because I see no other reasonable place to put them.  But they don't go on RtpTransceiver after that.  It's just passed in.  But I did change track-event to have a .transceiver, not a .receiver.

While updating addTrack, the big question that I had was: when does an m-line get reused?  The more I thought about it, the more thought that the only time we really care to have it reused is after setRemoteDescription(offer) ands an m-line and we want to call addTrack and then createAnswer and reused that m-line.  In other cases, we're don't have a compelling reason to reused m-lines.  I may be missing something and we may need to change what I wrote, but I erred on the side of "keep it simple".

stefhak:
I looked at #283. This PR is there said to add that functionality using the Transceiver, but I fail to see how unassigned media is covered here. @pthatcherg @martinthomson am I missing something?

pthatcherg:
I still need to add an example for receiving media before an answer
arrives.  I forgot about that when I wrote the example for "warm up".  But
it's quite simple:

var transceiver = pc.addTransceiver(kind);
var track = transceiver.receiver.track;
// ... render track
pc.createOffer().then(signalOffer);

Now this will render the track even if media arrives before the answer is
received.

stefhak:
What happens if you create two transceivers (of same type), and two tracks, and then receive one or two tracks (sort of like SSRCs) in RTP? How would that be wired up?
Also, #283 had an event fired when unknown media arrived at any time (i.e. also if there was no ongoing O/A). That is not supported here - are we fine with removing that? Asking @martinthomson.
Finally: if I were to implement this, I think I would need some more normative text. An example showing how the Transceiver can be used to handle early media is good, but the spec must be clear enough to implement.
To be clear: I think we should merge this PR, but knowing there is more work to be done speccing things out.

pthatcherg:
I don't remember what the plan for completely unknown media is.  I'd have
to go read the f2f minutes/notes to remember.  Martin, do you remember?

As for normative text, I'll add some in the addTransceiver method saying
that the receiver can receive media before setRemoteDescription is called
with an answer, along with the example which I'll add.

If you create two transceivers of the same type, they'll have different
MIDs, and they can be demuxed by MID.  Is that what you were asking?  It
wasn't completely clear.

On Mon, Oct 19, 2015 at 11:06 PM, stefan hakansson <notifications@github.com

alvestrand:
Den 20. okt. 2015 20:57, skrev pthatcherg:
> I don't remember what the plan for completely unknown media is. I'd have
> to go read the f2f minutes/notes to remember. Martin, do you remember?

For media that made it past SRTP verification but can't be associated to
an m-line?
I think it's safe to drop it, but YMMV.

> As for normative text, I'll add some in the addTransceiver method saying
> that the receiver can receive media before setRemoteDescription is called
> with an answer, along with the example which I'll add.
> 
> If you create two transceivers of the same type, they'll have different
> MIDs, and they can be demuxed by MID. Is that what you were asking? It
> wasn't completely clear.

I think that's right. That means MID must be known at addTransceiver
time (or at least before media arrives), which is OK.

stefhak:
Yes, I think this would work. Association is made by the MID field in the RTP header. 
This does not however address the case #283 addresses, namely firing an event if there is incoming (authenticated) media "out of the blue" (i.e. no ongoing negotiation). Personally I'm fine with that, just want to raise the question to @martinthomson and @fluffy

martinthomson:
As we discussed, I think that this is fine.  We can generate messages for the console if we find that diagnosing session errors is difficult.

stefhak:
Given that, I propose we merge this PR.

alvestrand:
Would be great to have a rebase...

webrtc.html
```diff
@@ -2875,6 +2891,30 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+	    add an m-line for the corresponding transceiver, as defined in
+	    [[!RTCWEB-JSEP]].
+
+            <p>If a track is passed in, the value of the <code>.sender.track</code> will
+            be set to that track and the MSID and media type generated
+            by <code>createOffer</code> will be that of the track.</p>
+
+            <p>If a kind is passed in, the value of the <code>.sender.track</code> will be
+            null and and media type generated by <code>createOffer</code> will be that of
+            the kind.  The MSID generated by <code>createOffer</code> will be selected by
```
alvestrand: According to the MSID spec, this will cause a track to be created at the remote side after an offer/answer exchange. I assume this is required as part of the "warmup" usage.
Don't we want to prevent this by passing "send:false"?

webrtc.html
```diff
@@ -3624,6 +3664,106 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+
+          <dt>boolean receive = true</dt>
+
+          <dd>
+            <p>If true, indicates that
+            the <a><code>RTCRtpTransceiver</code>'</a>s <a><code>RTCRtpReceiver</code></a>
+            will offer to receive RTP and receive RTP if the remote
+            peer accepts.  If false, indicates that
+            the <a><code>RTCRtpReceiver</code></a> will not offer to
+            receive RTP and will not receive RTP.</p>
```
alvestrand: Since we talk about "sendonly" and "sendrecv" elsewhere in the document, do we want to say that this causes the m-line to be "sendonly" or "Inactive"?

webrtc.html
```diff
@@ -3624,6 +3664,106 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        <dl class="idl" title="dictionary RTCRtpTransceiverInit">
+          <dt>boolean send = true</dt>
+
+          <dd>
+            <p>If true, indicates that
+            the <a><code>RTCRtpTransceiver</code></a>'s <a><code>RTCRtpSender</code></a>
+            will offer to send RTP and send RTP if the remote peer
+            accepts.  If false, indicates that
+            the <a><code>RTCRtpSender</code></a> will not offer to
+            send RTP and will not send RTP.</p>
```
alvestrand: as below, should we refer to "recvonly" or "inactive" here?

webrtc.html
```diff
@@ -3624,6 +3664,106 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+          </dd>
+
+          <dt>readonly attribute boolean stopped</dt>
+
+          <dd>
+            <p>The <code>RTCRtpTransceiver.stopped</code>
+            attribute indicates that the sender of this transceiver
+            will no longer send, and that the receiver will no longer
+            receive. It is true if
+            either <code>RTCRtpReceiver.stop</code> has been called or
```
alvestrand: This makes more sense if you mean RTCRtpTransceiver.stop here and below.

alvestrand:
I reviewed the PR a little more closely than before, and as always, found a few nits. If you can address them while rebasing, that would be great.

elagerway:
@pthatcherg can you have another look at this, feels like we are close.

pthatcherg:
I'll rebase and address the issues.

webrtc.html
pthatcherg: The MSID is only generated if needed, such as when "send: true".  I will make that more explicit.

webrtc.html
pthatcherg: Sure, I added that.

webrtc.html
pthatcherg: Sure, I added that.

webrtc.html
pthatcherg: Yep, fixed.

pthatcherg:
I've rebased and addressed the issues.

pthatcherg:
I also added an example of receiving media before signalling on the offerer
side.

Issue #4, opened 2014-10-17
Need API to read the CSRC on received tracks 


fluffy:
See section 4 of rtp-usage

fluffy:
@alvestrand - would stats be a reasonable place to put this or should it be RTCRtpReceiver ?

fluffy:
The use case for this BTW is that this is the best way to drive the active speaker indication in a conference roster list.

aboba:
It would extend RTCRtpReceiver, like this:

Contributing Sources

The getContributingSources method extends the RTCRtpReceiver object, providing information on the sources contributing to an incoming media stream.

partial interface RTCRtpReceiver {
    sequence<RTCRtpContributingSource> getContributingSources ();
};

Methods

getContributingSources
Retrieve the sequence of contributing sources.

No parameters.
Return type: sequence<RTCRtpContributingSource>

dictionary RTCRtpContributingSource

dictionary RTCRtpContributingSource {
             DOMHiResTimeStamp timestamp;
             unsigned long     csrc;
             byte              audioLevel;
};

Dictionary RTCRtpContributingSource Members

audioLevel of type byte,
The audio level of the contributing source. Value is between 0 and -127 representing the contributing source dBov value, as described in [RFC6465]

csrc of type unsigned long,
The contributing source.

timestamp of type DOMHiResTimeStamp,




On Jul 31, 2015, at 08:31, Cullen Jennings <notifications@github.com<mailto:notifications@github.com>> wrote:


@alvestrand<https://github.com/alvestrand> - would stats be a reasonable place to put this or should it be RTCRtpReceiver ?

aboba:
It can be used to determine a list of audio volume generators, but determining the dominant speakers is probably best done using other techniques. See:
https://jitsi.org/publications/nossdav2015lastn.pdf

Issue #276, opened 2015-08-25
Control to turn on Client-to-Mixer Audio Level in offers


fluffy:
Client-to-Mixer Audio Level is MTI for browsers but we need to decide when it is used. It is only used when negotiated in SDP but should the browser always include it in the offer or should we have a control surface for the JS to turn that on and off.

juberti:
For 1.0, I think it should always be offered by browsers. The remote side, if it is not a browser, can reject it; if it is a browser, it will be harmlessly accepted.

fluffy:
Justin's proposal works for me

alvestrand:
This doesn't need to be in the API, then. JSEP needs to say that we negotiate it; mention that we use it should go into JSEP or Transport, I think.

juberti:
I think this already is in the docs mentioned above.

alvestrand:
RTP-usage-25 section 5.2.2: "MUST be implemented". Nothing about usage.

JSEP-11:ssrc-audio-level is shown in the examples, but looking for the word "audio" does not locate it.

audio-08: No mention of this header extension.

What did I miss?

juberti:
JSEP, S 5.2.1 says this:

For each supported RTP header extension, an "a=extmap" line, as
      specified in [RFC5285], Section 5.  The list of header extensions
      that SHOULD/MUST be supported is specified in
      [I-D.ietf-rtcweb-rtp-usage], Section 5.2.  Any header extensions
      that require encryption MUST be specified as indicated in
      [RFC6904], Section 4.

alvestrand:
Change to JSEP, not to this document. Closing.

Issue #6, opened 2014-10-17
Need API to receive mixer to client audio level information for a track 


fluffy:
See info in 
https://tools.ietf.org/html/draft-ietf-rtcweb-rtp-usage-17#section-12.2.1

stefhak:
Isn't this covered in 5.3.2 of https://w3c.github.io/webrtc-pc/archives/20151006/webrtc.html ? Inclined to close this issue.

aboba:
Yes, it can be closed.

stefhak:
Closing.

PR #300, opened 2015-09-10
add API to get SSRC and audio levels

fluffy:
Fixes https://github.com/w3c/webrtc-pc/issues/4 and https://github.com/w3c/webrtc-pc/issues/6 and https://github.com/w3c/webrtc-pc/issues/276


fluffy:
Need to indicate which bugs this fixes in the PR

webrtc.html
```diff
@@ -2948,8 +2948,57 @@
...
             In the case where no a=mid attribute is present in the remote
             description, a random value will be generated.
-          </dd>
-        </dl>
+	  </dd>
+
+	  <dt> sequence&lt;RTCRtpContributingSource&gt;
+	    getContributingSources ()  </dt>
+	  <dd> This returns a RTCRtpContributingSource for each unique
+	  CSRC or SSRC received by this RTCRtpReceiver in the last 10
```
ekr: s/This r/R/

webrtc.html
```diff
@@ -2948,8 +2948,57 @@
...
+	  seconds.  </dd>
+	  
+	</dl>
+
+	<p> The RTCRtpContributingSource objects contian information
+	provided about a given contribution source included time most
```
ekr: s/contian/contain/ ->

webrtc.html
```diff
@@ -2948,8 +2948,57 @@
...
+	   <dt> readonly attribute unsigned long source</dt>
+	   <dd> The CSRC or SSRC value of the contributing source. </dd>
+
+	   <dt> readonly attribute byte? audioLevel</dt>
+	   <dd>
+	     The audio level contained in the last RTP packet received
+	   from the this source. If
+	   the source was a CSRC, this will be the value defined as
+	   the level in
+	   [[!RFC6464]] while if the source was set from a CSRC, this will
```
ekr: What's the difference between being from a CSRC and being set from a CSRC.

webrtc.html
```diff
@@ -2948,8 +2948,57 @@
...
+	   be the value as defined in level in [[!RFC6465]]. Those
+	   specification defile this level as a value from 0 to -127
+	   representing the audio level in decibels relative to the
```
ekr: s/defile/define/

ekr: s/value/integral value/

webrtc.html
```diff
@@ -2948,8 +2948,51 @@
...
+
+	  <dt> sequence&lt;RTCRtpContributingSource&gt;
+	    getContributingSources ()  </dt>
+	  <dd> Returns a RTCRtpContributingSource for each unique
+	  CSRC or SSRC received by this RTCRtpReceiver in the last 10
+	  seconds.  </dd>
+	  
+	</dl>
+
+	<p> The RTCRtpContributingSource objects contians information
```
jan-ivar: missed one.

fluffy: thanks - fixed

fluffy:
Folks, I think this represents all the comments from interim meeting and is ready to merge. (Perhaps there are some more typos but it is close enough to merge)

webrtc.html
```diff
@@ -2948,8 +2948,51 @@
...
+	received, the entries are updated for the SSRC of the RTP
+	packet and for any CSRC contained in it. </p>
+
+	 <dl class="idl" title="interface RTCRtpContributingSource">
+
+	   <dt> readonly attribute DOMHiResTimeStamp timestamp</dt>
+
+	   <dd> Time of reception of the most recent RTP packet
+	   containing the contributing source. </dd>
+
```
dontcallmedom: the actual definition of what the timestamp refers to was discussed [on the list](https://lists.w3.org/Archives/Public/public-webrtc/2015Sep/0058.html); shouldn't this be reflected here? (either as "to be determined" or with a specific proposal for review)

fluffy: I agree - I will update the PR to reflect this.

alvestrand:
Met with approval in Seattle. Should be merged.

alvestrand:
Some updates on the definition of time are needed.

fluffy:
This is ready to merge

alvestrand:
@fluffy can you go on the mailing list and reply to the "Potential problem with DOMHghResTimeStamp" thread and say that you think it is unproblematic to use this time differently from everything else in the web platform that uses it?
(in all other usages, DOMHighResTimeStamp are relative to document load time. Bernard's suggested that in this case, it should be relative to the Unix epoch. It's an offset, not an absolute time. Yes, this is a problem with stats too.)

fluffy:
I think we should use existing high res time exactly as defined - that is to say it would not be relative to Unix epoch. I think that is what we agreed to at meeting so if someone wants to make it some other sort of time, I'd rather merge this inn then look at PR that change that

martinthomson:
I agree with @fluffy.  Epoch can be document load time.

aboba:
Found some typos/inconsistencies, and created an updated PR (318):
https://github.com/w3c/webrtc-pc/pull/318

Take a look and see if this gets us closer.

alvestrand:
@aboba do you think you can update this PR instead? Having more PRs for the same thing confuses the trail.
@fluffy @martinthomson can you state your opinion on the list? I'm always hesitant to declare consensus when one opinion is stated on the list and another one is stated in a Github comment thread.

fluffy:
I merged @aboba PR as it looked good

aboba:
Thanks, Cullen.

One question.

For the case where the RTP packet contained CSRCs and a mixer-client
extension, but no client-mixer extension, is the intent for the browser to
include an entry for the SSRC and audio level as though the client-mixer
extension was present?  Wasn't sure if this was only relevant for the P2P
case (where there would be no CSRCs, just SSRC and the client-mixer
extension).

aboba:
Noticed a typo: 

+    information about a given contributing source, including the time
+    the most recent time a packet was received from the source. The

[BA] s/including the time/including/

stefhak:
I think we should merge this PR now, and have it part of a new Ed's draft for wider review.

PR #303, opened 2015-09-11
Replacetrack without renegotiation.

jan-ivar:
A rebase was needed from https://github.com/w3c/webrtc-pc/pull/237. PTAL.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2909,11 +2909,8 @@ sender.setParameters(params)
               </li>
 
               <li>
-                <p>If <var>withTrack's</var> <code><a href=
-                  "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
-                attribute differs from the </var><code><a href=
-                  "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
-                of this sender's
+                <p>If <var>withTrack's</var> <code>kind</code> attribute differs
+                from the <code>kind</code> of this sender's
                 <code><a href="#dom-rtpsender-track">track</a></code>,
                 then reject
                 <var>p</var> with <code>IncompatibleMediaStreamTrackError</code>,
--- a/webrtc.html
+++ b/webrtc.html
@@ -2912,8 +2912,7 @@ sender.setParameters(params)
                 <p>If <var>withTrack's</var> <code>kind</code> attribute differs
                 from the <code>kind</code> of this sender's
                 <code><a href="#dom-rtpsender-track">track</a></code>,
-                then reject
-                <var>p</var> with <code>IncompatibleMediaStreamTrackError</code>,
+                then reject <var>p</var> with <code>TypeError</code>,
                 return <var>p</var> and abort these steps.</p>
               </li>
 
@@ -2927,7 +2926,7 @@ sender.setParameters(params)
                     Ignore which <code>MediaStream</code> the track resides in and
                     the <code>id</code> attribute of the track in this
                     determination. If negotiation is needed, then reject
-                    <var>p</var> with <code>NegotiationNeededError</code>, return
+                    <var>p</var> with <code>InvalidModificationError</code>, return
                     <var>p</var> and abort these steps.</p>
                   </li>
                   <li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -2913,8 +2913,8 @@ sender.setParameters(params)
                   "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
                 attribute differs from the </var><code><a href=
                   "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
-                of
-                <code><a href="#dom-rtpsender-track">RTCRtpSender.track</a></code>,
+                of this sender's
+                <code><a href="#dom-rtpsender-track">track</a></code>,
                 then reject
                 <var>p</var> with <code>IncompatibleMediaStreamTrackError</code>,
                 return <var>p</var> and abort these steps.</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2890,6 +2890,79 @@
             </pre>
           </dd>
 
+          <dt>Promise&lt;void&gt; replaceTrack (MediaStreamTrack withTrack)</dt>
+          <dd>
+            <p>Attempts to replace the track being sent with another track
+            provided, without renegotiation.</p>
+
+            <p>When the <dfn id="dom-rtpsender-replacetrack"><code title=
+            "">replaceTrack()</code></dfn> method is invoked, the user agent MUST
```
martinthomson: drop title=""

webrtc.html
```diff
@@ -2890,6 +2890,79 @@
...
+
+              <li>
+                <p>Run the following steps in parallel:</p>
+              </li>
+              <ol>
+                <li>
+                  <p>Determine if negotiation is needed to transmit
+                  <var>withTrack</var> in place of the sender's existing track.
+                  Ignore which <code>MediaStream</code> the track resides in and
+                  the track's id in this determination. If negotiation is
```
martinthomson: Nit: "the <code>id</code> attribute of the track"

webrtc.html
```diff
@@ -2890,6 +2890,79 @@
...
+                  needed, then reject <var>p</var> with
+                  <code>NegotiationNeededError</code>, return
+                  <var>p</var> and abort these steps.</p>
+                </li>
+                <li>
+                  <p>Have the sender switch seamlessly to transmitting
+                  <var>withTrack</var> in place of what it is sending, without
+                  negotiating. On the remote receiving
+                  end, the track MUST maintain its existing id and stream
```
martinthomson: s/MUST/will/  This isn't a requirement, it's just mathematics.

webrtc.html
```diff
@@ -2890,6 +2890,79 @@
...
+                  associations until the connection ends. To accomplish this,
+                  the sender MUST keep track of the original id and stream
+                  associations, and keep using these in future renegotiations.
```
martinthomson: s/To accomplish this, .../To avoid causing track identifiers changing on the receiver, the server MUST retain the original track identifier and stream associates and use these in subsequent negotiations./

Note that using "keeping track" when you are talking about tracks is a little confusing, the rest just follows from the previous comment.

webrtc.html
jan-ivar: How so? Doesn't it need to be a requirement? The next sentence is "To accomplish this," so if it is self-evident, is that sentence redundant also?

webrtc.html
jan-ivar: Yeah, unintentional pun there.

webrtc.html
ekr: It's not a requirement because the receiver has no information on which it would not maintain its existing ID and stream assocs.

webrtc.html
```diff
@@ -2890,6 +2890,79 @@
...
+                </li>
+
+              </ol>
+              <li>
+                <p>Return <var>p</var>.</p>
+              </li>
+            </ol>
+
+            <div class="note">
+              <p>Things like dimensions and frame rate do not require negotiation.
```
martinthomson: s/do not/might not/  Changing frame rates to values outside of the negotiated imageattr bounds will require negotiation; changing frame rate might cause the block rate for the codec to be exceeded.  Sources that also encode (typically hardware encoders) might be unable to produce the negotiated codec; similarly, software sources might not implement the codec that was negotiated for an encoding source.

Please take another shot at this paragraph.

martinthomson:
A few nits to correct, but this is otherwise good.

webrtc.html
jan-ivar: but the end of this sentence is "until the connection ends", which means it would have to resist renegotiation, .. which, I see, this should be formulated as a requirement on the sender, not the receiver. Got it.

jan-ivar:
Thanks for the feedback, let me know if I hit everything and how the note reads now.

webrtc.html
```diff
@@ -2890,6 +2890,89 @@
...
+
+              </ol>
+              <li>
+                <p>Return <var>p</var>.</p>
+              </li>
+            </ol>
+
+            <div class="note">
+              <p>Changing dimensions and/or frame rates might not require
+              negotiation. Cases that may require negotiation are:
```
martinthomson: s/are/include/

webrtc.html
```diff
@@ -2890,6 +2890,89 @@
...
+              <ol>
+                <li>Changing a frame rate to a value outside of the negotiated
+                imageattr bounds.
```
martinthomson: Maybe cite `[[RFC6236]]`.

webrtc.html
```diff
@@ -2890,6 +2890,89 @@
...
+                  determination. If negotiation is needed, then reject
+                  <var>p</var> with <code>NegotiationNeededError</code>, return
+                  <var>p</var> and abort these steps.</p>
+                </li>
+                <li>
+                  <p>Have the sender switch seamlessly to transmitting
+                  <var>withTrack</var> in place of what it is sending, without
+                  negotiating. To avoid track identifiers changing on the remote
+                  receiving end, the sender MUST retain the original track
+                  identifier and stream associates and use these in subsequent
```
martinthomson: s/associates/associations/

martinthomson:
LGTM

webrtc.html
jan-ivar: I thought you wanted that. ;)

webrtc.html
```diff
@@ -2890,6 +2890,89 @@
...
+                <code><a href="#dom-rtpsender-track">track</a></code>,
+                then reject
+                <var>p</var> with <code>IncompatibleMediaStreamTrackError</code>,
+                return <var>p</var> and abort these steps.</p>
+              </li>
+
+              <li>
+                <p>Run the following steps in parallel:</p>
+              </li>
+              <ol>
```
dontcallmedom: the nesting is wrong here (the &lt;ol> should be inside the &lt;li>)

webrtc.html
```diff
@@ -2890,6 +2890,89 @@
...
+                <p>Let <var>p</var> be a new promise.</p>
+              </li>
+
+              <li>
+                <p>Let <var>withTrack</var> be the argument to this method.</p>
+              </li>
+
+              <li>
+                <p>If <var>withTrack's</var> <code><a href=
+                  "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
```
dontcallmedom: you need an absolute link here (getusermedia.html is no longer in this repo) (no link is also probably OK); this also applies to line 2915

webrtc.html
```diff
@@ -2890,6 +2890,89 @@
...
+                attribute differs from the </var><code><a href=
+                  "getusermedia.html#dom-mediastreamtrack-kind">kind</a></code>
+                of this sender's
+                <code><a href="#dom-rtpsender-track">track</a></code>,
+                then reject
+                <var>p</var> with <code>IncompatibleMediaStreamTrackError</code>,
```
dontcallmedom: I don't think we want to go with newly minted error names if we can avoid it; I would guess that TypeError, InvalidModificationError, or NotSupportedError would work here.

webrtc.html
```diff
@@ -2890,6 +2890,89 @@
...
+                <p>Run the following steps in parallel:</p>
+
+                <ol>
+                  <li>
+                    <p>Determine if negotiation is needed to transmit
+                    <var>withTrack</var> in place of the sender's existing track.
+                    Ignore which <code>MediaStream</code> the track resides in and
+                    the <code>id</code> attribute of the track in this
+                    determination. If negotiation is needed, then reject
+                    <var>p</var> with <code>NegotiationNeededError</code>, return
```
dontcallmedom: likewise, we should not mint a new error; probably should be using one of the options above (but different from the incompatible kind)

alvestrand:
Met with approval in Seattle (as #237, with proviso that renegotiation should not happen). Mergeable.

PR #304, opened 2015-09-14
Remove outdated issue on missing protocol spec for data channel

dontcallmedom:
Add inline ref to protocol for good measure


fluffy:
LGTM

Issue #299, opened 2015-09-10
operations array prose is broken


jan-ivar:
See https://github.com/w3c/webrtc-pc/pull/236#issuecomment-138804157 and please re-open that PR.

jan-ivar:
Tl;dr; the current operations prose was written for `void`-returning methods with callback-args, and is completely broken for `promise`-returning methods.

jan-ivar:
I was asked to clarify the problem here:

The existing text pushes objects on an array without returning anything to the caller, yet the caller is expecting a promise in return. That wont work.

Each method has a synchronous portion in its processing model, which needs to be executed synchronously, but isn't.

jan-ivar:
In other words, the old text doesn't return any promises is the problem, or allow the synchronous parts of the operations to run, and they need to run synchronously, since they're the ones creating the promises.

martinthomson:
I think that this is a fair summary.  If people are concerned about the use of promise-based language, which I can agree could be construed as being too complex, then I have no concern with adding prose to make the intent clearer.

If people are concerned about making the spec longer, that's a poor argument given its present length and lack of specificity (it needs to be longer).

Note: #236 was a precise description of the implementation of this in Firefox.  I'm sure that we can do better, but what we have isn't correct and should be fixed as per https://github.com/w3c/webrtc-pc/issues/299#issuecomment-142379985

jan-ivar:
Note that https://github.com/w3c/webrtc-pc/issues/307 is the latest PR with earlier feedback from @adam-be addressed.

adam-be:
The end of the title of #236 shouldn't have been "preserving existing behavior.". It should have been "to fix a lot of broken stuff". :) But let's fix it now.

alvestrand:
Solved by #314

PR #3, opened 2014-10-15
Change addStream to addTrack; add RTCRtpSender/RTCRtpReceiver

juberti:
Initial version of doohickeys. @alvestrand @stefhak @pthatcherg PTAL and see if you think this is going in the right direction. If so I will deal with the few formatting and link nits remaining.

Details:
* addStream/remoteStream/getSenders/getReceivers and associated APIs have now been moved to a new document section (#5). This allows section #4 to focus on the basics of the PC API, and #5 to extend it with media functionality and #6 to extend it with data channel functionality.
* addStream has become addTrack(MediaStreamTrack track, MediaStream... streams), as discussed in DC. addTrack returns a RTCRtpSender.
* removeStream has become removeTrack; removeTrack takes a RTCRtpSender.
* onaddstream has become onaddtrack; onremovestream has been removed (now, .onended is fired).
* getLocalStreams/getRemoteStreams -> getSenders/getReceivers.

PR #307, opened 2015-09-21
Fix broken RTCPeerConnection's operations array to work with promises.

jan-ivar:
I'm resubmitting this PR for consideration as a solution to https://github.com/w3c/webrtc-pc/issues/299, and I respectfully request that people take another look, as I don't see any competing PRs to solve this issue.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -542,8 +542,7 @@
             rejection handler, both of which MUST immediately decrement
             <var>connection</var>'s <var>operationsLength</var> by
             <code>1</code> and return <var>undefined</var> (this has no effect
-            on <var>p</var>). To ensure the decrement happens before any other
-            use of <var>p</var>, both handlers MUST not do anything else.</p>
+            on <var>p</var>).</p>
           </li>
 
           <li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -481,8 +481,9 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to represent a queue of
-            <code>operations</code> with an empty set.</p>
+            <p>On the <var>connection</var>, reserve an internal slot
+            <var>operations</var>, and initialize an internal slot
+            <var>operationsLength</var> to <code>0</code>.</p>
           </li>
 
           <li>
@@ -509,33 +510,60 @@
 
         <ol>
           <li>
-            <p>Append an object representing the current call being handled
-            (i.e. function name and corresponding arguments) to the
-            <code>operations</code> array.</p>
+            <p>Let <var>operation</var> be a function executing the steps of the
+            function in question.</p>
+          </li>
+          <li>
+            <p>If <var>connection</var>'s <a href=
+              "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+              signalingState</a> is <code>closed</code>, throw an
+              <code>InvalidStateError</code> exception and abort these steps.</p>
+          </li>
+          <li>
+            <p>If <var>connection</var>'s <var>operationsLength</var> is
+            <code>0</code>, let <var>p</var> be the result of executing
+            <var>operation</var>; otherwise, let <var>p</var> be the result of
+            transforming <var>connection</var>'s <var>operations</var> by a
+            fulfillment handler that runs the following steps:</p>
+            <ol>
+              <li>
+                <p>If <var>connection</var>'s <a href=
+                "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+                signalingState</a> is <code>closed</code>, return
+                <var>undefined</var> and abort these steps.</p>
+              </li>
+              <li>
+                <p>Return the result of executing <var>operation</var>.</p>
+              </li>
+            </ol>
           </li>
 
           <li>
-            <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            <p>Increment <var>connection</var>'s <var>operationsLength</var> by
+            <code>1</code>.</p>
+          </li>
+          <li>
+            <p>Set <var>connection</var>'s <var>operations</var> to the result
+            of transforming <var>p</var> by a fulfillment and
+            rejection handler, both of which decrement <var>connection</var>'s
+            <var>operationsLength</var> by <code>1</code> and return
+            <var>undefined</var> (this has no effect on <var>p</var>).
+            Both handlers MUST transform <var>p</var> directly, to ensure the
+            decrement happens before any other use of <var>p</var>.</p>
           </li>
 
           <li>
-            <p>When the asynchronous operation completes (either successfully
-            or with an error), remove the corresponding object from the
-            <code>operations</code> array. After removal, if the array is
-            non-empty, execute the first object queued asynchronously and
-            repeat this step on completion.</p>
+            <p>Return <var>p</var>.</p>
           </li>
         </ol>
 
         <p>The general idea is to have only one among <code>createOffer</code>,
-        <code>setLocalDescription</code>, <code>createAnswer</code> and
-        <code>setRemoteDescription</code> executing at any given time. If
+        <code>setLocalDescription</code>, <code>createAnswer</code>,
+        <code>setRemoteDescription</code> and <code>addIceCandidate</code>
+        executing at any given time. If
         subsequent calls are made while one of them is still executing, they
-        are added to a queue and processed when the previous operation is fully
-        completed. It is valid, and expected, for normal error handling
-        procedures to be applied.</p>
+        are added to a chain and processed upon fulfillment or rejection of the
+        previous operation's promise.</p>
 
         <p>Additionally, during the lifetime of the RTCPeerConnection object,
         the following procedures are followed when an ICE event occurs:</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -482,8 +482,8 @@
 
           <li>
             <p>On the <var>connection</var>, reserve an internal slot
-            <var>operations</var>, and initialize an internal slot
-            <var>operationsLength</var> to <code>0</code>.</p>
+            <var>operationsDone</var>, and initialize an internal slot
+            <var>operationsCount</var> to <code>0</code>.</p>
           </li>
 
           <li>
@@ -516,31 +516,27 @@
           <li>
             <p>If <var>connection</var>'s <a href=
               "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
-              signalingState</a> is <code>closed</code>, throw an
-              <code>InvalidStateError</code> exception and abort these steps.</p>
+              signalingState</a> is <code>closed</code>, return a promise
+              rejected with <code>InvalidStateError</code> and abort these steps.</p>
           </li>
           <li>
-            <p>If <var>connection</var>'s <var>operationsLength</var> is
+            <p>If <var>connection</var>'s <var>operationsCount</var> is
             <code>0</code>, let <var>p</var> be the result of executing
             <var>operation</var>; otherwise, let <var>p</var> be the result of
-            transforming <var>connection</var>'s <var>operations</var> by a
-            fulfillment handler that runs the following steps:</p>
-            <ol>
-              <li>
-                <p>Return the result of executing <var>operation</var>.</p>
-              </li>
-            </ol>
+            transforming <var>connection</var>'s <var>operationsDone</var> by a
+            fulfillment handler that returns the result of executing
+            <var>operation</var>.</p>
           </li>
 
           <li>
-            <p>Increment <var>connection</var>'s <var>operationsLength</var> by
+            <p>Increment <var>connection</var>'s <var>operationsCount</var> by
             <code>1</code>.</p>
           </li>
           <li>
-            <p>Set <var>connection</var>'s <var>operations</var> to the result
+            <p>Set <var>connection</var>'s <var>operationsDone</var> to the result
             of transforming <var>p</var> by a fulfillment and
             rejection handler, both of which MUST immediately decrement
-            <var>connection</var>'s <var>operationsLength</var> by
+            <var>connection</var>'s <var>operationsCount</var> by
             <code>1</code> and return <var>undefined</var> (this has no effect
             on <var>p</var>).</p>
           </li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -527,12 +527,6 @@
             fulfillment handler that runs the following steps:</p>
             <ol>
               <li>
-                <p>If <var>connection</var>'s <a href=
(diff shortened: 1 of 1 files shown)


alvestrand:
@ekr @adam-be @pthatcherg can you comment?

ekr:
This doesn't seem to me to be an improvement.

jan-ivar:
@ekr are you saying this does not fix https://github.com/w3c/webrtc-pc/issues/299?

ekr:
#299 does not contain a complete description.

jan-ivar:
I've updated https://github.com/w3c/webrtc-pc/issues/299 with a better summary.

webrtc.html
```diff
@@ -509,33 +510,54 @@
...
+            <p>Return <var>p</var>.</p>
           </li>
         </ol>
 
         <p>The general idea is to have only one among <code>createOffer</code>,
-        <code>setLocalDescription</code>, <code>createAnswer</code> and
-        <code>setRemoteDescription</code> executing at any given time. If
+        <code>setLocalDescription</code>, <code>createAnswer</code>,
+        <code>setRemoteDescription</code> and <code>addIceCandidate</code>
+        executing at any given time. If
```
pthatcherg: Mentioning addIceCandidate seems like a good thing.

webrtc.html
```diff
@@ -481,8 +481,9 @@
           </li>
 
           <li>
-            <p>Initialize an internal variable to represent a queue of
-            <code>operations</code> with an empty set.</p>
+            <p>On the <var>connection</var>, reserve an internal slot
+            <var>operations</var>, and initialize an internal slot
+            <var>operationsLength</var> to <code>0</code>.</p>
           </li>
```
pthatcherg: I still prefer the text speaking in terms of "queue of operations" rather than "reserve an internal slot".  Can you not keep the text in terms of "queue of operations" and also fix the deficiencies you are trying to address?

You argument seems to be "X has deficiencies, so we should replace it with Y", when just tweaking X a little bit might address the deficiencies without turning it into Y.

webrtc.html
```diff
@@ -509,33 +510,54 @@
...
-            <code>operations</code> array.</p>
+            <p>Let <var>operation</var> be a function executing the steps of the
+            function in question.</p>
+          </li>
+          <li>
+            <p>If <var>connection</var>'s <a href=
+              "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+              signalingState</a> is <code>closed</code>, throw an
+              <code>InvalidStateError</code> exception and abort these steps.</p>
+          </li>
```
pthatcherg: Mentioning what to do when the the signaling state is closed also seems like a good idea.

webrtc.html
```diff
@@ -509,33 +510,54 @@
 
         <ol>
           <li>
-            <p>Append an object representing the current call being handled
-            (i.e. function name and corresponding arguments) to the
-            <code>operations</code> array.</p>
+            <p>Let <var>operation</var> be a function executing the steps of the
+            function in question.</p>
```
pthatcherg: I like the idea of defining "operation" and using it later.

webrtc.html
```diff
@@ -509,33 +510,54 @@
...
+            <code>1</code>.</p>
+          </li>
+          <li>
+            <p>Set <var>connection</var>'s <var>operations</var> to the result
+            of transforming <var>p</var> by a fulfillment and
+            rejection handler, both of which MUST immediately decrement
+            <var>connection</var>'s <var>operationsLength</var> by
+            <code>1</code> and return <var>undefined</var> (this has no effect
+            on <var>p</var>). To ensure the decrement happens before any other
+            use of <var>p</var>, both handlers MUST not do any thing else.</p>
```
pthatcherg: This is the part where I think it gets more complicated than it needs to be.  Your argument seems to be "defining the behavior in terms of a queue of operations is impossible, therefore we must define it as a transformation of promises".  But I'm not convinced it's impossible to define the behavior in terms of a queue of operations.  And I still find this model of describing behavior much more difficult to understand.

webrtc.html
jan-ivar: We [seem to be using](http://w3c.github.io/webrtc-pc/#rtccertificate-interface)  "internal slot" now, instead of "internal variable". Open to discuss.

The conceptual difference I think you're pointing at is "chain of operations" vs. "queue of operations". Established vernacular seems to be that promise functions [chain](http://www.html5rocks.com/en/tutorials/es6/promises/#toc-chaining), they don't [queue](http://www.w3.org/TR/html5/webappapis.html#task-queue).

webrtc.html
jan-ivar: I'm sad to hear things need to be "impossible" before adopting a standard.

I started out "just tweaking", and it led me here when I realized I was reinventing [.then](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-performpromisethen), poorly. [1]

This was [9 months ago when I added this to Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1106675).

[1] Promises nail down behaviors and timings around asynchronous operations in ways that aren't easily replicable (since a microtask queue is used). Also, error handling needs to be done right.

jan-ivar:
Btw, here it is in code:
```js
_runOperation: function(operation) {
  this._throwIfClosed();
  let p = (!this._operationsLength)? operation() : this._operations.then(operation);
  this._operationsLength++;
  let decrement = () => this._operationsLength--;
  this._operations = p.then(decrement, decrement);
  return p;
},
```

dontcallmedom:
I like the rewrite; it looks more accurate and precise than the current text.

webrtc.html
pthatcherg: OK, I think I misunderstood it before.  I get it now.  It's really the part below I have a problem with.  This is just the init part.

pthatcherg:
OK, I've looked into this more deeply and I think it's trying to do 3 different things, 2 of which are good, 1 of which is not good.

1.  Update to deal with promises and closed signaling state (good).
2   Some general cleanup, like referring to addIceCandidate (good).
3.  Defining a queue of operations as a promise chain + queue length (not good).

To keep the first 2, but not the 3rd, I created https://github.com/w3c/webrtc-pc/pull/314/files.

I still think #3 is ugly, and I still don't see what it buys us.  As far as I can tell, it's only there to try and address this use case:

var senderX = pc.addTrack(trackX);
pc.createOffer();
var senderY = pc.addTrack(trackY);


But I don't see how this PR fixes that use case.  The fundamental problem with that use case is that addTrack isn't in the queue with the rest of the operations, not that the queue is defined in terms of a queue or a promise chain.

pthatcherg:
Yikes, somehow my text got garbled.  The three were supposed to be:


- Update to deal with promises and closed signaling state (good). 

- Some general cleanup, like referring to addIceCandidate (good).

- Defining a queue of operations as a promise chain + queue length (not good).

pthatcherg:
By the way, I think in this case:

var senderX = pc.addTrack(trackX);
pc.createOffer();
var senderY = pc.addTrack(trackY);

The answer should be that the offer has the first track but not the second.   Previously, you said that the "promise chain" approach would make it be the opposite, where the offer would include both.  I think that would be broken.

webrtc.html
pthatcherg: On Tue, Sep 22, 2015 at 7:25 PM, jan-ivar <notifications@github.com> wrote:

> In webrtc.html
> <https://github.com/w3c/webrtc-pc/pull/307#discussion_r40165975>:
>
> >            </li>
> >
> >            <li>
> > -            <p>If the length of the <code>operations</code> array is exactly 1,
> > -            execute the function from the front of the queue
> > -            asynchronously.</p>
> > +            <p>Increment <var>connection</var>'s <var>operationsLength</var> by
> > +            <code>1</code>.</p>
> > +          </li>
> > +          <li>
> > +            <p>Set <var>connection</var>'s <var>operations</var> to the result
> > +            of transforming <var>p</var> by a fulfillment and
> > +            rejection handler, both of which MUST immediately decrement
> > +            <var>connection</var>'s <var>operationsLength</var> by
> > +            <code>1</code> and return <var>undefined</var> (this has no effect
> > +            on <var>p</var>). To ensure the decrement happens before any other
> > +            use of <var>p</var>, both handlers MUST not do any thing else.</p>
>
> I'm sad to hear things need to be "impossible" before adopting a standard.
>
Before we throw something very complicated into a standard, we should first
see if it's possible accomplish the same thing with something more simple.​
 You are proposing replacing something simple with something complicated.
I think we should see if we can accomplish the same thing while keeping it
simple.

> I started out "just tweaking", and it led me here when I realized I was
> reinventing .then
> <https://people.mozilla.org/%7Ejorendorff/es6-draft.html#sec-performpromisethen>,
> poorly. [1]
>
> This was 9 months ago when I added this to Firefox
> <https://bugzilla.mozilla.org/show_bug.cgi?id=1106675>.
>
> [1] Promises nail down behaviors and timings around asynchronous
> operations in ways that aren't easily replicable (since a microtask queue
> is used). Also, error handling needs to be done right.
>
Can you please take a look at https://github.com/w3c/webrtc-pc/pull/314,
which I wrote to update the current approach (queue of operations) in terms
of promises?  Can you tell me under what circumstances the behavior would
be different than this PR (chain of promises)?  If there are real behavior
differences, then I'd like to attempt to achieve the same thing with a
queue of operations while keeping it simple.  If there are no behavior
differences, then we can keep the more simple model of a queue of
operations.

​

jan-ivar:
> I still think #3 is ugly, and I still don't see what it buys us.

It solves https://github.com/w3c/webrtc-pc/issues/299.

> As far as I can tell, it's only there to try and address this use case:
>
> var senderX = pc.addTrack(trackX);
> pc.createOffer();
> var senderY = pc.addTrack(trackY);

Not so. Compare to https://github.com/w3c/webrtc-pc/pull/222 to see what was added to solve that use-case.

> But I don't see how this PR fixes that use case.

It solves it by executing *operation* synchronously when *chainLength* is `0`.

> The fundamental problem with that use case is that addTrack isn't in the queue with the rest of the operations, not that the queue is defined in terms of a queue or a promise chain.

That sounds like another solution rather than a fundamental statement (`addTrack` is not the lone sync api). The working group spent all sprint to get to agreement that this was the desired behavior, so please lets not reopen that can of worms.

jan-ivar:
s/sprint/spring/

pthatcherg:
On Wed, Sep 23, 2015 at 6:36 AM, jan-ivar <notifications@github.com> wrote:

> I still think #3 <https://github.com/w3c/webrtc-pc/pull/3> is ugly, and I
> still don't see what it buys us.
>
> It solves #299 <https://github.com/w3c/webrtc-pc/issues/299>.
>
#299 isn't very clear, but it appears to say "current text doesn't address
promises" and "current text doesn't define 'execute asynchronously'".

I wrote the is PR: https://github.com/w3c/webrtc-pc/pull/314  to address
both of these while keeping it closer to the current text of "queue of
operations".  In other words I don't think "queue of operations" is needed
to address #299.

> As far as I can tell, it's only there to try and address this use case:
>
> var senderX = pc.addTrack(trackX);
> pc.createOffer();
> var senderY = pc.addTrack(trackY);
>
> Not so. Compare to #222 <https://github.com/w3c/webrtc-pc/pull/222> to
> see what was added to solve that use-case.
>
Can you please make it more clear?   I'm having a heard time reading the
difference between those two PRs and figure out how it address this case.

But I don't see how this PR fixes that use case.
>
> It solves it by executing *operation* synchronously when *chainLength* is
> 0.
>
​What do you mean it executes it synchronously?  ​The PeerConnection isn't
going to wait for the promise to fulfilled, is it?

I still don't see how the behavior of PR 307 is different than PR 314.
I'll ask again: can you please provide a case where it's different?
Because if it's not then I'd prefer PR 314 because it's much more simple.

> The fundamental problem with that use case is that addTrack isn't in the
> queue with the rest of the operations, not that the queue is defined in
> terms of a queue or a promise chain.
>
> That sounds like another solution rather than a fundamental statement (
> addTrack is not the lone sync api). The working group spent all sprint to
> get to agreement that this was the desired behavior, so please lets not
> reopen that can of worms.
>
​Sorry, I may have missed the discussion.  Can you summarize what the
agreement was?​

jan-ivar:
> #299 isn't very clear, but it appears to say "current text doesn't address promises" and "current text doesn't define 'execute asynchronously'".

It actually says: "the old text doesn't return any promises is the problem, or allow the synchronous parts of the operations to run, and they need to run synchronously, since they're the ones creating the promises."

Please let me know what is not clear about that.

pthatcherg:
Well, that's the 3rd comment down after a reference to a giant comment on a
PR, followed by a vague TL;DR.   I didn't get past the giant comment and
the vague TL;DR.  Your clear statement was lost in the noise, I'm afraid.
But I think I may understand now.

jan-ivar:
> ​What do you mean it executes it synchronously?

Each of the five *operation*s: `createOffer`, `setLocalDescription`, `createAnswer`, `setRemoteDescription`, and `addIceCandidate`, are *functions* that come with processing models.

Look at [`setLocalDescription`](http://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-setLocalDescription-Promise-void--RTCSessionDescription-description):
> 1. Let *p* be a new promise.
> 2. If this RTCPeerConnection object's signaling state is closed, the user agent MUST reject p with InvalidStateError, and jump to the step labeled Return.
> 3. If a local description contains a different set of ICE credentials, then the ICE Agent MUST trigger an ICE restart. When ICE restarts, the gathering state will be changed back to "gathering", if it was not already gathering. If the RTCPeerConnection ice connection state was "completed", it will be changed back to "connected".
> 4. The user agent MUST start the process to apply the RTCSessionDescription argument.
> 5. Return: Return *p*.

(I'm leaving out the remaining steps which are all about what step 4 launched in parallel).

Before it returns, this function:
1. Creates a new promise *p*.
2. Rejects and returns *p* if the pc is closed.
3. Starts stuff in parallel that will resolve or reject *p* later.
4. Returns *p*.


*p* is important, as without *p* you have no handle on when the asynchronous part set up by `setLocalDescription` completes.

*Without calling these steps you have no p*. Furthermore, in the typical call (where nothing is queued) the steps must be called *synchronously*, by which I mean that the steps must have executed *before* the call from content JS returns, and *p* must be returned to content JS.

That's what I mean by synchronous.

jan-ivar:
> But I think I may understand now.

Great. I'm striving to get better at boiling stuff down. I see you made additional comments elsewhere, so let me catch up.

jan-ivar:
Addressed feedback from @adam-be.

jan-ivar:
Simplified for fair comparison with https://github.com/w3c/webrtc-pc/pull/317.

jan-ivar:
Simplified for fair comparison with https://github.com/w3c/webrtc-pc/pull/314.

stefhak:
Closing as the discussion seems to have moved to #314 and in addition seems to have converged over there.

PR #311, opened 2015-09-22
Fixing syntax for required RTCCertificate arguments

### LINKED ISSUE #310: WebRTC/WebCrypto mismatch: must-be-supported AlgorithmIdentifier value is rejected by WebCrypto normalization process
The WebRTC spec states that keygenAlgorithm of generateCertificate uses the WebCrypto AlgorithmIdentifier type and that it must produce a non-error result when normalized according to the WebCrypto algorithm normalization process. It also states that the following value must be supported by the user agent:
{ name: "RSASSA-PKCS1-v1_5", modulusLength: 2048, publicExponent: 65537 }
http://w3c.github.io/webrtc-pc/archives/20150611/webrtc.html#widl-RTCPeerConnection-generateCertificate-Promise-RTCCertificate--AlgorithmIdentifier-keygenAlgorithm

The WebCrypto API on the other hand defines publicExponent as a BigInteger a.k.a. Uint8Array. Parsing the must-be-supported value produces an error result.
https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#RsaKeyGenParams-dictionary

This chromium issue shows the code and error message I got when I tried to normalize:
https://code.google.com/p/chromium/issues/detail?id=534416

martinthomson:
This uses the crappier, but correct form.

Fixes #310.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2434,7 +2434,7 @@
           <p>The following values MUST be supported by a <a>user agent</a>:
             <code>{ name:
             "<a href="https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#rsassa-pkcs1">RSASSA-PKCS1-v1_5</a>",
-            modulusLength: 2048, publicExponent: 65537 }</code>, and <code>{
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

PR #314, opened 2015-09-23
Update the operations queue to handle promises and closed signalling …

pthatcherg:
An alternative to https://github.com/w3c/webrtc-pc/pull/307/ where we update to handle promises, but we keep a queue of operations rather than a promise chain.


webrtc.html
```diff
@@ -509,33 +509,42 @@
...
-            <p>When the asynchronous operation completes (either successfully
-            or with an error), remove the corresponding object from the
-            <code>operations</code> array. After removal, if the array is
-            non-empty, execute the first object queued asynchronously and
-            repeat this step on completion.</p>
+            <p>If there is exactly one <code>operation</code> in the
+            queue, execute it asynchronously.</p>
+          </li>
+
+          </li>
```
dontcallmedom: should be `<li>`

pthatcherg: fixed

webrtc.html
```diff
@@ -509,33 +509,42 @@
...
 
           <li>
-            <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            <p>Append an <code>operation</code> to
+            the <var>operations</var> queue, where
+            an <code>operation</code> represents the asynchronous
+            execution of the called function that will fulfill the
+            promise returned by the function.</p>.
```
jan-ivar: The called function must be run synchronously, since it's creating the promise.

pthatcherg: The operation is the asynchronous execution that will fulfill the promise.  The code that sticks the operation in the queue is run synchronously, but surely the fulfillment of the promise isn't.

webrtc.html
```diff
@@ -509,33 +509,42 @@
...
           </li>
 
           <li>
-            <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            <p>If there is exactly one <code>operation</code> in the
+            queue, start the operation.</p>
```
jan-ivar: Start is vague.

pthatcherg:
Can you be more specific about which specifics that it is missing?  Because if I go implement this, I think I understand what I need to do.

webrtc.html
pthatcherg: Operation is previously defined as the asynchronous execution that fulfills the returned promise.  So "start" means start the asynchronous execution that fulfills the returned promise.    Would you prefer a different word?

webrtc.html
```diff
@@ -509,33 +509,42 @@
...
           <li>
-            <p>When the asynchronous operation completes (either successfully
-            or with an error), remove the corresponding object from the
-            <code>operations</code> array. After removal, if the array is
-            non-empty, execute the first object queued asynchronously and
-            repeat this step on completion.</p>
+            <p>When any <code>operation</code> completes and fulfills
+            the corresponding promise, remove it from the front of the
+            queue and start the next operation in queue if the queue
+            is non-empty.</p>.
```
jan-ivar: This doesn't handle errors. If an *operation* fails then this stalls.

pthatcherg: If the operation fails, then it still fulfills the promise, right?  How does that lead to a stall?

pthatcherg: Or is there a different promise-lingo that means "gives a result or an error"?

webrtc.html
```diff
@@ -509,33 +509,42 @@
...
           </li>
```
jan-ivar: This PR doesn't have enough specificity to implement code from. Straddling the difference between synchronous and asynchronous operations, is a mistake, and will lead to bugs.

webrtc.html
jan-ivar: From [setLocalDescription's processing steps](http://w3c.github.io/webrtc-pc/#dom-peerconnection-setlocaldescription):
> 1. Let *p* be a new promise.

Are you going to run that asynchronously?

webrtc.html
jan-ivar: See my earlier comment. I would prefer a definition of the operations chain that works without altering all the functions it operates on.

webrtc.html
jan-ivar: [States and fates](https://github.com/domenic/promises-unwrapping/blob/master/docs/states-and-fates.md).

webrtc.html
pthatcherg: I'm not sure what you mean.  You already have an implementation that does this, so clearly it's implementable.  The real question is: is it specific enough that implementations across browsers will have similar enough behavior such applications can rely on it?   And I'll ask again:  In what circumstances is the behavior described here different than PR 307?

webrtc.html
pthatcherg: Clearly, you prefer the chain of promises.  I prefer the queue of operations.  The question is: are the different ways of explaining the same behavior, or is there a real difference in behavior?  If so, what is it?

pthatcherg: I don't know what you mean by "without altering all the functions".  What functions are being altered?

jan-ivar: To be clear: your definition of *operation* as "the asynchronous execution of the called function that fulfills the returned promise", clashes with the fact that all the functions having both synchronous and asynchronous parts in their [processing steps](http://w3c.github.io/webrtc-pc/#dom-peerconnection-setlocaldescription) that need to execute, and your model is not executing the synchronous parts. If you're trying to define *operation* to mean *only* the asynchronous parts of said processing models, then please don't.

jan-ivar:
Feel free to add code showing what you mean.

pthatcherg:
OK, Jan-Ivar, I tweaked it in a way that you might like.  I like it better, because it's more simple:  it's a queue of promises to fulfill.  That way, everything is defined in terms of promises, which I think it the fundamental thing you want, while avoiding the queue being "impemented" as a chain+length, which is what I would like to avoid.

Please take a look and let me know if you still a have doubts of whether it's behavior is equivalent to #307 or not.

webrtc.html
```diff
@@ -509,33 +509,43 @@
...
-            <code>operations</code> array.</p>
+            <p>If <var>connection</var>'s <a href=
+              "#dom-peerconnection-signaling-state"><code>RTCPeerConnection</code>
+              signalingState</a> is <code>closed</code>, throw an
+              <code>InvalidStateError</code> exception and abort these steps.</p>
+          </li>
+
+          <li>
+            <p>Let <var>p</var> be the promise that the called
+            function will return.
```
jan-ivar: How is this possible? A promise is a tangible object that must be returned immediately in place of a synchronous result, that's the whole point of the pattern. How can you have it before having called the function?

pthatcherg: The called function puts it in the queue before returning it.  

I believe this is exactly what your PR #307 does, except "put it in the queue" means "put it at the end of a chain'.  Either way, it stores it way before returning it.

webrtc.html
```diff
@@ -509,33 +509,43 @@
...
+          </li>
+            <p>Add <code>p</code> to <var>operations</var>.</p>
           </li>
 
           <li>
-            <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            <p>If the length of <var>operations</var> is exactly 1,
+            execute the steps necessary to resolve <code>p</code>.</p>
```
jan-ivar: You've replacing "the function" with "the steps necessary to resolve *p*". This seems to intrude on the processing model of every method we hope to support queueing of, and requires each of them to be specially rewritten to work with this concept (and nothing else). That seems unnecessarily entangled in the face of the more encapsulated solution on the table that doesn't require this level of intimacy with all methods.

webrtc.html
```diff
@@ -509,33 +509,43 @@
...
           </li>
 
           <li>
-            <p>When the asynchronous operation completes (either successfully
-            or with an error), remove the corresponding object from the
-            <code>operations</code> array. After removal, if the array is
-            non-empty, execute the first object queued asynchronously and
-            repeat this step on completion.</p>
+            <p>When a promise in <code>operations</code> is resolved,
```
jan-ivar: settled,

webrtc.html
```diff
@@ -509,33 +509,43 @@
...
+            remove it from <code>operations</code>.  After removal,
```
jan-ivar: How?

webrtc.html
```diff
@@ -509,33 +509,43 @@
...
+            if <code>operations</code> is non-empty, execute the steps
+            necessary to resolve the first promise
+            in <code>operations</code>.
```
jan-ivar: This is an imprecise sentence involving a promise. What executes first, functions added by content as `.then` callbacks (there can be several) or the next operation? Instead of making up our own novel language around promises, we should lean on the standard semantics [established for promises](http://www.w3.org/2001/tag/doc/promises-guide) by the Tag.

Referring to other specs whenever possible, seems an accepted practice elsewhere in this working group. I don't know why we hesitate to apply it here.

webrtc.html
pthatcherg: "The function" could refer to two different things:

1.  The steps of checking the queue and then executing code if the queue is empty, but queueing the task and return if not (they synchronous part).

2.  The steps of doing the real work and resolving the promise (the asynchronous part).

I'm saying "the function" is the first and "the steps necessary to resolve the promise are the second.

I think in your mind, the second is the function and the first is some thing layered on top of the function.    That's probably what is causing the confusion between us.

webrtc.html
```diff
@@ -509,33 +509,43 @@
...
-        completed. It is valid, and expected, for normal error handling
-        procedures to be applied.</p>
+        <code>setLocalDescription</code>, <code>createAnswer</code>,
+        <code>setRemoteDescription</code>,
+        and <code>addIceCandidate</code> executing at any given
+        time. If subsequent calls are made while one of them is still
+        executing, they are added to a queue and processed when the
+        previous operation is fully completed. It is valid, and
+        expected, for normal error handling procedures to be
+        applied.</p>
```
jan-ivar: The sentence "It is valid, and expected, for normal error handling procedures to be applied." is vague and meaningless to this implementer, which is why my PR removes it in place of specifics that actually specify how error handling MUST be applied, as dictated by established standards.

webrtc.html
pthatcherg: I reworded it to be modeled more like it was with ("stuff on top of the function").  I think that will make it more clear.  Please take another look.

webrtc.html
pthatcherg: I reworded it to be modeled more like your way of thinking ("stuff on top of the function").  I think that will make it more clear.  Please take another look.

webrtc.html
pthatcherg: Those are some confusing terms (A promise is "settled" to make it "resolved"?;  So is the verb to settle or to resolve?), but I changed it to "settled" everywhere.

webrtc.html
pthatcherg: How what?  How do you remove something from a queue?

webrtc.html
pthatcherg: I removed it and replace it with more precise language involving promises, which I hope you will like.

webrtc.html
pthatcherg: Finally!  You've given me a case when all this chaining stuff you were writing might actually mean something.  Do you think you could create an example where this would matter, perhaps something like this?

console.log(pc.signalingState);
pc.setLocalDescription.then(=> console.log("set local"));
console.log(pc.signalingState);
pc.setRemoteDescription.then(=> console.log("set remote"));

That should print:
stable


This is exactly the problem with your PR 307, and relying 

  Since there's no way I'm going to actually implement PeerConnection's methods in terms of promise chaining, instead I'll have to go read the description and try to figure out what it means and then implement how I think it sounds.  But how it sounds does not convey this incredibly subtle meaning.  Which means the implementation wouldn't actually match the spec, because the spec was written with very subtle behavior expectations that would only make sense if you wrote it that on specific way.  In other words, by describing exactly how to do something rather than what to do, you made it so specific that unless the implementor followed your exact implementation, they would never

webrtc.html
pthatcherg: Ah wait, I see now that being settled doesn't necessarily  mean being resolved, which is all kind of confusing.    The more I learn about promises, the less I like them.

webrtc.html
pthatcherg: Finally! You've given me a case when all this chaining stuff you were writing might actually mean something.  Can you come up with an example where it actually matters?  Here's the best I could come up with:

pc.createOffer.then(offer => if (pc.signalingState == 'stable') pc.setLocalDescription(offer));
// Message comes in from the remote side which triggers this:
signaling.onoffer = offer => pc.setRemoteDescription(offer).then(=> pc.createAnswer(offer). then(answer => pc.setLocalDescription(answer)));

Does that set the local description first or the remote description first?  I think it should set the remote description first.  But I think according to your chaining rule, it would set the local description first.  

But I could be wrong.  The combination of your description of the chaining mixed with the rules of promising chaining make that distinction very subtle.  And that's one of the reasons I don't like PR 307.  It's so subtle and complex that unless someone implements it with exactly the setup you describe (which is rather particular to Firefox), it probably won't be correct.  And good luck to the poor javascript developer that has to understand what's supposed to happen.

webrtc.html
jan-ivar: A promise is either *settled* (*fulfilled* or *rejected*) or *pending* (neither *fulfilled* nor *rejected*).

`resolve` or `reject` is something the asynchronous code does to the promise (confusingly, both make it `resolved`). Because of chaining, a promise can be *resolved* and *pending* at the same time (e.g. it has had `resolve(p2)` called on it, where `p2` is another pending promise).

webrtc.html
jan-ivar: I mean you're not spelling out how an implementation should implement "When". There's [precise language available for that when dealing with promises](http://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting), that basically boil down to: `p.then(func, func)`. What you have here is imprecise and doesn't nail down timings behaviors, which can be subtle.

pthatcherg: Try the new text I just wrote.  It's even more precise (and now clearly different than #307).

pthatcherg:
I made the "queue or promises" even more precise in regards to when the queue executes compared to the promise handlers.  The good news is that we can choose which order we want, and it's very clear (because the steps are in order in the text).  The bad news is... promises... I'm not sure they are worth all the complexity involved.

webrtc.html
jan-ivar: It's not specific enough, so I can't call you on your claims that it works. Show me some code that you claim comes from this, and I can critique it. - The existing old wording hasn't worked in years. It was conceived back when all the methods returned void and basically no-one cared that error handling was totally busted.

webrtc.html
jan-ivar: >  It's so subtle and complex that unless someone implements it with exactly the setup you describe (which is rather particular to Firefox), it probably won't be correct.

Exactly! And with https://github.com/w3c/webrtc-pc/pull/307 browsers MUST implement it that way. That's the benefit of precision. Being vague here risks introducing timing differences in implementations, which I see no excuse for.

What's particular to Firefox?

jan-ivar: JavaScript developers have a much better shot if we follow the standard than if we reinvent things.

pthatcherg: The original problem of "only do one of these things at a time" being implemented with promises chaining is unique to Firefox.  It is not implemented that way in Chrome, and probably never will be.  

And precision is not an absolute good.  You could just copy an paste the Firefox code into the spec and that would be precise, but it wouldn't be good.  And in a way, that's what you're doing with #307.

I disagree that this is good for javascript developers.   


If we can come up with text that expresses sufficient precision and the behavior we want (I think #307 is NOT the behavior we want) with more simple text, we should go for more simple text.

webrtc.html
```diff
@@ -509,33 +509,47 @@
...
+              <code>InvalidStateError</code> exception and abort these steps.</p>
+          </li>
+
+          <li>
+            <p>If <code>operations</code> is empty, execute the
+              function and let <var>p</var> be the promise returned by
+              the function.  If <code>operations</code> is non-empty,
+              let <var>p</var> be a promise equivalent to the promise
+              that would be returned by calling the function (it will
+              be settled at the same time and with the same result).
```
jan-ivar: How can I make two promises be "equivalent" and "settled at the same time" in JavaScript? That doesn't sound implementable. Do you mean one that settles *after* the other and with the same result?

Even then, how am I supposed to do that? The spec should tell me, not leave it up to me, or we wont have the same  deterministic timing behavior across browsers.

If you could show me the code you have in mind for this (akin to https://github.com/w3c/webrtc-pc/pull/307#issuecomment-142479604), then that would help me, and we could work backwards from that to figure out how to describe it in a way we can get agreement on.

webrtc.html
```diff
@@ -509,33 +509,47 @@
...
+
+          <li>
+            <p>Add <code>p</code> to <var>operations</var>.</p>
           </li>
 
           <li>
-            <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            <p>When a promise in <code>operations</code> is settled,
```
jan-ivar: Again, I think the "when" needs to be replaced with steps that the dictate specific steps around promises that must be taken. I've found that people care about such timing behaviors being nailed down.

E.g. that's why a lot of input validation was recently moved to the synchronous parts of methods in media capture (people can inspect a promise returned from a function in the browser's JS debugger and tell that input validation failed from the fact that the promise already shows as rejected).

webrtc.html
jan-ivar: The argument that JavaScript developers benefit from simple implementations, I think is a red herring here. JavaScript Developers benefit from cross-browser deterministic behavior.

Simplicity for implementers is low in the [priority of constituencies](http://www.w3.org/TR/html-design-principles/#priority-of-constituencies).

> signaling.onoffer = offer => pc.setRemoteDescription(offer).then(=> pc.createAnswer(offer). then(answer => pc.setLocalDescription(answer)));

Btw, I think your example is non-deterministic, because it seems like a genuine race between `signaling.onoffer` and createOffer finishing.

Still, I think once one realizes that every `p.then(func)` indirection amounts to another (guaranteed) spin of the microtask queue, then one quickly becomes aware of timing issues like "what order will a and b run in given `p1.then(() => {}).then(a)` and `p1.then(b)`.

The answer may not always need to be obvious, but it better be deterministic, and hopefully across browsers, or debugging just got a lot harder.

webrtc.html
pthatcherg: Only one of the promises is seen by JS, so they don't have to be at the same time "in Javascript".  The one returned is simply a proxy for the one that doesn't exist yet.

It doesn't matter how you implement it as long as the order of the events is the same.  You can use promise chains if you want, and another browser will use a queue, or maybe something else entirely.  

I don't have time to write up example code right now.  Maybe later.

webrtc.html
pthatcherg: I disagree.  We should try our best to describe the behavior we want, not the exact way to implement it, especially when that exact way is so complicated no one can understand it.

webrtc.html
pthatcherg: I didn't say simple implementation.  I said a simple spec text, meaning you can read it and actually understand how the browser is supposed to behave.  

If you don't think this PR is deterministic enough, then please provide a example in which it wouldn't be.  Because I believe in the latest version, I have addressed all of our concerns you have expressed previously.

webrtc.html
jan-ivar: Great, looking forward to it. My point about "equivalent" and "settled at the same time", was that they seem like unnecessary requirements given the problem space. Promises have well-defined semantics already (even in c++ implementations), so we should be able to express our needs using them, without inventing new semantics.

webrtc.html
pthatcherg: In fact, I think this PR is already to prescriptive.  I'd like to make it less so, not more so.

jan-ivar: I urge you to read http://www.w3.org/2001/tag/doc/promises-guide

pthatcherg: By the way, I just realized the last push I did to git failed, so you weren't seeing the latest version, which is more precise, so take a look at the very latest.  You might like it more.

jan-ivar:
To compare, here's example code for the two PRs: https://jsfiddle.net/ezchqdcu/

pthatcherg:
Jan-Ivar asked for an implementation of the latest version of this PR in Javascript, so here it is, with a little test code at the bottom to show it working.  Someone who knows promises better could probably make it shorter, but this works.  Note two things:
1.  This is only a proof of concept, and only applicable if you're writing PeerConnection in JS.  Implementations of PeerConnection in C++ that I'm familiar with most likely wouldn't do it like this.  Most of the complexity here is just making Javascript Promises work how we want.
2.  It makes it easy to control whether a queue operation should be run before or after a completed operation's callback.  See the comment below.

```javascript
var pc = {
    operations: []
};
pc.run = function (operation) {
    var op = {};
    op.p = new Promise(resolve => {
        op.resolve = resolve;
    })
    op.execute = function () {
        op.result = operation();
        op.result.then(op.dequeAndResolve, op.dequeAndResolve);
    };
    op.dequeAndResolve = function () {
        pc.operations.shift();
        if (pc.operations.length > 0) {
            pc.operations[0].execute();
        }
        // If you want the returned promise to be settled before queued operations are executed, 
        // move this line up above the previous block.
        op.resolve(op.result);
    };
    pc.operations.push(op);
    if (pc.operations.length == 1) {
        op.execute();
    }
    return op.p;
};

pc.createOffer = () => pc.run(() => {
    console.log("Create offer.");
    return new Promise(resolve => setTimeout(resolve("An offer", 2000)));
});

pc.createAnswer = () => pc.run(() => {
    console.log("Create answer.");
    return new Promise((_, reject) => setTimeout(reject("Error creating answer", 1000)));
});

// Should print 
// Create offer
// Create answer
// An offer
// Error creating answer
var log = msg => console.log(msg);
pc.createOffer().then(log, null);
pc.createAnswer().then(null, log);
```

jan-ivar:
Looks like we had the same idea. :) I found one optimization which is that you don't need to push *p*.

In my link above I compare it to my PR, since you asked about differences. I note a difference in sequencing of errors.

Also, I still have concerns about the prose around promises being vague and non-Tag approved.

pthatcherg:
Here's my code in your JS Fiddle example, with an option to see "settle first" and "deque first"

https://jsfiddle.net/5xd1geb3/

pthatcherg:
I have updated this PR to have a much smaller change to the existing text while still maintaining the behavior that we want.  Here is some updated proof of concept code that reflects how the PR is now written:

```javascript
var pc3 = {
    operations: []
};
pc3.run = function (operation) {
    var operations = pc3.operations;
    var p = new Promise(function (resolve, reject) {
      var executeThenRunNextBeforeSettling = function() {
        operation().then(
          function (result) {
            operations.shift();
            if (operations.length > 0) {
             operations[0]();
            } 
            resolve(result);
          },
          function (err) {
            operations.shift();
            if (operations.length > 0) {
             operations[0]();
            } 
            reject(err);
          });
      };
      operations.push(executeThenRunNextBeforeSettling);
      if (operations.length == 1) {
        operations[0]();
      }
    });
    return p;
}
```

jan-ivar:
> Here's my code in your JS Fiddle example, with an option to see "settle first" and "deque first"
>
> https://jsfiddle.net/5xd1geb3/

The "dequeue first" case in this fiddle defers the initial createOffer, violating our spring consensus.

>  Here is some updated proof of concept code that reflects how the PR is now written:

This one is still "settle first", executing setLocal's error after setRemote's synchronous processing steps, which to me seems surprising at least, maybe wrong.

jan-ivar:
> Here is some updated proof of concept code that reflects how the PR is now written:

Here's an iteration on it (taking advantage of multiple `.then`s on the same promise executing in order):
```js
var runOperationA = operation => new Promise(function (resolve, reject) {
  var ready = () => (operations.shift(), operations.length && operations[0]());
  operations.push(() => {
    var p = operation();
    p.then(ready, ready);
    p.then(resolve, reject);
  });
  if (operations.length == 1) {
    operations[0]();
  }
});
```
Look how close we are:
```js
var runOperationB = operation => {
  var ready = () => operationsLength--;
  operationsLength++;
  var p = (operationsLength == 1)? operation() : operations.then(operation);
  operations = p.then(ready, ready);
  return p;
}
```
Comparing them here: https://jsfiddle.net/vrwofnmk/

jan-ivar:
> (taking advantage of multiple .thens on the same promise executing in order)

Clarification: they both chain off *p*, not off each-other. Multiples branch the chain so to speak.

jan-ivar:
Also, yours still has the error-out-of-order problem.

pthatcherg:
Can you provide a link to what the consensus was, or where the discussion
happened?  I'd like a little more detail/clarity on what the consensus was.

webrtc.html
```diff
@@ -517,6 +517,10 @@
 
         <ol>
           <li>
+            <p>Let <var>p</var> be a new promise</p>.
+          </li>
+
+          <li>
             <p>Append an object representing the current call being handled
```
jan-ivar: s/an object/a function/, because you execute it further down, and objects can't be executed.

webrtc.html
```diff
@@ -524,26 +528,31 @@
 
           <li>
             <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            execute the function from the front of the queue.</p>
+          </li>
+
+          <li>
+            <p>When the promise returned by the function is settled,
```
jan-ivar: "Upon fulfillment or rejection of the promise returned by the function," - [§ 5.3. Reacting to Promises](http://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting).

webrtc.html
```diff
@@ -524,26 +528,31 @@
...
+            remove the corresponding object from the
```
jan-ivar: remove the function from the

webrtc.html
```diff
@@ -524,26 +528,31 @@
...
+            <code>operations</code> array. After removal, if the array
+            is non-empty, execute the first object queued.  After
```
jan-ivar: first function queued.

pthatcherg:
That's a good idea using 2 .thens.  Here's an updated version of my code doing that.  But it's more readable, in my opinion, than yours.  It also shows that it's very easy to switch between "settle first" and "deque first", which ever the consensus is.

```javascript
var pc = {
    operations: []
};

pc.settleFirst = false;
pc.run = function (operation) {
  var operations = pc.operations;
  var runNext = function() {
    operations.shift();
    if (operations.length > 0) {
      operations[0]();
    } 
  };
      
  return new Promise(function (resolve, reject) {
    operations.push(function() {
      var p = operation();
      if (pc.settleFirst) {
        p.then(runNext, runNext)
        p.then(resolve, reject);
      } else {
        p.then(resolve, reject)
        p.then(runNext, runNext);
      }
    });
    if (operations.length == 1) {
      operations[0]();
    }
  });
};
```

webrtc.html
pthatcherg: Of course objects can be executed:  obj.execute().  That's what I had in one of my previous versions of the proof-of-concept code.  The fact that it's a function is an implementation detail.

webrtc.html
pthatcherg: Sounds good.

webrtc.html
pthatcherg: Again, implementation detail.

webrtc.html
pthatcherg: Again, implementation detail.

pthatcherg:
In general, I'm glad that we have found (with perhaps the minor detail of the error handling order), that these are behaviorally equivalent.  And I believe #314 is much easier to read and understand than #307 while achieving the same goals.   So, I think we only have two questions to answer:

1.  Do we do "deque first" or "settle first"?  In the only real-world example I have been able to think of where it matters, deque first made more sense.  But you keep referring to "spring consensus".  Fortunately, with the text in #314, it's very easy to switch.  You simply reorder two of the steps.

2.  Does the minor detail of the error handling order really matter?  I think it doesn't.  But I could be convinced otherwise if we have a real-world code example where it actually matters.  If we can come up with an example where it really matters, then we could attempt to fix #314.   But as I said, I don't think it matters.

webrtc.html
```diff
@@ -524,26 +528,31 @@
...
+            execution, settle <code>p</code> with the result from the
+            promise returned by the function.</p>
```
jan-ivar: This is where we need to decide about the order of callbacks.

That aside, by the § as above, this should read:

"Upon fulfillment of the promise returned by the function with value *v*, resolve *p* with value *v*.
Upon rejection of the promise returned by the function with reason *r*, reject *p* with *r*."

But that's a bit thick. Maybe if we said "with the result or rejection" it would be fine.

webrtc.html
jan-ivar: Where is obj.execute() defined?

Also, you say "execute the function from the front of the queue" below, which seems inconsistent. An object goes on the queue and a function comes off?

webrtc.html
jan-ivar: lets be consistent

webrtc.html
pthatcherg: Yes, this is where I mentioned we can re-order the two steps based on what the consensus is on "deque first" vs. "settle first".

As for the promises text, how about this?

"fulfill or reject p with the value or reason from the promise returned by the function"

webrtc.html
pthatcherg: The way to execute the object (obj.execute) is an implementation detail.

pthatcherg: As for "execute the function", that's what was already there.  All I did was remove the word "asynchronously".  Nevertheless, I will change it to "object" to make it more consistent.

webrtc.html
jan-ivar: or maybe "Upon fulfillment or rejection of the promise returned by the function, settle *p* correspondingly."

jan-ivar:
I observe two distinct problems in your "dequeue first":
 1. Methods get queued when nothing is queued (against [spring consensus](https://lists.w3.org/Archives/Public/public-webrtc/2015May/0046.html)). 
 1. Success/error callbacks come after subsequent queued action is fired off (not discussed)

I suspect order matters on both of them. For errors, I would expect errors to happen in the same order in the operations chain as they happen when using the peerConnection promise api.

adam-be:
How is the battle of the algorithms going? :)

This is my understanding (correct me if I'm wrong):
We are converging and using promises seems like the way to go. But on one side we have a promise chain, "linked" by then()'s, with the length kept in a variable on the side. On the other side we have an array of promises with the length "built in".

If we go with the chain, I think we need to document that particular behaviour a bit more. I was a bit confused that the variable 'operations' is a promise and not a list. Now I think of it like 'headOfChain'.

jan-ivar:
@adam-be note that with chains e.g. `var p = foo().then(bar);` *p* points to the promise returned by *bar* not foo, so it's the tail, not the head.

There's no dictate on naming, but I like: someStateIWant = reachSaidState(). e.g."loaded" [from the promise guide](http://www.w3.org/2001/tag/doc/promises-guide#one-time-events). That way I can write: `var loaded = load();` , `loaded.then(resource => { /* use resource */ });` or `await loaded;`.

*operationsDone* maybe.

jan-ivar:
> On the other side we have an array of promises with the length "built in".

I prefer to think of the array as redundant, because promises chain inherently. That's what I meant to illustrate with the comparison [above](https://github.com/w3c/webrtc-pc/pull/314#issuecomment-142966141).

jan-ivar:
@adam-be We also need to decide whether success/error callbacks should happen before or after the  synchronous processing steps of a subsequent operation. Leaving that up to implementations seems like a bad idea.

martinthomson:
I think that the only consideration is that the callbacks from the first operation be called before the callbacks from the second, even if the second operation completes immediately (it will never complete synchronously, but on a microtask).  I think that we can guarantee that with something as simple as:

```
p1 = pc.first().then(success1, error1);
p2 = pc.second().then(success2, error2);
// where pc.second is:
pc.second = _ => Promise.resolve();
```
Even in this case, the first set of callbacks run before `pc.second()` because you get:
```
p1 = pc.first().then(success1, error1);
p2 = pc.first().then(_ => Promise.resolve()).then(success2, error2);
```

pthatcherg:
Actually, the latest version of my PR (#314) is no longer an "array of
promises".  I reworked it to minimize the difference between it and the
current spec text.  So it's still the same "queue of functions" that is in
the current spec.  The only difference from the current spec is that now
there is a Promise returned that is a proxy for the promise that will
return from the queued function when it is finally run and the promise it
returns is fulfilled or rejected.  So a better way of describing the
difference between PR #307 and PR #314 is "chain of promises with a queue
length" vs "queue of functions with a proxy promise".

That said, the two are converging in the sense that they represent (almost)
the exact same behavior, meaning the choice to make between them is to pick
the one that is more readable and understandable.  Obviously, I've of the
opinion that #314 is more readable and understandable.


But no matter which is chosen (#307, #314, or some other idea), there is
still one question to answer: when it's time to fulfill or reject a promise
returned by one of the functions, do we deque and run the next function
before or after?  On that question, the opinions so far are these:

- Me:  I can't find a real world use case where it matters, but the closest
example I could come up with lead me to believe it should be "before".
- Jan-Ivar:  Spring consensus (
https://lists.w3.org/Archives/Public/public-webrtc/2015May/0046.html)
indicates it should be "after".  (However, I think that consensus is
related to the relationship between addTrack and createOffer, which is a
separate matter not related to this one, so I don't think it really
indicates one way or the other)
- Martin: Either (as long as the callbacks will fire in the correct order
even if the functions fulfill their promises immediately, which is true for
both "before" and "after").

The remaining question may be much ado about nothing, but we have to pick
one.  Luckily PR #314 makes it very easy to pick one way or the other :)

jan-ivar:
Peter, spring consensus doesn't affect callback vs. start of subsequent operation IMHO. I separate the two issues in https://github.com/w3c/webrtc-pc/pull/314#issuecomment-143014501 for that reason (your "dequeue first" solution violates both).

>  - Me: I can't find a real world use case where it matters, but the closest example I could come up with lead me to believe it should be "before".

Care to share? It would have to involve some state that affects the synchronous steps, and the only one seems to be `closed`, as the other steps are all input validation (nevermind broken `addIceCandidate`).

So here's an example:

    pc.setRemoteDescription(offer).catch(() => pc.close());
    pc.createAnswer().catch(() => { /* called? */ };

Is the `createAnswer` failure callback called or not?

The spec [says](http://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-createAnswer-Promise-RTCSessionDescription--RTCAnswerOptions-options) "If this RTCPeerConnection object is closed before the SDP generation process completes, the USER agent MUST suppress the result and not resolve or reject the returned promise."

Unfortunately, the spec lacks processing steps for `createAnswer` which has left both Chrome and Firefox throwing `InvalidStateError` on `closed` on entry, like it does for all the other methods:

> Uncaught InvalidStateError: Failed to execute 'createOffer' on 'RTCPeerConnection': The RTCPeerConnection's signalingState is 'closed'.

So right now it's:
 1. not even clear if `pc.close(); pc.createAnswer().catch(() => { /* called? */ };` is called,
 2. But if # 1 is called, whether it's called in my queued example earlier, depends on our timing question.

martinthomson:
FWIW, I think that the text recommending that the promise be left hanging should be removed.  A reject is far superior.

But I'm unclear on how that helps resolve the issue of ordering.  Can't we address that in a separate issue?

jan-ivar:
@martinthomson It might be the only observable difference. Are we happy to say the order doesn't matter? (not sure I like the change you suggest, but agree we should discuss it elsewhere.)

Note that a related issue (improvement?) with both PRs that just occurred to me is that they check for closed state on entry, even for `createOffer` and `createAnswer` which have no such language in the spec today (even though both Chrome and Firefox check this).

webrtc.html
```diff
@@ -517,6 +517,10 @@
 
         <ol>
           <li>
+            <p>Let <var>p</var> be a new promise</p>.
```
jan-ivar: move period inside paragraph.

webrtc.html
```diff
@@ -524,26 +528,31 @@
...
+            or reason from the promise returned by the function.</p>
           </li>
 
           <li>
-            <p>When the asynchronous operation completes (either successfully
-            or with an error), remove the corresponding object from the
-            <code>operations</code> array. After removal, if the array is
-            non-empty, execute the first object queued asynchronously and
-            repeat this step on completion.</p>
+            <p>Return <code>p</code></p>
```
jan-ivar: missing period

jan-ivar:
> I think that the only consideration is that the callbacks from the first operation be called before the callbacks from the second

Since *operation* callbacks are effectively queued, this cannot happen, I'm pretty sure.

martinthomson:
If we are debating about something that isn't observable, then I'm not interested in having this discussion continue and we can flip a coin or something to decide.

Here are the questions I think that we need to resolve, which goes a little beyond the previous, since I realized that an operation might have observable effects.

Do the callbacks to *operation1* fire before the callbacks to *operation2*?  I say yes.
Do the callbacks to *operation1* fire before *operation2* itself? I say yes.

If new microtasks are created in a callback, then I think we are under no obligation with respect to the ordering of those new microtasks with respect to the pending operations.  Other than that, I would like to maintain the above as invariants.  If we can do that, then fix the text any way you please, optimizing for readability, as @pthatcherg has suggested.

jan-ivar:
> Do the callbacks to operation1 fire before operation2 itself? I say yes.

My PR satisfies this, but Peter's doesn't ("dequeue first" not being on the table for [violates spring consensus](https://github.com/w3c/webrtc-pc/pull/314#issuecomment-143014501)).

webrtc.html

webrtc.html

pthatcherg:
I still don't see how "dequeue first" violates spring consensus.  I see nothing in the spring consensus mailing list thread related to "dequeue first" vs. "settle first".  But if it makes you feel better, I have updated the PR to be "settle first", which means it it doubly clearly satisfies what Martin wants and doubly clearly doesn't violate spring consensus and we can put aside the issue of "dequeue first" vs "settle first" and just focus on which PR is more clear and readable.

jan-ivar:
Thanks. To explain, your fiddle https://jsfiddle.net/5xd1geb3/ produces:

    Settle First:         Deque First:
    here1                 here1
    createOffer           here2
    here2                 here3
    here3                 here4
    here4                 createOffer
    createOffer done      createOffer done
    ...                  ...

The Dequeue First case is bad because createOffer is not invoked synchronously, but later, which means:

    pc.addTrack(x);
    pc.createOffer().then(offer => { /* contains x and y * } );
    pc.addTrack(y);

when it should only contain *x*. To implement only containing *x*, createOffer MUST execute synchronously, so it can snapshot the current state (x, not y).

Of course none of the browsers comply yet, which doesn't help discussion. Basically, createOffer will need to put this state (the tracks) into the object it queues.

jan-ivar:
Peter, Martin wants it to be that "the callbacks to operation1 fire before operation2 itself", but neither version in your fiddle https://jsfiddle.net/5xd1geb3/ appears to accomplish this. I get:

    Settle First:                          Deque First:
    here1                                  here1
    createOffer                            here2
    here2                                  here3
    here3                                  here4
    here4                                  createOffer
    createOffer done                       createOffer done
    setLocal                               setLocal
    setLocal done                          setLocal done
    setRemote                              setRemote
    ReferenceError: bogus is not defined   ReferenceError: bogus is not defined
    setRemote done                         setRemote done

The ReferenceError is from `setLocal`, and comes after `setRemote` has launched in both cases.

Compare to my PR (`B` in https://jsfiddle.net/vrwofnmk/ ) where it comes before `setRemote`.

pthatcherg:
That can be fixed with a small tweak to the JS proof-of-concept.  Now the proof-of-concept for PR 314 runs exactly the same as the one for PR 307.

```javascript
var pc = {
    operations: []
};

pc.run = function (operation) {
  var operations = pc.operations;
  var runNext = function() {
    operations.shift();
    if (operations.length > 0) {
      operations[0]();
    } 
  };

  return new Promise(function (resolve, reject) {
    operations.push(function() {
      operation().then(resolve, reject).then(runNext, runNext);
    });
    if (operations.length == 1) {
      operations[0]();
    }
  });
};
```

jan-ivar:
Looks good. If we agree this is desired behavior then should we update this PR?

pthatcherg:
In what way do you think it needs updated?  Specific text for how to say "fulfill or reject the promises all the way through all its callbacks" in the proper promises lingo is welcome if you think it's necessary.

jan-ivar:
I don't know about "all the way through"  but nailing down the order we just discussed takes words.

Since `.then(resolve, reject)` falls into the [promise constructor anti-pattern](http://stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it), the [promise guide](http://www.w3.org/2001/tag/doc/promises-guide) has no great suggestions for language here.

Maybe if we broke the step into two:

* Upon fulfillment or rejection of the promise returned by the function, fulfill or reject *p* with the corresponding value or reason.
*  Upon fulfillment of the previous step (not *p*), remove the corresponding object from the *operations* array. After removal, if the array is non-empty, execute the first object queued.

pthatcherg:
Actually, the more I think about it, the more I think we don't need any additional text.  I think the text as-is is sufficient.  What I had was really just a bug in the proof-of-concept.  But the proof-of-concept isn't the spec and doesn't go in the spec.  The spec as-is seems sufficiently clear to me, and I don't think we need to break it up into separate steps.  

And if you don't like the proof-of-concept because of some "anti-pattern", you can write your own non-anti-pattern JS for your implementation.  It's just a proof-of-concept.

jan-ivar:
When nothing in the spec prevents you from writing it wrong, then by definition it's insufficient.

jan-ivar:
I don't think there's a non-anti-pattern way of implementing your PR unfortunately, since it reinvents chaining with an array.

martinthomson:
I think that the current text is fine, however I do find that the "next operation" step is a little big and could be split into not two, but three sub steps.  @jan-ivar's suggestion is perhaps a little overwrought though, I would just split on the sentence boundaries.

pthatcherg:
I have broken the big step into 3 smaller steps, per Martin's suggestion.

martinthomson:
OK, @jan-ivar just explained the problem to me.  You do three things after an operation completes:
1. fulfill *p*
2. pop *operation* from the queue
3. call the next operation

The concern is that if you do this all at once, then what happens is that the next operation starts running *before* the handlers/callbacks are called, because those are called on the next microtask cycle as a result of step 1.  You need to yield in order to ensure that the callbacks are run first.

You can do that anywhere, but there are some risks.

If you yield before step 2 then there is a risk that the code will run operations twice.  If the current microtask cycle or any microtask that is enqueued before the new microtask you create includes another call to an enqueued function, then the operation that is left on the front of the queue will be invoked again.

If you yield before step 3, then there is also a risk that you run operations twice.  In this case, if the callbacks (or anything between the callbacks and your operation microtask, which should be nothing), invoke an operation that hits the queue again, the head of the queue will be invoked again.

An easy fix is to mark the head of queue as "enqueued" and not execute it if it is marked thus.

pthatcherg:
In reply to "When nothing in the spec prevents you from writing it wrong, then by definition it's insufficient.":  It sounds like you're saying that the spec isn't sufficient until it's there is no possible way for an implementation to have a bug.  That seems like an impossible goal. 

In reply to "I don't think there's a non-anti-pattern way of implementing your PR unfortunately": I think you're making the assumption that an implementation will be implemented in Javascript.  That may be true of one implementation, but that is not true of all implementations.  Clearly a C++ implementation will not run afoul of some Javascript-specific anti-pattern.

pthatcherg:
Whoops.  Wrong button.

pthatcherg:
I see how yielding between steps 2 and 3 can cause a function to be executed twice, but I don't see how yielding between steps 1 and 2 can cause a function to be executed twice.  Can you write a more specific example that would make it more clear to me?

By the way, here's the example I came up with that's a problem for yielding between steps 2 and 3:

pc.f1().then(() => pc.f2());

That would cause f2 to be executed twice if an implementation chose to yield between steps 2 and 3.  But, again, I don't see a problem if yielding occurs between steps 1 and 2.

jan-ivar:
> If you yield before step 2 then there is a risk that the code will run operations twice. If the current microtask cycle or any microtask that is enqueued before the new microtask you create includes another call to an enqueued function, then the operation that is left on the front of the queue will be invoked again.

I agree with Peter, wont anyone enqueueing a function in this window just come in, push their object, see that length is > 1, and leave?

martinthomson:
No, both of you are correct, that means that the yield before step 2 fixes the problem.  (I had the test as `queue.empty()` in my mind rather than `queue.size() == 1`).

martinthomson:
That leaves this with the one remaining bug.  Or are we disagreeing that it is a bug?

pthatcherg:
What's the one remaining bug?

pthatcherg:
By the way, another thought I had about Jan-Ivar's comment saying "I don't think there's a non-anti-pattern way of implementing your PR unfortunately":  since this PR is behaviorally identical to #307, then wouldn't the implementation of #307 also be an implementation of this PR?  In other words, wouldn't the Firefox implementation have to do nothing to match this PR, since it already does?

martinthomson:
The bug is the one that I explained.  As described, the following code:

```
pc.setRemoteDescription(desc).then(_ => doSomethingWith(pc));
pc.addIceCandidate(candidate);
```

Calls `addIceCandidate()` before `doSomethingWith(pc)` is called because fulfillment of the `setRemoteDescription()` promise causes the `.then()` arguments to be called on the next microtask, whereas `addIceCandidate()` is run synchronously.

This doesn't matter, unless `addIceCandidate()` has observable synchronous behaviour.  Whether it does depends on how it is implemented, but Firefox is actually entirely synchronous when it comes to managing the lists of candidates that it manages.

jan-ivar:
https://github.com/w3c/webrtc-pc/pull/307 orders the callbacks [correctly](https://jsfiddle.net/ezchqdcu/), in prose and execution. This PR does not (its remaining bug).

pthatcherg:
Jan-Iver, the proof-of-concept code I posted earlier works.  If you change your fiddle to this:

```javascript
var runOperationA = function (operation) {
  var runNext = function() {
    operations.shift();
    if (operations.length > 0) {
      operations[0]();
    } 
  };

  return new Promise(function (resolve, reject) {
    operations.push(function() {
      operation().then(resolve, reject).then(runNext, runNext);
    });
    if (operations.length == 1) {
      operations[0]();
    }
  });
};
```

Then it's the same.

jan-ivar:
Agreed. We need to change the prose to make the other implementation non-compliant.

martinthomson:
This isn't about the code.  The code is fine because there is a dispatch/yield/new microtask between the resolve step and the call to `runNext`:

```
      operation().then(resolve, reject).then(runNext, runNext);
```

The problem is that the text doesn't say that.

pthatcherg:
Martin, according to the current text of this PR, and the current proof-of-concept code I just posted, addIceCandidate() will be called after doSomethingWith(pc), not before.

martinthomson:
@pthatcherg, I disagree.  If the following are executed synchronously, then it will not:

>              <li><p>Fulfill or reject <code>p</code> with the
                corresponding value or reason.</p></li>
              <li><p>Remove the corresponding object from
                the <code>operations</code> array.</p></li>
              <li><p>If the array is non-empty, execute the first
                object queued.</p></li> 

Fulfilling *p* (this should be `<var>p</var>` btw) causes the callback to happen on the *next* microtask.  The current prose causes the *current* microtask to run the next operation.

The reason that your code is OK is that it has a `.then` between calling `resolve()` and calling `operations[0]()`.

pthatcherg:
It sounds like you'd like to have something in the text that says "The callbacks of <code>p</code> must be executed before the next queued operation is executed."  Would that be sufficient?

martinthomson:
That would achieve the goal I guess.  Or you could embrace the microtask paradigm and say that steps 2 and 3 are part of a new microtask.

jan-ivar:
I think we need to do both, otherwise the steps say one thing and this another, confusing implementers about how to achieve this (it's complicated to get right after all).

martinthomson:
How about:

>             <p>Upon fulfillment or rejection of the promise returned
            by the function, fulfill or reject <var>p</var> with the
                corresponding value or reason and execute the following steps:</p>
            <ol>
              <li><p>Remove the corresponding object from
                the <code>operations</code> array.</p></li>
              <li><p>If the array is non-empty, execute the first
                object queued.</p></li>

pthatcherg:
Martin, I think I'd be happy with that text, but I don't really see how it's different.

martinthomson:
My bad, it isn't.  I meant to add one thing...

>         <p>Upon fulfillment or rejection of the promise returned
        by the function, fulfill or reject <var>p</var> with the
            corresponding value or reason and execute the following steps **in parallel**:</p>
        <ol>
          <li><p>Remove the corresponding object from
            the <code>operations</code> array.</p></li>
          <li><p>If the array is non-empty, execute the first
            object queued.</p></li>

pthatcherg:
Execute the 2 following steps in parallel with each other or the two following steps in parallel with the fulfilling/rejecting of the promise?

martinthomson:
Oh, this is an established pattern, ... specs usually reference https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel when they use the phrase "in parallel"

jan-ivar:
I don't think that's right. To imply a microtask, I think you need the "Upon filfillment" language in the promise-guide, which is direct code for 'p.then', like I suggested in https://github.com/w3c/webrtc-pc/pull/314#issuecomment-144440627

martinthomson:
Yeah, that would be more precise.  So maybe:

> Upon fulfillment or rejection of the promise returned
    by the function, fulfill or reject <var>p</var> with the
        corresponding value or reason.  Then, upon fulfillment of a freshly resolved promise, run the following steps:

jan-ivar:
That would work.

pthatcherg:
What's a "freshly resolved promise"?

jan-ivar:
`Promise.resolve()`

jan-ivar:
From http://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting : "Upon fulfillment of p with value v is shorthand for calling p.then(onFulfilled)"

pthatcherg:
I added text that says to do steps 2 and 3 "Upon fulfillment of p".  Is that sufficient?

martinthomson:
That should work, but you need to include "or rejection" in this case.  By that text *p* rejects because the operation failed, then the code won't run.

pthatcherg:
I added "or rejection of p".  Now is it sufficient?

jan-ivar:
:+1:

martinthomson:
WFM

webrtc.html
```diff
@@ -524,26 +528,36 @@
...
           <li>
             <p>If the length of the <code>operations</code> array is exactly 1,
-            execute the function from the front of the queue
-            asynchronously.</p>
+            execute the object from the front of the queue.</p>
+          </li>
+
+          <li>
+            <p>Upon fulfillment or rejection of the promise returned
+            by the function, fulfill or reject <code>p</code> with the
```
jan-ivar: nit: s/code/var/

webrtc.html
```diff
@@ -524,26 +528,36 @@
...
+            corresponding value or reason.  Upon fulfillment or
+            rejection of <code>p</code>, execute the following
```
jan-ivar: s/code/var/

webrtc.html
```diff
@@ -524,26 +528,36 @@
...
+            </ol>
           </li>
 
           <li>
-            <p>When the asynchronous operation completes (either successfully
-            or with an error), remove the corresponding object from the
-            <code>operations</code> array. After removal, if the array is
-            non-empty, execute the first object queued asynchronously and
-            repeat this step on completion.</p>
+            <p>Return <code>p</code>.</p>
```
jan-ivar: s/code/var/

pthatcherg: All the <code>s have been changed to <var>s.

martinthomson:
@pthatcherg, do you want to drop a message onto the list to let them know that we've agreed on the contents of the PR and that we'd like it to be merged now please?

adam-be:
great news :)

stefhak:
:+1:

PR #318, opened 2015-09-25
Update to PR 300

aboba:
Some fixes to PR 300 to fix typos, fix inconsistencies in the time to keep entries (10 seconds vs. 1 second) and clarify when dictionary entries are updated.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3378,36 +3378,38 @@ sender.setParameters(params)
 
           <dt> sequence&lt;RTCRtpContributingSource&gt;
             getContributingSources ()  </dt>
-          <dd> <p>Returns a RTCRtpContributingSource for each unique
+          <dd> <p>Returns an <code><a>RTCRtpContributingSource</a></code> for each unique
               CSRC or SSRC received by this RTCRtpReceiver in the last 10
-            seconds. </p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3378,36 +3378,38 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
 
 	 <dl class="idl" title="interface RTCRtpContributingSource">
 
-	   <dt> readonly attribute DOMHighResTimeStamp timestamp</dt> <dd> Time of
-	   reception of the most recent RTP packet containing the contributing
-	   source. </dd>
+	   <dt> readonly attribute DOMHighResTimeStamp timestamp</dt> 
+	   <dd> <p>The timestamp of type DOMHighResTimeStamp [HIGHRES-TIME], indicating the time of 
+	   reception of the most recent RTP packet containing the source.  
+	   The timestamp corresponds to the local clock.</p> </dd>
```
martinthomson: This last sentence might be confusing.

alvestrand: Actually that's not what I interpreted it as, nor is it what was my "alternative C" on the list - that was the time of *reception* of the packet, not the receiver's interpretation of the time it was sent.

PR #325, opened 2015-10-09
Adding additional members to RTCIceCandidate dictionary.

taylor-b:
These attributes aren't used in addIceCandidate; they're just provided
in candidates signaled by onicecandidate, or retrieved from an
RTCIceCandidatePair. This will allow interested applications to get more
information about an ICE candidate, without parsing the candidate
string.

This is the change Peter suggested, which we agreed to at the Redmond f2f.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2429,10 +2429,7 @@
           <h4>The RTCIceTcpCandidateType</h4>
 
           <p>The <dfn>RTCIceTcpCandidateType</dfn> represents the type of the ICE TCP
-          candidate, as defined in [[!RFC6544]]. Browsers MUST gather <code>active</code>
-          TCP candidates and only <code>active</code> TCP candidates. Servers and other
-          endpoints MAY gather <code>active</code>, <code>passive</code> or <code>so</code>
-          candidates.</p>
+          candidate, as defined in [[!RFC6544]].</p>
 
           <dl class="idl" title="enum RTCIceTcpCandidateType">
             <dt>active</dt>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1271,7 +1271,9 @@
             <code>none</code>. This call will result in a change to the
             connection state of the ICE Agent, and may result in a change to
             media state if it results in different connectivity being
-            established.</p>
+            established. Note that the only members of the candidate attribute
+            used by this method are <var>candidate</var>, <var>sdpMid</var> and
+            <var>sdpMLineIndex</var>; the rest are ignored.</p>
 
             <ol>
               <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1271,7 +1271,9 @@
             <code>none</code>. This call will result in a change to the
             connection state of the ICE Agent, and may result in a change to
             media state if it results in different connectivity being
-            established.</p>
+            established. Note that the only members of the candidate attribute
```
pthatcherg: I think you can remove the "note that" and just say "The only ...

webrtc.html
```diff
@@ -2354,11 +2356,126 @@
...
+
+        <section>
+          <h4>The RTCIceTcpCandidateType</h4>
+
+          <p>The <dfn>RTCIceTcpCandidateType</dfn> represents the type of the ICE TCP
+          candidate, as described in [[!RFC6544]]. Browsers MUST gather <code>active</code>
+          TCP candidates and only <code>active</code> TCP candidates. Servers and other
+          endpoints MAY gather <code>active</code>, <code>passive</code> or <code>so</code>
+          candidates.</p>
+
```
pthatcherg: I don't think we need the text about gathering in this part of the spec.

webrtc.html
```diff
@@ -2354,11 +2356,126 @@
...
+            <dd>A server reflexive candidate.</dd>
+
+            <dt>prflx</dt>
+
+            <dd>A peer reflexive candidate.</dd>
+
+            <dt>relay</dt>
+
+            <dd>A relay candidate.</dd>
```
pthatcherg: This should reference [[!ICE]].

webrtc.html
taylor-b: Where should it go then? Or is this explained in another spec that can simply be referenced?

pthatcherg: I'm guessing it's in JSEP or is a TODO of JSEP.  In either case, it wouldn't be appropriate in this PR since this PR is about providing more information, not about defining behavior.

webrtc.html
```diff
@@ -1271,7 +1271,9 @@
             <code>none</code>. This call will result in a change to the
             connection state of the ICE Agent, and may result in a change to
             media state if it results in different connectivity being
-            established.</p>
+            established. The only members of the candidate attribute
+            used by this method are <var>candidate</var>, <var>sdpMid</var> and
+            <var>sdpMLineIndex</var>; the rest are ignored.</p>
```
alvestrand: If sdpMid is given and sdpMLineIndex is not, what will happen?

webrtc.html
```diff
@@ -2354,11 +2356,124 @@
...
+
+          <dt>DOMString foundation</dt>
+
+          <dd>A unique identifier that allows ICE to correlate candidates that
+          appear on multiple <code><a>RTCIceTransport</a></code>s.</dd>
+
+          <dt>unsigned long priority</dt>
+
+          <dd>The assigned priority of the candidate. This is automatically
+          populated by the browser.</dd>
```
alvestrand: Didn't we agree that all the attributes are read-only? I don't want anything to give the impression that modifying these values makes any sense at all.

webrtc.html
pthatcherg: The same thing as before this PR, I think.

webrtc.html
pthatcherg: The only place where this has any effect on behavior is in addIceCandidate, and that says it ignores these values.

martinthomson: If `addIceCandidate()` ignores these values, then they don't need to be added here.  They can be added as attributes on a new dictionary that inherits from `RTCIceCandidate`, a new type that would be an argument to onicecandidate events only.  That avoids the implication that these parameters are required, or that they might need to be consistent with the main `candidate` string.

pthatcherg: I like that idea.  What would you call this new type?

martinthomson: Jerry.

pthatcherg: Then I proposed we make RTCIceCandidate the "full" one coming from onicecandidate and RTCSignalledIceCandidate the "signalled" one that goes into addIceCandidate.

But if anyone can come up with a better pair of names, I'd welcome them.

pthatcherg: Err.... I propose :)

PR #326, opened 2015-10-12
Options to guide codec selection

martinthomson:
This is pretty primitive, but I think that it is sufficient.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3323,7 +3323,7 @@
         </dl>
 
         <dl class="idl" title="interface RTCRtpCodecCapability">
-          <dt>DOMString mimeType</dt>
+          <dt>readonly attribute DOMString mimeType</dt>
           <dd>
             <p>The codec MIME type. Valid types are listed in [IANA-RTP-2].
             </p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -3312,7 +3312,7 @@
           </dd>
         </dl>
 
-        <dl class="idl" title="dictionary RTCRtpCodecCapability">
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -359,6 +359,16 @@
...
+
+          <dt>sequence&lt;RTCRtpCodecCapability&gt; codecs</dt>
+          <dd>
+            <p>This establishes what codecs the browser is permitted to
+            negotiate.  Codec capabilities that do not appear in this list
+            cannot be negotiated.</p>
+            <p>The order of this sequence determines the preference order.
+            Codecs that appear first in this list MUST be selected by the user
+            agent in preference to those that appear after.</p>
+          </dd>
```
alvestrand: How does this work if this option is not specified?
I think the text should say that "if left off, browser can negotiate everything".

webrtc.html
```diff
@@ -3312,8 +3322,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           </dd>
         </dl>
 
-        <dl class="idl" title="dictionary RTCRtpCodecCapability">
-          <dt>DOMString mimeType</dt>
+        <dl class="idl" title="interface RTCRtpCodecCapability">
+          <dt>readonly attribute DOMString mimeType</dt>
```
alvestrand: Please explain the reasons for the dictionary -> interface change here.

martinthomson: If this is not an interface, the application can construct preferred codecs that the browser didn't agree to supporting.

martinthomson:
I'm going to drop this change, and await the merging of #293 before generating a new PR.

PR #327, opened 2015-10-12
Adding sha-256 to the certificate management options for RSA

### LINKED ISSUE #322: WebRTC/WebCrypto mismatch: AlgorithmIdentifier is missing 'hash', WebCrypto normalization unhappy
The following AlgorithmIdentifier value must be supported by the user agent for generateCertificate:
{ name: "RSASSA-PKCS1-v1_5", modulusLength: 2048, publicExponent: new Uint8Array([1, 0, 1]) }

But it must also produce a non-error result when normalized according to the WebCrypto normalization process with an operation name of generateKey. However, see https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#rsassa-pkcs1. Normalizing with generateKey takes parameters RsaHashedKeyGenParams, which are the expected parameters (name, modulusLength, publicExponent) plus "HashAlgorithmIdentifier hash".

Is this a WebCryptoAPI or WebRTC bug? In order to pass the normalization process one could add hash:"SHA-1" to our AlgorithmIdentifier, but that would force WebRTC implementations to use a certain hash algorithm (I don't know how or when this applies, but current chromium webrtc code does not have a hash alg parameter of any kind).

(This issue is similar/related to https://github.com/w3c/webrtc-pc/issues/310 which updated publicExponent from an integer to Uint8Array.)

martinthomson:
Closes #322.  This concedes that @alvestrand is right.  It is also what I just landed in Firefox.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2607,7 +2607,7 @@
           <p>The following values MUST be supported by a <a>user agent</a>:
             <code>{ name:
             "<a href="https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#rsassa-pkcs1">RSASSA-PKCS1-v1_5</a>",
-            modulusLength: 2048, publicExponent: new Uint8Array([1, 0, 1]) }</code>, and <code>{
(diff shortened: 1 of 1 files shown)


alvestrand:
LGTM

PR #333, opened 2015-10-15
Bunch of (mostly) editorial suggestions

dontcallmedom:
I've started reading through the spec, and have created commits of fixes I identified as I went.

I think most of the commits are editorial and uncontroversial.

I may or may not keep working on my review tomorrow (and thus add commits here), and won't be able to do so next week, so thought I would already send this in. I'm sending this as a long list of commits rather than a long list of pull requests — I suggest to cherry pick from it as needed, but if another format for this is preferred, let me know and I'll do my best to accommodate it.

Also, I pushed this as a branch in the main repo so that other committers can update this pull request as needed


stefhak:
I looked at a random selection of commits, and they all looked good to me. I propose we merge this one.

PR #342, opened 2015-10-16
Using DOMTimestamp for RTCCertificate::expires

### LINKED ISSUE #324: Let RTCCertificate.expires have type DOMTimeStamp
http://w3c.github.io/webrtc-pc/#rtccertificate-interface

Is it the intention that this attribute return the same object every time, or a new object every time? It is generally frowned upon to have APIs where `obj.attr==obj.attr` doesn't hold true, so using `[SameObject]` would be most neutral choice here.

martinthomson:
Closes #324

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2747,9 +2747,10 @@
         peer.</p>
 
         <dl class="idl" title="interface RTCCertificate">
-          <dt>readonly attribute Date expires</dt>
+          <dt>readonly attribute DOMTimeStamp expires</dt>
           <dd>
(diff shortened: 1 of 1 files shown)


jan-ivar:
two words, really? https://en.wikipedia.org/wiki/Timestamp

alvestrand:
@martinthomson it seems you changed the mode of every file as a side effect.
Can you back out and try again?

martinthomson:
Damn, I didn't even notice.  I hate it how git on windows does that.

alvestrand:
Now it looks good!

Issue #257, opened 2015-07-24
ICE Candidate should have accessors for protocol-relevant items


alvestrand:
The ICECandidate object is:

interface RTCIceCandidate {
                attribute DOMString       candidate;
                attribute DOMString?      sdpMid;
                attribute unsigned short? sdpMLineIndex;
    serializer = {attribute};
};

The stats doc has:

dictionary RTCIceCandidateAttributes : RTCStats {
             DOMString                ipAddress;
             long                     portNumber;
             DOMString                transport;
             RTCStatsIceCandidateType candidateType;
             long                     priority;
             DOMString                addressSourceUrl;
};

Somewhat more easily accessible.

Suggestion: Add the same fields as accessors on RTCIceCandidate.

martinthomson:
Easy problem: the serializer will then pick up the redundant information.

Bigger problem: now this has redundant information.  Do the extra attributes need to change when attributes are modified?  That implies custom setters, which blows.

I'd prefer to remove the RTCIceCandidate interface entirely and just use the dictionary.

alvestrand:
That's the cute thing about having an interface rather than a dictionary.
If we list the attributes we want in the serialization in the "serializer" attribute, what we specify happens.

WRT modification: I'd prefer to have all the attributes readonly - I think not making the existing ones readonly in the first place was a mistake. Modifying them for local candidates can cause nothing but confusion, and I don't see an use case for having them on remote candidates.
I could be persuaded either way about whehter there is a variant of the constructor arguments that has the "decomposed" form rather than the SDP string.

ekr:
Let's  just make the existing ones readonly too if we can. It's not like there's anything to
do by modifying them, right?

stpeter:
+1 to readonly attributes. I agree with @alvestrand's reasoning here.

stefhak:
@alvestrand In the latest version (has maybe not made it to an Ed's draft yet) there are many fields added to the candidate interface. Does that address this issue?

alvestrand:
I'll take a look.

alvestrand:
https://github.com/w3c/webrtc-stats/pull/32 tries to change webrtc-stats to conform to webrtc-pc. PTAL.

alvestrand:
I think we're OK now.

PR #349, opened 2015-10-21
Defining RTCIceCandidateInformation, which inherits from RTCIceCandidate.

taylor-b:
This makes it more clear that the only parameters relevant to
addIceCandidate() are candidate, sdpMid and sdpMLineIndex.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -625,7 +625,7 @@
 
                 <p>Add the candidate to <var>connection</var>'s
                 <code><a>localDescription</a></code> and create a
-                <code><a>RTCIceCandidate</a></code> instance to represent the
+                <code><a>RTCIceCandidateFull</a></code> instance to represent the
                 candidate. Let <var>newCandidate</var> be that object.</p>
               </li>
 
@@ -1259,7 +1259,7 @@
             since the offer or answer was created.</p>
           </dd>
 
-          <dt>Promise&lt;void&gt; addIceCandidate (RTCSignaledIceCandidate candidate)</dt>
+          <dt>Promise&lt;void&gt; addIceCandidate (RTCIceCandidate candidate)</dt>
 
           <dd>
             <p>The <dfn id=
@@ -1757,7 +1757,7 @@
             argument.</p>
           </dd>
 
-          <dt>void addIceCandidate (RTCSignaledIceCandidate candidate, VoidFunction
+          <dt>void addIceCandidate (RTCIceCandidate candidate, VoidFunction
           successCallback, RTCPeerConnectionErrorCallback failureCallback)</dt>
 
           <dd>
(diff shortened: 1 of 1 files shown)


martinthomson:
I'd prefer if you kept the shorter name for what you currently call `RTCIceCandidate`.  Maybe call the larger one `RTCIceCandidateFull`.

taylor-b:
I just realized, the smaller dictionary *needs* to be called `RTCIceCandidate`, so that existing applications can call its constructor... I suppose the constructor could just be moved to the larger dictionary, but that seems unnecessarily confusing. `RTCIceCandidateFull` sounds good to me; I'll update the PR.

webrtc.html
```diff
@@ -2356,7 +2357,13 @@
 
           <dd>This indicates the index (starting at zero) of the m-line in the
           SDP this candidate is associated with.</dd>
+        </dl>
+        <div class="note">
+          <p>The constructor on <code><a>RTCIceCandidate</a></code>
+          exists for legacy compatibility reasons only.</p>
+        </div>
 
+        <dl class="idl" title="dictionary RTCIceCandidateFull : RTCIceCandidate">
```
alvestrand: I must admit to a dislike for the name "RTCIceCandidateFull" (only partly because Norwegian associates the word "Full" with "drunk").
Can we make it RTCIceCandidateInformation, to emphasize that this is used to give information about a candidate rather than initializing from it?
(The way dictionaries work, sending an RTCIceCandidateInformation into an interface that expects an RTCIceCandidate will Just Work, btw).
</bikeshed>

elagerway:
@taylor-b please review (cc @pthatcherg )

taylor-b:
I was waiting to see if anyone else had a name preference. My only nit with `RTCIceCandidateInformation` is that it makes it less clear that an `RTCIceCandidateInformation` "is an" `RTCIceCandidate`; it sounds a bit like a supplementary, independent type. But maybe that's not important. Since no one else has commented, I'll go ahead and change it.

pthatcherg:
Since we can't change the existing RTCIceCandidate, the only name options
seem awkward in one way or another.  RTCIceCandidateInformation seems as
good as any that we've come up with.

martinthomson:
RTCIceCandidateInformation,  RTCIceCandidateDetails, ...

alvestrand:
Fixes #257

juberti:
RTCIceCandidate2? RTCIceCandidateInformation suffers from being both overly verbose and also unclear that it is derived from RTCIceCandidate.

taylor-b:
http://strawpoll.me/5871800/r

pthatcherg:
Nice.  I threw one in for RTCIceCandidate2 :)

martinthomson:
You missed RTCIceCandidateDetails from https://github.com/w3c/webrtc-pc/pull/349#issuecomment-151685385

taylor-b:
Fixed.

elagerway:
@taylor-b Seeing some conflicts here, could you address?

webrtc.html
```diff
@@ -2483,11 +2486,11 @@
           <dt>Constructor(DOMString type, RTCPeerConnectionIceEventInit
           eventInitDict)</dt>
 
-          <dt>readonly attribute RTCIceCandidate? candidate</dt>
+          <dt>readonly attribute RTCIceCandidateInformation? candidate</dt>
```
alvestrand: Do dictionary-valued attributes work in WebIDL now?

adam-be:
An earlier change, that made RTCIceCandidate a dictionary, broke the RTCPeerConnectionIceEvent. It has a candidate attribute which now is a Dictionary and that's not allowed by WebIDL. We might want to address that issue before we continue to build upon RTCIceCandidate as a dictionary.

taylor-b:
If we need an `RTCIceCandidate` interface, then I suggest we go back to having `dictionary RTCIceCandidateInit` and `interface RTCIceCandidate`. Then we can put the extra fields on the interface, and there's no confusion about what fields are needed in `addIceCandidate` because the only way to construct an `RTCIceCandidate` is with an `RTCIceCandidateInit` dictionary. This also means we don't have to decide on a new name. :)

adam-be:
We probably need to bring back the interface. If we want to keep the feature of passing a simple object to RTCPeerConnection.addIceCandidate we could make its argument a union type and accept the init dictionary as well.

elagerway:
Closing this, let's try again.

Issue #348, opened 2015-10-21
RTCRtpSender section doesn't include the definition for RTCRtpSender


martinthomson:
This is likely a consequence of using respec, but you will find `RTCRtpSender` in Section 5.2.10, entitled "Dictionary RTCRtpHeaderExtensionCapability Members"

cc @pthatcherg

dontcallmedom:
I believe this was fixed by 7a269b2bda2aac021f659b5d13f42841b10c9ba

aboba:
It does seem to be fixed in the latest Editor's draft: 
http://w3c.github.io/webrtc-pc/#rtcrtpsender-interface

PR #350, opened 2015-10-21
RTCRtpSender support objects -> interfaces

martinthomson:
There are very few things in here that are actually mutable as it turns out.  Having them as dictionaries is contrary to some of the advice I've received internally on the design of web APIs.

Most of the entire tree can be safely turned into interfaces with readonly attributes.  I think that ideally we should move `getParameters` to an attribute, but I don't want to re-open that discussion again.  I get the point of having an atomic `setParameters` though I think that we'll find that it isn't necessary in practice.

One consequence of this is that it is much more clear now on what can be changed and what cannot.  The SSRC was the only one that was previously mutable, which might have been inadvisable with a=ssrc in such heavy use; with the move to RID-based identification, I think that I can accept this being changed.

The main consequence of this change is that you have to take a copy of the parameters if you want to change them.  I think that this is an acceptable cost.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3410,8 +3410,8 @@
           </dd>
         </dl>
 
-        <dl class="idl" title="dictionary RTCRtpCapabilities">
-          <dt>sequence&lt;RTCRtpCodecCapability&gt; codecs</dt>
+        <dl class="idl" title="interface RTCRtpCapabilities">
+          <dt>readonly attribute FrozenArray&lt;RTCRtpCodecCapability&gt; codecs</dt>
 
           <dd>
             <p>
@@ -3419,7 +3419,7 @@
             </p>
           </dd>
 
-          <dt>sequence&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensions</dt>
+          <dt>readonly attribute FrozenArray&lt;RTCRtpHeaderExtensionCapability&gt; headerExtensions</dt>
           <dd>
             <p>
               Supported RTP header extensions.
--- a/webrtc.html
+++ b/webrtc.html
@@ -3208,13 +3208,13 @@
         the encoding is changed appropriately.</p>
 
         <dl class="idl" title="interface RTCRtpParameters">
-          <dt>readonly attribute FrozenArray&lt;RTCRtpEncodingParameters&gt; encodings</dt>
+          <dt>attribute sequence&lt;RTCRtpEncodingParameters&gt; encodings</dt>
 
           <dd>
             <p>Parameters for RTP encodings of media.</p>
           </dd>
 
-          <dt>readonly attribute FrozenArray&lt;RTCRtpHeaderExtensionParameters&gt; headerExtensions</dt>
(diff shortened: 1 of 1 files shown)


martinthomson:
Just thinking through the copy thing and I realized that it isn't good advice, this should do:

```js
var params = sender.getParameters();
sender.setParameters({ codecs: massageCodecs(params.codecs) });
```

But that raises the question: how does the browser treat an absent sequence differently to a present, but empty one?  The original specification isn't that clear on the point, and with the suggested usage model, it's unlikely that an absent member is often encountered.

I think that we could make this much easier to use (as above), by insisting that absent members result in no change to the current state of the corresponding attribute.

martinthomson:
Travis reports a validation error on the new `FrozenArray` usages that I added.  For some reason, the IDL doesn't include the part in `<>`, though the description of the attribute does.

martinthomson:
Oh, and editors: this section is totally screwed up.  See #348 for the tip of the iceberg.  When these were dictionaries, it looked a little better, now it's just wholly messed up.

alvestrand:
@martinthomson we need to switch to plain IDL.... Respec can do that now.

ShijunS:
It seems we don't have an issue yet to track the PR.  It'd be helpful to open an issue to state the problem so the group can decide whether and how to resolve the issue based on the group consensus.

martinthomson:
I'll let the chairs decide, but I believe that a PR has all the same features as an issue.

alvestrand:
The minutes from TPAC need studying over this one. Consulting with @dontcallmedom .

adam-be:
Is the idea that when setting an interface type, it will serialize to the corresponding dictionary type (that only has the writable attributes represented as members)?

So when you get a copy of an interface type, it will complain if you try to set something that's readonly. That's the early error you want I assume. But with the approach from the example above, where you set a new object ```{ codec: ... }``` you can still include the readonly attribute names as members with new desired values. Those will simply be throw away without any error. That's perhaps not ideal.

martinthomson:
Yes, the idea is that an interface type will be accepted as a dictionary as long as it has all the necessary attributes (and their types match, etc...).  Attributes that are readonly on the interface and absent in the dictionary are silently dropped.

And yes, if you set things that aren't in the dictionary with the expectation that they will have some effect, nothing happens.  That's an important forward-compatibility property.

adam-be:
So to be fully forward-compatible where we allow new params to be added as well as making currently "readonly" params writable, it's problematic with interfaces with readonly attributes as well.

One possible way to achieve that would be to use the current approach with dictionaries, but instead of throwing a ReadOnlyError, when the script tries to set a non-changeable param, we could just do a console warning.

elagerway:
Looks like we need to close this and try again.

martinthomson:
I'm not sure what basis this was closed on.  I think that I disagree with @adam-be here.  I don't believe that we need to warn about someone attempting to set a parameter that doesn't exist in the dictionary.

stefhak:
@martinthomson I think we may have a better discussion if we first file an Issue that explains the problem we want to solve. Is that something you could do?

PR #356, opened 2015-10-29
Mark sender and receiver in Transceiver as same object

dontcallmedom:
Per discussions at Sapporo F2F


dontcallmedom:
Closing to avoid dup with #357

PR #357, opened 2015-10-29
Making RTCRtpTransceiver properties SameObject

martinthomson:


martinthomson:
Hmm, looks like @dontcallmedom beat me to it with #356.  Pick one, close t'other.

PR #361, opened 2015-10-29
Adding setCodecPreferences to RTCRtpTransceiver

martinthomson:


alvestrand:
lgtm

stefhak:
Perhaps a warning that using this method to remove codecs may result in no common codec and hence no media?
Anyway, something we can add later. LGTM, and we should merge IMO.

PR #362, opened 2015-10-30
Fix for issue #21

dontcallmedom:
I grepped for lowercase "must" in the docs — the 3 instances all seemed to NOT be meant as RFC2119, so I have reworded them to avoid using "must".


fluffy:
LGTM

PR #365, opened 2015-10-30
Adding an accessor for the browser-configured ICE servers

### LINKED ISSUE #263: over riding default ice servers
Some browsers are setting a default TURN and / or STUN servers. There has to be a clear way to override this so they are not used. 

var configuration =  { "iceServers": [] };
pc = new RTCPeerConnection(configuration);

is a fine way to that but the spec would need to be updated to make it clear that over wrote any default servers.  Thoughts on if this is the right way to do it?

martinthomson:
As discussed at TPAC.

Closes #263.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1364,6 +1364,16 @@
             this value is <code>null</code>.
           </dd>
 
+          <dt>static sequence&lt;RTCIceServer&gt; getDefaultIceServers()</dt>
+
+          <dd>
+            <p>This method returns a list of ICE servers that are configured into
+            the browser.  A browser might be configured to use local or private
+            STUN or TURN servers.  This method allows an application to learn about
+            these servers and optionally use them.
+            </p>
+          </dd>
+
           <dt>RTCConfiguration getConfiguration()</dt>
 
           <dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1364,13 +1364,14 @@
             this value is <code>null</code>.
           </dd>
 
-          <dt>static sequence&lt;RTCIceServer&gt; getDefaultIceServers()</dt>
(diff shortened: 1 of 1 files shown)


juberti:
lgtm

fluffy:
LGTM

ekr:
LGTM

webrtc.html
```diff
@@ -1364,6 +1364,16 @@
...
 
+          <dt>static sequence&lt;RTCIceServer&gt; getDefaultIceServers()</dt>
+
+          <dd>
+            <p>This method returns a list of ICE servers that are configured into
+            the browser.  A browser might be configured to use local or private
+            STUN or TURN servers.  This method allows an application to learn about
+            these servers and optionally use them.
+            </p>
+          </dd>
```
pthatcherg: I think this would make more sense on IceTransport.  It will still be valuable in WebRTC NV and it's ICE-specific.

webrtc.html
```diff
@@ -1364,6 +1364,16 @@
             this value is <code>null</code>.
           </dd>
 
+          <dt>static sequence&lt;RTCIceServer&gt; getDefaultIceServers()</dt>
```
jan-ivar: Can't this be an attribute? e.g. `static readonly attribute FrozenArray<RTCIceServer> defaultIceServers;`

alvestrand:
Waiting for FrozenArray to be fixed.

dontcallmedom:
FrozenArray is now fixed in ReSpec, but this has only uncovered another bug in ReSpec (in WebIDL oldschool mode): it doesn't know static attributes; I'll work on fixing that.

dontcallmedom:
I've submitted https://github.com/w3c/respec/pull/527 that should make this finally Travis-happy.

dontcallmedom:
Travis now happy, un-assigning myself.

PR #371, opened 2015-11-02
Add RtcpMuxPolicy

pthatcherg:
It looks like we forgot to add this after we added BundlePolicy.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -169,6 +169,12 @@
             <p>Indicates which <a href="#target-bundle-policy">media-bundling policy</a> to use when gathering ICE candidates. </p>
           </dd>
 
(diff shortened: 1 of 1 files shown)


alvestrand:
LGTM after IETF discussion that concluded that we can't remove this from the API surface.
Note that JSEP is adding more text about default and allowing apps to not support non-mux.

webrtc.html
```diff
@@ -169,6 +169,12 @@
             <p>Indicates which <a href="#target-bundle-policy">media-bundling policy</a> to use when gathering ICE candidates. </p>
           </dd>
 
+          <dt>RTCRtcpMuxPolicy rtcpMuxPolicy = "required"</dt>
```
alvestrand: Can you make this "require" (or change the other one to "required"?

pthatcherg: Done

PR #380, opened 2015-11-04
Another bunch of editorial fixes

dontcallmedom:
in the same spirit and process as #333


webrtc.html
```diff
@@ -3043,6 +3035,39 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           </dd>
 
         </dl>
+        <dl class="idl" title="dictionary RTCRtpTransceiverInit">
```
alvestrand: I don't get why this has moved. Shouldn't it stay with RTCRtpTransceiver?

dontcallmedom: I moved it closer to where it is being used (in `addTransceiver`); at the moment, it is with `RTCRtpTransceiver`, but nothing in that interface makes use of `RTCRtpTransceiverInit`, with the result that the reader is left puzzled as to what it is used for.

alvestrand:
@adam-be says he can merge once you've rebased it if you want a second pair of eyes.
Otherwise, just merge.

dontcallmedom:
rebased; assigning to @adam-be  for a last check.

adam-be:
Ops. I missed this. Now we might as well merge this one at the editor's meeting later today.

alvestrand:
Oops - we let other PRs disturb it again. Please rebase and merge immediately!

Issue #393, opened 2015-11-17
Why throw on empty iceServers list?


jan-ivar:
In [setConfiguration](http://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-setConfiguration-void-RTCConfiguration-configuration) under step 6 it says:

> "2. If the *server.urls* dictionary member an empty list, then throw an `InvalidAccessError` and abort these steps."

This seems superfluous, inconsistent, surprising and problematic.

Superfluous, because it provides no invariant (omitting `iceServers` produces an empty list).

Inconsistent, because an empty list is a valid value (the default value in fact).

Surprising, because `iceServers` is not a required field.

Problematic, because some browsers (Firefox) allow end-users to configure a non-empty default, meaning explicitly providing an empty list has purpose.

An example of the latter is an app that wants to limit communication to the LAN (i.e. communicate exclusively with host candidates). `new RTCPeerConnection({ iceServers: [] })` seems the most intuitive way to communicate this intent (to turn off any user-configured defaults in browsers like Firefox).

stefhak:
Closing since #394 has been merged.

PR #394, opened 2015-11-17
don't throw on empty iceServers list.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/393.


martinthomson:
Yep, this is good.

PR #397, opened 2015-11-19
Make degradationPreference per-sender instead of per-encoding.

pthatcherg:
This is what we decided to do at TPAC 2015.


aboba:
This is a useful change - and one we have already made in the ORTC API specification.

alvestrand:
LGTM (layer-dropping then becomes a reasonable thing to say)

alvestrand:
@pthatcherg can you rebase? @aboba will merge.

PR #398, opened 2015-11-20
Make RtpTransceiver.mid nullable and remove RtpSender.mid and RtpReceiver.mid.

pthatcherg:
This is what we decided to do at TPAC 2015.


alvestrand:
LGTM

stefhak:
LGTM.

Issue #377, opened 2015-11-04
is `audioLevel` in `RTCRtpContributingSource` positive or negative?


dontcallmedom:
The spec describes `audioLevel` as ranging from 0 to -127, but refers to RFC6464 which uses positive values to represent that range (by expressing it in the unit -dBov).

fluffy:
Great question. My preference is it returns a negative number 

Sent with my thumbs - sorry

dontcallmedom:
This was closed when #403 was merged

PR #403, opened 2015-11-25
Fix for Issue 377 AudioLevel

aboba:
https://github.com/w3c/webrtc-pc/issues/377

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3836,8 +3836,8 @@ sender.setParameters(params)
        value as if it had come from RFC 6464 and use that.  If the
        source was set from a CSRC, this will be the level value
        defined in [[!RFC6465]].  RFC 6464 and 6465 define the level as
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3836,8 +3836,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
        value as if it had come from RFC 6464 and use that.  If the
        source was set from a CSRC, this will be the level value
        defined in [[!RFC6465]].  RFC 6464 and 6465 define the level as
-       a integral value from 0 to -127 representing the audio level in
-       decibels relative to the loudest signal that they system could
+       a integral value from 0 to 127 representing the audio level in
+       negative decibels relative to the loudest signal that they system could
```
alvestrand: Is this term readily understood? We could add "Thus, 0 represents the loudest signal the system could possibly encode, and 127 represents silence."

PR #405, opened 2015-11-25
Add note about addtrack event in 'dispatch a receiver' steps

adam-be:


alvestrand:
Waiting for w3c/mediacapture-main#281

Issue #35, opened 2014-11-10
addIceCandidate shall throw a TypeError exception if the candidate parameter is malformed


zqzhang:
In 4.3.2.3 [1], there is a paragraph:

> If the candidate parameter is malformed, throw a SyntaxError exception and abort these steps.

`malformed` shall mean 'not conforming to a standard type'. For example, if the candidate parameter is an empty string `""`, `addIceCandidate` shall throw a `TypeError` exception rather than a `SyntaxError` per the WebIDL dictionary definition [2]. At least the spec [1] is not clear on this aspect.

[1] http://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-addIceCandidate-void-RTCIceCandidate-candidate-VoidFunction-successCallback-RTCPeerConnectionErrorCallback-failureCallback
[2] http://heycam.github.io/webidl/#es-dictionary

alvestrand:
Handling should be consistent with what happens when invalid SDP is passed to Set*Description.

jan-ivar:
`addIceCandidate("")` will throw `TypeError` (on implementations relying on a webidl-compiler for sure). We don't have to spec that since it's implicit in WebIDL. I think the prose here is about candidate validity in ways unique to candidates and undetectable by WebIDL binding code.

About the choice of `SyntaxError`, my understanding is that this error is reserved for JS parsing errors, i.e. when your JS is not JS. Yet the webrtc spec returns it in a couple of places. Should we revisit that? Thoughts, @annevk or @bzbarsky?

bzbarsky:
That was my understanding of SyntaxError too, fwiw.

annevk:
Yes you should revisit that. `TypeError` is usually a safe choice.

adam-be:
I think we can all agree on what @jan-ivar is saying about the WebIDL type checking. Regarding SyntaxError, I think it's a pretty good match from it's definition [1].

"SyntaxError" 	The string did not match the expected pattern.

We expect the candidate string to follow a certain pattern. SyntaxError is for example used by the WebSocket constructor if the argument url is malformed.

[1] https://heycam.github.io/webidl/#syntaxerror

bzbarsky:
That's a link to DOMException with .name == "SyntaxError".  That's not the same thing as a SyntaxError.  Yes, this is unfortunate...  WebSocket throws the DOMException, fwiw.

[...]

PR #407, opened 2015-11-30
Bringing back the RTCIceCandidate interface, with 'init' dictionary.

taylor-b:
This is necessary since RTCIceCandidate is an attribute of an event
interface. This is also convenient since the new attributes on a
candidate (ip, port, foundation, etc.) can now become readonly
attributes.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1333,31 +1333,28 @@
 
             <ol>
               <li>
-                <p>Let <var>p</var> be a new promise.</p>
-              </li>
-
-              <li>
                 <p>If this <code><a>RTCPeerConnection</a></code> object's
                 <a href="#dom-peerconnection-signaling-state">signaling
-                state</a> is <code>closed</code>, the user agent MUST reject
-                <var>p</var> with <code>InvalidStateError</code>, and
-                jump to the step labeled <em>Return</em>.</p>
+                state</a> is <code>closed</code>, the user agent MUST return
+                a promise rejected with an <code>InvalidStateError</code>.</p>
               </li>
 
               <li>
                 <p>If <var>candidate</var> is missing values for both <var>sdpMid</var> and
-                <var>sdpMLineIndex</var>, reject <var>p</var> with a <code>TypeError</code>
-                and jump to the step labeled <em>Return</em>.</p>
+                <var>sdpMLineIndex</var>, return a promise rejected with a
+                <code>TypeError</code>.</p>
               </li>
 
               <li>
-                <p>If the candidate could not be successfully applied, reject
-                <var>p</var> with a <code>DOMError</code> object whose
-                <code>name</code> attribute has the value TBD
-                and jump to the step
-                labeled <em>Return</em>.</p>
-                <p class="issue">TODO: define names for DOMError (
-                InvalidCandidate and InvalidMidIndex (see also <a href="https://github.com/w3c/webrtc-pc/issues/319">Issue 319</a>)</p>
+                <p>If the candidate could not be successfully applied, return a promise
+                rejected with a <code>DOMError</code> object whose <code>name</code>
+                attribute has the value TBD.</p>
+                <p class="issue">TODO: define names for DOMError
+                (InvalidCandidate and InvalidMidIndex (see also <a href="https://github.com/w3c/webrtc-pc/issues/319">Issue 319</a>))</p>
+              </li>
+
+              <li>
+                <p>Let <var>p</var> be a new promise.</p>
               </li>
 
               <li>
@@ -1366,7 +1363,7 @@
               </li>
 
               <li>
-                <p><em>Return:</em> Return <var>p</var>.</p>
+                <p>Return <var>p</var>.</p>
               </li>
             </ol>
 
--- a/webrtc.html
+++ b/webrtc.html
@@ -2460,6 +2460,8 @@
           candidate, the <dfn>relatedPort</dfn> is the port of the candidate that
           it is derived from. For host candidates, the <code>relatedPort</code>
           is <code>null</code>.</dd>
+
+          <dt>serializer = { candidate, sdpMid, sdpMLineIndex }</dt>
         </dl>
 
         <dl class="idl" title="dictionary RTCIceCandidateInit">
--- a/webrtc.html
+++ b/webrtc.html
@@ -2399,7 +2399,7 @@
           <a class="idlType" href="#idl-def-RTCIceCandidate"><code>RTCIceCandidate</code></a>
           object. If <var>sdpMid</var> or <var>sdpMLineIndex</var> is not present in
           <var>candidateInitDict</var>, the corresponding attribute will be initialized to
-          null.</dd>
+          <code>null</code>.</dd>
 
           <dt>readonly attribute DOMString candidate</dt>
 
--- a/webrtc.html
+++ b/webrtc.html
@@ -1329,8 +1329,7 @@
             media state if it results in different connectivity being
             established. The only members of <code>candidate</code>
             used by this method are <var>candidate</var>, <var>sdpMid</var> and
-            <var>sdpMLineIndex</var>; the rest are ignored. A value for either
-            <var>sdpMid</var> or <var>sdpMLineIndex</var> MUST be provided.</p>
+            <var>sdpMLineIndex</var>; the rest are ignored.</p>
 
             <ol>
               <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1327,9 +1327,10 @@
             <code>none</code>. This call will result in a change to the
             connection state of the ICE Agent, and may result in a change to
             media state if it results in different connectivity being
-            established. The only members of the candidate attribute
+            established. The only members of <code>candidate</code>
             used by this method are <var>candidate</var>, <var>sdpMid</var> and
-            <var>sdpMLineIndex</var>; the rest are ignored.</p>
+            <var>sdpMLineIndex</var>; the rest are ignored. A value for either
+            <var>sdpMid</var> or <var>sdpMLineIndex</var> MUST be provided.</p>
```
adam-be: It would be great to have this requirement as a part of the algorithm steps below and throw a TypeError if not fulfilled. The resulting error type can be discussed, but we have earlier used TypeError when we put custom requirements on dictionaries to align with the new "required member" feature.

webrtc.html
```diff
@@ -2385,82 +2386,95 @@
...
+          <dt>Constructor (RTCIceCandidateInit candidateInitDict)</dt>
+
+          <dd>The <dfn id="dom-icecandidate"><code>RTCIceCandidate()</code></dfn> constructor
+          takes a dictionary argument, <var>candidateInitDict</var>, whose content is used
+          to initialize the new
+          <a class="idlType" href="#idl-def-RTCIceCandidate"><code>RTCIceCandidate</code></a>
+          object. If <var>sdpMid</var> or <var>sdpMLineIndex</var> is not present in
+          <var>candidateInitDict</var>, the corresponding attribute will be initialized to
+          <code>null</code>.</dd>
+
```
adam-be: The above "action" should also go into the addIceCandidate() algorithm to make that complete.

webrtc.html
```diff
@@ -2385,82 +2386,95 @@
...
           <dd>If present, this contains the identifier of the "media stream
           identification" as defined in [[!RFC5888]] for the media component this
           candidate is associated with.</dd>
 
-          <dt>unsigned short sdpMLineIndex</dt>
+          <dt>readonly attribute unsigned short? sdpMLineIndex</dt>
 
-          <dd>This indicates the index (starting at zero) of the <a>media description</a> in the
-          SDP this candidate is associated with.</dd>
+          <dd>If present, this indicates the index (starting at zero) of the
```
adam-be: s/present/not null/ ?

adam-be:
This looks overall good. A few comments above.

webrtc.html
```diff
@@ -2385,82 +2386,95 @@
...
 
-          <dt>DOMString sdpMid</dt>
+          <dt>readonly attribute DOMString? sdpMid</dt>
 
           <dd>If present, this contains the identifier of the "media stream
           identification" as defined in [[!RFC5888]] for the media component this
           candidate is associated with.</dd>
 
-          <dt>unsigned short sdpMLineIndex</dt>
+          <dt>readonly attribute unsigned short? sdpMLineIndex</dt>
```
adam-be: Before we went from interface to dictionary, the candidate, sdpMid and sdpMLineIndex attributes were writable. I don't remember exactly why, but we should probably not change that without list discussion.

webrtc.html
```diff
@@ -2385,82 +2386,95 @@
...
+          <dt>readonly attribute DOMString? relatedAddress</dt>
 
           <dd>For a candidate that is derived from another, such as a relay or reflexive
           candidate, the <dfn>relatedAddress</dfn> is the IP address of the candidate that
           it is derived from. For host candidates, the <code>relatedAddress</code>
-          is not present in the dictionary.</dd>
+          is <code>null</code>.</dd>
 
-          <dt>unsigned short relatedPort</dt>
+          <dt>readonly attribute unsigned short? relatedPort</dt>
```
adam-be: I believe we need a serializer [1] for this interface. The question is if it should be for all attributes or just the three "old" ones.

[1] http://heycam.github.io/webidl/#idl-serializers

webrtc.html
taylor-b: I assumed they were writable because the RTCIceCandidateInit argument to the constructor was optional. So it would be possible to do:

`var candidate = new RTCIceCandidate();`
`candidate.sdpMid = "audio";`
`/*...*/`

As opposed to:

`var candidate = new RTCIceCandidate({sdpMid:"audio",/*...*/});`

At some point, the argument to the constructor became mandatory, breaking the first example. I brought this up on the list and haven't seen a response yet.

webrtc.html
taylor-b: In my opinion it would be best to stick with the three old ones, since that's what current applications are expecting, and anything more would be redundant information.

webrtc.html
adam-be: Please remove this since it duplicates what the algorithm now says.

webrtc.html
```diff
@@ -1345,6 +1346,12 @@
               </li>
 
               <li>
+                <p>If <var>candidate</var> is missing values for both <var>sdpMid</var> and
+                <var>sdpMLineIndex</var>, reject <var>p</var> with a <code>TypeError</code>
+                and jump to the step labeled <em>Return</em>.</p>
```
adam-be: This is just a formatting thing. A nice way to handle error cases without having to jump to a "return label" is to express it like: If bad condition is true, return a promise rejected with BadThingError. Then the steps bails at that point and you can define p much later in the steps. See example in the promise writing guide [1].

[1] https://www.w3.org/2001/tag/doc/promises-guide#example-validated-delay

webrtc.html
adam-be: When I read it again, I see that my previous comment wasn't right.

But the constructor needs to throw as well. A straight forward way to define it would be to set each property (sdpMid and sdpMLineIndex) to null if not present in the init dict. And follow up with a step that throws if both are null.

webrtc.html
adam-be: It's perhaps best to, at least, keep the candidate attribute readonly. Otherwise, we could end up in a situation where the candidate line says one thing, and the information attributes (below) says something else. If we don't parse the candidate every time it's updated and update the information attributes that is.

adam-be:
Thanks for updating @taylor-b

webrtc.html
taylor-b: Alright; I'll make steps 2 and 4 use this style as well.

webrtc.html
taylor-b: I don't know if it's *really* needed, since `addIceCandidate` will throw an exception anyway, ultimately. But I guess it can't hurt.

webrtc.html
taylor-b: Another formatting question: When do we say "this attribute MUST return x" as opposed to simply "this attribute returns x"? There seems to be some inconsistency.

adam-be:
LGTM now

adam-be:
Correct me if I'm wrong, but as I understand it, constructing a new RTCIceCandidate() with an sdp string will populate all the new attributes with information. That would involve parsing the sdp string, which could fail. That brings back the problem from issue #35. But let's do that in a follow-up PR.

webrtc.html
```diff
@@ -2385,82 +2388,98 @@
...
+        <dl class="idl" data-merge="RTCIceCandidateInit" title="interface RTCIceCandidate">
+          <dt>Constructor (RTCIceCandidateInit candidateInitDict)</dt>
+
+          <dd>The <dfn id="dom-icecandidate"><code>RTCIceCandidate()</code></dfn> constructor
+          takes a dictionary argument, <var>candidateInitDict</var>, whose content is used
+          to initialize the new
+          <a class="idlType" href="#idl-def-RTCIceCandidate"><code>RTCIceCandidate</code></a>
+          object. If either <var>sdpMid</var> or <var>sdpMLineIndex</var> is not present in
+          <var>candidateInitDict</var>, the corresponding attribute will be initialized to
+          <code>null</code>. If neither is present, an <code>InvalidParameter</code> exception
```
adam-be: I'll do a s/InvalidParameter/TypeError/ before landing. To be consistent with the similar error in addIceCandidate() and how we treat these errors in general.

pthatcherg:
I think the synchronous parsing of the SDP line is fine.  In fact, it would be kind of nice for the JS to use, and it should be minimal implementation complexity.  If it fails, we can throw an InvalidParameter just like we do if sdpMid or sdpMLineIndex is missing.

Issue #408, opened 2015-12-02
Clarify component for ICETransport


stefhak:
To me it seems unspecified what "component" will be when rtcp-mux is used.

pthatcherg:
Same as for the RTP component: 1.

stefhak:
Fixed, closing.

PR #411, opened 2015-12-02
Fix for Issue 408: Clarify component for IceTransport when RTP/RTCP mux is used

aboba:
https://github.com/w3c/webrtc-pc/issues/408


stefhak:
This goes a bit, so I think we could merge. But the ORTC doc also specifically mentions the data channel (and SCTP) here to cover the case with only data channels. I think we should too.

PR #416, opened 2015-12-03
RTCRtpSender.replaceTrack() fixes

adam-be:
* Use early promise return syntax (from promise writing guide)
* Handle cases when the sender has been removed or if the associated RTCPeerConnection is closed (introduced RTCRtpSender stopped)
* Queue a task to update the track attribute (we can't do that from "in parallel")

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3518,9 +3518,10 @@ sender.setParameters(params)
                     negotiations.</p>
                   </li>
                   <li>
-                    Set <code><var>sender</var>.<a href="#dom-rtpsender-track">track</a></code>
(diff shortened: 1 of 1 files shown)


alvestrand:
@pthatcherg  you may want to look at this.

stefhak:
LGTM. But I expect that we will need more text in the spec to define situations where a sender is stopped.

adam-be:
Yes. This PR only fixes the most obvious problems. We also have the case where a sender has not "started" yet. I.e., addTrack() is done, but the sender is not negotiated.

PR #421, opened 2015-12-04
Require sdp in RTCSessionDescription{,Init}

### LINKED ISSUE #420: RTCSessionDescription should not allow nullable sdp
Unless I missed something, I see no reason why the sdp attribute of `RTCSessionDescription` is marked as nullable.

From what I can tell, it was made nullable when `RTCSessionDescriptionInit` was introduced (commit fcda53c460a3f8e32db683f6284dfabed3688a3a), and that was probably needed to be compatible with dictionaries when they could not have required members.

But `RTCSessionDescriptionInit` has now `type` as required, and I think it should also have `sdp` as required, and the sdp attribute in `RTCSessionDescription` marked as non-nullable.

Happy to provide the PR if I get confirmation the reasoning is valid.

dontcallmedom:
close #420

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2306,10 +2306,9 @@
 
           <dd>The type of this RTCSessionDescription.</dd>
 
-          <dt>attribute DOMString? sdp</dt>
+          <dt>attribute DOMString sdp</dt>
 
-          <dd>The string representation of the SDP [[!SDP]] or
-          a <code>null</code> value.</dd>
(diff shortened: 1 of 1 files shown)


martinthomson:
Only the change to the interface is valid if we want to continue to support offer rollback.

dontcallmedom:
@martinthomson This PR doesn't change anything about the type attribute and member, only about the sdp one. I guess you're suggesting that the type member should no longer be marked as required in the dictionary?

PR #422, opened 2015-12-07
Remove confusing paragraph on fourth party interception

dontcallmedom:
Per https://lists.w3.org/Archives/Public/public-webrtc/2015Dec/thread.html#msg13


stefhak:
LGTM

PR #424, opened 2015-12-07
Distinguish states from their attribute representation

dontcallmedom:


adam-be:
This is a great improvement.

The current mix with "connection.signalingState" in some sections and "RTCPeerConnection object's RTCPeerConnection signaling state" in others, is far from ideal. There might be other versions too.

dontcallmedom:
Re other fixes needed, I'm hoping to come up with a credible plan to transition us from WebIDL-oldschool to WebIDL contiguous for ReSpec in the upcoming few weeks; after that transition, I think I'll take a more thorough stab at distinguishing between the abstract model and the concrete API.

adam-be:
Merged master. One conflict in the new addIceCandidate() code that referred to  signalling state).

Will merge PR when Travis gives thumbs up.

PR #429, opened 2015-12-09
Adding expires attribute to generateCertificate.

### LINKED ISSUE #386: Certificates expiry?
When generating certificates or obtaining them automatically from the peer connection object, the developer has no control on the expiry date which seems to be implementation dependent.

Why is that? Shouldn't generateCertificate take a parameter to set the expiry time? 
Should there be a common default expiry date for automatically generated certificates (I think both Chrome and Firefox uses a month)?

martinthomson:
As I said on the list, and as @steelyglint has noted, I'm not sure that this is the best approach, but it's what Firefox currently does.

Closes #386.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2752,6 +2752,17 @@
           linked to a user or <a>user agent</a>.  Randomized values for
           distinguished name and serial number SHOULD be used.</p>
 
+          <p>An optional <code>expires</code> attribute MAY be added to
+          the <var>keygenAlgorithm</var> parameter.  If this contains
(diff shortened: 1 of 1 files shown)


alvestrand:
Commenting on a closed CL:

IDL-formally, I think it's a bit sad that the AlgorithmIdentifier is a typedef for "DOMString or object", since that means that we can't record in any IDL place that the "expires" field is added to it.

No change needed.

martinthomson:
@alvestrand, I totally agree.  I really wanted to do something more concrete.  I'm not very happy with the WebCrypto API in many other respects either.

PR #434, opened 2015-12-10
Change setParameters call to be Async 

### LINKED ISSUE #308: RTCRtpSender.get/setParameters is underspecified
There are a number of questions that occur to me:
* Are changes to the parameters signaled to the other side? If so, which ones, and does onnegotiationneeded fire?
* How does content know when its preferences have been applied (or have failed)?
* It is unclear whether getParameters() is intended to get the current preferences expressed by content, or the current state of the encoder. If it is the latter, there are more questions:
   * After a call to setParameters, it is unclear when the new value should be obtainable by a call to getParameters; does the operation need to actually complete? If renegotiation is required, does this need to finish?
   * If an attempted change to parameters fails (because the remote side prevented it, or some other reason), will the PC fire onnegotiationneeded if it thinks that would allow the preferences to be applied? Will the PC continue trying to apply the preferences on each subsequent renegotiation (ie; are these settings "sticky")?

fluffy:
Fixes #308


fluffy:
Have a look at this and see if it is ready for merge

webrtc.html
```diff
@@ -3462,6 +3462,16 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             renegotiated.  After the codecs are renegotiated, they are
             set to the value negotiated, and <code>setParameters</code> needs to be
             called again to re-apply codec preferences.</p>
+
+            <p> The setParameters does not cause renegotiation at the SDP level
+            and can only be used to change what the media stack is sending or
+            receiving withint the envelope negotiated by the Offer/Answer. The
```
martinthomson: within*t*

webrtc.html
```diff
@@ -3462,6 +3462,16 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            atributes in the <code>RTCRtpParameter</code> dictionary are
```
martinthomson: atributes (missing a t)

webrtc.html
```diff
@@ -3462,6 +3462,16 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            designed to not enable this so things like <code>ssrc</code> that
+            can not be changed are read only and things like bitrate are only
+            controllable with a <code>maxBitrate</code> which is merged with
+            bitrate constraints that may be signaled in SDP. </p>
```
martinthomson: I think that "merged" needs to be better defined (or state that it's not merged, but applied separately as another limit on bitrate, separate to that imposed by the peer via SDP).

fluffy:
Martin - thanks for the review. I pushed some changes that I think fix all them. Have a read of my new line around where I used to have the "merged" text and see if that looks better or just propose something and I will use that.

webrtc.html
```diff
@@ -3462,6 +3462,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             renegotiated.  After the codecs are renegotiated, they are
             set to the value negotiated, and <code>setParameters</code> needs to be
             called again to re-apply codec preferences.</p>
+
+            <p> The setParameters does not cause renegotiation at the SDP level
+            and can only be used to change what the media stack is sending or
+            receiving within the envelope negotiated by the Offer/Answer. The
+            attributes in the <code>RTCRtpParameter</code> dictionary are
+            designed to not enable this so things like <code>ssrc</code> that
+            can not be changed are read only. Other things, like bitrate, are
```
martinthomson: cannot

martinthomson:
LGTM

stefhak:
LGTM. 
It is not clear to me what would happen if a future re-negotiation leaves the parameters set outside the envelope of the (newly) negotiated session, but that is beyond the scope of this PR (and perhaps not a practical problem).

fluffy:
> On Dec 16, 2015, at 7:33 AM, stefan hakansson <notifications@github.com> wrote:
> 
> LGTM. 
> It is not clear to me what would happen if a future re-negotiation leaves the parameters set outside the envelope of the (newly) negotiated session, but that is beyond the scope of this PR (and perhaps not a practical problem).
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/w3c/webrtc-pc/pull/434#issuecomment-165144964>.
> 

I think we decided that you need to redo the setParameters after each re-negotiation and they are all effectively cleared on negotiation

fluffy:
I think this is ready to be merged

alvestrand:
Seems to me that the text doesn't say where the promise is resolved, so it's not complete. Where should the text saying "resolve the promise" go?

alvestrand:
the other worrisome item raised was what is returned for getParameters() while a setParameters() is in flight (before you resolve the promise).
It's natural to assume that consecutive setParameters() are queued, so that you're guaranteed that the last one invoked is the last one to take effect, and that the resolutions happen in the same order, but it would be nice to have text that says so rather than making the assumption.

fluffy:
Is there a use case you have in mind where this matters to help drive the design ?

alvestrand:
A situation where the user wants to reduce bandwidth usage at the same time as an automatic response to detected congestion is carried out by JS. The user modifies (get - modify - set) encodings[0].maxBitrate, while the automatic process modifies (get - modify - set) encodings[1].active.

If the set operation is synchronous, both things will happen, no matter which order they're done in.
If the set operation is asynchronous, it's not clear what will happen.

If both of these things eventually happen, all is good. If one of these things happens and the other does not, and there's no way to predict which one will be applied, the system becomes hard to reason about.

fluffy:
This is NOT ready to merge ... needs a few updates. 

@alvestrand - could you look a the text around the issue from the interim and propose chances to what I wrote ?

alvestrand:
I promised to give a text proposal based on what I suggested at the interim.
Basically:

- getParameters' returned dictionary includes a "transactionToken" value fetched from the RTPSender
- setParameters (asynchronous part) checks that the transactionToken in its parameters is present and is the same as in the RTPSender. If it is, it applies the change and changes the RTPSender's "transactionToken" to a new, previously unused value (for example current+1).

That's it. If RTPEncodingParameters had been an interface not a dictionary, we could even have made it a hidden property, but since it's a dictionary, it has to be out in the open.

elagerway:
@fluffy My recollection on this is that @alvestrand would make a PR for the diff based on his comments above. Please have a look and if it seems fine you can re-assign to Harald.

alvestrand:
ping @fluffy - please say "OK" if you're OK - since this is a branch in the main repo, I can just pull & update it.

elagerway:
@fluffy I know you are busy, assigning this to @alvestrand for a PR, please review when you have a moment.

alvestrand:
OK, here's my proposal. It turned out to be easiest to add it as a processing model for setParameters, leaving the earlier description as commentary.

Note that the description refers to "parameters marked read-only" - there are no such parameters now, and they don't exist in dictionaries anyway. That's another problem that needs solving.

jan-ivar:
Could someone point me to the list discussion on why `setParameters` needs to be async?

dontcallmedom:
https://www.w3.org/2015/10/29-webrtc-minutes#item03
> harald: #308 RTCRTPSender.get/setParameters is underspecified
> 
> peter: I can't think of anything that would go outside the negotiated params.
> 
> adam, fluffy: we have several things that constraint the bitrate.
> 
> fluffy: if we can detect an error, please detect it.
> 
> martin: this is only one of the issue in this bug.
> ... on the setting of parameters, shall we use getParameters? Is it done as a promise resolve?
> ... I think that the other adam (B) has something for that, but .... I'm not sure we can close this one today.
> 
> harald: so setting the parameters is immediate, but the action triggered by the setting can take some time
> 
> dom: so if you set and get, you get back the values you've just set?
> 
> peter: if we really wan tot know if things were done, we need to return a promise, but ....
> 
> martin: so what is the conclusion? The parameters are set right away, but taken into account only at a future (undefined) time?
> 
> xxxx: yes
> 
> peter: the question remains if we want a promise to be resolved at that undefined time.
> 
> martin: codec setting example
> 
> dan: so what happen if you make a set, and then a get, before the set finished?
> 
> adam: you get whatever is there at the get time, even if there is a pending operation.
> 
> martin, ekr: then you lost the ability to get the old parameters.
> 
> peter: that seems for a lot of complexity, while we have only one use case.
> 
> adam: the other problem is that it is an expending set.
> 
> peter: we are close to 1.0 ...
> ... the set can't expand forever.
> 
> ekr: do we reject at set time? Why don t we make it async, and that s it.
> 
> harald: ok, set return a promise.
> 
> martin: i m not in favor of having another queue. I don t want get to return a promise.
> 
> harald: who is taking responsibility for this?
> 
> jib: note: you can t throw and return a promise
> 
> martin: i think we should ask Byron to identify what we missed.
> 
> peter: and also that we really do not have anything going out.
> 
> martin: I did answer that.
> 
> adam: I'll take care of the PR

jan-ivar:
I see I was in the tail of that exchange, and I still don't get it. They're just settings, so what's a case where we care exactly when they're acted on? Is there a case where validation of the settings need to be async?

It also sounded like we were reaching out for more info. Do we have more info?

alvestrand:
We discussed it again at the recent interim (that's where I volunteered to tweak Cullen's PR).

stefhak:
@aboba can you review?

stefhak:
LGTM.

elagerway:
LGTM

PR #439, opened 2015-12-16
Complete security considerations based on security questionnaire and IP address discussions

dontcallmedom:
This pull request:
* gives more structure to the security considerations section
* adds privacy to its scope
* documents fingerprinting surfaces
* links to the IP address policy proposal
* adds security and privacy impact in-line in the API where it matters

PR #444, opened 2015-12-28
Add languages on how to choose IP addresses.

guoweis:


webrtc.html
```diff
@@ -7698,7 +7698,9 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
     session.</p>
 
     <p>Revealing IP addresses can leak location and means of
-    connection; this can be sensitive.</p>
+    connection; this can be sensitive but necessary for establishing peer to
+    peer connections. The choice of which addresses to expose is controlled by
```
pthatcherg: Would it make sense to change "which addresses to expose" into "which addresses to expose and when to expose them"?

pthatcherg: Would "is determined by" be more accurate than "is controlled by"?

guoweis: PTAL.

Thanks,
Guowei

alvestrand:
Thanks - this is also covered in #439 - @dontcallmedom and @guoweis - please make sure language is compatible;M it would be nice to merge this only once.

guoweis:
I see that dontcallmedom's write up has covered what I intended to say. I
can close #444 as a dup.

Thanks,
Guowei

On Tue, Jan 5, 2016 at 11:05 AM, Harald Alvestrand <notifications@github.com

guoweis:
This is a dup with #439

PR #445, opened 2015-12-29
Make DataChannels Transferable and explain the implications

ekr:
@martinthomson PTAL

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4342,8 +4342,8 @@ sender.setParameters(params)
           </dd>
 
         <dt>RTCDataChannel createDataChannel([TreatNullAs=EmptyString]
-        DOMString label, optional RTCDataChannelInit dataChannelDict)</dt>
-
+        DOMString label, optional RTCDataChannelInit dataChannelDict)
+        </dt>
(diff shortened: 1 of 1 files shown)


stefhak:
The actual transfer algorithm is not described. Look at e.g. https://html.spec.whatwg.org/multipage/scripting.html#transferCanvasProxy or https://html.spec.whatwg.org/multipage/comms.html#transferMessagePort (the transfer for ArrayBuffer also describes some error handling for the case when the ArrayBuffer was already neutered).

alvestrand:
Tagging @annevk because I think the comment originated from him.

annevk:
Yeah, you will need to define what transferring actually means (presumably you create a new instance within the other realm and move internal slot) and how it affects the original object (presumably some operations will start failing). https://github.com/whatwg/html/pull/444 has another example of an object that is being made transferable.

stefhak:
This PR (and #317 ) needs work. If no one does the work it can't be part of 1.0.

alvestrand:
Closing while preparing for 1.0. It can be reopened later.

PR #446, opened 2016-01-05
non-nullable RTCTrackEvent args means Init dict members are required.

jan-ivar:
Mozilla's webidl compiler actually requires this or the binding code wont compile (it's undefined what happens when no args are passed). It's either this or make the RTCTrackEvent args nullable and default to `null`.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4307,7 +4307,7 @@ sender.setParameters(params)
           associated with the <code><a>RTCRtpReceiver</a></code> identified by
           <code>receiver</code>.</p>
         </dd>
-        <dt>required sequence&lt;MediaStream&gt; streams</dt>
+        <dt>sequence&lt;MediaStream&gt; streams = []</dt>
         <dd>
           <p>The <code>streams</code>
            attribute returns an array of <code><a>MediaStream</a></code> objects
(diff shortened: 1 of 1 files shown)


martinthomson:
This is the best option.  I wouldn't make streams required though.  Prose can indicate that absence == [].

adam-be:
Based on @martinthomson's idea with streams. Can't we use [] as a default value instead of describing it in prose? It seems valid by the grammar.

jan-ivar:
I like that.

jan-ivar:
Does Travis dislike `[]` as a default? Does anyone know a way around this?

```
5: at '}': expected ';'
error found in IDL #54
dictionary RTCTrackEventInit : EventInit {
    required RTCRtpReceiver        receiver;
    required MediaStreamTrack      track;
             sequence<MediaStream> streams = [];
*** };[ Constructor (DOMString type, RTCTrackEventInit eventInitDict)]
interface RTCTrackEvent : Event {
    readonly        attribute RTCRtpReceiver           receiver;
    readonly        attribute MediaStreamTrack         track;
    readonly        attribute FrozenArray<MediaStream> streams;
};------------------------------
make: *** [check] Error 64
```

dontcallmedom:
I'll look into it; probably a bug in my WebIDL parser.

dontcallmedom:
fixed, travis now happy

PR #449, opened 2016-01-06
Clarify flow of SDP exchanges

### LINKED ISSUE #388: Section 10.1 (Simple Peer-to-peer Example) should explicitly handle "answer" type
The sample, while functionally correct, doesn't actually look for the type "answer" on the SDP; instead, it assumes that anything that's not an "offer" is an "answer". While this may or may not be true in the context of this specific example, it makes the example unclear -- it took me a while, I'm afraid, to realize that these two lines were in fact handling the "answer":

```
        } else
            pc.setRemoteDescription(desc).catch(logError);
```

I propose changing this to:

```
        } else if (desc.type == "answer") {
            pc.setRemoteDescription(desc).catch(logError);
        } else {
            log("Unsupported SDP type. Your code may differ here.");
        }
```

This would make the example easier and faster to comprehend and actually makes the entire onmessage handler much more obvious.

dontcallmedom:
close #388

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6765,8 +6765,11 @@ signalingChannel.onmessage = function (evt) {
                 signalingChannel.send(JSON.stringify({ "desc": pc.localDescription }));
             })
             .catch(logError);
-        } else
+        } else if (desc.type == "answer") {
             pc.setRemoteDescription(desc).catch(logError);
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM

Issue #383, opened 2015-11-05
Keep event handler attribute descriptions simle (don't specify when called)


adam-be:
We should not have any text in the event handler attribute descriptions saying that this event handler is fired when this or that happens. The event handler attributes are fired as part of the event flow procedure and that's redundant to describe for every single attribute handler.

elagerway:
@adam-be PR ?

adam-be:
Proposed fix: #451

PR #451, opened 2016-01-07
Clean up event handler attribute descriptions (fixes:  #383)

adam-be:
Text were removed from the following event handler attribute descriptions. The comment hints about where the functionality is specified in the document (so nothing is lost). Also fixed a minor bug in the Event Summary section (s/Event/RTCDTMFToneChangeEvent).

signalingstatechange: Event firing specified in "set an RTCSessionDescription" steps 3.2.5

iceconnectionstatechange: Event firing specified by "update the ICE connection state" steps (referenced by several algorithms in the Operations section)

onicegatheringstatechange: Event firing specified by "update the ICE gathering state" steps (referenced by several algorithms in the Operations section)

tonechange: Event firing specified in insertDTMF() steps 6.1 and 6.5. Info about tone is documented by the event.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1586,30 +1586,18 @@
           <dt>attribute EventHandler onsignalingstatechange</dt>
 
           <dd>The event type of this event handler is <code><a href=
-          "#event-signalingstatechange">signalingstatechange</a></code>. It is called any
-          time the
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM

Issue #338, opened 2015-10-16
specify what parts of the SDP can be changed between the createOffer and setLocalDescription


dontcallmedom:
The spec as a "TODO" that states:
> specify what parts of the SDP can be changed between the createOffer and setLocalDescription

so raising it as an issue. 

It may just need to reference to [JSEP section 6](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-11#section-6).

alvestrand:
Should reference JSEP section 6 - "Configurable SDP attributes".

stefhak:
#456 was just merged, closing this issue.

PR #456, opened 2016-01-07
SDP changes between the createOffer and setLocalDescription

aboba:
In Section 4.3.2.3, insert reference to JSEP Section 6 and remove the note on Issue 338.  

Issue: https://github.com/w3c/webrtc-pc/issues/338

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1111,13 +1111,13 @@
             until a final answer is received, at which point the
             <code><a>RTCPeerConnection</a></code> can fully adopt the pending local
             description, or rollback to the current description if the remote side
-            rejected the change. </p>
+            rejected the change.</p>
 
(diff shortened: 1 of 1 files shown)


aboba:
Thank you, Dom for the fix to the JSEP reference.  How were you able to patch the PR (thought we had a permissions issue)?

dontcallmedom:
@aboba since this pull request comes from the shared repo (in the w3c org rather than your own account), anybody with write access to the repo can edit it.

dontcallmedom:
I think instead of "just" linking to JSEP from `setLocalDescription`, this link should be added as one of the "error" step of *set[ting] an `RTCSessionDescription`*, and it should tell what error is sent back  — I suggest [`InvalidModificationError`](http://heycam.github.io/webidl/#invalidmodificationerror).

PR #460, opened 2016-01-08
getRemoteCertificates() behavior in "new" and "connecting" states

aboba:
Issue https://github.com/w3c/webrtc-pc/issues/461

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4071,7 +4071,11 @@ sender.setParameters(params)
 
           <dd>
             <p>Returns the certificate chain in use by the remote side, with each certificate
-            encoded in binary Distinguished Encoding Rules (DER) [[!X690]].</p>
+            encoded in binary Distinguished Encoding Rules (DER) [[!X690]]. If the remote peer offers
+            multiple certificates, then <code>getRemoteCertificates()</code> will return null until
+            the DTLS negotiation selects the remote certificate, which is guaranteed to occur once
+            <code><a>RTCDtlsTransportState</a></code> transitions to "connected", but could occur earlier
+            (e.g. in "connecting").</p>
           </dd>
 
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4071,7 +4071,11 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
 
           <dd>
             <p>Returns the certificate chain in use by the remote side, with each certificate
-            encoded in binary Distinguished Encoding Rules (DER) [[!X690]].</p>
+            encoded in binary Distinguished Encoding Rules (DER) [[!X690]]. If the remote peer offers
+            multiple certificates, then <code>getRemoteCertificates()</code> will return null until
```
martinthomson: `null` or the empty list?  Returning `null` would require that the return value is nullable, which it isn't.

webrtc.html
```diff
@@ -4068,10 +4068,13 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-
+          
           <dd>
-            <p>Returns the certificate chain in use by the remote side, with each certificate
-            encoded in binary Distinguished Encoding Rules (DER) [[!X690]].</p>
+            <p>Returns the certificate chain in use by the remote side, with each certificate encoded
+            in binary Distinguished Encoding Rules (DER) [[!X690]]. If the remote side offers multiple
+            certificates, then <code>getRemoteCertificates()</code> will return an empty list prior to
+            selection of the remote certificate, which will be completed by the time
+            <code><a>RTCDtlsTransportState</a></code> transitions to "connected".</p>
```
alvestrand: If it's in "new"  (before the remote certificates are presented), what will be returned? (Suggest empty list)
The text seems to only return empty when multiple certificates are offered, which doesn't make sense for the "new" state.

Issue #316, opened 2015-09-24
Don't use Interface.attribute to refer to properties of instances


dontcallmedom:
egreping through the document, I find the following culprits:
* RTCPeerConnection.createOffer
* RTCPeerConnection.setLocalDescription
* RTCPeerConnection.createAnswer
* RTCPeerConnection.setLocalDescription
* RTCPeerConnection.addIceCandiddate
* RTCPeerConnection.getStats
* RTCPeerConnection.close
* RTCPeerConnectionIceEvent.candidate
* RTCRtpEncodingParamters.priority
* RTCPeerConnection.generateCertificate
* Crypto.subtle
* RTCRtpSender.setParameters
* RTCOutboundRTPStreamStats.targetBitrate
* RTCRtpSender.track
* RTCRtpSender.transport
* RTCRtpSender.track
* RTCRtpSender.transport
* RTCRtpSender.rtcpTransport
* RTCRtpSender.transport
* RTCRtpSender.rtcpTransport
* RTCRtpSender.rtcpTransport
* RTCRtpSender.rtcpTransport
* RTCRtpSender.transport
* RTCRtpSender.mid
* RTCRtpSender.getCapabilities
* RTCRtpSender.setParameters
* RTCRtpSender.track
* RTCRtpSender.getParameters
* RTCRtpSender.track
* RTCRtpSender.setParameters
* RTCRtpReceiver.track
* RTCRtpReceiver.transport
* RTCRtpReceiver.track
* RTCRtpReceiver.transport
* RTCRtpReceiver.rtcpTransport
* RTCRtpReceiver.transport
* RTCRtpReceiver.rtcpTransport
* RTCRtpReceiver.rtcpTransport
* RTCRtpReceiver.rtcpTransport
* RTCRtpReceiver.transport
* RTCRtpReceiver.mid
* RTCRtpReceiver.getCapabilities
* RTCDtlsTransport.transport
* RTCTrackEvent.track
* RTCDataChannel.label
* RTCDataChannel.ordered
* RTCDataChannel.maxPacketLifeTime
* RTCDataChannel.maxRetransmits
* RTCDataChannel.protocol
* RTCDataChannel.negotiated
* RTCDataChannel.id
* RTCDataChannel.priority
* RTCDataChannel.readyState
* RTCStats.id

(ignoring the ones in the changelog)

martinthomson:
[...]

PR #462, opened 2016-01-12
Add PeerConnection.activateSender() and update early media example.

pthatcherg:


webrtc.html
```diff
@@ -3998,6 +3998,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             the <a><code>RTCRtpReceiver</code></a> to be stopped.</p>
           </dd>
 
+          <dt>void activeSender()</dt>
+
```
dontcallmedom: Shouldn't this be "activateSender"? (that's what the description below uses)

webrtc.html
```diff
@@ -3998,6 +3998,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+          <dd>
+            <p>The <dfn id="dom-rtptransceiver-activesender"><code>activateSender</code></dfn>
+            method actives the <code>RTCRtpTransceiver.sender</code> such that future calls
```
dontcallmedom: s/actives/activates/

webrtc.html
```diff
@@ -3998,6 +3998,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            to <code>createOffer</code> and <code>createAnswer</code> to mark the corresponding
```
dontcallmedom: s/to mark/mark

webrtc.html
```diff
@@ -6866,19 +6879,16 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
     pc.ontrack = function (evt) {
         if (evt.track.kind === "audio") {
           if (answerer) {
-            // TODO: Figure out how to reuse an existing m-line with something like:
-            // audio = evt.transceiver;
-            // audio.sender.activate();
-            // This avoids the need for an extra re-offer.
-            // Similarly for video.
-            audio = pc.addTransceiver(evt.track.kind);
+            audio = evt.transceiver;
```
dontcallmedom: as far as I can tell, the `RTCTrackEvent` interface doesn't have a `transceiver` attribute.

webrtc.html
pthatcherg: Fixed.

webrtc.html
pthatcherg: Yes, thanks for catching that.  It is fixed now.

webrtc.html
pthatcherg: Fixed.

webrtc.html
pthatcherg: Good point.  I added it to RTCTrackEvent.

webrtc.html
```diff
@@ -3998,6 +3998,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+          <dd>
+            <p>The <dfn id="dom-rtptransceiver-activatesender"><code>activateSender</code></dfn>
+            method activates the <code>RTCRtpTransceiver.sender</code> such that future calls
+            to <code>createOffer</code> and <code>createAnswer</code> mark the corresponding
+            media description as <code>sendrecv</code> or <code>sendonly</code>, as defined in
+            [[!JSEP]].
+
+            <p>If the sender is not already active and the PeerConnection is in the stable state,
+            this will trigger <code>onnegotiationneeded</code>.
+          </dd>
```
alvestrand: Shouldn't this say "this will set the negotiation-needed flag", and leave off mentioning the stable state?
I assume that this sequence:

pc.createOffer() => offer
pc.setLocal(offer)
track.activateSender()
pc.setRemote(answer)  // state is now stable

expects a "negotiationneeded" to be fired after the setRemote() is called.

Also, "trigger onnegotiationneeded" is language we're not using - it's either "set negotiation-needed" or "fire a negotiationneeded event" (I think the former).

pthatcherg: Fixed

webrtc.html
```diff
@@ -3998,6 +3998,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             the <a><code>RTCRtpReceiver</code></a> to be stopped.</p>
           </dd>
 
+          <dt>void activateSender()</dt>
+
+          <dd>
+            <p>The <dfn id="dom-rtptransceiver-activatesender"><code>activateSender</code></dfn>
+            method activates the <code>RTCRtpTransceiver.sender</code> such that future calls
```
dontcallmedom: we're moving away from using Foo.bar to refer to attribute (see #316); could use just use `<code>sender</code>`?

aboba: Line 4011:  Extraneous </code> is causing a Travis build failure.  Should it be </p>?

webrtc.html
```diff
@@ -3998,6 +3998,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            to <code>createOffer</code> and <code>createAnswer</code> mark the corresponding
+            media description as <code>sendrecv</code> or <code>sendonly</code>, as defined in
+            [[!JSEP]].
+
+            <p>If the sender is not already active and the PeerConnection is in the stable state,
+            this will set the negotiation-needed flag.</code>.
```
alvestrand: You have a </code>. that looks completely out of plae. Remove it?

aboba:
Yes, it should be removed.

elagerway:
@pthatcherg would you have time for this one ?

burnburn:
Editorial.  I'll fix and merge.

PR #463, opened 2016-01-13
Add more explicit language about transceivers, MIDs and rollbacks.

pthatcherg:


webrtc.html
```diff
@@ -2251,6 +2251,10 @@
             state. Note the local or remote SDP descriptions in the
 	    previous stable state could be null if there has
             not yet been a successful offer-answer negotiation.</p>
+
+            <p>If the value of a <code>RTCRtpTransceiver.mid</code> was set to a non-null value by
+              an SDP description that is rolled back, the <code>RTCRtpTransceiver.mid</code> will be
+              set to null, as defined by [[!JSEP]].</p>
```
dontcallmedom: This should use the new mechanism to refer to JSEP: `<span data-jsep="[id of jsep section]">[[!JSEP]]</span>` (likewise for the two other JSEP references)

pthatcherg: OK, I added it to them.  Two of them had a clear ID.  The other.... I just made one up: "rollbacks".  Is that OK?

aboba:
There is a "Rollback" section in JSEP (Section 4.1.4.2).  So is the solution to change data-jsep="rollbacks" to data-jsep="rollback"? 

travis

dontcallmedom:
well, except that 4.1.4.2 doesn't say anything about mid at this point in JSEP.

pthatcherg:
I changed it from "rollbacks" to "rollback".  Even though JSEP doesn't say anything about rollbacks right now, that's probably the section it will go in later.  If we change the section in JSEP, we can always change it here.  

If you (or Travis) still don't like that, I'll remove it.

webrtc.html
```diff
@@ -2251,6 +2251,10 @@
             state. Note the local or remote SDP descriptions in the
 	    previous stable state could be null if there has
             not yet been a successful offer-answer negotiation.</p>
+
+            <p>If the value of a <code>RTCRtpTransceiver.mid</code> was set to a non-null value by
+              an SDP description that is rolled back, the <code>RTCRtpTransceiver.mid</code> will be
+              set to null, as defined by <span data-jsep="rollback">[[!JSEP]]</span>.</p>
```
stefhak: Isn't this a bit ambiguous in the sense that if an SDP sets the mid to one value (x), and a subsequent SDP sets it to another value (y), at rollback of the latest SDP I assume the mid should roll back to x and not to null?

pthatcherg: I don't think setting a remote description can change the .mid, except between null and a value.  I don't think it can change it from A to B.  Of course, we haven't written that part of JSEP yet, so I may be wrong.

stefhak:
@adam-be to comment.

webrtc.html
```diff
@@ -3018,6 +3022,12 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                 with <code>.sender.track</code> set to <var>track</var>, add it to
                  <var>connection</var>'s set of transceivers, and
                  return <code>.sender</code> to the caller.</p>
+
+                <p>The initial value of <code>RTCRtpTransciever.mid</code> is
+                null.  <code>setLocalDescription</code> and <code>setRemoteDescription</code> may
```
adam-be: Please link these function names.

webrtc.html
```diff
@@ -2251,6 +2251,10 @@
             state. Note the local or remote SDP descriptions in the
 	    previous stable state could be null if there has
             not yet been a successful offer-answer negotiation.</p>
+
+            <p>If the value of a <code>RTCRtpTransceiver.mid</code> was set to a non-null value by
```
adam-be: We have made a few runs to try to get rid of the "static like" syntax when referring to properties. Please don't add new ones.

alvestrand:
Adam: Can you pull it in and fix it?

alvestrand:
Assigning to Jan-Ivar for comment. It's a task within a task.

jan-ivar:
Sorry for seeing this late, @alvestrand what did you want me to comment on? What task?

alvestrand:
@jan-ivar I think I made the comment on the wrong PR. The right one is #494

PR #464, opened 2016-01-13
Re-add RTCRtpTransceiverInit.sendEncodings

dontcallmedom:
Got lost in moving the description (in 66b382a4c7c2da68405955d08ff73d5d757166b9)


aboba:
sendEncodings.active is presumably determined by directionality, no? Seems to me that we need to be careful about adding potential conflicting settings.

Issue #375, opened 2015-11-04
true as default values for dictionary is bad practice


dontcallmedom:
The [WebIDL spec](http://heycam.github.io/webidl/#idl-dictionaries) warns that:
> [it] is strongly suggested not to use of true as the default value for boolean-typed dictionary members, as this can be confusing for authors who might otherwise expect the default conversion of undefined to be used (i.e., false). 

We have true as default of:
* `ordered` in `RTCDataChannelInit`
* `voiceActivityDetection` in `RTCOfferAnswerOptions`
* `send` and `receive` in `RTCRtpTransceiverInit`

This means that e.g. `createOffer({voiceActivityDetection:undefined})` keeps voice activity detection on, which is indeed confusing.

I'm not sure if we still have room for change here; one option would be invert the property names (e.g. unordered, disableVoiceActivityDetection, receiveOnly, sendOnly).

alvestrand:
We also have some places that exhort against negative truth values (don't use "disable"). And receiveOnly = true, sendOnly = true as a synonym for "inactive" would just look weird, so it would have to be full inversion and say "disableSend" / "disableReceive".

Can we make undefined (as opposed to "missing") a TypeError somehow, so that createOffer({voiceActivityDetection: undefined}) fails fast?

martinthomson:
An `undefined` value and omission from the dictionary are undetectably identical.

One approach would be to use enumerated strings rather than booleans:

```js
createOffer({voiceActivityDetection: 'disabled'})
RTCRtpTransceiverInit('sendonly'||'recvonly'||'sendrecv'||'inactive')
```

Given that each of these is only set to something other than a default rarely, I think that this would be OK.  Also, an enumerated value will make the API use more clearly self-documenting.

dontcallmedom:
I've created a pull request for `RTCRTCTransceiverInit` and `RTCOfferOptions`, where I think the backwards compat risk is low.

But I'm not sure we can still afford to change `RTCDataChannelInit.ordered` — I assume this one is already in wide use.

martinthomson:
@jesup would know better than I, but I think that ordered could be dealt with in a backwards-compatible fashion by defining a union of bool and a new enumeration.  Or we can leave browsers to deal with the mess and just move to a new enumeration.

dontcallmedom:
in PR #471 I've used a union of Boolean and enum in the init dictionary, and set the type of the attribute in the interface to the enum.

alvestrand:
[...]

PR #467, opened 2016-01-15
Use enum for voiceActivityDetection

dontcallmedom:
part of #375, based on @martinthomson's [suggestion](https://github.com/w3c/webrtc-pc/issues/375#issuecomment-163433990)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -385,7 +385,7 @@
         the offer/answer creation process.</p>
 
         <dl class="idl" title="dictionary RTCOfferAnswerOptions">
-          <dt>boolean voiceActivityDetection = true</dt>
+          <dt>RTCVoiceActivityOption voiceActivityDetection = "enabled"</dt>
 
           <dd>
             <p>Many codecs and systems are capable of detecting "silence" and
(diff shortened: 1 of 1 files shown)


alvestrand:
As we go through the consequences of eliminating true defaults, it gets uglier. Comments wanted.

adam-be:
If vad is a feature that should be default on because that's what would benefit most people, would it be so bad to have an option called 'disableVoiceActivityDetection'?

dontcallmedom:
I rather like @adam-be 's suggestion.

alvestrand:
I don't like negative-valued options - "disableVoiceActivityDetection: false" is a double negative.

webrtc.html
```diff
@@ -398,6 +398,14 @@
           </dd>
         </dl>
 
+        <dl class="idl" title="enum RTCVoiceActivityOption">
+          <dt>enabled</dt>
+          <dd>Request voice activity detection to be enabled</dd>
+
+          <dt>disabled</dt>
+          <dd>Request voice activity detection to be disabled</dd>
```
pthatcherg: Should we have periods at the end of the sentences?

pthatcherg: Otherwise, I'm fine with this.

webrtc.html
```diff
@@ -385,7 +385,7 @@
         the offer/answer creation process.</p>
 
         <dl class="idl" title="dictionary RTCOfferAnswerOptions">
-          <dt>boolean voiceActivityDetection = true</dt>
+          <dt>RTCVoiceActivityOption voiceActivityDetection = "enabled"</dt>
```
pthatcherg: Can we do the same "or boolean" trick here as with PR 471 to make it more convenient?

alvestrand:
Closed based on list call-for-consensus results.

PR #468, opened 2016-01-15
Replace DOMError with DOMException

### LINKED ISSUE #335: DOMError is removed from the platform; use DOMException instead

dontcallmedom:
close #335

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1011,7 +1011,7 @@
             session description SHALL contain an appropriate assertion.  If the
             identity provider is unable to produce an identity assertion, the
             call to <code>createOffer</code> MUST be rejected with a
-            <code>DOMError</code> that has a name of <code>IdpError</code>.</p>
+            <code>DOMException</code> that has a name of <code>IdpError</code>.</p>
 
             <p>If this <code>RTCPeerConnection</code> object is closed before
             the SDP generation process completes, the user agent MUST suppress
@@ -1033,7 +1033,7 @@
 
             <p>If the SDP generation process failed for any other reason, the
             user agent MUST reject the returned promise with an
-            <code>DOMError</code> object of type TBD as its argument.</p>
+            <code>DOMException</code> object of type TBD as its argument.</p>
           </dd>
 
           <dt>Promise&lt;RTCSessionDescription&gt; createAnswer (optional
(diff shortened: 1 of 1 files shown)


burnburn:
This is not a pre-defined DOMException, and per the discussion on the gUM spec we are not allowed to add to the list of defined DOMException values.  We need to use a pre-defined one or define a new error object as in gUM.  Off the top of my head I don't know which pre-defined value would be appropriate.

alvestrand:
Dan has promised to do a pass over the error handling to bring it in line with the (accepted) approach taken in getUserMedia. We'll hold this until that.

dontcallmedom:
note that I have a pull request that goes over the errors to align them with existing error names: https://github.com/w3c/webrtc-pc/pull/472

alvestrand:
RTCSdpError was intended as the result error from SDP failures in Set{local/remote}Description.
Is it gone from there?

dontcallmedom:
it was never added there (error management has been marked at TBD in setLD and setRD for ever)

dontcallmedom:
(I should say it's also telling that RTCSdpError was also never implemented, nor have we seen lots of complaints about it as far as I can tell)

alvestrand:
Given the headaches that implementing the name field in ConstraintError gave me, I can't say that I'm eager to defend SDPError's continued existence.
Interim topic?

dontcallmedom:
I think the case for SDPError is also weaker than ConstrainError, since it's very unlikely that anyone can take programmatic actions based on the line number where the error was detected in the SDP.

I'll bring it to the list (since after all that's part of the changes I'm proposing); depending on the reaction, making this an interim topic may or may not be needed.

burnburn:
I understand now that all you are doing here is to rename DOMError to DOMException and then adjusting the individual errors in PR #472.  Merging.

foolip:
Did you forget to push this to gh-pages? https://w3c.github.io/webrtc-pc/ doesn't include the fix and I spent time worrying about it in https://codereview.chromium.org/1661493002/ before finding this :(

alvestrand:
We don't auto-push to gh-pages, only when publishing a new editors' draft (every few weeks).
The reasons for this are convoluted.

PR #481, opened 2016-01-26
Add missing ReSpec configuration options related to current specStatus

vivienlacourba:
specStats: 'WD' requires the previousPublishDate and previousMaturity options to be set.
This is needed to set up automatic publication with link such as:
http://w3c.github.io/webrtc-pc/webrtc.html?specStatus=WD

### DIFF
--- a/webrtc.js
+++ b/webrtc.js
@@ -56,6 +56,10 @@ var respecConfig = {
   //      company: "Your Company", companyURL: "http://example.com/" }
   //],
 
+  // Requirements of the current specStatus: 'WD'
+  previousMaturity: "WD",
+  previousPublishDate: "2015-02-10",
+
   // name of the WG
   wg:           "Web Real-Time Communications Working Group",
 


dontcallmedom:
@vivienlacourba my understanding is that previousMaturity and previousPublishDate is set automatically upon publication by echidna (via the specgenerator), so I don't think that patch is necessary.

vivienlacourba:
Well without that change I was not able to see the expected result when accessing: http://w3c.github.io/webrtc-pc/webrtc.html?specStatus=WD (see the ReSepec warning).
Which also prevented me from using the [Specberus publication checker](https://labs.w3.org/pubrules/) on that URI (something I wanted to do before setting automatic publication by echidna).

dontcallmedom:
you should use http://w3c.github.io/webrtc-pc/webrtc.html?specStatus=WD;previousMaturity=WD;previousPublishDate=2015-02-10 for that :)

vivienlacourba:
Ok that works. You can close this merge request if you think the benefit is too limited. Thx.

PR #485, opened 2016-01-28
Update SOTD as the document is now quite stable and the group is look…

vivienlacourba:
…ing for wide review.

This updated text is used in the latest WebRTC spec published on /TR and was drafted with @dontcallmedom 's help:
  https://www.w3.org/TR/2016/WD-webrtc-20160128/

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -29,26 +29,11 @@
   </section>
 
   <section id="sotd">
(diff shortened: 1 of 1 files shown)


alvestrand:
Slightly optimistic, but let's go.

Issue #488, opened 2016-01-29
getSenders/getReceivers: optional kind parameter?


fippo:
while playing with simulcast I noticed that neither [getSenders](http://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-getSenders-sequence-RTCRtpSender) nor [getReceivers](http://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-getReceivers-sequence-RTCRtpReceiver) (and getTransceivers?) takes a 'kind' member specifying the media type.

What I was originally looking to do was along the lines of
```
pc.getSenders('video')[0].setParameters(...)
```
to configure simulcast on the first video track. Trivial to work around by using .filter on the array of senders returned:
```
pc.getSenders().filter(function(sender) { return sender.track.kind === 'video'; })[0]
.setParameters(...)
```

But since [getCapabilities](http://w3c.github.io/webrtc-pc/#widl-RTCRtpReceiver-getCapabilities-RTCRtpCapabilities-DOMString-kind) takes an optional kind parameter that seems slightly inconsistent.

jan-ivar:
This seems supefluous to me. What's wrong with:

    pc.getSenders().filter(sender => sender.track.kind == 'video');

?

jan-ivar:
> But since getCapabilities takes an optional kind parameter that seems slightly inconsistent.

I don't think so. If we were talking about mediaStreamTrack.getCapabilities() having a `kind` parameter, then yes I agree it would be consistent with `pc.getSenders()` having one, since those would both be simple dictionary filters.

But `receiver.getCapabilities()` is not that. It's a selector between two value-sets of the same type, since video has `RTCRtpCapabilities`, and audio also has `RTCRtpCapabilities`, but with different values.

jan-ivar:
(mediaStreamTrack.getCapabilities() doesn't really make sense either, but take getSupportedConstraints() where we in fact did have a kind parameter for a brief while, before we realized it was a simple dictionary-filter.

jan-ivar:
In other words:
* the `kind` in `receiver.getCapabilities` is an index
* the `kind` in `pc.getSenders` is a filter

fippo:
at least in Edge, the getCapabilities behaviour seems more like a filter to me.
```
RTCRtpSender.getCapabilities()
```
returns a set of audio and video codecs which could be filtered by looking at their kind. But if you do that, should there not be a consistent way to do this filtering?

[...]

PR #489, opened 2016-01-29
getSenders/getReceivers: optional kind parameter

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/488


fippo:
exactly what I was looking for!

aboba:
As noted in the issue discussion, equivalent functionality can be obtained via a filter() callback.  So before merging this PR, we need WG consensus that it is needed.

alvestrand:
Declaring consensus against merging this.

Issue #492, opened 2016-02-05
avoid risk of reentrancy


ShijunS:
Open an issue based on discussions with Bernard.  We should avoid any risk of reentrancy, and meanwhile clarify a few places for firing events in the related algorithm steps.

Section 4.3.1
----------
 ...
 If the process to apply description fails for any reason, then user agent must queue a task runs the following steps:
1. If connection's signaling state is closed, then abort these steps.
2. If the content of description is invalid or if description's type is wrong for the current signaling state of connection, then reject p with an InvalidSessionDescriptionError and abort these steps.
3. If description cannot be applied at the media layer, but the User Agent recovered, possibly by rolling back to the previous configuration, then reject p with IncompatibleSessionDescriptionError and abort these steps.
This occurs, for example, when the version of description is older than the one currently being used.
4. Otherwise, set connection's signaling state to closed.
5. Fire a simple event named signalingstatechange at connection. 
**[BA] Allows the signalingstatechange to execute code that could invalidate objects.** 
6. Reject p with InternalError.
...
...
5. If connection's signaling state changed above, fire a simple event named signalingstatechange at connection.
**[BA] Allows code to execute in the signalingstatechange event handler (e.g. PC.close) that could invalidate objects.**
6. If description is set as a local description, connection's ICE gathering state is new, and description contains media, then update connection's ICE gathering state to gathering.
**[BA] ICE gathering state is updated but the state change event isn't fired. Is a task queued for this?** 
7. If the process to apply description resulted in an ICE restart JSEP, then run the following steps:
  If connection is not already gathering, update connection's ICE gathering state to gathering.
  **[BA] ICE gathering state is updated but the state change event isn't fired. Is a task queued for this?** 
  If connection's ICE connection state is completed, update connection's ICE connection state to connected.
  **[BA] ICE connection state is updated but the state change event isn't fired. Is a task queued for this?**
8. If description is set as a remote description with new media descriptions [JSEP], the User Agent must dispatch a receiver for all new media descriptions.
9. If connection's signaling state is now stable, and the negotiation-needed flag is set, the User Agent must queue a task to fire a simple event named negotiationneeded at connection and clear the negotiation-needed flag.
10. Resolve p with undefined.
...

Section 6.2
------
...
2. Set channel's readyState attribute to closed.
3. If the transport was closed with an error, fire an NetworkError event at channel.
**[BA] The NetworkError event should be queued to avoid having an error handler execute prior to the next step.**
[...]

PR #494, opened 2016-02-09
proposed fix to issue #492

ShijunS:
The proposal is to queue a task to fire the corresponding events to avoid risk of reentrancy (#492).

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -720,7 +720,7 @@
                   </li>
 
                   <li>
-                    <p>Fire a simple event named <code><a href=
+                    <p>Queue a task to fire a simple event named <code><a href=
                     "#event-signalingstatechange">signalingstatechange</a></code>
                     at <var>connection</var>.</p>
                   </li>
@@ -866,7 +866,7 @@
 
                   <li>
                     <p>If <var>connection</var>'s <a>signaling state</a>
-                    changed above, fire a simple event named <code><a href=
+                    changed above, queue a task to fire a simple event named <code><a href=
                     "#event-signalingstatechange">signalingstatechange</a></code>
                     at <var>connection</var>.</p>
                   </li>
@@ -4685,7 +4685,7 @@ sender.setParameters(params)
         <li>
           <p>If the <a href="#dfn-underlying-data-transport">transport</a> was
           closed <dfn id="data-transport-closed-error">with an error</dfn>,
-          fire an NetworkError event at <var>channel</var>.</p>
+          queue a task to fire an NetworkError event at <var>channel</var>.</p>
         </li>
 
         <li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -720,7 +720,7 @@
                   </li>
 
                   <li>
-                    <p>Queue a task to fire a simple event named <code><a href=
+                    <p>Fire a simple event named <code><a href=
                     "#event-signalingstatechange">signalingstatechange</a></code>
                     at <var>connection</var>.</p>
                   </li>
@@ -4685,7 +4685,7 @@ sender.setParameters(params)
         <li>
           <p>If the <a href="#dfn-underlying-data-transport">transport</a> was
           closed <dfn id="data-transport-closed-error">with an error</dfn>,
-          queue a task to fire an NetworkError event at <var>channel</var>.</p>
+          fire an NetworkError event at <var>channel</var>.</p>
         </li>
 
         <li>


webrtc.html
```diff
@@ -720,7 +720,7 @@
                   </li>
 
                   <li>
-                    <p>Fire a simple event named <code><a href=
+                    <p>Queue a task to fire a simple event named <code><a href=
```
jan-ivar: We are already in a queued task here, so reentering that task is impossible.

Nor is there any risk of garbage collection, since the peer connection is held alive by the stack.

This change alters the order in which a signalingstatechange event-handler and a promise onRejected function are called. Is that desirable?

webrtc.html
```diff
@@ -4685,7 +4685,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
         <li>
           <p>If the <a href="#dfn-underlying-data-transport">transport</a> was
           closed <dfn id="data-transport-closed-error">with an error</dfn>,
-          fire an NetworkError event at <var>channel</var>.</p>
+          queue a task to fire an NetworkError event at <var>channel</var>.</p>
```
jan-ivar: Same here.

webrtc.html
```diff
@@ -866,7 +866,7 @@
 
                   <li>
                     <p>If <var>connection</var>'s <a>signaling state</a>
-                    changed above, fire a simple event named <code><a href=
+                    changed above, queue a task to fire a simple event named <code><a href=
```
jan-ivar: This one makes sense I think, though I observe that elsewhere (like for ICE gathering state directly below) we tend to queue a task to set the state *and* fire the event, though here it may matter more that the state is set early? Should we worry that the state may have changed again by the time the event fires? I'm unsure.

alvestrand: We have had issues with the state being overwritten by a second event after the task was queued but before the event was fired, so yes, I think we should set the state within the queued task if we queue it (otherwise, the user observes a surprising state transition).

stefhak: @alvestrand are you arguing we should leave things as they are (i.e. we queue a task that potentially changes the state, and if it does it fires the event rather than queuing another task to fire the event)?

webrtc.html
stefhak: (maybe this is what you're saying @jan-ivar ) this change would mean we change the order of the NetworkError and close events.

webrtc.html
adam-be: We want the state update to happen synchronously as the event that tells the script about it, so we don't want to queue an other task here. @jan-ivar, does it matter if we fire the event first or reject the promise? I.e. change the order of step 6 and 7 in this algorithm.

webrtc.html
alvestrand: Looking back, I think it's right as it is (no change). The event should fire after the state change, but before any possible later state change.
(It took me a while to realize that "fire an event" means "call the event handler NOW", not "put an event in a queue and call the event handler later").

adam-be: Note that all operations like "update connection's ICE gathering/connection state" are queued in the steps that are liked to.

The intention here is to have the state updated and receivers dispatched for all new incoming media before the promise "then-handler" is run. So signalingstatechange needs to be fired synchronously as the state is updated. The question is we need to the synchronous work later in the steps.

webrtc.html
adam-be: This one is a bit strange. Is the "NetworkError event" even defined somewhere? I think we could do with one event here.

webrtc.html
jan-ivar: @adam-be the difference would only be observable in a JS debugger, because a script can only learn of a rejection through functions passed to `.then` or `.catch`, which are guaranteed to always execute on a microtask.

webrtc.html
jan-ivar: @adam-be But as you just mentioned, "update connection's ICE gathering/connection state" are queued. That means they already happen *after* functions passed to `.then` execute, because promises operate on the microtask queue, whereas "queue a task" refers to the main event loop.

jan-ivar: Also, of the three changes in this patch, I think @ShijunS has a point on this one that it's perhaps unwise to dictate substantive steps involving non-stack objects after requiring the UA to synchronously execute content script code, as it's a window to potentially trip up assumptions in the remaining steps in some implementations. For instance, an implementation that didn't properly root the peer connection object on the stack in this queued task, might not take kindly to content script removing its last peer connection reference. I believe that's what he's referring to as reentrancy.

jan-ivar: Although I just spoke with our DOM team and they're saying that "In practice, web algorithms are treated as closing over everything they might care about that they don't get from somewhere explicitly".

webrtc.html
ShijunS: It looks like there is no objection to @jan-ivar's comment.   Meanwhile, the Mozilla team has confirmed offline that there is no concern from browser implementation perspective.  I will remove this specific change from the proposal.

webrtc.html
ShijunS: Based on comment from @Jan-Ivar and feedback from the Mozilla team, I will remove the specific change from the proposal.  @adam-be, it might be better to open a new issue to address the question on the NetworkError event.

adam-be:
> Based on comment from @Jan-Ivar and feedback from the Mozilla team, I will remove the specific change from the proposal. @adam-be, it might be better to open a new issue to address the question on the NetworkError event.

I got this comment from @ShijunS in my email inbox, but I can't find it on github so I'm quoting it. @ShijunS, please do create a new issue on the RTCDataChanel NetworkError event.

ShijunS:
@adam-be, issue #526 is opened to track the question on NetworkError event.

adam-be:
Thank you @ShijunS

webrtc.html
adam-be: If we want to queue a new task here I think we want to do the state update in that task as well. But I didn't really understand from @jan-ivar if it was necessary. What do you say @jan-ivar?

alvestrand:
@jan-ivar this was the PR I wanted your comment on - the proposed 1-line change would queue a task from within a task that's already been queued. You tend to have thought about this kind of thing.

jan-ivar:
I don't know if a change is necessary. @ShijunS seemed to feel strongly about it though. It is certainly true that, after having made a synchronous call to content JS, one must either re-check that one still has everything one needs, or make sure to follow a model where things are well-rooted (reference-counted) at the start of the task, such that reference-counts wont go to zero and GC'ed objects could be encountered. To me, that would only seem to be problematic in poorly written implementations, e.g. C++ code that stores a raw pointer on the stack before the synchronous call to JS, and relies on using it again after.

It would be easier if we didn't make this change, because even though I agree with @adam-be that it'd make sense to want to do the state update in that new task as well, doing so would update the state too late to be observable in the success callback, because a "queued task" happens on the next JS event loop, which is later than `.then` success callbacks which happen on the micro-task queue. (The micro-task queue is completely emptied at the end of the current run-to-completion in most browsers).

jan-ivar:
TL;DR: If I go by our DOM team, then no change is necessary (they say "in practice, web algorithms are treated as closing over everything they might care about that they don't get from somewhere explicitly").

ShijunS:
There are events queued in steps 6 and 7 in the same algorithm.  I expect it will be safer to queue the event at step 5 as well.  All the events in the specific algorithm will at least be queued in the same place so stay in the same order.  @Alvestrand, it seems some browser implementation might have the problem described by @Jan-Ivar.  Meanwhile, the Firefox folks have confirmed there is no problem with the Firefox implementation.  So, I don't feel as strongly as previously about the potential risk.  It can be considered as an implementation issue, unless other browser vendor has different preference.

adam-be:
I believe the current text reflects the groups desire to have the signalling state updated by the time the promise resolved (as @jan-ivar talks about above). Should we close this for now and reopen if people have a hard time implementing it as written?

jan-ivar:
wfm.

alvestrand:
Conclusion is that we should close this without merging. Branched-off discussions have been moved to other bugs.

Issue #497, opened 2016-02-11
RID as read-only in setParameters()


aboba:
The current text does not indicate that the RID should be considered read-only within setParameters().

adamroach:
I don't think that's where we settled on the topic. Looking at the minutes from Sapporo (https://www.w3.org/2015/10/28-webrtc-minutes), I see:

  peter: ok JS picks, no more than 16 characters
  martin: not unicode please
  peter: A~&, 0~9, both cases
  all: good

This implies that the JS must have the ability to write the RID.

aboba:
@adamroach - you are correct with respect to addTransceiver()  where init.sendEncodings enables setting of the RID value prior to calling createOffer().  However, the question has arisen as to whether setParameters() can be used to change the RID value.  The text recently added in PR https://github.com/w3c/webrtc-pc/pull/434 states: 

The setParameters does not cause renegotiation at the SDP level
and can only be used to change what the media stack is sending or
receiving within the envelope negotiated by the Offer/Answer.

So the question is whether setParameters() can change the RID value without requiring a new O/A exchange.

adamroach:
@aboba -- Ah, yes, I missed the context here. I agree that this can't be changed once the session is established, since that would require renegotiation.

aboba:
@adamroach - can you look at PR https://github.com/w3c/webrtc-pc/pull/498 to see if it makes sense? 

Also, whether the logic described here is accurate: 
https://lists.w3.org/Archives/Public/public-webrtc/2016Feb/0028.html

adamroach:
@aboba -- The logic in the email is sound, but the proposed change has an issue. I've commented in PR #498 about the issue.

jan-ivar:
@aboba so currently in Firefox Nightly, users can [do this](https://jsfiddle.net/kv14z7r6/):

    var sender = pc.addTrack(stream.getVideoTracks()[0], stream);
    sender.setParameters({ encodings: [{ rid: "A",  maxBitrate: 300000 },
                                       { rid: "B",  maxBitrate: 100000 }] });
    pc.createOffer().then(d => pc1.setLocalDescription(d));

Works without having to use `addTranceiver`. Did we decided this should not work?

jan-ivar:
Note to self: probably need to queue createOffer behind setParameters here now that it is async.

aboba:
I agree with Peter's comment that:  
[...]

PR #498, opened 2016-02-11
RID as read-only in setParameters()

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/497


adamroach:
This isn't quite right. The RTCRtpEncodingParameters are used in the RTCRtpTransceiverInit dictionary, which is where the application is going to set the value initially.

I think what you need to do here is either decide whether changing the rid results in an error or whether it is silently ignored, and then specify that behavior in the RTCRtpSender.setParameters().

aboba:
Closing this pull request in order to create a new one addressing the mailing list discussion.

Issue #496, opened 2016-02-10
Certificate API: Do we need certificate introspection?


alvestrand:
At the moment, the certificate API we provide is this:

interface RTCCertificate {
    readonly        attribute DOMTimeStamp expires;
};

A question was raised on whether we could tell if a given cert was RSA or ECDSA - there doesn't seem to be one.

Would it be reasonable to require an introspection API that is able to tell something about it (possibly a subset of its keygen parameters)?

henbos:
Getting an AlgorithmIdentifier, same format as the input for RTCPeerConnection.generateCertificate, makes sense to me. That is an established format and allows you to tell the difference between, say, RSA-1024 and RSA-2048.

martinthomson:
I couldn't think of a reason that this would be needed in practice, hence my initial patch not including anything.  It's easy to implement, I think, but much harder to un-implement.  Does anyone have a use case?

henbos:
I'm thinking it would probably be most useful for debugging or unit tests. I don't think this is strictly necessary, just that it might be nice to have in some cases. Like if you're generating multiple certificates it might be nice to be able to tell them apart based on inspection, not just based on which promise was resolved. Or debugging, being able to print "this is the cert we use" and that actually meaning something. And its funny you can do getConfiguration to get the list of certificates, but you can't tell anything about them.

If you don't specify which certificate(s) to use, RTCPeerConnection constructor creates certificate(s) for you. In this case, getConfiguration to get a set of certificates tells you nothing about which certificates are used. But I suppose you shouldn't have to know?

aboba:
The use case would probably involve situations where the same certificate is used in multiple calls so that it might be handy not to have to store the associated keygenAlgorithm values. 

An example came up in ORTC to support forking, where multiple DtlsTransport objects would need to use the same certificate/fingerprint.  This lead to RTCCertificate being extended with the fingerprint (RSA vs. ECDSA wasn't relevant in that case): 

partial interface RTCCertificate {
    readonly        attribute RTCDtlsFingerprint fingerprint;
};

dictionary RTCDtlsFingerprint {
             DOMString algorithm;
             DOMString value;
};

martinthomson:
Regarding fingerprint, I think that's a fine thing to add.  However, note that the same comments apply to that attribute as I made on #499.

martinthomson:
Oh, and the formulation Bernard describes doesn't work if we have to entertain the notion of hash agility.

aboba:
[...]

PR #499, opened 2016-02-11
Certificate API: add getAlgorithm method

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/496

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2789,6 +2789,12 @@
             <p>Note that this value might not be reflected in
             a <code>notAfter</code> parameter in the certificate itself.</p>
           </dd>
+          <dt>readonly attribute AlgorithmIdentifier keygenAlgorithm</dt>
+          <dd>
+          	<p>The value of <var>keygenAlgorithm</var> passed in the call
+          	to <code>generateCertificate()</code> that returned the
+          	<code>RTCCertificate</code> object.</p>
+          </dd>	
         </dl>
 
         <p>For the purposes of this API, the
--- a/webrtc.html
+++ b/webrtc.html
@@ -2789,11 +2789,12 @@
             <p>Note that this value might not be reflected in
             a <code>notAfter</code> parameter in the certificate itself.</p>
           </dd>
-          <dt>readonly attribute AlgorithmIdentifier keygenAlgorithm</dt>
+          <dt>AlgorithmIdentifier getAlgorithms()</dt>
           <dd>
-          	<p>The value of <var>keygenAlgorithm</var> passed in the call
-          	to <code>generateCertificate()</code> that returned the
-          	<code>RTCCertificate</code> object.</p>
+          	<p>
+          		Returns the value of <var>keygenAlgorithm</var> passed in the call
+          		to <code>generateCertificate()</code>.
+          	</p>
           </dd>	
         </dl>
 
--- a/webrtc.html
+++ b/webrtc.html
@@ -2789,7 +2789,7 @@
             <p>Note that this value might not be reflected in
             a <code>notAfter</code> parameter in the certificate itself.</p>
           </dd>
-          <dt>AlgorithmIdentifier getAlgorithms()</dt>
+          <dt>AlgorithmIdentifier getAlgorithm()</dt>
           <dd>
           	<p>
           		Returns the value of <var>keygenAlgorithm</var> passed in the call


martinthomson:
Unfortunately, I think that this doesn't reflect current best practice.  I hear all sorts of complaints about performance and other nasty side effects of returning dictionaries as attribute values.  I think that what we actually want here is something more prescriptive:

```
partial interface RTCCertificate {
  [SameObject]
  readonly attribute RTCCertificateAlgorithm algorithm; 
  // I've shortened the name a little here, that's just a taste thing
};

interface RTCCertificateAlgorithm {
  readonly attribute DOMString name;
  readonly attribute DOMString namedCurve; // ECDSA
  readonly attribute ArrayBuffer publicExponent; // RSA
  readonly attribute unsigned long modulusLength; // RSA
  readonly attribute DOMString hash; // RSA
  // other attributes as we get EdDSA and other things.
};
```

alvestrand:
@martinthomson do you have a public cite for the problem of returning dictionaries?
I'd like to check whether that's a Firefox problem or a generic problem - if it's generic, the getstats API isn't what it should be.

martinthomson:
Just so I have all the things in the same place.

[WebIDL](http://heycam.github.io/webidl/#idl-attributes) says:
>  The type of the attribute, after resolving typedefs, MUST NOT be a nullable or non-nullable version of any of the following types: a sequence type; a dictionary; a union type that has a nullable or non-nullable sequence type or dictionary as one of its flattened member types.

The reason being that every access to the attribute would then be forced to create a new instance of the dictionary.  That's non-performant to the point of being unimplementable.

It's also good to remember [this advice](https://lists.w3.org/Archives/Public/public-script-coord/2014JanMar/0201.html):
> Dictionaries are good for things like option arguments or multiple value return records where the actual aggregation has little or no domain significance. When you start using them in place of designing real domain object interfaces you are going off the rails.

cc @pthatcherg

jan-ivar:
Good links. FWIW there are not attributes in the `getStats` API. That said, it should likely follow the second advice and return interfaces instead dictionaries on account of relying on polymorphic inheritance.

martinthomson:
I don't see any evidence of list discussion here.

foolip:
Filed https://github.com/w3c/webcrypto/issues/161 because `getAlgorithm()` seems a bit odd. It says "Returns the value of keygenAlgorithm passed in the call to generateCertificate()" but if it's really the same value there doesn't seem to be any point in making this a method instead of a readonly attribute.

martinthomson:
> The type of the attribute, after resolving typedefs, must not be a nullable or non-nullable version of any of the following types:
>
>  *  a sequence type
> *  a dictionary type
> *   a union type that has a nullable or non-nullable sequence type or dictionary as one of its flattened member types

-- https://heycam.github.io/webidl/#idl-attributes

foolip:
I think the main reason for that is that an attribute should return the same value every time, while returning a dictionary would create a new object every time. But `AlgorithmIdentifier` is `(object or DOMString)`, not involving a dictionary type, so strictly speaking it would be allowed. Not really suggesting making it an attribute, though.

Whether or not https://github.com/w3c/webcrypto/issues/161 is resolved, it probably makes sense to behave as if there were a dictionary, which means creating a new object every time. However it's done, some prose will be needed. An option might be to always normalize to an Algorithm dictionary and return that?

@mwatson2, any suggestions?

jan-ivar:
@foolip Returning an object returns a reference to the exact same object the implementation has, unless we add a reverse [normalize-an-algorithm](https://w3c.github.io/webcrypto/Overview.html#dfn-normalize-an-algorithm) step, which means `pc.getAlgorithm().name = "foo"` would change what the implementation is using. Is that desirable?

Creating an object every time for an attribute would mean that `a.algorithm != a.algorithm`, which breaks most expectations.

jan-ivar:
Also note that based on the algorithm I linked to, it's not the same object as was passed in, rather a reference to the internal copy the implementation produced using that algorithm.

foolip:
https://w3c.github.io/webrtc-pc/#dom-rtccertificate-getalgorithm currently says "Returns the value of keygenAlgorithm passed in the call to generateCertificate()." which I take to mean that the following web-platform-tests should pass:
```JavaScript
promise_test(() => {
  var pc = new RTCPeerConnection();
  var algorithm = { name: "ECDSA", namedCurve: "P-256" };
  return pc.generateCertificate(algorithm).then(certificate => {
    assert_equals(certificate.getAlgorithm(), algorithm);
    algorithm.name = 'banana';
    assert_equals(certificate.getAlgorithm().name, 'banana');
  });
});
```

Was this really intended? I don't quite know what would be most consistent with Web Crypto.

jan-ivar:
@foolip I agree that's how it's written. Note however that the implementation won't see `'banana'` here, due to the [normalize-an-algorithm](https://w3c.github.io/webcrypto/Overview.html#dfn-normalize-an-algorithm) algorithm. You're modifying the original object, not the copy that the implementation made. Was that the intent? I dunno.

foolip:
A test to compare this with `crypto.subtle.generateKey`: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/4582

Here both Chrome and Firefox return something other than the original object. (In Chrome it's a new object every time, which seems like a bug: https://bugs.chromium.org/p/chromium/issues/detail?id=656997)

I wonder if the most reasonable thing here might be to make a copy, freeze it, and then return that. Then it won't be possible to modify it.

ericroman920:
Can we use the same mechanism as CryptoKey.algorithm?
In other words:

* Instead of getAlgorithm() accessors, you have a read-only property algorithm (which addresses concerns over always re-creating the object).

* The property type can be a KeyAlgorithm (from WebCrypto), or if WebRTC needs to reflect other parameters it can invent something equivalent.

alvestrand:
This PR was merged 6 months ago. Can we move the discussion to a new bug?

FWIW, I don't see much point in modifying the interface again, but marking getAlgorithm() with [SameObject] seems uncontroversial (no change to client code needed).

jan-ivar:
@alvestrand I don't see how it can return (object or DOMString) and be [SameObject]. What happens when a new object or string is set? Agree we should open a new bug. @foolip want to do the honors?

martinthomson:
@alvestrand:
>The [SameObject] extended attribute must not be used on anything other than a read only attribute whose type is an interface type or object.

Also, I think that the *value* being the same does not imply that it is the same object.  If it were an interface that were returned, then maybe.  As for webcrypto returning a value that is writeable, we should not repeat their mistakes.

alvestrand:
I misspoke. [sameObject] wasn't what I was trying to say, but that cert.getAlgorithm() === cert.getAlgorithm() (principle of least surprise), which is what @foolip expressed better upthread.

martinthomson:
An alternative design would be this:

```
interface RTCCertificateAlgorithm {
  readonly DOMString name;
  ...
};
partial interface RTCCertificate {
  [SameObject] readonly RTCCertificateAlgorithm algorithm;
};
```

Of course, that means porting all potentially applicable `AlgorithmIdentifier` fields into an interface, or defining a whole hierarchy.  That would be foolish because we would probably find that a constant maintenance headache.  Now, if *webcrypto* were to define that interface (and use it, since they clearly need to), then we're in business.

martinthomson:
Forgot to note: you shouldn't return the same dictionary.  Because you would then potentially be lying about state if someone set `name` to "banana".

jan-ivar:
Sameness matters more for attributes, because it affects an object's sameness in a comparison by value (i.e. enumerating all its attributes and comparing each by value). With getMethods we have more leeway to return copies like we do e.g. with getTracks. That said I prefer attributes because they work well with es6 destructuring.

Issue #473, opened 2016-01-18
Section 5.2.4: Definition of Active for an RTCRtpReceiver


aboba:
In Section 5.2.4, active is defined as follows: 

active of type boolean
Indicates that this encoding is actively being sent. Setting it to false causes this encoding to no longer be sent. Setting it to true causes this encoding to be sent.

This does not describe the meaning of active for a receiver (if any).

aboba:
Issue partially addressed in PR https://github.com/w3c/webrtc-pc/pull/466

aboba:
Additional description of the issue: 
https://lists.w3.org/Archives/Public/public-webrtc/2016Feb/0029.html

aboba:
The basic issue has been reviewed - next step is review of each of the PRs submitted to address it.

aboba:
I think we can close this now.

PR #505, opened 2016-02-16
Add activateReceiver method to RTCRtpTransceiver

aboba:
Add activateReceiver method to the RTCRtpTransceiver

Related to Issue https://github.com/w3c/webrtc-pc/issues/473

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4043,6 +4043,19 @@ sender.setParameters(params)
             <p>If the sender is not already active and the PeerConnection is in the stable state,
             this will set the <code>negotiationneeded</code> flag.
           </dd>
+          
+          <dt>void activateReceiver()</dt>
+
+          <dd>
+            <p>The <dfn id="dom-rtptransceiver-activatereceiver"><code>activateReceiver</code></dfn>
+            method activates the <code>RTCRtpTransceiver.receiver</code> such that future calls
+            to <code>createOffer</code> and <code>createAnswer</code> mark the corresponding
+            media description as <code>sendrecv</code> or <code>recvonly</code>, as defined in
+            [[!JSEP]].
+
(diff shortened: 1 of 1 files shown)


aboba:
@pthatcherg Can you review?

webrtc.html
```diff
@@ -4043,6 +4043,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+          <dd>
+            <p>The <dfn id="dom-rtptransceiver-activatereceiver"><code>activateReceiver</code></dfn>
+            method activates the <code>RTCRtpTransceiver.receiver</code> such that future calls
+            to <code>createOffer</code> and <code>createAnswer</code> mark the corresponding
+            media description as <code>sendrecv</code> or <code>recvonly</code>, as defined in
+            [[!JSEP]].
+
+            <p>If the receiver is not already active and the PeerConnection is in the stable state,
+            this will set the <code>negotiationneeded</code> flag.
+          </dd>
```
pthatcherg: I was hoping we would need only one "set horrible" (activateSender()) and not two (activateReceiver()).  

Does this accomplish anything other than a way to tell the sending side "please stop sending me this track" without rejecting the whole m-line?

aboba:
Implementing various "hold" scenarios appears to require both "set horrible" methods. 

For example, if addTransceiver(init) is called with init.send=true and init.receive=false (e.g. sendonly), activateReceiver() can be used to go from sendonly to sendrecv, taking the remote peer off hold (after another O/A exchange).

See the "hold" example, in http://tools.ietf.org/html/rfc4317 Section 3.1.

webrtc.html
pthatcherg: If we need it, this PR looks good.

Issue #336, opened 2015-10-16
No definition of "applying an ICE candidate"


dontcallmedom:
The description of `addIceCandidate` refers to "successfully apply[ing]" an ICE candidate, but I couldn't find a definition of what was meant neither in this spec, nor in JSEP.

I guess the first part of the method description is probably the closest thing to what is meant: "add[ing the candidate] to the remote description", and "[sending] connectivity checks", but if that's indeed what is meant, it should be made explicit, and it should also probably be turned into an algorithmic form, among other things to distinguish what is synchronous and what isn't.

adam-be:
This needs some work indeed. I'll propose something (tomorrow).

adam-be:
PR #507

alvestrand:
#507 solved most of this. There may need to be a JSEP reference, @adam-be will raise an issue.

PR #507, opened 2016-02-17
Make the definition of addIceCandidate() more explicit (issue: #336)

adam-be:
This is a start. We might want to update JSEP as well.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1278,16 +1278,13 @@
           <dd>
             <p>The <dfn id=
             "dom-peerconnection-addicecandidate"><code>addIceCandidate()</code></dfn>
-            method provides a remote candidate to the <a>ICE Agent</a>. In addition to
-            being added to the remote description, connectivity checks will be
-            sent to the new candidates as long as the <a href=
-            "#ice-transports-setting">ICE Transports setting</a> is not set to
-            <code>none</code>. This call will result in a change to the
-            <a>ICE connection state</a>, and may result in a change to
-            media state if it results in different connectivity being
-            established. The only members of <code>candidate</code>
-            used by this method are <var>candidate</var>, <var>sdpMid</var> and
-            <var>sdpMLineIndex</var>; the rest are ignored.</p>
+            method provides a remote candidate to the <a>ICE Agent</a>. The only
+            members of the argument used by this method are <code><a href=
+            "#widl-RTCIceCandidate-candidate">candidate</a></code>, <code>
+            <a href="#widl-RTCIceCandidate-sdpMid">sdpMid</a></code> and <code>
+            <a href=
+            "#widl-RTCIceCandidate-sdpMLineIndex">sdpMLineIndex</a></code>; the
+            rest are ignored.</p>
 
             <ol>
               <li>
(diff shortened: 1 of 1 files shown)


alvestrand:
@pthatcherg @martinthomson can you review this for ICE handling correctness?
@juberti note that there's a comment about JSEP in there.

webrtc.html
```diff
@@ -1313,8 +1308,55 @@
               </li>
 
               <li>
-                <p>If the candidate is successfully applied, resolve
-                <var>p</var> with <code>undefined</code>.</p>
+                <p>In parallel, start the process to apply
+                <var>candidate</var>.</p>
+
+                <div class="note">We probably want to define the process to
+                'apply an ICE candidate' in JSEP and link to it above.</div>
```
alvestrand: Can we make this link now, or do we have to wait for JSEP updates?

webrtc.html
```diff
@@ -1313,8 +1308,55 @@
...
+                        <var>remoteDescription</var>.</p>
+                      </li>
+
+                      <li>
+                        <p>If the <a>ICE Agent</a> is not currently checking
+                        candidate pairs, the <a>ICE Agent</a> MUST start
+                        checking candidate pairs and <a data-lt=
+                        "update the ICE connection state">update
+                        <var>connection</var>'s ICE connection state</a> to
+                        <code>checking</code>.</p>
```
alvestrand: Is this state change allowed in all previous states? What about "closed"?
(since we're firing a task to do this, things can have happened between this and the code above).

webrtc.html
adam-be: We could make a generic link to JSEP, but we can't use the nice feature and point to a specific section (that doesn't exist yet).

webrtc.html
adam-be: Yes. That might be a problem.

adam-be:
Added closed state checks in the reject and resolve tasks (similar to how setLocal/RemoteDescription() does it)

adam-be:
If the m-line information in the candidate (sdpMid or sdpMLineIndex) don't match an existing m-line we need to fail. The question is if the check should  be in the addIceCandidate() steps or if it's part of the "apply the candidate" operation which in turn will report about the error.

webrtc.html
```diff
@@ -1297,15 +1294,13 @@
...
+                <p>Let <var>candidate</var> be the methods argument.</p>
               </li>
 
               <li>
-                <p>If the candidate could not be successfully applied, return a promise
-                rejected with a <code>DOMException</code> object whose <code>name</code>
-                attribute has the value <code>OperationError</code>.</p>
+                <p>If <var>candidate</var> is missing values for both <var>sdpMid</var> and
+                <var>sdpMLineIndex</var>, return a promise rejected with a
+                <code>TypeError</code>.</p>
```
adam-be: The above text duplicates what's said in the RTCIceCandidate constructor. We could specify it as 'construct an RTCIceCandidate' and don't have to care about what the argument was from this point on. I noted that the RTCSessionDescription constructor has a note about it being deprecated. Perhaps the RTCIceCandidate constructor has the same status.

webrtc.html
adam-be: @alvestrand, do think the RTCSingnalingState closed-state check (already added) is enough or should we also check the RTCIceConnectionState for closed?

alvestrand: Where's the transition state diagram for the ICE connection state?
Currently (editors' draft) I find two references to "update the ICE connection state", one to "connected" and the other one to either of "completed" or "failed". there is no transition to "closed" or "disconnected" that I can find.
If it can only transition to "closed" together with the RTCSignalingState, we're OK, but then *that* needs to be documented.

adam-be: Perhaps we can create an issue to revisit this when we have cleared out the uncertainties you mention with the ICE connection state.

PR #515, opened 2016-02-22
Remove state check when ICE Agent is initialized (state is predictable)

adam-be:
The actions described in the modified text used to be triggered by "an ICE event". It's now triggered by the initialization of the ICE Agent so the state is known at creation.

This fix should have been part of PR #510.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -435,12 +435,10 @@
 
         <ul>
           <li>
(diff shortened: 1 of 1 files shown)


adam-be:
The fix to #522 will overwrite this change anyhow.

Issue #478, opened 2016-01-21
Clarify relation between a local track's settings and how it is encoded by PeerConnection


stefhak:
As discussed in the thread https://lists.w3.org/Archives/Public/public-webrtc/2016Jan/0058.html there is currently no text about the relation between how a local MediaStreamTrack has been set up using constraints (to control things like framerate, width, height and so on) and how it would be encoded when transmitted.
I think everyone has been assuming that the properties should be maintained as far as possible, but the spec does not say anything about it.

alvestrand:
Big issue. Might want to file more issues for detailed handling. Stefan will write a PR with the prose part.

stefhak:
I'm wondering if we can't close this one now that #519 has landed?

alvestrand:
Seems good. Closing.

elagerway:
LGTM

PR #519, opened 2016-02-24
Relation between local track and outgoing encoding (Issue #478)

stefhak:
Could we address #478 with something like this?

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2849,6 +2849,16 @@
       Each track to be sent is associated with exactly one <code><a>RTCRtpSender</a></code>, and
       each track to be received is associated with exactly one <code><a>RTCRtpReceiver</a></code>.</p>
 
+      <p>The basic principle is that the encoding and transmission of each
+      <code>MediaStreamTrack</code> is made such that the characteristics are as far as possible
+      maintained by the track created on the remote side. In other words, the
+      <code>MediaTrackSettings</code> width, height, frameRate for a video track, and volume, sampleSize,
+      sampleRate and channelCount for an audio track should be maintained. There are situations when this
+      does not apply, there may for example be resource constraints (HW Codec, CPU, network throughput
+      etc.) that makes this impossible, or there may be <code><a>RTCRtpSender</a></code> settings applied
+      that instruct the implementation to deviate from this principle.</p> 
+
+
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2849,6 +2849,16 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       Each track to be sent is associated with exactly one <code><a>RTCRtpSender</a></code>, and
       each track to be received is associated with exactly one <code><a>RTCRtpReceiver</a></code>.</p>
 
+      <p>The basic principle is that the encoding and transmission of each
+      <code>MediaStreamTrack</code> is made such that the characteristics are as far as possible
+      maintained by the track created on the remote side. In other words, the
```
jan-ivar: This sentence structure confused me. I started reading "the characteristics are... " and expected the rest of the sentence to reveal what they are, but instead I think with "the characteristics" you mean the characteristics of the MediaStreamTrack just mentioned. Maybe "it's characteristics"?

webrtc.html
```diff
@@ -2849,6 +2849,16 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+      <code>MediaTrackSettings</code> width, height, frameRate for a video track, and volume, sampleSize,
+      sampleRate and channelCount for an audio track should be maintained. There are situations when this
+      does not apply, there may for example be resource constraints (HW Codec, CPU, network throughput
+      etc.) that makes this impossible, or there may be <code><a>RTCRtpSender</a></code> settings applied
```
jan-ivar: "Impossible" sounds like a high bar. I could see UAs faced with lots of complicated decisions about how to best allocate bandwidth between competing demands. If we removed "that makes this impossible", then the sentence still works I think fwiw.

jan-ivar: Should we also point out that these characteristics may even vary over time? For instance, this comes up as an issue during remote recording, where not all players cope well with variable-resolution recordings, so it might be useful to mention that there's no guarantee of characteristics being fixed (whatever they are).

stefhak:
Thanks @jan-ivar for good input. I'll do an update soon. But I take it that you think that having something like this in this part of the spec makes sense?

jan-ivar:
It seems worth mentioning.

Issue #513, opened 2016-02-18
Handling the stopping of a track should be described


alvestrand:
See https://github.com/w3c/mediacapture-main/issues/314

alvestrand:
#520 hotlink

alvestrand:
#520 solved this

PR #520, opened 2016-02-25
Add text about 'remote sources' and how they are stopped (issue #513)

adam-be:
This is work in progress.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6615,7 +6615,8 @@ function processStats() {
 
       <p>When a <a>remote source</a> is notified that a <code>
       <a>MediaStreamTrack</a></code>, using the source, has <code>ended</code>
-      [[!GETUSERMEDIA]] it must TBD.</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6597,6 +6599,10 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
       <p class="issue">How do you know when it has stopped? This seems
       like an SDP question, not a media-level question. (Suggestion: when the
       track is ended, either through port 0, or removing the a=msid attrib)</p>
+
+      <p>When a <a>remote source</a> is notified that a <code>
+      <a>MediaStreamTrack</a></code>, using the source, has <code>ended</code>
+      [[!GETUSERMEDIA]] it must TBD.</p>
```
alvestrand: "it MAY choose to stop using resources on the incoming stream, for instance by turning off the decoder."
Do we want to say anything about the possibility of negotiating it away in the next offer/answer (state sendonly or inactive for the media section)?

Issue #483, opened 2016-01-28
Signaling a=end-of-candidates


martinthomson:
Trickle ICE has a signal that says that there are no more candidates coming.  I can never remember whether it is per-component or per-agent, I think that it's the latter.

We currently do not have a way of providing that signal.  I believe that an `RTCIceCandidate` with a null `candidate` attribute has been suggested and maybe even implemented by some.  That doesn't require any new code paths in (most) clients, which would seem to be a virtue.

Current the `candidate` attribute is both mandatory and not nullable.

(I couldn't find an open issue on this, so here 'tis.)

stefhak:
@martinthomson there is some related discussion in #442.

aboba:
Section 4.9.2.2 permits the candidate attribute to be null: 

"The candidate attribute is the RTCIceCandidate object with the new ICE candidate that caused the event.  This attribute is set to null when an event is generated to indicate the end of candidate gathering."

However, as noted in Section 4.3.2.3, the candidate attribute supplied to addIceCandidate() cannot be null. 

Is this the missing functionality you are proposing to add?

pthatcherg:
There's something for sending out an "end of candidates", but nothing for taking it in.  But it's not clear whether that's needed for anything anyway.  In fact, it's not clear whether the "end of candidates" signal coming out is needed for anything, since the application can just get the same information from the gathering state.

But this is one of the topics for the upcoming virtual interim (as part of the discussion of the failed ICE connection state).  Hopefully we can resolve it then.​

elagerway:
@aboba 9 days to virtual interim, if there is a new proposal forthcoming please edit the slides https://docs.google.com/presentation/d/1R7lQeo4LNGu0KKW9GLdkwJ5x8FUmR4qrnhfPFZO3WS8/edit#slide=id.g828c698ba_0_25

alvestrand:
Merged #527

Issue #442, opened 2015-12-17
Impossible to know if ICE agent is "finished checking", for "failed" and "completed" states.


taylor-b:
Focusing first on "failed", it's defined as: "The ICE Agent is finished checking all candidate pairs and failed to find a connection for at least one component."

However, even after finishing gathering and checking all current candidate pairs, a new remote candidate could be applied at any time via addIceCandidate, creating new candidate pairs.

The trickle ICE spec addresses this, saying a check list should only go to the Failed state when "the remote agent has sent an end-of-candidates indication for that check list as described in Section 9.3.". The RTCPeerConnectionIceEvent does signal "end-of-candidates", but there are two problems with this:

1. It's only signaled once, but it needs to be signaled once per check list (once per m-line).
2. There's no way to pass the end-of-candidates indication into the browser.

I would propose adding a way to indicate "end of remote candidates" to the browser, either by overloading addIceCandidate or by adding a new method. This indication would ensure an accurate "failed" state (that doesn't flip back to "checking"), and it could also be used to optimize the ICE algorithm, as discussed in trickle ICE.

If we *don't* do that, we should at least make the spec clear in the fact that "failed" is temporary, and it could return to "checking" if a new candidate pair is created.

ekr:
I was going to say that this just worked, but it seems that addIceCandidate doesn't take a nullable candidate, which is what you would want for this.

pthatcherg:
Another option could be to define "failed" as "will only be connected again
with an ICE restart" (different from disconnected which is "may be
connected again without an ICE restart").  For example, when consent is not
refreshed.  Right now, that happens if you don't get any STUN requests or
responses for 30 seconds in the middle of a call.  Could we just make that
the same for the start of a call?  No connection after 30 seconds since
last created candidate pair == failed?  That would be a lot easier to
implement, and would actually be useful if an implementation chooses to
never end candidate gathering.

aboba:
On Jan 6, 2016, at 15:49, Peter Thatcher <notifications@github.com> wrote:
> 
> Another option could be to define "failed" as "will only be connected again
> with an ICE restart" (different from disconnected which is "may be
> connected again without an ICE restart"). For example, when consent is not
> refreshed. Right now, that happens if you don't get any STUN requests or
> responses for 30 seconds in the middle of a call.

[...]

PR #527, opened 2016-02-29
Enable trickling of end-of-candidates through addIceCandidate.

taylor-b:
A null parameter in addIceCandidate now represents end-of-candidates
for all media sections.

The definitions of the `failed` and `completed` ICE connection states
have been updated so that they only occur when there are no more
remote candidates.

Also updated the example code that trickled candidates, such that the
null candidate is trickled rather than ignored.

Fixes issues #483 and #442.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1272,12 +1272,14 @@
 </p>
           </dd>
 
-          <dt>Promise&lt;void&gt; addIceCandidate ((RTCIceCandidateInit or RTCIceCandidate) candidate)</dt>
+          <dt>Promise&lt;void&gt; addIceCandidate ((RTCIceCandidateInit or RTCIceCandidate)? candidate)</dt>
 
           <dd>
             <p>The <dfn id=
             "dom-peerconnection-addicecandidate"><code>addIceCandidate()</code></dfn>
-            method provides a remote candidate to the <a>ICE Agent</a>. The only
+            method provides a remote candidate to the <a>ICE Agent</a>. This
+            method can also be used to indicate the end of remote candidates
+            when called with a <code>null</code> value for <code>candidate</code>. The only
             members of the argument used by this method are <code><a href=
             "#widl-RTCIceCandidate-candidate">candidate</a></code>, <code>
             <a href="#widl-RTCIceCandidate-sdpMid">sdpMid</a></code> and <code>
@@ -1294,6 +1296,25 @@
 
               <li>
                 <p>Let <var>candidate</var> be the methods argument.</p>
+              </li>
+
+              <li>
+                <p>If <var>candidate</var> is <code>null</code>, then run the
+                following steps:</p>
+                <ol>
+                  <li>
+                    <p>For each media section in the last successfully applied
+                    remote description, perform the processing for an
+                    end-of-candidates indication for said media section as
+                    defined in [[TRICKLE-ICE]], unless an ICE restart occurred
+                    for said media section since setting the last remote
+                    description.</p>
+                  </li>
+                  <li>
+                    <p>Return a promise resolved with
+                    <code>undefined</code>.</p>
+                  </li>
+                </ol>
               </li>
 
               <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1297,6 +1299,25 @@
...
+                <p>If <var>candidate</var> is <code>null</code>, then run the
+                following steps:</p>
+                <ol>
+                  <li>
+                    <p>For each media section in the last successfully applied
+                    remote description, perform the processing for an
+                    end-of-candidates indication for said media section as
+                    defined in [[TRICKLE-ICE]], unless an ICE restart occurred
+                    for said media section since setting the last remote
+                    description.</p>
```
taylor-b: This caveat exists to avoid the race condition where one peer sets a local offer triggering an ICE restart at the same time as another peer signals end-of-candidates. In this situation, an "end-of-candidates" received before the remote answer refers to the previous "generation" of ICE candidates, so it should be ignored.

webrtc.html
```diff
@@ -2056,6 +2074,26 @@
...
+          </li>
+          <li>
+            <p>Setting a remote description with
+            <code>a=end-of-candidates</code> as defined in [[TRICKLE-ICE]].</p>
+          </li>
+          <li>
+            <p>Passing <code>null</code> into <code><a href=
+            "#dom-peerconnection-addicecandidate">addIceCandidate()</a></code>.</p>
+          </li>
+        </ul>
```
taylor-b: Should this be moved to JSEP (at least, the first two bullets)? I'm just not sure where in JSEP it would fit.

webrtc.html
```diff
@@ -1297,6 +1299,25 @@
               </li>
 
               <li>
+                <p>If <var>candidate</var> is <code>null</code>, then run the
+                following steps:</p>
+                <ol>
+                  <li>
+                    <p>For each media section in the last successfully applied
```
adam-be: Please use 'media description' (instead of 'media section') as it's defined with a link to the RFC.

alvestrand:
Looks good, but needs rebase.

taylor-b:
Alright, rebased.

webrtc.html
```diff
@@ -1297,6 +1299,25 @@
...
+                    remote description, perform the processing for an
+                    end-of-candidates indication for said media section as
+                    defined in [[TRICKLE-ICE]], unless an ICE restart occurred
+                    for said media section since setting the last remote
+                    description.</p>
+                  </li>
+                  <li>
+                    <p>Return a promise resolved with
+                    <code>undefined</code>.</p>
```
pthatcherg: Should it resolve immediately, or once the processing is done?  It seems like it shouldn't resolve until it's done.

webrtc.html
pthatcherg: Is ignoring it the right thing to do?  If the local description for the ICE restart is rolled back, we'd still want the end-of-candidates to have applied. 

You already specified that the end-of-candidates applies to the last applied remote description, so that part seems fine.  But perhaps we should specify that it applies to all successfully applied remote descriptions, not just the last one.  That way, if there are two concurrent remote gathering generations, both will be affected by the end-of-candidates.

webrtc.html
```diff
@@ -2073,40 +2094,45 @@
...
           <dt>disconnected</dt>
 
           <dd>Liveness checks have failed for one or more components. This is
           more aggressive than <code>failed</code>, and may trigger
           intermittently (and resolve itself without action) on a flaky
-          network.</dd>
+          network. Alternatively, the <a>ICE Agent</a> has finished checking
+          all existing candidates pairs and failed to find a connection for
+          at least one component, but is still gathering and/or waiting for
+          additional remote candidates.</dd>
```
pthatcherg: Does this mean that when we have more than one component, we will go new -> disconnected -> completed or new -> disconnected -> failed?

webrtc.html
pthatcherg: Yes, it seems like the first two bullet points should be,  if not all three.

alvestrand:
@taylor-b ping

webrtc.html
taylor-b: No, there would be an intermediary state of "checking".

webrtc.html
pthatcherg: Perhaps a better way of saying this is:  We need to make it clear what happens when this is done:

pc.setRemoteDescription(remote_offer_with_ice_restart);
pc.addIceCandidate(null);
pc.setRemoteDescription({type: rollback});

or this:

pc.setLocalDescription(local_offer_with_ice_restart);
pc.addIceCandidate(null);
pc.setLocalDescription({type: rollback});

Can the pc now go to completed/failed?

taylor-b: I created two JSEP issues for these scenarios. https://github.com/rtcweb-wg/jsep/issues/251 for the scenario above, and https://github.com/rtcweb-wg/jsep/issues/250 for the one below.

webrtc.html
taylor-b: You're right. I fixed this.

webrtc.html
taylor-b: Done. The link to the RFC is in the "Terminology" section which this now links to.

webrtc.html
taylor-b: It's still not clear to me where in JSEP this should go. Should it be a standalone section, or spread across "Applying a Remote Description" and "addIceCandidate"? For now I'm at least adding a note.

fluffy: I think it should be spread across the sections you mentioned

taylor-b: Ok. I removed the stuff that should go in JSEP, and created this PR: https://github.com/rtcweb-wg/jsep/pull/254

Note that there's still a TODO for describing "applying a candidate" process in JSEP.

aboba:
LGTM.

elagerway:
LGTM

Issue #529, opened 2016-03-02
MediaStreamTrack clarifications needed(?)


stefhak:
The spec says nothing about e.g.
* A track sourced by the PeerConnection should have its ```remote``` attribute set to true
* If the ```readonly``` attribute should be true (should it not? I guess in principle downscaling could be supported though)
* What settings like facingMode, latency ... should read (the SDP does not provide this data, so it should be the default value?)
* How settings like width, height, frameRate, channelCount are populated (presumably from observing the incoming RTP)

dontcallmedom:
The first half would be moot if we remove remote/readonly as per https://github.com/w3c/mediacapture-main/issues/320

stefhak:
I think we can close this issue now that #531 has been merged.

PR #531, opened 2016-03-02
Added initialization of readonly and remote attributes for incoming t…

stefhak:
…racks. Proposed fix to a part of #529

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3209,6 +3209,16 @@
                 </li>
 
                 <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3209,6 +3209,16 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                 </li>
 
                 <li>
+                  <p>Initialize <code><var>track</var>.readonly</code>
```
alvestrand: Proper English has "the" after "Initialize".

Issue #457, opened 2016-01-08
Non-normative ICE state transition diagram


aboba:
Figure 2 does not show transitions between 'checking' and 'disconnected'  (and back).

The transition from 'disconnected' to 'checking' can occur if a new interface comes up, thereby providing new candidates (and pairs) to check.

A transition from 'checking' to 'disconnected' can occur if the new candidate pairs do not result in a successful connectivity check, but gathering is still going on, so 'failed' has not been reached. 

Related ORTC Issue: https://github.com/openpeer/ortc/issues/332

aboba:
For reference, here is the non-normative ICE state transition diagram from ORTC. 
<img width="432" alt="icestates-cont" src="https://cloud.githubusercontent.com/assets/808811/12187131/e4db72f0-b55d-11e5-87c1-e5798d731890.png">

taylor-b:
I don't see how you could transition from "disconnected" to "checking" or vice versa (at least, for a single IceTransport). As I understood it, "disconnected" means "was previously connected, but a liveness check failed", and "checking" means "was *never* connected, and are still checking".

So, if a new interface comes up while in "disconnected", I would expect the state to remain in "disconnected" until a connection is found, then transition to "connected" and eventually "completed".

And if the ICE agent finishes checking current candidate pairs, but hasn't finished gathering, I would expect it to remain in "checking", since the definition of "checking" includes "may also still be gathering."

This kind of confusion is why I think we need a more well-defined algorithm for determining the ICE state, similar to what Peter did for RTCPeerConnectionState. I'll work on a PR to show what I mean.

aboba:
The ICE agent can find a successful candidate pair and use it to send media before completing checking.  In this situation, RTCIceTransport.getSelectedCandidatePair() will return the selected candidate pair.  If RTCIceTransport.state is "gathering" then the application won't know that outgoing media is permitted. To clarify this, ORTC uses the following state definitions: 

checking: In this state the RTCIceTransport is checking candidate pairs but has not yet found a successful candidate pair, or liveness checks have failed (such as those in [CONSENT]) on a previously successful candidate pair.

connected: The RTCIceTransport has received a response to an outgoing connectivity check, or has received incoming DTLS/media after a successful response to an incoming connectivity check, but is still checking other candidate pairs to see if there is a better connection. In this state outgoing media is permitted.

completed: all appropriate candidate pairs have been tested and at least one functioning candidate pair has been found.

[...]

PR #533, opened 2016-03-03
Add RTCIceTransportState Enum

aboba:
Add definitions for RTCIceTransport.state

Relates to Issue https://github.com/w3c/webrtc-pc/issues/457

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4338,54 +4338,56 @@ sender.setParameters(params)
         	<dt>new</dt>
         	<dd>
         		<p>
-                            The <code><a>RTCIceTransport</a></code> object is gathering candidates and/or waiting for 
-                            remote candidates to be supplied.
-                        </p>
-                </dd>
-                <dt>checking</dt>
-                <dd>
-                        <p>
-                            The <code><a>RTCIceTransport</a></code> has received at least one remote candidate,
-                            and is checking candidate pairs but has not yet found a successful candidate pair, 
-                            or consent checks [[!RFC7675]] have failed on a previously successful candidate pair.
-                            In addition to checking, it may also still be gathering. 
+        			The <code><a>RTCIceTransport</a></code> object is gathering candidates and/or waiting for
+        			remote candidates to be supplied.
+        		</p>
+        	</dd>
+        	<dt>checking</dt>
+        	<dd>
+        		<p>
+        			The <code><a>RTCIceTransport</a></code> has received at least one remote candidate,
+        			and is checking candidate pairs but has not yet found a successful candidate pair,
+        			or consent checks [[!RFC7675]] have failed on a previously successful candidate pair.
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4331,6 +4331,65 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        			The <code><a>RTCIceTransport</a></code> has received at least one remote candidate,
+        			and is checking candidate pairs but has not yet found a successful candidate pair,
+        			or consent checks [[!RFC7675]] have failed on a previously successful candidate pair.
+        			In addition to checking, it may also still be gathering. 
+                        </p>
+                </dd>
+                <dt>connected</dt>
+                <dd>
+                        <p>
+                        	The <code><a>RTCIceTransport</a></code> has received found a usable connection
```
nils-ohlmeier: "has found"

webrtc.html
```diff
@@ -4331,6 +4331,65 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                        	an indication that there are no more remote candidates.  However, all candidate pairs tested thus
+                        	far have failed (or consent checks [[!RFC7675]] once successful, have now failed).  Since other
+                        	candidate pairs may become available for testing, the "failed" state has not been reached.
+                        </p>
+                </dd>
+                <dt>failed</dt>
+                <dd>
+                        <p>
+                        	The <code><a>RTCIceTransport</a></code> has finished gathering, received an indication that there are
+                        	no more remote candidates, finished checking all candidate pairs, and failed to find a connection.
```
nils-ohlmeier: The descriptions for "disconnected" and "failed" read from the consent checks perspective as:
- if the current state is "connected" (but not yet "completed") and consent checks fail I should switch to to "disconnected"
- if the current state is "completed" and consent checks fail I should switch to....? I'm assuming "failed", as "completed" says gathering is finished, and "disconnected" explicitly includes that gathering has not finished yet. If failed is the desired state it would be good to add one more sentence about failed consent checks to be 100% clear.

aboba:
@nils-ohlmeier Added clarifications responding to your comments.

elagerway:
@taylor-b would mind taking a look and providing some comment here?

taylor-b:
I just made PR #557 which does what I talked about in the virtual interim: move the ICE state definitions into the RTCIceTransport section, and define RTCIceConnectionState in the same fashion that RTCPeerConnectionState is defined.

Could we try to merge that PR before merging this one? That way, when viewing the diff for this one, it will be clear how the ICE state definitions are changing. My PR doesn't change the definitions, it just moves them around.

aboba:
@taylor-b PR https://github.com/w3c/webrtc-pc/pull/557 covers everything in this PR other than the handling of ICE consent failure.  So it makes sense to merge 557 first and then just focus on the additional consent-related text.

PR #535, opened 2016-03-04
update MID to be random values when not received in offer

### LINKED ISSUE #227: .mid attribute - random or Null when MID is not present in SDP?
https://github.com/w3c/webrtc-pc/pull/193 added a .mid property that, if there is not a value present in the SDP, will be set to a random value.

Using a random value rather than letting the attribute be null is still controversial; @juberti argued in favor of the random value.

fluffy:
Fixes #227


webrtc.html
```diff
@@ -4096,8 +4096,15 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
             attribute is the <code>mid</code> negotatiated and present
             in the local and remote descriptions as defined in
-            <span data-jsep="initial-offers initial-answers">[[!JSEP]]</span>.  Before negotiation is complete,
-            the <code>mid</code> value may be null.  After rollbacks,
+            <span data-jsep="initial-offers
+	    initial-answers">[[!JSEP]]</span>.
+	    Before negotiation is complete,
+	    the <code>mid</code> value may be null.  If there is no
+	    MID value in the the remote SDP, a random value will be
+	    created for the <code>mid</code> as described in
```
stefhak: What happens if the offerer includes a MID in the offer, the answer comes back without a MID? According to above this should mean that a new (random) value is created, but should not the original MID be retained?

fluffy: Good point - updated to deal with that.

fluffy:
@dontcallmedom - can you take a look at this and figure out why CI is failing. CI also fails if I add the data-jsep to the ref (See the TODO). Thanks

webrtc.html
```diff
@@ -4125,13 +4125,20 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-            "dom-rtptransceiver-mid"><code>mid</code></dfn>
+            <p>The <dfn id="dom-rtptransceiver-mid"><code>mid</code></dfn>
             attribute is the <code>mid</code> negotatiated and present
             in the local and remote descriptions as defined in
-            <span data-jsep="initial-offers initial-answers">[[!JSEP]]</span>.  Before negotiation is complete,
-            the <code>mid</code> value may be null.  After rollbacks,
-            the value may change from a non-null value to null.</p>
+            <span>[[!JSEP]]</span>.
+            <!--
+            TODO add data-jsep="initial-offers initial-answers" to the span
```
fluffy: the CI seems to fail if I add this. I did update the map but did not help

dontcallmedom:
@fluffy this is not an issue with your pull request but with the current state of the repo; hopefully it will be fixed soon

fluffy:
> On Mar 16, 2016, at 6:46 AM, Dominique Hazael-Massieux <notifications@github.com> wrote:
> 
> @fluffy this is not an issue with your pull request but with the current state of the repo; hopefully it will be fixed soon
> 
> —
> You are receiving this because you were mentioned.
> 

thank you

dontcallmedom:
The fix has landed in https://github.com/w3c/webrtc-pc/commit/4fa52fad30574830e3cd5c8c8167cb01ca499985 ; you'll need to rebase if you want to benefit from it. I would suggest removing the change to the JSEP mapping since it doesn't help and in fact might create issues down the line.

fluffy:
OK - I think this is ready to merge

alvestrand:
@adam-be please review the ICECandidate validity change, since you recently touched this.

adam-be:
It would be nice to squash this history and rebase to get rid of the diffs that are not part of this change.

webrtc.html
```diff
@@ -2389,9 +2389,9 @@
           takes a dictionary argument, <var>candidateInitDict</var>, whose content is used
           to initialize the new
           <a class="idlType" href="#idl-def-RTCIceCandidate"><code>RTCIceCandidate</code></a>
-          object. If either <var>sdpMid</var> or <var>sdpMLineIndex</var> is not present in
+          object. If the <var>sdpMLineIndex</var> is not present in
           <var>candidateInitDict</var>, the corresponding attribute will be initialized to
-          <code>null</code>. If neither is present, a <code>TypeError</code> exception
+          <code>null</code>. If <var>sdpMid</var> is not present, a <code>TypeError</code> exception
           will be thrown.</dd>
```
adam-be: It it safe to require sdpMid here? We accept remote descriptions without mid and generate one in that case.

This behaviour (throw if not present) can also be expressed with the "required" WebIDL keyword on the member. And should we keep sdpMLineIndex if we require sdpMid?

webrtc.html
```diff
@@ -4125,13 +4125,20 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            <span>[[!JSEP]]</span>.
+            <!--
+            TODO add data-jsep="initial-offers initial-answers" to the span
+            -->
+            Before negotiation is complete, the <code>mid</code> value may be
+            null.  If there is no MID value in the the remote SDP, and no MID
+            value was previously assigned, a random value will be created for
+            the <code>mid</code> as described in <span
+            data-jsep="ice-candidate-format">[[!JSEP]]</span> when the remote
+            SDP is set. After rollbacks, the value may change from a non-null
```
adam-be: It would be nice to integrate the generation of mid into the 'set an RTCSessionDescription' algorithm since it would define when it time it should be done in relation to all the state updating and event dispatching.

fluffy:
This merges cleanly and it not easy to rebase. I do not plan to rebase it as it is simply a waste of time.

webrtc.html
fluffy: > On Mar 24, 2016, at 3:27 AM, Adam Bergkvist <notifications@github.com> wrote:
> 
> In webrtc.html:
> 
> >            <var>candidateInitDict</var>, the corresponding attribute will be initialized to
> > -          <code>null</code>. If neither is present, a <code>TypeError</code> exception
> > +          <code>null</code>. If <var>sdpMid</var> is not present, a <code>TypeError</code> exception
> >            will be thrown.</dd>
> 
> It it safe to require sdpMid here? We accept remote descriptions without mid and generate one in that case.
> 
> This behaviour (throw if not present) can also be expressed with the "required" WebIDL keyword on the member. And should we keep sdpMLineIndex if we require sdpMid?
> 
> —


I don't think we can require it here because it should not be retuned if it was not in remote description as that would cause something to fetched it to think the remote side supported mid when it does not. 

We have to keep the mLineIndiex

webrtc.html
fluffy: > On Mar 24, 2016, at 3:28 AM, Adam Bergkvist <notifications@github.com> wrote:
> 
> In webrtc.html:
> 
> >              attribute is the <code>mid</code> negotatiated and present
> >              in the local and remote descriptions as defined in
> > -            <span data-jsep="initial-offers initial-answers">[[!JSEP]]</span>.  Before negotiation is complete,
> > -            the <code>mid</code> value may be null.  After rollbacks,
> > -            the value may change from a non-null value to null.</p>
> > +            <span>[[!JSEP]]</span>.
> > +            
> > +            Before negotiation is complete, the <code>mid</code> value may be
> > +            null.  If there is no MID value in the the remote SDP, and no MID
> > +            value was previously assigned, a random value will be created for
> > +            the <code>mid</code> as described in <span
> > +            data-jsep="ice-candidate-format">[[!JSEP]]</span> when the remote
> > +            SDP is set. After rollbacks, the value may change from a non-null
> 
> It would be nice to integrate the generation of mid into the 'set an RTCSessionDescription' algorithm since it would define when it time it should be done in relation to all the state updating and event dispatching.

I don't think it is really generated it then - its more when the ICE stuff happens

webrtc.html
adam-be: Throwing a TypeError if a member is not present, as the prose says, is the same behaviour as requiring it. We can also have a default value on the sdpMLineIndex member to have it initialized to null if not present.
```
required DOMString sdpMid;
unsigned short? sdpMLineIndex = null;
```

webrtc.html
adam-be: We should figure out when and document it.

PR #537, opened 2016-03-08
[WIP] Trigger echidna automatic publications from Travis

vivienlacourba:
Those changes that will trigger Echidna's automatic publication should only appear in the `gh-pages` branch as otherwise they would disable or continuous integration tests that run on the  `master` branch.

I am checking with @deniak to see if he knows of another solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,18 +17,11 @@ install:
 script:
  - make check
 
-branches:
-  only:
-    - gh-pages
-
 env:
   global:
     - URL="https://w3c.github.io/webrtc-pc/W3CTRMANIFEST"
     - DECISION="https://lists.w3.org/Archives/Public/public-media-capture/2015Dec/0031.html"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -16,3 +16,19 @@ install:
...
  - make check
+
+branches:
+  only:
+    - gh-pages
+
+env:
+  global:
+    - URL="https://w3c.github.io/webrtc-pc/W3CTRMANIFEST"
+    - DECISION="https://lists.w3.org/Archives/Public/public-media-capture/2015Dec/0031.html"
```
dontcallmedom: I think we need a decision that would have been sent to public-webrtc

dontcallmedom:
Except for the decision URL, LGTM

vivienlacourba:
I am recreating this PR against the master branch as it seems possible to use a single `.travis.yml` version for running both tests on the `master` branch and automatic publication from the `gh-pages` branch.

PR #538, opened 2016-03-08
Trigger echidna automatic publications using Travis

vivienlacourba:
As pointed out by @dontcallmedom for the DECISION uri
> I think we need a decision that would have been sent to public-webrtc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ script:
 env:
   global:
     - URL="https://w3c.github.io/webrtc-pc/W3CTRMANIFEST"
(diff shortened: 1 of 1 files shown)


dontcallmedom:
LGTM (but waiting for an accurate  decision URI before merging)

vivienlacourba:
I think we should wait till March 18th (deadline to send objections to the [CfC](https://lists.w3.org/Archives/Public/public-webrtc/2016Mar/0031.html)) before merging.

vivienlacourba:
This will be merged after the 18th.

Issue #521, opened 2016-02-25
RTCPeerConnectionIceErrorEvent errorCode not defined


stefhak:
https://www.w3.org/2011/04/webrtc/wiki/images/5/50/WebRTCWG-2016-02-25.pdf#9 lists the error codes, but the spec does not provide that info (just says the error is a string).

robin-raymond:
````
4.9.3.2 Attributes
If the server could not be reached, errorCode will be set to a TBD value in the 7XX range, as this does not conflict with the STUN error code range.
````

I'm not sure this particular error is wise. In any implementation (especial IPV6), it's very likely many host candidates will not be able to reach the TURN server or STUN server because they don't route appropriately. So you will warm/flag these as being failed but with other host candidates they actually succeed. The programmer will think things are going horribly wrong when this is probably normal.

Further, if you don't indicate that this URI is actually succeeding then the programmer is going to be somewhat in the dark as to what's going on.

I would only use the "cannot contact server" error if NONE of the host IPs could reach the server.

alvestrand:
Fixed in #525

PR #540, opened 2016-03-10
Removal of "server cannot be reached" error

aboba:
Removed "server cannot be reached" error, responding to Robin Raymond's comment on Issue https://github.com/w3c/webrtc-pc/issues/521


alvestrand:
There are issues with reporting "server could not be reached" errors that need addressing first. Closing this for now.

PR #544, opened 2016-03-14
remove "public" from ice transport policy

### LINKED ISSUE #541: "public" is deprecated
The "public" value of RTCIceTransportPolicy has disappeared from Section 4.1.1 of https://tools.ietf.org/draft-ietf-rtcweb-jsep-13

fluffy:
Fixes #541

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -288,11 +288,7 @@
         candidates to the application, and only use those candidates for
         connectivity checks.</p>
         <dl class="idl" title="enum RTCIceTransportPolicy">
-          <dt>public</dt>
-
-          <dd>The <a>ICE agent</a> MUST filter out candidates with private
-          IP addresses [[RFC1918]].</dd>
-
+        
           <dt>relay</dt>
 
           <dd>The <a>ICE agent</a> MUST only use media relay candidates such as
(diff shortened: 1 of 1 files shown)


fluffy:
I don't think the travis error on this are from this PR.

aboba:
@fluffy  Instead of deleting "public", should we replace it with "reserved" and note that this value should be ignored?  Concerned about the potential backward compatibility effects of deleting "public"/"none" entirely.

fluffy:
> On Mar 14, 2016, at 8:32 PM, aboba <notifications@github.com> wrote:
> 
> @fluffy Instead of deleting "public", should we replace it with "reserved" and note that this value should be ignored? Concerned about the potential backward compatibility effects of deleting "public"/"none" entirely.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/w3c/webrtc-pc/pull/544#issuecomment-196621686
> 

The browsers or the pollyfill might decide to do something backward compatible like map this to one of the other values but I don't think we need to do anything in the spec. Worst case is it is ignored and you get the default which is the backwards compatible things to do.

Issue #337, opened 2015-10-16
Interfacing between WebRTC spec and JSEP


dontcallmedom:
When reading through the spec, there are a number of explicit and many more implicit references made to processes defined in JSEP, making it hard to read the doc from top to bottom, and making it hard to read e.g. with a view of creating test cases (I assume it also makes it harder to implement the spec).

That is even more so since JSEP has conformance requirements that overlap with ones set in the WebRTC spec.

I think a cleaner interfacing of the specs would be to set all the conformance requirements in the WebRTC 1.0 spec, and have JSEP define algorithms that the WebRTC spec can then reference explicitly. For instance, JSEP would define an algorithm with the name "apply an ICE candidate" (cf #336), would define an algorithm called "verify user-supplied SDP" which would be referenced from setLD/setRD, etc.

I realize that this is a potentially major set of editorial work across two specs and groups, so it may not be realistic to achieve it pre-CR, but it might be something achievable as part of the "rolling" CR process.

stefhak:
At TPAC we decided that @dontcallmedom and @ekr would look into the cross ref to JSEP.

dontcallmedom:
Once #423 gets merged, at least the following references need to be made more specific:
* [ ] `If <var>description</var> is set as a remote description
                    with new media descriptions [[!JSEP]], the User Agent
                    MUST <a href="#dispatch-receiver">dispatch a receiver</a> for
                    all new media descriptions.`
* [ ] `<p>Doing so will cause future calls to <code>createOffer</code>
                  and <code>createAnswer</code> to mark the corresponding <a>media description</a>
                  as <code>sendrecv</code> or <code>sendonly</code>, as defined in
                  [[!JSEP]].`
* [ ] `            <p>Stops sending media from <var>sender</var>.  The <code><a>RTCRtpSender</a></code> will still
            appear in <code>getSenders</code>. Doing so will cause future calls
            to <code>createOffer</code> to mark the <a>media description</a> for the corresponding
            transceiver as <code>recvonly</code> or <code>inactive</code>, as defined in
            [[!JSEP]].`
* [ ] `            <p>Adding a transceiver will cause future calls to <code>createOffer</code> to
              add a <a>media description</a> for the corresponding transceiver, as defined in
              [[!JSEP]].`
* [ ] `            <p>If <code>init.sendEncodings</code> is set, then subsequent calls
               to <code>createOffer</code> will be configured to send with multiple RTP encodings as
[...]

PR #549, opened 2016-03-16
add missing detail to JSEP references

burnburn:
These are commits to address issue #337.

Quite a few of these references are to a section which is likely to be the correct one but where the specific text implied by the reference does not yet exist.  To give one specific example, JSEP section 5.2.2 (subsequent-offers) does not include much of the text it will eventually need based on the references in the WebRTC spec.


alvestrand:
Good progress, but we're not finished with #337. Some JSEP changes are needed.
@burnburn to rebase and merge.

Issue #364, opened 2015-10-30
Browser calls close() on failed rollback


jan-ivar:
In http://w3c.github.io/webrtc-pc/#dom-peerconnection-setlocaldescription it says:

> If rollback was not necessary or was completed successfully, let reason be IncompatibleSessionDescriptionError. If rollback was not possible, let reason be InternalError and set connection's signaling state to closed.

AFAICT this is the lone case in the specification where the browser rather than content JS is calling `.close`. This case seems worth reconsidering, given how it reads like a "throw up our hands" last resort.

jan-ivar:
If the browser can call close, then it raises at least one issue which is how would JS learn about it.

juberti:
Does this make sense? Why not just fail the rollback and return the error via the promise? We already need to do this when a rollback is not possible (e.g. nothing to do), so unclear why we would self-destruct in this case.

martinthomson:
I agree with @juberti here.

jan-ivar:
wfm

stefhak:
I can't find that text in the current version of the spec. Can we close this Issue? @jan-ivar

jan-ivar:
It has been refactored, but is still there. See http://w3c.github.io/webrtc-pc/#set-description

3.4. "Otherwise, set connection's signaling state to closed."

alvestrand:
This was modelled on websockets' behavior: "if we messed up royally, we bail out by closing".
May or may not be appropirate.

alvestrand:
Related to #389 and #530 - the two other places where the PC is closed by the UA.
@adam-be is working on those places.

adam-be:
Proposed fix #552

alvestrand:
Merged #552

PR #552, opened 2016-03-17
Never close the RTCPeerConnection if setting a local/remote description fails (issue #364)

adam-be:
* Removed text about rollback since the 'process to apply description' should either succeed or fail. And in case of failure, the description should not be updated.
* Added a "catch all/rest" error case.


stefhak:
LGTM

jan-ivar:
lgtm.

Issue #530, opened 2016-03-02
RTCPeerConnectionState is a better home for "closed" than RTCSignalingState is


adam-be:
enum RTCSignalingState {
    "stable",
    "have-local-offer",
    "have-remote-offer",
    "have-local-pranswer",
    "have-remote-pranswer",
    "closed" // <-- do I belong here?
};

enum RTCPeerConnectionState {
    "new",
    "connecting",
    "connected",
    "disconnected",
    "failed"
};

alvestrand:
Let's make a PR to move.

alvestrand:
Duplicate of #389

PR #553, opened 2016-03-17
Move 'closed' state from RTCSignalingState to RTCPeerConnectionState (issue #530)

adam-be:
Added [[isClosed]] internal slot to keep track of the closed state that is exposed as an RTCPeerConnectionState.

Signaling state drawing is not updated yet.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -475,6 +475,11 @@
           </li>
 
           <li>
+            <p>Let <var>connection</var> have an [[<dfn>isClosed</dfn>]]
+            internal slot, initialized to <code>false</code>.</p>
+          </li>
+
+          <li>
             <p>Set <var>connection</var>'s <a>
             signaling state</a> to <code>stable</code>.</p>
           </li>
@@ -586,8 +591,8 @@
           </li>
 
           <li>
-            <p>If <var>connection</var>'s
-            <a>signaling state</a> is <code>closed</code>, abort these steps.</p>
+            <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+            <code>true</code>, abort these steps.</p>
           </li>
 
           <li>
@@ -648,10 +653,9 @@
 
         <ol>
           <li>
-            <p>If <var>connection</var>'s
-            <a>signaling
-            state</a> is <code>closed</code>, the user agent MUST return a
-            promise rejected with an <code>InvalidStateError</code>.</p>
+            <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+            <code>true</code>, the user agent MUST return a promise rejected
+            with an <code>InvalidStateError</code>.</p>
           </li>
 
           <li>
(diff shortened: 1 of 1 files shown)


alvestrand:
Asking @aboba and @jan-ivar to take a look.

webrtc.html
```diff
@@ -1677,7 +1674,8 @@
 
         <dl class="idl" title="partial interface RTCPeerConnection">
 
-          <dt>void createOffer (RTCSessionDescriptionCallback successCallback,
+          <dt>void createOffer (RTCSessionDescriptionCallback successCallb
+          ack,
```
adam-be: Fix this!

burnburn:
LGTM

webrtc.html
```diff
@@ -1910,8 +1908,8 @@
 
         <p>An <code><a>RTCPeerConnection</a></code> object MUST not be garbage
         collected as long as any event can cause an event handler to be
-        triggered on the object. When the object's
-        <a>signaling state</a> is <code>closed</code>, no such event handler can be
+        triggered on the object. When the object's [[<a>isClosed</a>]]
+        internal slot is <code>true</code>, no such event handler can be
```
jan-ivar: Not a problem with this patch, but this reminds me this section on garbage collection is quite wrong. All closed peer connections are not "safe to garbage collect", as this sentence runs on saying. Nevermind any outstanding JS references, but I thought we agreed one can call `getStats` on closed peer connections as well. Is this section even needed, or is being derived from `EventTarget` perhaps sufficient (I don't know, just guessing it might be)?

webrtc.html
```diff
@@ -3030,10 +3024,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
               </li>
 
               <li>
-                <p>If <var>connection</var>'s <a>signaling state</a>
-                is <code>closed</code>, throw an
-                <code>InvalidStateError</code> exception and abort these
-                steps.</p>
+                <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+                <code>true</code>, throw an <code>InvalidStateError</code>
+                exception and abort these steps.</p>
```
jan-ivar: comma (wasn't there before, but seems needed) before and.

jan-ivar:
lgtm.

jan-ivar:
Careful when merging this, as the section on rollback appears to overlap with https://github.com/w3c/webrtc-pc/pull/552 ?

webrtc.html
adam-be: I interpret that section as containing additional rules that come into play when no JS references are left. Otherwise it's not OK to attempt to gc in the first place.

adam-be:
Yes, it's was easier to let this change share history with #552. I'll rebase it now when #552 has landed. Speaking of #552, it was merged quite quickly. It would be great if you could take a look at it @jan-ivar?

PR #557, opened 2016-03-17
Splitting apart RTCIceConnectionState and RTCIceTransportState.

taylor-b:
The fundamental state definitions, example state transitions, and
non-normative state transition diagram now all move to the
RTCIceTransportState section, and are defined in terms of
RTCIceTransports instead of an "ICE Agent".

The RTCIceConnectionState section now contains explicit rules for how
multiple RTCIceTransport states are aggregated, similar to how the
RTCPeerConnectionState is defined. So, situations that were previously
ambiguous (such as "what if one transport is 'new' and the other
is 'connected'?") are now well-defined.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2072,120 +2072,51 @@
         <dl class="idl" title="enum RTCIceConnectionState">
           <dt>new</dt>
 
-          <dd>The <a>ICE Agent</a> is gathering addresses and/or waiting for remote
-          candidates to be supplied, and has not yet started checking.</dd>
+          <dd>Any of the <code><a>RTCIceTransport</a></code>s are in the
+          <code>new</code> state and none of them are in the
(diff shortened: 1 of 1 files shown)


taylor-b:
This PR should be exactly what we talked about in the last virtual interim. The only thing in question is how states get aggregated from multiple RTCIceTransports, since this is something that was ambiguous before now.

I wrote the PR with the following priorities for states: 

1. Failed
2. Disconnected
3. Checking
4. New
5. Connected
6. Completed
7. Closed

Meaning that if RTCIceTransports have different states, the one with the highest priority wins. So if one transport is "disconnected" while the other is "failed", the aggregate state is "failed". If one is "new" while another is "connected", the aggregate state is "new". And so on.

I'm very open to suggestions on how this priority order should be changed. It's just an initial pass.

aboba:
LGTM.

PR #558, opened 2016-03-18
Revert "Added initialization of readonly and remote attributes for in…

### LINKED ISSUE #556: Remove track attributes "remote" and "readonly"
As a consequence of them having been removed in mediacapture-main

stefhak:
…coming tracks."

This reverts commit 5d1a60576687910cc61eaebe97482d922db35ca6, meaning we remove initialization of remote and readonly when creating MediaStreamTracks on PeerConnection. The reason is that these attributes have been removed from the definition of MediaStreamTrack in mediacapture-main.

Close #556 .

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3318,16 +3318,6 @@
                 </li>
 
                 <li>
-                  <p>Initialize <code><var>track</var>.readonly</code>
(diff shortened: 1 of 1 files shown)


adam-be:
LGTM

PR #560, opened 2016-03-21
Changing from callback interface to dictionary for RTCIdentityProvider

### LINKED ISSUE #559: Redesign RTCIdentityProvider to not be a callback interface
We only have 3 callback interfaces in the platform and all of them are historical mistakes. Let's not introduce a fresh one.

See also https://github.com/heycam/webidl/issues/100.

martinthomson:
Closes #559.


jan-ivar:
You missed two "RTCIdentityProvider interface" in https://rawgit.com/martinthomson/webrtc-pc/rtcidentity-dict/webrtc.html#conformance

PR #587, opened 2016-04-21
First half hearted shot at #529.

stefhak:


webrtc.html
```diff
@@ -6777,6 +6777,26 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
+
+      <section>
+        <h4>MediaTrackSupportedConstraints, MediaTrackCapabilities, 
+        MediaTrackConstraints and MediaTrackSettings</h4>
+
+        <p>The basics of <code><a>MediaTrackSupportedConstraints</a></code>, 
+        <code><a>MediaTrackCapabilites</a></code>,
+        <code><a>MediaTrackConstraints</a></code> and
+        <code><a>MediaTrackSettings</a></code> is outlined in [[!GETUSERMEDIA]].
+        However, the <code><a>MediaTrackSettings</a></code> for a
```
alvestrand: Suggest deleting "however". Don't make excuses :-)

stefhak:
I agree, "however" should probably go. However (sic), that sentence is informative rather than normative, I was thinking it should perhaps be a note of some sort?

PR #592, opened 2016-04-21
latest respec2html needs node >= 5.0

vivienlacourba:
and it can be installed with nvm only from the .travis.yml file

This is based on:
https://github.com/travis-ci/travis-ci/issues/3900#issuecomment-102249706
http://stackoverflow.com/a/36033744

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ addons:
       - libcss-dom-perl
       - python-lxml
       - cmake
-      - cmake-data
       - gcc-4.8
       - g++-4.8
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,7 @@ addons:
       - gcc-4.8
       - g++-4.8
 before_install:
-  - export CXX="g++-4.8" CC="gcc-4.8"
-  - nvm install 5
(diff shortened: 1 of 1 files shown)


vivienlacourba:
@dontcallmedom, update: it is now using gcc-4.8 as expected and I believe building the node dependencies correctly.
Unfortunately calling Respec does not generate `build/output.html` and does not produce any output either (see Travis [build #1058 line 1319](https://travis-ci.org/w3c/webrtc-pc/builds/124815066#L1319)) whereas locally I have the following warning message from Respec:
```
😳 ReSpec warn: Defining WebIDL in `dl` elements is deprecated. Please use Contiguous IDL instead: https://www.w3.org/respec/guide.html#contiguous-idl
```

dontcallmedom:
great progress; I know what's needed to fix this, will get to it tomorrow

vivienlacourba:
FYI I added an `ls -l` of the file generated by respec, which confirms that respec did not generated any output (ans also makes the Makefile stops sooner). I believe that when it fails resepec should return an error code.

Also I was wondering if switching to gcc 4.9 would help here (it is also available in the `ubuntu-toolchain-r-test` apt source repo).

Issue #584, opened 2016-04-15
currentLocalDescription: incongruous explanation


mparis:
Hello,
if I am not wrong, the currentLocalDescription is the one that _was successfully negotiated_ and NOT the last updated by setLocalDescription because you cannot have it without having negotiated with the other peer, isn't it?

See this paragraph:
The currentLocalDescription attribute represents the local RTCSessionDescription that was successfully negotiated the last time theRTCPeerConnection transitioned into the stable state plus any local candidates that have been generated by the ICE Agent since the offer or answer was created. This attribute is updated by setLocalDescription(). [1]

Refs
[1] http://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-currentLocalDescription

mparis:
The same for currentRemoteDescription

taylor-b:
So specifically, the problem is the sentence "This attribute is updated by setLocalDescription()."? Since it can be updated by setLocalDescription *or* setRemoteDescription? I agree, this should be corrected.

jan-ivar:
@taylor-b It is my understanding it is updated by the transition to `stable` state.

alvestrand:
Fixed by #580, I think.

jan-ivar:
@alvestrand I don't think so. https://github.com/w3c/webrtc-pc/pull/580 still says that `currentLocalDescription` and `currentRemoteDescription` are updated by `setLocalDescription` and `setRemoteDescription` respectively, when I believe - and @mparis let me know if you disagree - it is more correct to say they are updated by the transition to `'stable'` state.

mparis:
@jan-ivar @alvestrand, what do you think about removing "This attribute is updated by setLocalDescription()" like in #/594?

PR #594, opened 2016-04-25
webrtc.html: fix currentLocalDescription and currentRemoteDescription…

mparis:


jan-ivar:
wfm.

mparis:
This PR is related to issue #584

adam-be:
There was a suggestion to say something about these attributes being updated in the transition to the 'stable' state. That would only be for information/documentation purposes, but it helps to understand what the attribute represents.

PR #595, opened 2016-04-25
Fix rtcpTransport description

mparis:
This PR depends on #594

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4392,7 +4392,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               which media from <code>track</code> is sent in the form of RTP
               packets. When BUNDLE is used, many
               <code><a>RTCRtpSender</a></code> objects will share one
-              <code>transport</code> and will all send RTP over the same
+              <code>transport</code> and will all send RTP and RTCP over the same
               transport. When RTCP mux is used, <code>rtcpTransport</code> will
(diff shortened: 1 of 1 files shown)


alvestrand:
When you make two PRs that build on top of each other like this, please make a note of that in the description. It's otherwise confusing to see unrelated changes in the diff.

aboba:
The PR removes statements describing the effect of BUNDLE which were put in to respond to another issue.

alvestrand:
This seems to be the text Bernard's noting: "and will all send and receive RTCP over the same transport."

Needs a rebase now that its parent is merged.

alvestrand:
Don't see the point here - it's not just a langauge cleanup, but it's not clear what it improves.
Unless a better reason is found, we'll close this.

mparis:
With this patch, I want to clarify that the management of the rtcpTransport is the same in the case of BUNDLE that for rtcp-mux (it will be null), because BUNDLE mandates the usage of rtcp-mux.

alvestrand:
It loses the explicit note that with BUNDLE, all the RTCP for all the flows will go over the same transport channel.
(I think that's what @aboba intended to say).

alvestrand:
@mparis this is waiting for your response.

aboba:
Replaced by rebased PR https://github.com/w3c/webrtc-pc/pull/677

Issue #600, opened 2016-04-27
Operations queue: What is run synchronously before the operation is queued?


adam-be:
The current spec text indicates that the whole called to a queued operation is "wrapped" to be executed when the operation is picked up from the queue. We should probably run, for example, WebIDL type checking before the operation is queued as the Firefox implementation does.

adam-be:
Closed by #721, #744, #751 and #756

PR #601, opened 2016-04-27
Specify the synchronous and queued steps for addIceCandidate (issue #600)

adam-be:
This PR is only for reviewing since it doesn't update all queued operations; only addIceCandidate.

Issue #600

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -474,6 +474,12 @@
           </li>
 
           <li>
+            <p>Let <var>connection</var> have an [[<dfn>operations</dfn>]]
+            internal slot, representing an <a>operations queue</a>, initialized
+            to an empty list.</p>
+          </li>
+
+          <li>
             <p>Set <var>connection</var>'s <a>
             signaling state</a> to <code>stable</code>.</p>
           </li>
@@ -523,39 +529,45 @@
           </li>
         </ol>
 
-        <p>Once the RTCPeerConnection object has been initialized, for every
-        call to <code>createOffer</code>, <code>setLocalDescription</code>,
-        <code>createAnswer</code>, <code>setRemoteDescription</code>,
-        and <code>addIceCandidate</code>,
-        execute the following steps:</p>
+        <p>An <code><a>RTCPeerConnection</a></code> object has an
+        <dfn>operations queue</dfn>, [[<a>operations</a>]], which enforces that
+        only one operation in the queue is executed at a given time. If
+        subsequent calls are made while the returned promise of a previous call
+        is still unsettled, they are added to the queue and executed when all
+        the previous calls are executed and their promises are settled.</p>
+
+        <p>To <dfn id="enqueue-an-operation">enqueue an operation</dfn>, run the
+        following steps:</p>
 
         <ol>
+          <li>
+            <p>Let <var>operation</var> be the operation to be enqueued.</p>
+          </li>
+
           <li>
             <p>Let <var>p</var> be a new promise.</p>
           </li>
 
           <li>
-            <p>Append an object representing the current call being handled
-            (i.e. function name and corresponding arguments) to the
-            <var>operations</var> array.</p>
+            <p>Append <var>operation</var> to [[<a>operations</a>]].</p>
           </li>
 
           <li>
-            <p>If the length of the <var>operations</var> array is exactly 1,
-            execute the object from the front of the queue.</p>
+            <p>If the length of [[<a>operations</a>]] is exactly 1,
+            execute <var>operation</var>.</p>
           </li>
 
           <li>
             <p>Upon fulfillment or rejection of the promise returned
-            by the function, fulfill or reject <var>p</var> with the
+            by <var>operation</var>, fulfill or reject <var>p</var> with the
             corresponding value or reason.  Upon fulfillment or
             rejection of <var>p</var>, execute the following
             steps:</p>
             <ol>
-              <li><p>Remove the corresponding object from
-                the <var>operations</var> array.</p></li>
-              <li><p>If the array is non-empty, execute the first
-                object queued.</p></li>
+              <li><p>Remove the first element of [[<a>operations</a>]].</p></li>
+              <li><p>If [[<a>operations</a>]] is non-empty, execute the
+              operation represented by the first element of
+              [[<a>operations</a>]].</p></li>
             </ol>
           </li>
 
@@ -563,15 +575,6 @@
             <p>Return <var>p</var>.</p>
           </li>
         </ol>
-
-        <p>The general idea is to have only one among <code>createOffer</code>,
-        <code>setLocalDescription</code>, <code>createAnswer</code> and
-        <code>setRemoteDescription</code>
-        and <code>addIceCandidate</code> executing at any given
-        time. If subsequent calls are made while the returned promise
-        of a previous call is still unsettled, they are added to a
-        queue and executed when all the previous calls are executed
-        and their promises are settled.</p>
 
         <p>When a new ICE candidate is available or when the ICE gathering process is done
         , the user agent MUST queue a task to run the
@@ -1262,7 +1265,30 @@
             <a href="#widl-RTCIceCandidate-sdpMid">sdpMid</a></code> and <code>
             <a href=
             "#widl-RTCIceCandidate-sdpMLineIndex">sdpMLineIndex</a></code>; the
-            rest are ignored.</p>
+            rest are ignored. When the method is invoked, the User Agent MUST
+            run the following steps:</p>
+
+            <ol>
+              <li>
+                <p>Let <var>candidate</var> be the methods argument.</p>
+              </li>
+
+              <li>
+                <p>If <var>candidate</var> is not <code>null</code> but is
+                missing values for both <var>sdpMid</var> and
+                <var>sdpMLineIndex</var>, return a promise rejected with a
+                <code>TypeError</code>.</p>
+              </li>
+
+              <li>
+                <p>Return the result of <a href=
+                "#enqueue-an-operation">queuing</a> <a>addIceCandidate's queued
+                steps</a>.</p>
+              </li>
+            </ol>
+
+            <p>The following steps defines <dfn>addIceCandidate's queued
+            steps</dfn>:</p>
 
             <ol>
               <li>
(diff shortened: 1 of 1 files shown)


adam-be:
@jan-ivar, could you take a look at this please.

webrtc.html
```diff
@@ -474,6 +474,12 @@
           </li>
 
           <li>
+            <p>Let <var>connection</var> have an [[<dfn>operations</dfn>]]
+            internal slot, representing an <a>operations queue</a>, initialized
+            to an empty list.</p>
```
jan-ivar: You wanna remove the corresponding old mention of *operations* in step 10.

webrtc.html
```diff
@@ -523,39 +529,45 @@
           </li>
         </ol>
 
-        <p>Once the RTCPeerConnection object has been initialized, for every
-        call to <code>createOffer</code>, <code>setLocalDescription</code>,
-        <code>createAnswer</code>, <code>setRemoteDescription</code>,
-        and <code>addIceCandidate</code>,
-        execute the following steps:</p>
+        <p>An <code><a>RTCPeerConnection</a></code> object has an
+        <dfn>operations queue</dfn>, [[<a>operations</a>]], which enforces that
```
jan-ivar: Maybe "ensures"?

webrtc.html
```diff
@@ -523,39 +529,45 @@
...
+        only one operation in the queue is executed at a given time. If
+        subsequent calls are made while the returned promise of a previous call
+        is still unsettled, they are added to the queue and executed when all
```
jan-ivar: I would say "not settled" to make sure we use the promise terms verbatim.

webrtc.html
```diff
@@ -523,39 +529,45 @@
...
         </ol>
 
-        <p>Once the RTCPeerConnection object has been initialized, for every
-        call to <code>createOffer</code>, <code>setLocalDescription</code>,
-        <code>createAnswer</code>, <code>setRemoteDescription</code>,
-        and <code>addIceCandidate</code>,
-        execute the following steps:</p>
+        <p>An <code><a>RTCPeerConnection</a></code> object has an
+        <dfn>operations queue</dfn>, [[<a>operations</a>]], which enforces that
+        only one operation in the queue is executed at a given time. If
```
jan-ivar: Not a must, but how about: "only one asynchronous operation in the queue is executed concurrently." ?

webrtc.html
```diff
@@ -523,39 +529,45 @@
...
+        subsequent calls are made while the returned promise of a previous call
+        is still unsettled, they are added to the queue and executed when all
+        the previous calls are executed and their promises are settled.</p>
```
jan-ivar: have settled?

webrtc.html
```diff
@@ -523,39 +529,45 @@
...
+
+        <p>To <dfn id="enqueue-an-operation">enqueue an operation</dfn>, run the
+        following steps:</p>
 
         <ol>
           <li>
+            <p>Let <var>operation</var> be the operation to be enqueued.</p>
```
jan-ivar: Alternatively, might this be another opportunity to elaborate on a definition of "operation"? Soemthing like "Let *operation* be the asynchronous function to be enqueued"?

webrtc.html
```diff
@@ -1262,30 +1265,42 @@
...
-            rest are ignored.</p>
+            rest are ignored. When the method is invoked, the User Agent MUST
+            run the following steps:</p>
 
             <ol>
               <li>
-                <p>Let <var>connection</var> be the <code>
-                <a>RTCPeerConnection</a></code> object on which the method was
-                invoked.</p>
+                <p>Let <var>candidate</var> be the methods argument.</p>
```
jan-ivar: method's

webrtc.html
```diff
@@ -1262,30 +1265,42 @@
...
+                <p>If <var>candidate</var> is not <code>null</code> but is
+                missing values for both <var>sdpMid</var> and
+                <var>sdpMLineIndex</var>, return a promise rejected with a
+                <code>TypeError</code>.</p>
               </li>
 
               <li>
-                <p>Let <var>candidate</var> be the methods argument.</p>
+                <p>Return the result of <a href=
+                "#enqueue-an-operation">queuing</a> <a>addIceCandidate's queued
```
jan-ivar: We probably want to use "enqueuing" here to be consistent (so the document makes sense even without hyperlinks).

webrtc.html
```diff
@@ -1262,30 +1265,42 @@
...
+                steps</a>.</p>
               </li>
+            </ol>
+
+            <p>The following steps defines <dfn>addIceCandidate's queued
+            steps</dfn>:</p>
```
jan-ivar: The naming seems redundant and not used anywhere else. Couldn't we just say

 "Return the result of [enqueuing](https://rawgit.com/adam-be/webrtc-pc/queued-operations-sync/webrtc.html#enqueue-an-operation) the following steps:" ?

jan-ivar: and indent them one level. That seems consistent with e.g. how we run things in parallel.

webrtc.html
```diff
@@ -1262,30 +1265,42 @@
...
 
+            <ol>
               <li>
-                <p>If <var>candidate</var> is not <code>null</code> but is
-                missing values for both <var>sdpMid</var> and
-                <var>sdpMLineIndex</var>, return a promise rejected with a
-                <code>TypeError</code>.</p>
+                <p>Let <var>connection</var> be the <code>
+                <a>RTCPeerConnection</a></code> object on which the method was
+                invoked.</p>
```
jan-ivar: I think it would make sense to move this step out of the enqueued steps, to the steps above, to be next to `candidate` and the others. Otherwise the context for this text seems wrong.

jan-ivar:
@adam-be, my numerous nits aside, I think this is great!

webrtc.html
adam-be: Missed that one

webrtc.html
adam-be: I'm a bit hesitant to use 'function call' since that somewhat indicates that we're enqueuing a function, when we're actually queueing a part of a function.

webrtc.html
adam-be: That makes sense.

webrtc.html
adam-be: Yes.

adam-be:
Thanks for reviewing @jan-ivar. New version coming up.

webrtc.html
adam-be: That's better.

From this point on, no comment on review point == fixed

webrtc.html
```diff
@@ -523,39 +524,46 @@
...
-        <p>Once the RTCPeerConnection object has been initialized, for every
-        call to <code>createOffer</code>, <code>setLocalDescription</code>,
-        <code>createAnswer</code>, <code>setRemoteDescription</code>,
-        and <code>addIceCandidate</code>,
-        execute the following steps:</p>
+        <p>An <code><a>RTCPeerConnection</a></code> object has an
+        <dfn>operations queue</dfn>, [[<a>operations</a>]], which ensures that
+        only one asynchronous operation in the queue is executed concurrently. If
+        subsequent calls are made while the returned promise of a previous call
+        is still not settled, they are added to the queue and executed when all
```
alvestrand: what does "settled" mean? "resolved" or "rejected"? (is this a known term?)

jan-ivar: It's from [states and fates](https://github.com/domenic/promises-unwrapping/blob/master/docs/states-and-fates.md) at least.

alvestrand:
@pthatcherg can you take a look at this?
I'm not sure what the current Chrome implementation does here.

adam-be:
Link to preview:
http://rawgit.com/adam-be/webrtc-pc/queued-operations-sync/webrtc.html#widl-RTCPeerConnection-addIceCandidate-Promise-void--RTCIceCandidateInit-RTCIceCandidate-candidate

webrtc.html
jan-ivar: How does one enqueue part of a function?

webrtc.html
```diff
@@ -1289,97 +1283,110 @@
...
                       </li>
 
                       <li>
-                        <p>If the <a>ICE Agent</a> is not currently checking
-                        candidate pairs, the <a>ICE Agent</a> MUST start
-                        checking candidate pairs and <a data-lt=
-                        "update the ICE connection state">update
-                        <var>connection</var>'s ICE connection state</a> to
-                        <code>checking</code>.</p>
+                        <p>If <var>candidate</var> could not be successfully,
```
jan-ivar: Lose the comma

webrtc.html
```diff
@@ -1289,97 +1283,110 @@
...
-                    </ol>
+                    <p>Let <var>p</var> be a new promise.</p>
                   </li>
 
                   <li>
-                    <p>If <var>candidate</var> is applied successfully, the
-                    User Agent MUST queue a task that runs the following
-                    steps:</p>
+                    <p>In parallel, start the process to apply
+                    <var>candidate</var>.</p>
```
jan-ivar: I'd prefer the more pedantic "run the following steps:", but maybe a colon at the end at least?

webrtc.html
```diff
@@ -1289,97 +1283,110 @@
...
 
+                    <ol>
                       <li>
-                        <p>Add <var>candidate</var> to
-                        <var>remoteDescription</var>.</p>
+                        <p>If <var>candidate</var> is <code>null</code>, the User
+                        Agent MUST queue a task that runs the following steps:</p>
+                        <ol>
+                          <li>
+                            <p>For each <a>media description</a> in the last
```
jan-ivar: You probably want to check [[isClosed]] here too.

jan-ivar:
A few more nits, otherwise lgtm!

jan-ivar:
@adam-be What's blocking this? I had to look up today whether *createOffer* is supposed to throw on close and I couldn't find the answer in the current spec, and was surprised this never landed.

webrtc.html
adam-be: We have an algorithm, that defines the behavior of a function, and we're saying that these steps are run synchronously, and these steps are queued.

webrtc.html
adam-be: True

webrtc.html
adam-be: We're "calling" a JSEP-process here and waiting for the result. Compare with the step 5.1 in [1].

[1] https://www.w3.org/2001/tag/doc/promises-guide#example-add-bookmark

adam-be:
@jan-ivar, the move to the new WebIDL style and reflow of the document made this one hard to rebase. It's now replaced by #721 which we should merge as soon as possible.

adam-be:
Replaced by #721

Issue #581, opened 2016-04-15
No error defined for addTransceiver() with a bogus track kind


adam-be:
We need an error for:
pc.addTransceiver("bogus");

Some possible types from the WebIDL spec:

"InvalidAccessError" 	The object does not support the operation or argument.
"SyntaxError" 	The string did not match the expected pattern.
"NotSupportedError" 	The operation is not supported.
"InvalidModificationError" 	The object can not be modified in this way.
"OperationError" 	The operation failed for an operation-specific reason.

dontcallmedom:
I would add "TypeError" to that list; if MediaStreamTrack.kind was an enum rather than a generic DOMString, this would be what would be triggered.

adam-be:
True

alvestrand:
TypeError is consistent with other behaviors. I think (not sure) that we've specified this elsewhere too.

alvestrand:
Seems that the conclusion is that TypeError is uncontroversial.

alvestrand:
Fixed in #602

PR #602, opened 2016-04-28
addTransceiver(): Throw a TypeError on a bogus track kind (issue: #581)

adam-be:
Added text to the addTransceiver() description.

We probably want to create explicit steps here further on so that the order of error checks and such are defined.

Issue: #581


stefhak:
LGTM.

webrtc.html
```diff
@@ -3119,6 +3119,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             be set to that track and the MSID and media type generated
             by <code>createOffer</code> will be that of the track.</p>
 
+            <p>If a kind is passed in and the value is not <code>audio</code>
+            or <code>video</code>, throw a <code>TypeError</code>.</p>
```
alvestrand: Can we refer to "a legal value of the track.kind" or something like that instead of hard-coding "audio & video"?

elagerway:
lgtm

Issue #542, opened 2016-03-10
Server cannot be reached - Issues with IPv6


aboba:
In Section 4.9.3.2 it is stated: 
If the server could not be reached, errorCode will be set to a TBD value in the 7XX range, as this does not conflict with the STUN error code range.

This will be a problem for IPv6 where the browser can have interfaces (such as tunnel interfaces) with routability problems. 

From Robin Raymond:
"I'm not sure this particular error is wise. In any implementation (especial IPV6), it's very likely many host candidates will not be able to reach the TURN server or STUN server because they don't route appropriately. So you will warn/flag these as being failed but with other host candidates they actually succeed. The programmer will think things are going horribly wrong when this is probably normal.

Further, if you don't indicate that this URI is actually succeeding then the programmer is going to be somewhat in the dark as to what's going on.

I would only use the "cannot contact server" error if NONE of the host IPs could reach the server. (Although that could just indicate a general connectivity problem and not really a server problem)."

fluffy:
We need someway to debug which TURN serves could not be reach.

PR #610, opened 2016-05-02
Server cannot be reached - Issues with IPv6

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/542

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2641,7 +2641,7 @@
           <dd>
             <p>The <code>errorCode</code> attribute is the numeric STUN
             error code returned by the STUN or TURN server [[STUN-PARAMETERS]].</p>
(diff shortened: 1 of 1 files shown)


aboba:
@taylor-b Can you review?t.

webrtc.html
```diff
@@ -2641,9 +2641,10 @@
           <dd>
             <p>The <code>errorCode</code> attribute is the numeric STUN
             error code returned by the STUN or TURN server [[STUN-PARAMETERS]].</p>
-            <p>If the server could not be reached, <code>errorCode</code> will
+            <p>If the server could not be reached from all host candidates, <code>errorCode</code> will
```
taylor-b: I think this would be more clear if worded as "If no host candidate could reach the server, ..."

Issue #517, opened 2016-02-22
Clarify ICE consent freshness feedback


nils-ohlmeier:
When implementing [RFC 7665](https://tools.ietf.org/html/rfc7675) it is not very well defined how the different state transitions from that RFC get reported back through the PeerConnection APIs.

nils-ohlmeier:
I think it is kind of obvious to use the RTCIceConnectionState to communicate the events.

As RFC 7665 mandates that loosing consent after the 30s timer is permanent for a given pair translating loosing consent for a single component into the 'failed' state appears appropriate.
The open question for me is if each STUN transaction timeout from 7665 should get translated into 'disconnected', followed by potentially switching back to 'connected' or 'completed' once consent has been re-established.

alvestrand:
@nils-ohlmeier can you check #457 to see if this is part of the same problem?
Also check the slides for the interim here:
https://docs.google.com/presentation/d/1R7lQeo4LNGu0KKW9GLdkwJ5x8FUmR4qrnhfPFZO3WS8/edit#slide=id.gd4bd98a32_0_387

@elagerway opinions?

nils-ohlmeier:
In some yes it is the same problem, as it also touches the same state diagram.

Whether [CONSENT] final 30s timeout leads to 'failed' or to 'disconnected' (or 'checking' as proposed on the slides for the checking state) does not really matter to me. It just needs to be defined and written down. That should be easy.
The open questions which I don't have a good answer to (yet) is if a timeout on a single STUN transaction from [CONSENT] should result in a (temporary) state change (and if so to which state). From reading through #457 that does not appear as question which has been raised. But I'll happily add it to the discussion in #457, because it makes sense to clean that up together.

aboba:
The outcome of each individual consent check should not be reflected in either RTCIceTransportState or RTCIceConnectionState, because there is nothing a developer can or should do about the failure of individual consent checks.  Consent request/response statistics is another matter. 

My suggestion is that a consent failure in the "completed" state should cause a transition to "failed" for both RTCIceTransportState and RTCIceConnectionState.

aboba:
@nils-ohlmeier @alvestrand  I have submitted RTCIceTransportState definitions that at least partially takes consent check failures into account:  PR https://github.com/w3c/webrtc-pc/pull/533

aboba:
Related to https://github.com/w3c/webrtc-pc/issues/457

alvestrand:
Depends on #457

[...]

PR #611, opened 2016-05-02
Clarify ICE consent freshness feedback

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/517

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4482,7 +4482,7 @@ sender.setParameters(params)
           <li>(<code>checking</code>, found usable connection):
           <code>connected</code></li>
 
-          <li>(<code>checking</code>, checks fail but gathering still in progress):
+          <li>(<code>checking</code>, checks fail, but gathering still in progress):
           <code>disconnected</code></li>
 
           <li>(<code>checking</code>, gave up): <code>failed</code></li>
@@ -4493,8 +4493,8 @@ sender.setParameters(params)
           <li>(<code>connected</code>, finished all checks):
           <code>completed</code></li>
 
-          <li>(<code>completed</code>, lost connectivity):
-          <code>disconnected</code></li>
+          <li>(<code>completed</code>, consent lost):
+          <code>failed</code></li>
 
           <li>(any state, ICE restart occurs): <code>new</code></li>
 
--- a/webrtc.html
+++ b/webrtc.html
@@ -4426,7 +4426,10 @@ sender.setParameters(params)
           there is a better connection. It may also still be gathering and/or
           waiting for additional remote candidates. If consent checks [[!RFC7675]]
           fail on the connection in use, and there are no other successful
-          candidate pairs available, then the state transitions to "disconnected".</dd>
+          candidate pairs available, then the state transitions to "checking" 
+          (if there are candidate pairs remaining to be checked) or "disconnected"
+          (if there are no candidate pairs to check, but the peer is still gathering
+          and/or waiting for additional remote candidates).</dd>
 
           <dt>completed</dt>
 
--- a/webrtc.html
+++ b/webrtc.html
@@ -4414,7 +4414,7 @@ sender.setParameters(params)
           <dt>checking</dt>
 
           <dd>The <code><a>RTCIceTransport</a></code> has received at least one
-          remote candidate, and is checking candidate pairs but has not yet
+          remote candidate, and either is checking candidate pairs but has not yet
           found a connection, or consent checks [[!RFC7675]] have failed on a
           previously successful candidate pair.  In addition to checking, it
           may also still be gathering.</dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -4415,8 +4415,8 @@ sender.setParameters(params)
 
           <dd>The <code><a>RTCIceTransport</a></code> has received at least one
           remote candidate, and either is checking candidate pairs but has not yet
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4415,8 +4415,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
 
           <dd>The <code><a>RTCIceTransport</a></code> has received at least one
           remote candidate, and is checking candidate pairs but has not yet
-          found a connection. In addition to checking, it may also still be
-          gathering.</dd>
+          found a connection, or consent checks [[!RFC7675]] have failed on a
```
taylor-b: Not a big deal, but it's not 100% clear that the boolean logic is "A and (B or C)" and not "(A and B) or C". Maybe adding "either" would help: "... has received at least remote candidate, and either is checking ..."

webrtc.html
```diff
@@ -4429,7 +4430,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
 
           <dd>The <code><a>RTCIceTransport</a></code> has finished gathering,
           received an indication that there are no more remote candidates,
-          finished checking all candidate pairs and found a connection.</dd>
+          finished checking all candidate pairs and found a connection.
+          If consent checks [[!RFC7675]] subsequently fail, the state 
+          transitions to "failed".</dd>
 
           <dt>failed</dt>
```
taylor-b: Don't you need to add something in the "failed" section"?

webrtc.html
```diff
@@ -4429,7 +4430,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
 
           <dd>The <code><a>RTCIceTransport</a></code> has finished gathering,
           received an indication that there are no more remote candidates,
-          finished checking all candidate pairs and found a connection.</dd>
+          finished checking all candidate pairs and found a connection.
+          If consent checks [[!RFC7675]] subsequently fail, the state
```
taylor-b: If we're going to add this sentence in "completed", should we do the same in "connected"? "If consent checks fail, the state transitions to disconnected"?

taylor-b:
Besides the comments, I have a broader question about this change. If the candidate pair currently being used loses consent, but another candidate pair still has consent, can the IceTransport switch to it? I believe that's what Chrome would currently do. In that case, the state should only be "failed" if _every_ pair either failed checking or lost consent.

aboba:
@taylor-b Agree that state should only be "failed" if every pair either failed checking or lost consent. Also, if state was "connected" and consent is lost on the candidate pair currently being used, but another candidate pair still has consent, state is still "connected", right? (e.g. no transition to "disconnected").

webrtc.html
aboba: If consent checks fail in "connected" the state transitions to "disconnected" only if there are no successful candidate pairs left.

webrtc.html
```diff
@@ -4414,37 +4414,43 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           <dt>checking</dt>
 
           <dd>The <code><a>RTCIceTransport</a></code> has received at least one
-          remote candidate, and is checking candidate pairs but has not yet
-          found a connection. In addition to checking, it may also still be
-          gathering.</dd>
+          remote candidate, and either is checking candidate pairs but has not yet
```
taylor-b: It looks like I suggested putting the "either" in the wrong spot; my apologies. The intended logic is "received candidate AND is checking AND (no connection found OR consent checks failed)"

webrtc.html
```diff
@@ -4414,37 +4414,43 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
 
           <dt>connected</dt>
 
           <dd>The <code><a>RTCIceTransport</a></code> has found a usable
           connection, but is still checking other candidate pairs to see if
           there is a better connection. It may also still be gathering and/or
-          waiting for additional remote candidates.</dd>
+          waiting for additional remote candidates. If consent checks [[!RFC7675]]
+          fail on the connection in use, and there are no other successful
+          candidate pairs available, then the state transitions to "disconnected".</dd>
```
taylor-b: I guess it could transition to "checking" as well, depending on whether or not the transport is currently checking.

taylor-b:
@aboba I agree that if a pair loses consent but the IceTransport can seamlessly switch to a different pair, the state should stay "connected". The only necessary indication for this is a `selectedcandidatepairchange` event.

webrtc.html
aboba: If consent checks fail in "connected" and there are no successful candidate pairs left, the state transitions to "disconnected" if there are no candidate pairs left to check (but gathering is not complete or the last remote candidate has not been received), or to "checking" if there are candidate pairs left to check.

webrtc.html
```diff
@@ -4484,8 +4493,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           <li>(<code>connected</code>, finished all checks):
           <code>completed</code></li>
 
-          <li>(<code>completed</code>, lost connectivity):
-          <code>disconnected</code></li>
```
taylor-b: The "lost connectivity" transition to "disconnected" is still possible, so I don't think this needs to be removed.

taylor-b:
lgtm with a minor comment (take it or leave it)

nils-ohlmeier:
lgtm

elagerway:
lgtm

PR #616, opened 2016-05-04
Convert to WebIDL contiguous mode

dontcallmedom:
See https://lists.w3.org/Archives/Public/public-webrtc-editors/2015Dec/0015.html for context


adam-be:
One PR to break them all! :)

Issue #597, opened 2016-04-25
Calling RTCRtpReceiver.track.stop()


aboba:
What happens if RTCRtpReceiver.track.stop() is called? 

Related ORTC Issue: https://github.com/openpeer/ortc/issues/498

alvestrand:
@aboba what do you think should happen?

aboba:
receiver.track.stop() is final, so that RTCRtpReceiver.track cannot be rendered after that. 

However, receiver.track.stop() does not stop the RTCRtpReceiver so that the decoder remains operational and Receiver Reports continue to be sent, even though track cannot be rendered.

mparis:
Hello,
if the track cannot be rendered and it is not reversible, it makes not sense wasting resources (CPU for encoding, decoding, etc.) and network bandwidth.
So, I think that:

- The decoder should be stopped
- A renegotiation should be performed (set negotiation-needed flag , fire a negotiationneeded event, etc.)

In SDP negotiation this media should be set as "sendonly" or "Inactive" to notify the remote peer (_sender_) that the local peer (_receiver_) will not use the media any more.
So the _sender_ can:

- stop the encoder
- stop sending media through the network.

Best!!

pthatcherg:
I agree that the negotiated state of the RtpTransceiver and RtpReceiver should not be affected by stopping a track.  If an application wants to change negotiated state, it should use the RtpTransceiver.  That's what it's for.

However, I don't think we need to specify that the decoder must stay active. An implementation could choose to stop using CPU or other resources if it knows that nothing will use the decoded frames. I think we could just leave that part about the decoder being active out of the spec.

alvestrand:
The observable part of behavior after track stop is that the RTP session continues to send RTCP saying "I got your packets, thank you". Whether the implementation bit-buckets those packets before or after the decoder should be completely up to the implementation; it's not observable.

mparis:
@pthatcherg, if the app uses the RtpTransceiver.stop(), it stops the RtpReceiver and the RtpSender [1].
So, if it only wants to stop one of them, I think that we should add a stop() method in RtpReceiver and in RtpSender.

Refs
[1] http://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-stop

pthatcherg:
[...]

PR #617, opened 2016-05-04
Effect of RTCRtpReceiver.track.stop()

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/597

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3959,9 +3959,16 @@ sender.setParameters(params)
           <dt>readonly attribute MediaStreamTrack track</dt>
 
           <dd>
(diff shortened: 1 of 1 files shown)


aboba:
@adam-be Can you review?

stefhak:
Trying to wrap my head around this. Is `RTCRtpReceiver.track` always a single track? What happens if you do `RTCRtpReceiver.track.clone()`? Will the first one be a "master" track?
It would have made me feel better if this was analogous to mediacapture-main in the sense that all tracks from a source (in this case a RTCRtpReceiver) are created equal and the source is stopped when all associated tracks are.

Issue #619, opened 2016-05-04
setDirection()


aboba:
@pthatcherg had said (at the end of the thread relating to PR https://github.com/w3c/webrtc-pc/pull/466):

"I thought about this some more, and I think that if we have {direction: "sendrecv"} in the construction of an RtpTransceiver, we probably ought to have .setDirection instead of .activateSender. That would be easier to specify and would allow the application to change direction in any way (sender/receiver active/inactive).

martinthomson:
`setActiveDirection(s)` perhaps?  After all a transceiver is omnidirectional naturally.

pthatcherg:
Then we'd want to call it {activeDirection: "sendrecv"} in addTransceiver
as well.  I'd prefer just calling it "direction" and "setDirection".
Unless we can come up with a better name than "direction".

alvestrand:
Fixed by #620

PR #620, opened 2016-05-04
setDirection()

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/619

(not ready to merge yet, still under discussion).

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4147,7 +4147,7 @@ sender.setParameters(params)
             <p>
             	If the <code>RTCRtpTransceiver.sender</code> or <code>RTCRtpTransceiver.receiver</code>
             	is not already active and the PeerConnection is in the stable state, this will set the
-            	<code>negotiationneeded</code> flag.
+            	negotiation-needed flag.
             </p>
           </dd>
 
--- a/webrtc.html
+++ b/webrtc.html
@@ -4132,31 +4132,22 @@ sender.setParameters(params)
             the <a><code>RTCRtpReceiver</code></a> to be stopped.</p>
           </dd>
 
-          <dt>void activateSender()</dt>
+          <dt>void setDirection(RTCRtpTransceiverDirection direction)</dt>
 
           <dd>
-            <p>The <dfn id="dom-rtptransceiver-activatesender"><code>activateSender</code></dfn>
-            method activates the <code>RTCRtpTransceiver.sender</code> such that future calls
-            to <code>createOffer</code> and <code>createAnswer</code> mark the corresponding
-            media description as <code>sendrecv</code> or <code>sendonly</code>, as defined in
-            <span data-jsep="subsequent-offers subsequent-answers">[[!JSEP]]</span>.
+            <p>
+            	The <dfn id="dom-rtptransceiver-setdirection"><code>setDirection</code></dfn>
+            	method sets the direction of the <code>RTCRtpTransceiver.sender</code> and
+            	<code>RTCRtpTransceiver.receiver</code> such that future calls
+            	to <code>createOffer</code> and <code>createAnswer</code> mark the corresponding
+            	media description as <code>sendrecv</code>, <code>sendonly</code>, <code>recvonly</code>
+            	or <code>inactive</code> as defined in
+            	<span data-jsep="subsequent-offers subsequent-answers">[[!JSEP]]</span>.
+            </p>
 
-            <p>If the sender is not already active and the PeerConnection is in the stable state,
-            this will set the <code>negotiationneeded</code> flag.
+            <p>If the PeerConnection is in the stable state, this will set the <code>negotiationneeded</code> flag.</p>
           </dd>
-          
-          <dt>void activateReceiver()</dt>
-
-          <dd>
-            <p>The <dfn id="dom-rtptransceiver-activatereceiver"><code>activateReceiver</code></dfn>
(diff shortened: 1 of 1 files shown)


aboba:
@pthatcherg Can you review?

webrtc.html
```diff
@@ -4132,30 +4132,23 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-            to <code>createOffer</code> and <code>createAnswer</code> mark the corresponding
-            media description as <code>sendrecv</code> or <code>recvonly</code>, as defined in
-            <span data-jsep="subsequent-offers subsequent-answers">[[!JSEP]]</span>.
-
-            <p>If the receiver is not already active and the PeerConnection is in the stable state,
-            this will set the <code>negotiationneeded</code> flag.
+            <p>
+            	The <dfn id="dom-rtptransceiver-setdirection"><code>setDirection</code></dfn>
+            	method sets the direction of the <code>RTCRtpTransceiver.sender</code> and
+            	<code>RTCRtpTransceiver.receiver</code> such that future calls
```
pthatcherg: I think it sets the direction of the transceiver, not the sender and receiver, which changes whether the sender or receiver are active.

webrtc.html
```diff
@@ -4132,30 +4132,23 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            	to <code>createOffer</code> and <code>createAnswer</code> mark the corresponding
+            	media description as <code>sendrecv</code>, <code>sendonly</code>, <code>recvonly</code>
+            	or <code>inactive</code> as defined in
+            	<span data-jsep="subsequent-offers subsequent-answers">[[!JSEP]]</span>.
+            </p>
+            <p>
+            	If the <code>RTCRtpTransceiver.sender</code> or <code>RTCRtpTransceiver.receiver</code>
+            	is not already active and the PeerConnection is in the stable state, this will set the
+            	negotiation-needed flag.
```
pthatcherg: I think this needs to say that the negotiation-needed flag will fire whenever the direction changes.

adam-be:
In addition to this text, perhaps we want to store the direction value in an internal slot. That way it would be easy to specify a corresponding getter which I think we want at some point.

The addTransceiver() call with a direction dictionary member should write to the same internal slot in that case.

adam-be:
If we want to both get and set the direction, should it be a writable attribute then perhaps?

webrtc.html
aboba: A direction change initiated via setDirection() will cause the negotiation-needed flag to fire in order to make sure that the peers are in sync.  However, since the spec does not characterize the RTCRtpEncodings.active attribute as read-only  an application can set sender.encodings[0].active or receiver.encodings[0].active without causing the negotiation-needed flag to fire, correct?

aboba:
@adam-be @pthatcherg  Assuming that setDirection() changes whether the sender or receiver are active, can't we use transceiver.sender.getParameters() and transceiver.receiver.getParameters() to ascertain the current direction? (Example:  direction is "sendrecv" when both transceiver.sender.getParameters.encodings[0].active and transceiver.receiver.getParameters.encodings[0].active are "true").  Or am I missing something?

webrtc.html
pthatcherg: Changing the .active of a given encoding doesn't require the negotiation-needed flag because no negotiation is needed, because no SDP is changed.  All that's happening is that a certain encoding is no longer sent.

The direction is different because it does affect the SDP.

pthatcherg:
The direction (whether a sender and receiver are active) are not tied to whether a certain encoding is active.  They are two separate things.  You can't learn one from the other.

By the way, I think setDirection() is better than having an attribute that sets the direction, for the same reason we have setParameters instead of an attribute that sets the parameters:  it gives us more flexibility to fail, to add more parameters in the future, to have a return value (or a promise), etc.

adam-be:
So there should be no way to retrieve the value? I think the direction is already set apart from the parameters when it has it's own setter function.

aboba:
@adam-be @pthatcherg 
If the "active" property of an RtpReceiver/RtpSender cannot be deduced from getParameters.encodings[].active, then it would seem that there needs to be a way to retrieve the value.  Also, can the "active" property only be changed via setDirection() or do other methods (e.g. transceiver.stop()) also affect it?

pthatcherg:
RtpSender/RtpReceiver do not have an "active" property, and the direction
of the transceiver cannot be deduced by RtpParameters.  Therefore, I agree
it would make sense to have a read-only property for direction.

I don't think other methods can affect the direction.  But
RtpTransceiver.stop is interesting, because a stopped transceiver
effectively has no direction at all.  But it might be useful to leave the
direction on changed when stopped, similar to how we leave properties on
PeerConnection unchanged when it is closed:  so you can see the values it
had before being closed/stopped.

adam-be:
+1 for leaving the direction as is when the transceiver is stopped.

Regarding 'setDirection() + readonly attribute direction' vs. 'attribute direction', I think we have already limited ourselves to 'simple setter behavior' for direction when we sometimes initialize it with a dictionary member (as below):
```
let transceiver = pc.addTransceiver("video", { direction: 'sendrecv' });
log(transceiver.direction); // needs to have a value here
```

aboba:
@adam-be @pthatcherg Have revised the PR based on this thread. Is it better now?

PR #624, opened 2016-05-06
Upscale allowed

### LINKED ISSUE #305: Describe what happens when media changes
Sometimes, a change will cause the media to go outside of the parameters that were negotiated.  We could stop sending at that point, but that would be bad.
1. Scale up or down to meet resolution constraints.
2. Letterbox or pillar to comply with aspect ratio constraints.
3. Reduce bitrate to comply with maxfs or bitrate constraints.
4. Resample to comply with sample rate constraints.

Much of this is already done in practice, but never actually specified.

fluffy:
Fixes #305

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4964,7 +4964,6 @@ sender.setParameters(params)
             cease and an error generated since since the negotiated frame rate
             could not be satisfied. </p>
           </dd>
-            </dd>
           </dl>
         </section>
       </div>
--- a/webrtc.html
+++ b/webrtc.html
@@ -4894,7 +4894,7 @@ sender.setParameters(params)
              RTCRtcpParameters                         rtcp;
              sequence&lt;RTCRtpCodecParameters&gt;           codecs;
              RTCDegradationPreference                  degradationPreference = "balanced";
-             boolean                                   upscaleAllowed = false;
+             RTCRtpScalingPolicyType                   scalingPolicy = "down-only";
 };</pre>
         <section>
           <h2>Dictionary <a class="idlType">RTCRtpParameters</a> Members</h2>
@@ -4947,26 +4947,56 @@ sender.setParameters(params)
               <code>degradationPreference</code> indicates which is
               preferred.</p>
             </dd>
-            <dt><dfn><code>upscaleAllowed</code></dfn> of type
-            <span class="idlMemberType"><a>boolean</a></span>,
-            defaulting to <code>false</code></dt>
+            <dt><dfn><code>scalingPolicy</code></dfn> of type
+            <span class="idlMemberType"><a>RTCRtpScalingPolicyType</a></span>,
+            defaulting to <code>"down-only"</code></dt>
            <dd>
-             <p>When, and only when, this value is set true the browser MAY
-            upscale video temporally or spatially, or upscale the audio sampling
-            rate, or upscale the number of bits in the audio sample that this
-            <code><a>RtpSender</a></code> sends over RTP.
-            Things like image size can be negotiated to be at lease a minimum
-            size.  However, the sources camera may not be capable of producing
-            an image of that size.
-            If this value is set to true, the browser MAY upscale the 
-            video to produce a larger image. If the value was set to false, or the
-            browser did not implement up scaling, then sending RTP would need to
-            cease and an error generated since since the negotiated frame rate
-            could not be satisfied. </p>
+	     <p> Controls certain aspects of how the browser
+	     can manipulate the media from
+	     a camera or microphone that is sent over RTP by this
+	     <code><a>RtpSender</a></code>.
+	     The aspects this controls are: </p>
+	     <ul>
+	       <li>
+	       <p>changing the width and height of the video frames, </p>
+	       </li>
+	       <li>
+	       <p>changing the same rate of audio, </p>
+	       </li>
+	       <li>
+	       <p>changing the same size of audio. </p>
+	       </li>
+	     </ul>
           </dd>
           </dl>
         </section>
       </div>
+      <div>
+      </div>
+  <pre class="idl">enum RTCRtpScalingPolicyType {
+      "down-only",
+      "up-and-down"
+};</pre>
+        <table data-link-for="RTCRtpScalingPolicyType" data-dfn-for=
+          "RTCRtpScalingPolicyType" class="simple">
+            <tbody>
+              <tr>
+                <th colspan="2">Enumeration description</th>
+              </tr>
+              <tr>
+                <td><dfn><code>"down-only"</code></dfn></td>
+                <td> The browser MUST NOT increase the video width, 
+		  video height, audio sample rate, or the audio sample
+		  size. It MAY decrease any of theses.  </td>
+              </tr>
+              <tr>
+		<td><dfn><code> "up-and-down"</code></dfn></td>
+		<td> The browser MAY change the the video image size,
+		  audio sample rate, and audio sample sizes.  </td>
+		</tr>
+            </tbody>
+          </table>
+
       <div>
         <pre class="idl">dictionary RTCRtpEncodingParameters {
              unsigned long       ssrc;
--- a/webrtc.html
+++ b/webrtc.html
@@ -4951,10 +4951,19 @@ sender.setParameters(params)
             <span class="idlMemberType"><a>boolean</a></span>,
             defaulting to <code>false</code></dt>
            <dd>
-            <p>When, and only when, this value is set true the browser MAY
+             <p>When, and only when, this value is set true the browser MAY
             upscale video temporally or spatially, or upscale the audio sampling
             rate, or upscale the number of bits in the audio sample that this
-            <code><a>RtpSender</a></code> sends over RTP.  </p>
+            <code><a>RtpSender</a></code> sends over RTP. Typically things like
+            frame rate would be negotiated to be inside an envelope such as 0 to
+            30 frames per second. However, consider the case where the envole
+            has been negotiated to be from 25 to 30 fps, and the sources camera
+            changed from producing 30 fps to 10 fps due to perhaps lighting
+            changes. If this value was set to true, the browser MAY interpolate
+            frames to produce 30 fps. If the value was set to false, or the
+            browser did not implement up scaling, then sending RTP would need to
+            cease and an error generated since since the negotiated frame rate
+            could not be satisfied. </p>
           </dd>
             </dd>
           </dl>
--- a/webrtc.html
+++ b/webrtc.html
@@ -4894,6 +4894,7 @@ sender.setParameters(params)
              RTCRtcpParameters                         rtcp;
              sequence&lt;RTCRtpCodecParameters&gt;           codecs;
              RTCDegradationPreference                  degradationPreference = "balanced";
+             boolean                                   upscaleAllowed = false;
 };</pre>
         <section>
           <h2>Dictionary <a class="idlType">RTCRtpParameters</a> Members</h2>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3714,6 +3714,26 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            <code><a>RtpSender</a></code> sends over RTP. Typically things like
+            frame rate would be negotiated to be inside an envelope such as 0 to
+            30 frames per second. However, consider the case where the envole
+            has been negotiated to be from 25 to 30 fps, and the sources camera
+            changed from producing 30 fps to 10 fps due to perhaps lighting
+            changes. If this value was set to true, the browser MAY interpolate
+            frames to produce 30 fps. If the value was set to false, or the
+            browser did not implement up scaling, then sending RTP would need to
+            cease and an error generated since since the negotiated frame rate
+            could not be satisfied. </p>
```
fluffy: What error would be generated ?

stefhak:
Is this really needed? For the frameRate example:
- if you can accept it going outside the preferred range do not use a mandatory frameRate constraint on the track feeding to PeerConnection
- if you can't accept it going outside the preferred range use a mandatory frameRate constraint on the track feeding to PeerConnection, and get an "overconstrained" error
And I can't see any mentioning of frameRate being part of the negotiation in JSEP.

aboba:
@stefhak 
Don't understand the framerate statement either.  framerate is not negotiated in JSEP because attributes like max-fr only provide the maximum framerate that the receiver can handle - and it isn't clear whether implementations will even provide max-fr (making it tricky for applications to determine what maxFramerate can be set to).

alvestrand:
Awaiting responses. May need list discussion.

fluffy:
The "negotiation" comment was confusing. What I mean was if the JS sets a mandatory constraint of min frame rate of 25 fps, and that camera drops below that rate, it need to stop sending and cause an error. I agree folks most apps will not want to set a min frame rate.

stefhak:
"Negotiation" comes from the proposed text itself:
```Typically things like frame rate would be negotiated to be ...```
But if you are only talking about a JS setting a constraint on the track itself this seems unneeded. If it is mandatory to stay above X, set a mandatory constraint and the track would stop delivering data (and fire an overconstrained event). If not mandatory, do not set a mandatory constraint and things will work.

**Comment by fluffy** [comment]
We agreed that the browser can not create fake data. What we are doing here is inserting an option to remove that constraint and allow the the browser to create fake data. I am confused on what the chairs would like to see clarified on this

**Comment by stefhak** [comment]
I think both Bernard and I pointed out that the example (frame rate) given in the PR does not make sense since there is no "negotiated window" for it.
I think we first need an example of where this would actually be useful.

**Comment by fluffy** [comment]
So is what you want just to remove the part that says 

Typically things like
+            frame rate would be negotiated to be inside an envelope such as 0 to
+            30 frames per second. However, consider the case where the envole
+            has been negotiated to be from 25 to 30 fps, and the sources camera
+            changed from producing 30 fps to 10 fps due to perhaps lighting
+            changes. If this value was set to true, the browser MAY interpolate
+            frames to produce 30 fps. If the value was set to false, or the
+            browser did not implement up scaling, then sending RTP would need to
+            cease and an error generated since since the negotiated frame rate
+            could not be satisfied.

**Comment by alvestrand** [comment]
The thing with frame rate is that in SDP (as far as I know) there is no way to specify a minimum frame rate; in constraints, there is. But constraints aren't negotiated, they're set; the term "negotiate" is only applied to the SDP offer/answer model in this document.

And this is a property on the RTP sender, which isn't in the right place in the model to be affected by constraints - constraints on a track are conceptually applied before the data gets to the RTP sender.

So it's not clear to me (and I assume not to Stefan or Bernard either) what situation it's possible to get into where the RTP sender needs to upscale the framerate. We're confused.

With width and height, it's possible (using imageattr, if JSEP's description of how one handles incoming imageattr values) to end up in a situation where width and height can vary within a range. If it's possible to restate your proposal using that example, it would be easier to ask the WG to have an opinion on whether it's a feature we need or not.

**Comment by fluffy** [comment]
Ok - the example has been changed from the one causing confusion

**Comment by stefhak** [comment]
In my view, the relation between the local media and what is sent over RTP is already fluffy enough to allow upscale in this situation. There is no text in the spec AFAIK that says the UA can't upscale.

If we were spec out more precisely what is allowed to go out in RTP given track constraints and sender settings there could be a need, but for now I think all the spec says is

```The encoding and transmission of each MediaStreamTrack SHOULD be made such that its characteristics (width, height and frameRate for video tracks; volume, sampleSize, sampleRate and channelCount for audio tracks) are to a reasonable degree retained by the track created on the remote side. There are situations when this does not apply, there may for example be resource constraints at either endpoint or in the network or there may be RTCRtpSender settings applied that instruct the implementation to act differently.```

**Comment by jan-ivar** [comment]
Thus this PR should be renamed "forbid upscaling unless specifically asked for".

**Comment by fluffy** [comment]
@stefhak - no we have consensus that the browsers can't manufacture fake data. This was an attempt to relax that. If that consensus is not reflected in the spec that needs to be fixed separately.

**Comment by jan-ivar** [comment]
@fluffy wasn't that consensus made in mediacapture wrt cameras and constraints? Would that implicitly apply to peer connection as well?

**Comment by stefhak** [comment]
@fluffy my faint memory matches @jan-ivar's, i.e. that discussion and consensus was for mediacapture-main, and that is also reflected quite well when it is said that the ```overconstrained``` event must fire and the track must be muted if the UA ```is not able to configure the source within the limitations established by the intersection of imposed constraints.```

But as said it is a faint memory. Perhaps you could to this PR add some wording that says that the UA is not allowed (unless ```upscale allowed``` is set) to transmit media if it can't be made within 'the intersection of track constraints, sender settings and negotiated session' (I'm guessing these are the 'knobs' that are relevant) or something like that, and we can see if people are OK with that?

**Comment by fluffy** [comment]
JSEP https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-15#section-3.6.2 has 

  If the original resolution is less than the size limits in the
   attribute, upscaling is needed, but this may not be appropriate in
   all cases.  To address this concern, the application can set an
   upscaling policy for each sent track.  For this case, if upscaling is
   permitted by policy, the sender SHOULD apply upscaling in order to
   provide the desired resolution.  Otherwise, the sender MUST NOT apply
   upscaling.  The sender SHOULD NOT upscale in other cases, even if the
   policy permits it. 

That been there for a long time and came out of WG meeting. What we are trying to do here is allow a way to set upscale policy in the API

**Comment by stefhak** [comment]
I was not aware of that section in JSEP. This is something we should talk about at TPAC. @fluffy will you prepare a slide or two?

File: webrtc.html [hunk 2]
```diff
@@ -4894,6 +4894,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
              RTCRtcpParameters                         rtcp;
              sequence&lt;RTCRtpCodecParameters&gt;           codecs;
              RTCDegradationPreference                  degradationPreference = "balanced";
+             boolean                                   upscaleAllowed = false;
```
stefhak: If we decide to add a flag for upscaling I think it should be named something like "spatialUpscaleAllowed" or so since the only upscale where any document (in this case JSEP) has text forbidding it as default is for video width and height.

webrtc.html
```diff
@@ -4946,6 +4947,23 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            upscale video temporally or spatially, or upscale the audio sampling
+            rate, or upscale the number of bits in the audio sample that this
+            <code><a>RtpSender</a></code> sends over RTP.
+            Things like image size can be negotiated to be at lease a minimum
+            size.  However, the sources camera may not be capable of producing
+            an image of that size.
+            If this value is set to true, the browser MAY upscale the 
+            video to produce a larger image. If the value was set to false, or the
+            browser did not implement up scaling, then sending RTP would need to
+            cease and an error generated since since the negotiated frame rate
```
stefhak: we do not have any error for this(?). It feels like this is a similar situation to circuit breakers stopping the flow.

aboba:
If we had RTCRtpEncodingParameters.scaleFrameRateDownBy, we could have a discussion of whether such an attribute could have a value < 1.0 (implying upscaling).  But presently we don't have such an attribute, just maxFramerate.

webrtc.html
pthatcherg: The meeting notes reflect that we decided to have the default be true.  Also, there's the whole "don't default to a falsy value" thing, so even we wanted to have not allowed be the default, we'd need to make it an enum.  Perhaps we should just make it an enum now, with a more clear name:

boolean resolutionScalingPolicy  = "up-and-down"  // Other option: "down-only"

webrtc.html
```diff
@@ -4946,6 +4947,23 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
               preferred.</p>
             </dd>
+            <dt><dfn><code>upscaleAllowed</code></dfn> of type
+            <span class="idlMemberType"><a>boolean</a></span>,
+            defaulting to <code>false</code></dt>
+           <dd>
+             <p>When, and only when, this value is set true the browser MAY
+            upscale video temporally or spatially, or upscale the audio sampling
+            rate, or upscale the number of bits in the audio sample that this
+            <code><a>RtpSender</a></code> sends over RTP.
```
pthatcherg: I thought the only consensus we had in the WG was for resolution (spatial) and for video.  I don't recall any discussion about audio or temporal upscale.     

And even if we do have controls for those things, I  think they should be separate controls, not one big control for 3 different things.

fluffy: Definitely included audio same rate upscale because we had a bunch of bugs with FF doing that breaking our ultrasonic audio pairing stuff we do.  But I don't think we talked about one control or more. What are you thinking with are use cases for separate policy on this?

webrtc.html
fluffy: Hmm - can you point me at the notes on this because I had in my notes default false but perhaps I wrote that down wrong.  Making it an enum seem like reasonable or switching it to upscaleNotAllowed = true

webrtc.html
pthatcherg: I don't remember that conversation about audio sample rate.  Can you point me to minutes about it? Maybe I just zoned out during that time.

If we do have consensus to add that as well, I think something like this would be more clear:

ResolutionScalingPolicy resolutionScalingPolicy = "down-only";  // or "up-and-down"
SampleRatePolicy sampleRatePolicy = "change-allowed";  // "maintain-rate"

fluffy:
Notes from TPAC meetings. 

Need to change to be an enum. 

Need to make explicitly that this applies to exactly video size, audio same size, audio sample rate. 

(also note because this on a Sender, it would end up being set independently for audio and video)

webrtc.html
```diff
@@ -4946,10 +4947,57 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+	     can manipulate the media from
+	     a camera or microphone that is sent over RTP by this
+	     <code><a>RtpSender</a></code>.
+	     The aspects this controls are: </p>
+	     <ul>
+	       <li>
+	       <p>changing the width and height of the video frames, </p>
+	       </li>
+	       <li>
+	       <p>changing the same rate of audio, </p>
```
stefhak: s/same/sample/ ?

webrtc.html
```diff
@@ -4946,10 +4947,57 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+	       </li>
+	       <li>
+	       <p>changing the same size of audio. </p>
```
stefhak: s/same/sample/ ?

Further: I did not find any SDP attribute mentioned in JSEP about audio sample size, so it seems there would be no reason for the UA to ever upscale that in any way.

webrtc.html
```diff
@@ -4946,10 +4947,57 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+          "RTCRtpScalingPolicyType" class="simple">
+            <tbody>
+              <tr>
+                <th colspan="2">Enumeration description</th>
+              </tr>
+              <tr>
+                <td><dfn><code>"down-only"</code></dfn></td>
+                <td> The browser MUST NOT increase the video width, 
+		  video height, audio sample rate, or the audio sample
+		  size. It MAY decrease any of theses.  </td>
```
stefhak: Would we need to mention somewhere that other things, like frameRate, are _never_ allowed to upscale?

alvestrand:
@fluffy can we discuss this in Korea? Is there relevant stuff in JSEP?

alvestrand:
@fluffy this is waiting for a response to the September comments.

PR #625, opened 2016-05-06
unifiy DTMF time with rtcweb WG

fluffy:
See email thread on rtcweb list with subject "Confirmation of DTMF issue"


stefhak:
LGTM.

PR #631, opened 2016-05-11
Clarification on 'apt' parameter

aboba:
Fix for Issue https://github.com/openpeer/ortc/issues/539 (filed against ORTC, but also applies to WebRTC 1.0).


taylor-b:
This sounds right to me. I'm assuming no browser would have a reason to support re-transmitting only a subset of the supported codecs.

Issue #634, opened 2016-05-12
RTCRtpSender.replaceTrack: 'old track' may be null so we can't check its kind


adam-be:
> 5. If withTrack.kind differs from the sender.track.kind, return a promise rejected with a TypeError.

PR #636, opened 2016-05-12
replaceTrack: Use 'transceiver kind' instead of track.kind (track may be null) (issue: #634)

adam-be:
Fixes issue: #634

First commit it shared with PR #635


elagerway:
lgtm

PR #637, opened 2016-05-12
mimeType clarification

aboba:
Fix for Issue https://github.com/openpeer/ortc/issues/529 (Opened against ORTC, but applies to WebRTC 1.0 as well).

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3887,7 +3887,7 @@ sender.setParameters(params)
           <dt>DOMString mimeType</dt>
 
           <dd>
-            <p>The codec MIME type. Valid types are listed in [[IANA-RTP-2]].</p>
+            <p>The codec MIME subtype. Valid types are listed in [[IANA-RTP-2]].</p>
           </dd>
 
           <dt>unsigned long clockRate</dt>
@@ -3929,7 +3929,7 @@ sender.setParameters(params)
         <dl class="idl" title="dictionary RTCRtpCodecCapability">
           <dt>DOMString mimeType</dt>
           <dd>
-            <p>The codec MIME type. Valid types are listed in [[IANA-RTP-2]].
+            <p>The codec MIME subtype. Valid types are listed in [[IANA-RTP-2]].
             </p>
           </dd>
         </dl>
--- a/webrtc.html
+++ b/webrtc.html
@@ -3929,8 +3929,7 @@ sender.setParameters(params)
         <dl class="idl" title="dictionary RTCRtpCodecCapability">
           <dt>DOMString mimeType</dt>
           <dd>
-            <p>The codec MIME subtype. Valid types are listed in [[IANA-RTP-2]].
-            </p>
+            <p>The codec MIME subtype. Valid types are listed in [[IANA-RTP-2]].</p>
           </dd>
         </dl>
 


elagerway:
lgtm

alvestrand:
I protest. I think the use of the subtype without the top level type was a grave architectural mistake that we should not repeat.
I suggest that we add a definition of "mime type" that says explicitly it's type/subtype, and link to that from here.

aboba:
One might argue that calling getCapabilites(kind) where kind is the MIME media type enables the application developer to retrieve both the MIME media type and subtype.  However, there may be some holes in that argument because not every MIME media type is a valid value of "kind":
https://lists.w3.org/Archives/Public/public-ortc/2016May/0027.html

taylor-b:
Certainly, if the set of supported MIME types is different than the set of supported "kind"s, this variable should be "type/subtype".

But is there any reason why we can't assume kind == MIME type? RFC 4566 section 8.2.1 says:

> where possible the same name should be registered for SDP as for MIME

With "audio", "video", "text", "application", "message", and "image", it seems like this is the case so far, and likely will continue to be. Though I guess you could make the philosophical point that it's *possible* for them to diverge in the future.

alvestrand:
The top level MIME type matches one-to-one with "Kind".
This discussion is about the *subtype*. If it's represented at all, it needs (in my opinion) to be represented as type/subtype, *not* as subtype by itself.

ibc:
While I don't disagree with having  `codec.name` (ORTC) and `codec.mimeType` (WebRTC) as the full MIME type ("type/subtype") I would like to understand what the problem is if just the "subtype" is used.

Note that codec names are always used in a context in which `kind` has already been defined (`getCapabilities()` requires `kind` as argument and both `RtpSender` and `RtpReceiver` have a `MediaStreamTrack` which also implies a specific `kind`).

alvestrand:
On 05/16/2016 08:47 PM, Iñaki Baz Castillo wrote:
>
> While I don't disagree with having |codec.name| (ORTC) and
> |codec.mimeType| (WebRTC) as the full MIME type ("type/subtype") I
> would like to understand what the problem is if just the "subtype" is
> used.
>
> Note that codec names are always used in a context in which |kind| has
> already been defined (|getCapabilities()| requires |kind| as argument
> and both |RtpSender| and |RtpReceiver| have a |MediaStreamTrack| which
> also implies a specific |kind|).
>

If an RTPSender is not attached to a MediaStreamTrack, it needs to have
its "kind" linked from elsewhere.

That nit aside .... assuming that any given type "will only be used in a
context in which <top level type> has already been defined" has proved
to be a Bad Idea in the past (if SDP hadn't done this particular
mistake, we could have skipped all of BUNDLE). We shouldn't repeat it.

> —
> You are receiving this because you were assigned.
> <https://github.com/w3c/webrtc-pc/pull/637#issuecomment-219511107>
>


-- 
Surveillance is pervasive. Go Dark.

ibc:
> If an RTPSender is not attached to a MediaStreamTrack, it needs to have its "kind" linked from elsewhere.

AFAIK in ORTC there is no such corner case (a `RTPSender` is always attached to a `MediaStreamTrack`). That's why I expected the same for WebRTC.


> That nit aside .... assuming that any given type "will only be used in a context in which <top level type> has already been defined" has proved to be a Bad Idea in the past (if SDP hadn't done this particular
mistake, we could have skipped all of BUNDLE). We shouldn't repeat it.

Good argument.

ibc:
@aboba I don't think 0428e57123b4a2c1a31211e29a7f522f8422f580 fixes this issue. Such a commit defines `codec.mimeType` as the IANA MIME **subtype**, but @alvestrand is providing solid rationale in favor of full `type/subtype` syntax.

Shouldn't this be reopened?

PR #641, opened 2016-05-13
Add reference to FILEAPI for definition of Blob

dontcallmedom:
thanks @tidoust for noticing the missing normative reference


stefhak:
LGTM

PR #642, opened 2016-05-13
Make last arg (a dictionary) of addTransceiver optional

### LINKED ISSUE #639: addTransceiver: Dictionary argument should be optional (no required members)
https://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface-extensions

dontcallmedom:
Per http://heycam.github.io/webidl/#dfn-optional-argument
> If the type of an argument is a dictionary type or a union type that has a dictionary type as one of its flattened member types, and that dictionary type and its ancestors have no required members, and the argument is either the final argument or is followed only by optional arguments, then the argument MUST be specified as optional

close #639


stefhak:
LGTM

Issue #622, opened 2016-05-05
RTCRtpTransceiverInit: receive parameters


mparis:
Hello,
I see that a transceiver can be configured with parameters for sending [1].

I think that having parameters for receiving is also interesting. Wiht these parameters we can configure for example:

- Max recv bandwidth (exposed in the SDP as "b=AS:X")
- Congestion Control options
- JitterBuffer configuration like minSize, maxSize, etc.
- etc.

There are some comments in discuss-webrtc asking for a way of configure this kind of things.
What do you think about this idea?

Thanks!!

Refs
[1] http://w3c.github.io/webrtc-pc/#widl-RTCRtpTransceiverInit-sendEncodings

aboba:
RTCTransceiverInit dictionary includes sendEncodings because the RTCRtpEncodingParameter dictionary is largely for use by RtpSender objects, not RtpReceiver objects.  Based on what is in Section 5.2.4, here is a table of which attributes are relevant to senders and receivers: 


Attribute	                                          Type	                      Receiver/Sender              Read/Write
-------------------------                 -------------------------                -------------------------            ------------------------
ssrc	                                       unsigned long	                      Sender                             Read-only
fec	                                       RTCRtpFecParameters	      Receiver/Sender              Read-only
rtx	                                       RTCRtpRtxParameters	      Receiver/Sender              Read-only
active	                               boolean	                              Sender                             Read/Write
priority	                               RTCPriorityType	              Sender                             Read/Write
maxBitrate	                       unsigned long	                      Sender                             Read/Write
maxFramerate	                       unsigned long	                      Sender                             Read/Write
scaleResolutionDownBy	       double	                              Sender                             Read/Write
rid                                           DOMString	                      Sender                             Read/Write

alvestrand:
The transceiver doesn't seem to be the right surface to control receiver options.
If there are specific controls that need exposing for a particular usage, please propose them as separate issues, and we'll discuss what the most appropriate placement is.

mparis:
@alvestrand, if it not this is not the right place, why does RTCRtpTransceiverInit have sendEncodings params?
To keep the API symetric, send and recv params should be in the same place, don't they?

pthatcherg:
[...]

PR #629, opened 2016-05-11
Table of RTCRtpEncodingParameters for RtpSender/RtpReceiver

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/622
Partial fix for Issue https://github.com/w3c/webrtc-pc/issues/609

This PR is a work-in-progress (needs review).  Please do not merge yet.

Issue #609, opened 2016-04-29
Split RTCRtpEncodingParameters in RTCRtpEncodingParameters and RTCRtpDecodingParameters 


aboba:
@murillo128

This issue was opened in the ORTC CG, but since it affects WebRTC 1.0, it has been closed there and opened here: 
https://github.com/openpeer/ortc/issues/440

aboba:
PR 629 above clarifies which attributes are used by senders and receivers and whether they are read or read/write.  So it is not a direct response to this issue but hopefully it makes things more clear.

alvestrand:
The chairs don't see any support for changing the IDL as proposed. The table added in #629 seems sufficient.

PR #646, opened 2016-05-16
Table of RTCRtpEncodingParameters for RtpSender/RtpReceiver

aboba:
Related to Issue https://github.com/w3c/webrtc-pc/issues/622
Partial fix for Issue https://github.com/w3c/webrtc-pc/issues/609

Rebase of PR https://github.com/w3c/webrtc-pc/pull/629

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4836,7 +4836,7 @@ sender.setParameters(params)
                             <td>
                                 <code>unsigned long</code>
                             </td>
-                            <td>Sender</td>
+                            <td>Receiver/Sender</td>
                             <td>Read-only</td>
                         </tr>
                         <tr id="def-fec*">
@@ -4900,7 +4900,7 @@ sender.setParameters(params)
                             <td>
                                 <code>DOMString</code>
                             </td>
-                            <td>Sender</td>
+                            <td>Receiver/Sender</td>
                             <td>Read-only</td>
                         </tr>
                     </tbody>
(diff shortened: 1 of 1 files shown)


mparis:
Hello @aboba,
from my point of view, the issue #622 is not fixed with this PR.
There, I meant how we can specify some parameters interesting for receiving like:

-     Max recv bandwidth (exposed in the SDP as "b=AS:X")
-     Congestion Control options
-     JitterBuffer configuration like minSize, maxSize, etc.
-     etc.

In the table I do not see them, even the maxBitrate is only for RtpSender.

Reading discuss-webrtc we can see that they are demanded, and as implementer I have the need of exposing them to allow the developer to configure the internal behaviour.

Best!!

aboba:
@mparis Agree that this PR only clarifies what the currently specification does and does not support.

stefhak:
Adding a table such as this is helpful IMO (I've not carefully reviewed its content though but I trust you @aboba !)

mparis:
@aboba, if it is related with the **current** spec I agree with you,
but we shouldn't say that this fixes the issue #622, should we?

aboba:
@mparis Sure - Changed description to say "Related to Issue https://github.com/w3c/webrtc-pc/issues/622"

pthatcherg:
I like the table, but why not have .ssrc and .rid be in the Receiver parameters as well?

aboba:
@pthatcherg Added .ssrc and .rid to the Receiver parameters.

Issue #548, opened 2016-03-16
RTX/RED/FEC handling


aboba:
Currently, we have: 

partial dictionary RTCRtpParameters {
             sequence<RTCRtpCodecParameters>           codecs;
};

dictionary RTCRtpCodecParameters {
             unsigned short payloadType;
             DOMString      mimeType;
             unsigned long  clockRate;
             unsigned short channels = 1;
             DOMString      sdpFmtpLine;
};

My assumption is that RTX (or RED or FEC, for that matter) are included in codecs[], and therefore that fmtp parameters such as "apt" and "rtxtime" will be included within sdpFmtpLine.  It may be useful to make that statement explicit.

ibc:
For a future re-design (and as already commented in several issues/threads):

* RTX is per media codec so it should be a property of a codec (which means that RTX is not a codec).
* FEC is per stream and can carry any other PT defined within the SSRC stream it's protecting to. Therefore it may make sense that FEC is a codec.
* AFAIU, same for RED.

robin-raymond:
To further explain the trouble:
RTX codecs have an "apt" value which cross links the source codec to the RTX codec. This forces a requirement that not only an RTX codec exist, and that a custom "parameter" property exist on the RTX codec to define the RTX "apt" linkage and rtxTime. The RTX's properties (like clock rate, and channels) match exactly in the ftmp line.

The simpler solution would be to have an RTX payload type as part of the original codec. Otherwise you get things like:
- what happens if two RTX "apt" codecs point to the same media codec?
- what happens if RTX clock rate/channel do not match the pointed apt codec?
- what happens if RTX codec is missing but encodings says to use RTX?
- capabilities return a list of codecs, but then require a RTX capability PER codec to define the PT for the RTX per codec with the correct fmtp

This adding a sub "rtx" dictionary in `RTCRtpCodecParameters` with:
````
partial dictionary RTCRtpCodecParameters {
   ...
   RTCRtpRtxCodecParameters rtx;
   ...
}

dictionary RTCRtpRtxCodecParameters {
   unsigned short payloadType;
   unsigned long rtxTime;
}
````

I also disagree with @ibc FEC / RED that FEC must be defined codecs any more than RTX must be a codec. The same philosophy can exist as RTX. The difference would be that the payloadType set in the the FEC codec could match other codecs because FEC can be shared, whereas with RTX they cannot.

[...]

PR #647, opened 2016-05-16
Clarification on RTX in Codec Capabilities/Parameters

aboba:
Fix for Issue https://github.com/openpeer/ortc/issues/539 (filed against ORTC, but also applies to WebRTC 1.0).

Rebase of PR https://github.com/w3c/webrtc-pc/pull/631

This PR could be invalidated by a future proposal on handling of RTX/RED/FEC - please do not merge until Issue https://github.com/w3c/webrtc-pc/issues/548 is resolved.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4714,11 +4714,11 @@ sender.setParameters(params)
             <dd>
             <p>
               A sequence containing the media codecs that an <code><a>RTCRtpSender</a></code>
-              will choose from, as well RTX, RED and FEC entries. Corresponding to each media
-              codec where retransmission via rtx is enabled, there will be an entry in
-              <code>codecs[]</code> with a <code>mimeType</code> value indicating
-              retransmission via rtx, and an <code>sdpFmtpLine</code> attribute (providing
-              the "apt" and "rtx-time" parameters).
+              will choose from, as well as entries for RTX, RED and FEC mechanisms.
+              Corresponding to each media codec where retransmission via rtx is enabled,
+              there will be an entry in <code>codecs[]</code> with a <code>mimeType</code>
+              value indicating retransmission via rtx, and an <code>sdpFmtpLine</code>
+              attribute (providing the "apt" and "rtx-time" parameters).
             </p>
             </dd>
             <dt><dfn><code>degradationPreference</code></dfn> of type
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4712,9 +4712,13 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
             "idlMemberType">sequence&lt;<a>RTCRtpCodecParameters</a>&gt;</span></dt>
             <dd>
-              <p>A sequence containing the codecs that an
-              <a><code>RTCRtpSender</code></a> will choose from in order to
-              send media.</p>
+            <p>
+              A sequence containing the codecs (including RTX/RED/FEC) that
+              an <a><code>RTCRtpSender</code></a> will choose from in order
+              to send media. There will be an "rtx" entry in <code>codecs[]</code>
+              for each media codec that can be retransmitted, each with their own
```
alvestrand: English nit: A codec isn't retransmitted. "for each media codec where retransmission via rtx can be used"?
It's still not clear - if VP8 and H264 are supported, each with RTX, are there 4 elements, what are the names of the last two, and how do we tell them apart (if we need to)? An example may be useful.

alvestrand:
I think we didn't reach a conclusion on this at the interim. Next interim topic (again)?

webrtc.html
```diff
@@ -4712,9 +4712,15 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-              <a><code>RTCRtpSender</code></a> will choose from in order to
-              send media.</p>
+            <p>
+              A sequence containing the codecs (including RTX/RED/FEC) that
+              an <code><a>RTCRtpSender</a></code> will choose from in order
+              to send media. Corresponding to each media codec where retransmission
+              via rtx is enabled, there will be an entry in <code>codecs[]</code>
+              with a <var>mimeType</var> value indicating retransmission via rtx,
+              but with a distinct <var>sdpFmtpLine</var> attribute (providing an
+              "apt" parameter).
```
pthatcherg: Instead of relying on the APT in the sdpFmtpLine, why don't we just have an associatedPayloadType field on the RTCRtpCodecParameters?  That would make it much easier for applications to read it.

PR #648, opened 2016-05-16
mimeType clarification

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/650

Rebase of https://github.com/w3c/webrtc-pc/pull/637

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4970,8 +4970,8 @@ sender.setParameters(params)
             <dt><dfn><code>mimeType</code></dfn> of type <span class=
             "idlMemberType"><a>DOMString</a></span></dt>
             <dd>
-              <p>The codec MIME subtype. Valid subtypes are listed in
-              [[IANA-RTP-2]].</p>
+              <p>The codec MIME media time/subtype. Valid media types
+              and subtypes are listed in [[IANA-RTP-2]].</p>
             </dd>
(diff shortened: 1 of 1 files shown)


pthatcherg:
I like the idea of making the value just the subtype, but it's then silly to have the name of the field "mimeType", isn't it?  Should we call it "subType".  Or better yet, "name".  "Name" is what everyone says in real life.  No one says "the codec sub type".  We always say "the codec name".

pthatcherg:
Given that we realized that the media type doesn't always match the kind, I withdraw my suggestion about having the "name".

webrtc.html
```diff
@@ -4970,8 +4970,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             <dt><dfn><code>mimeType</code></dfn> of type <span class=
             "idlMemberType"><a>DOMString</a></span></dt>
             <dd>
-              <p>The codec MIME subtype. Valid subtypes are listed in
-              [[IANA-RTP-2]].</p>
+              <p>The codec MIME media time/subtype. Valid media types
```
alvestrand: Oops. Spelling corrector? time -> type

webrtc.html
```diff
@@ -5027,8 +5027,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
```
alvestrand: And here.

PR #662, opened 2016-05-19
Effect of RTCRtpReceiver.track.stop()

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/597

Rebase of https://github.com/w3c/webrtc-pc/pull/617

@adam-be Can you review?


pthatcherg:
I agree that the negotiated state of the RtpTransceiver and RtpReceiver should not be affected by stopping a track.

However, I don't think we need to specify that the decoder must stay active.  An implementation could choose to stop using CPU or other resources if it knows that nothing will use the decoded frames.  I think we could just leave that part about the decoder being active out of the spec.

stefhak:
LGTM now.

PR #668, opened 2016-05-24
ICE Transport State Diagram

aboba:
Ice Transport State Diagram. 

Related to Issue https://github.com/w3c/webrtc-pc/issues/457


aboba:
@taylor-b Can you review?

taylor-b:
Sorry for the delay in reviewing. Here are the only issues I noticed:

1. There shouldn't be a transition from `failed` to `checking`. Aside from an ICE restart, there's no way to get out of `failed`; `failed` means the transport has finished checking and there are no more ICE candidates on the way.
2. For the same reason as above, there shouldn't be a transition from `completed` to `connected`.
3. Technically it's possible to go directly from `checking` to `completed`, if there's only one candidate pair. Though since the diagram is non-normative, it's probably ok to leave this out if it makes the diagram cleaner.

Also, just to confirm: I assume that the reason there are transitions from `completed`->`failed` and `connected`->`checking` without passing through `disconnected` is that the current candidate pair may suddenly become unusable, from something *other* than consent expiring. In which case there would be no time for "liveness checks to fail". Is that correct?

aboba:
@taylor-b  An authenticated STUN response with error code Forbidden (403) indicates that consent has been revoked.  That could cause transitions from completed to failed or from connected to checking without first going through disconnected (due to failure of "liveness checks" prior to a consent failure).

aboba:
Is this closer to being right? 

webrtc-icetransport-transitions

taylor-b:
Looks perfect to me now, thanks.

Issue #644, opened 2016-05-13
Fob on RTCRtpEncodingParameters to turn on and off sending CN / DTX 


fluffy:
In the case where SDP negotiated CN allowed int he envelope, folks still want a FOB to turn on and off sending the RTP for it. (Relates to JSEP issue 202 and 233 )

alvestrand:
Hm. Doesn't look like an appropriate RTPSender surface. More like an encoding parameter, or even an audio track thingy. 

But sure, people do ask for it.

alvestrand:
Seems like multiple people want this. Encoding parameter "feels" right.
Proposal needed.

aboba:
In the meeting today, it was pointed out that it is possible to turn CN on/off by adding/removing the CN codec from parameters.codecs[] and calling setParameters().  However, this does not work for DTX, because in WebRTC 1.0 (unlike in ORTC) there is no useDtx parameter than be set for a codec such as Opus. 

So it occurs to me that an alternative way of addressing this need would be to add such an Opus parameter.

juberti:
Having the CN codec and the FMTP param just means that DTX is possible, not that it is active. Same for the use of FEC (i.e., with negotiation of FEC 'codecs').

I think we want RtpEncodingParameters.vad or some such, similar to RtpEncodingParameters.fec.

alvestrand:
Closed by #734

PR #675, opened 2016-05-26
RTCRtpEncodingParameters attribute to turn on/off sending CN / DTX

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/644

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4868,7 +4868,6 @@ sender.setParameters(params)
         </table>
       </div>
       <div>
-      <div>
         <pre class="idl">enum RTCDegradationPreference {
     "maintain-framerate",
     "maintain-resolution",
--- a/webrtc.html
+++ b/webrtc.html
@@ -4778,12 +4778,12 @@ sender.setParameters(params)
               <p>
                 For an <code><a>RTCRtpSender</a></code>, indicates whether
                 discontinuous transmission will be used. Setting it to 
-                <code>disabled</code> causes discontinuous transmission to
-                be turned off. Setting it to <code>enabled</code> causes
-                discontinuous transmission to be turned on if it was
-                negotiated (either via a codec-specific parameter or via
-                negotiation of the CN codec). This attribute is ignored by
-                a video sender.
+                <code>disabled</code> causes discontinuous transmission
+                to be turned off. Setting it to <code>enabled</code>
+                causes discontinuous transmission to be turned on if
+                it was negotiated (either via a codec-specific parameter
+                or via negotiation of the CN codec). This attribute is
+                ignored by a video sender.
               </p>
             </dd>
             <dt><dfn><code>active</code></dfn> of type <span class=
--- a/webrtc.html
+++ b/webrtc.html
@@ -4846,6 +4846,7 @@ sender.setParameters(params)
         "disable",
         "enableIfNegotiated"
         };</pre>
+        <section>
         <table data-link-for="RTCVadStatus" data-dfn-for="RTCVadStatus" class="simple">
           <tbody>
             <tr>
@@ -4865,6 +4866,7 @@ sender.setParameters(params)
             </tr>
           </tbody>
         </table>
+      </section>
       </div>
       <div>
         <pre class="idl">enum RTCDegradationPreference {
@@ -4872,6 +4874,7 @@ sender.setParameters(params)
     "maintain-resolution",
     "balanced"
 };</pre>
+    <section>
         <table data-link-for="RTCDegradationPreference" data-dfn-for=
         "RTCDegradationPreference" class="simple">
           <tbody>
@@ -4898,6 +4901,7 @@ sender.setParameters(params)
             </tr>
           </tbody>
         </table>
+    </section>
       </div>
       <div>
         <pre class="idl">dictionary RTCRtpRtxParameters {
--- a/webrtc.html
+++ b/webrtc.html
@@ -4776,11 +4776,11 @@ sender.setParameters(params)
             "idlMemberType"><a>boolean</a></span></dt>
             <dd>
               <p>
-                For an <code><a>RTCRtpSender</a></code>, indicates that voice
-                activity detection, if negotiated, is actively being used.
+                For an <code><a>RTCRtpSender</a></code>, indicates whether voice
+                activity detection (if negotiated), will be used.
                 Setting it to false causes voice activity detection to be
                 turned off. Setting it to true causes voice activity detection 
-                to be turned on (if negotiated). For an <code><a>RTCRtpReceiver</a></code>,
+                to be turned on. For an <code><a>RTCRtpReceiver</a></code>,
                 this attribute is unset.
               </p>
             </dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -4843,11 +4843,10 @@ sender.setParameters(params)
       </div>
       <div>
         <pre class="idl">enum RTCVadStatus {
-    "disable",
-    "enableIfNegotiated"
-};</pre>
-        <table data-link-for="RTCVadStatus" data-dfn-for=
-        "RTCVadStatus" class="simple">
+        "disable",
+        "enableIfNegotiated"
+        };</pre>
+        <table data-link-for="RTCVadStatus" data-dfn-for="RTCVadStatus" class="simple">
           <tbody>
             <tr>
               <th colspan="2">Enumeration description</th>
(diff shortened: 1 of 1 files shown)


fluffy:
Like this but think it might need a few more things. We need to say how it interacts wth policy set, I think it needs to say it just overrides any policy. It also needs to be clear that if the SDP did not negotiate the ability to send CN/ DTX, then it will not be sent regardless of how this is set.

pthatcherg:
Would a name of ".comfortNoise" make sense with an enum value of active/inactive make sense?

I agree it should go on the encoding parameters.

stefhak:
I think we need to sort out what should happen if CN/DTX has *not* been negotiated, and this flag is set to true, or if CN/DTX has been negotiated but is negotiated away later (with this flag true). Will we report errors?

pthatcherg:
Well, perhaps that's why an enum might be good.  Then we could have "disable" and "enableIfNegotiated" or something like that.

stefhak:
@pthatcherg I agree, that would simplify.

aboba:
@pthatcherg Like the enum idea.

aboba:
@pthatcherg In ORTC, DTX is controlled by setting a codec parameter (e.g. in the case of Opus, parameters.useDtx).  Since CN usage can be controlled by adding/removing the CN codec via  setParameters(), should we consider a similar parameter for Opus in WebRTC 1.0, rather than an encoding parameter?

aboba:
@stefhak @dontcallmedom Are there any good tutorials on the new respec webidl formatting?  Clearly I do not understand this well enough.

webrtc.html
```diff
@@ -4830,11 +4842,39 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
         </section>
       </div>
       <div>
+        <pre class="idl">enum RTCVadStatus {
+        "disable",
+        "enableIfNegotiated"
+        };</pre>
```
dontcallmedom: you use "disable" and "enable" here, but "disabled" and "enabled" in the table below (the latter sounds better)

dontcallmedom:
The reference documentation on the new WebIDL is at https://www.w3.org/respec/examples/webidl-contiguous.html

Note that you're using it right, the problem here is a consistency between the IDL declaration and the names used in the description.

fluffy:
the PR  LGTM but wanted to comment on name comfortNoise ... there are actually two things going on here. 1) is when no one is speaking, do you send the background noise or suppress it out and send nothing - that is Voice Activity Detection. 2) the other things is when the receiver receives the evidence of suppressed packets - so it effectively it has received a digital zero - does it it generate a synthetic background noise to insert into the audio it is playing. That is comfort noise.  When phones first switched to digital signals, the background noise was so low compared to analog lines that people thought the call had disconnected to "comfort noise" was added to the call to simulate the noise on analog phone lines. For high def audio, applications are slowly migrating away from comfort noise but continuing to do voice activity detection. I think the thing we want to control here is voice activity detection not comfort noise.

pthatcherg:
@fluffy How about .discontinuousTransmission instead of .comfortNoise?  That more closely matches what opus calls it, and it covers all the use cases you are describing.  I'd favor that over "voiceActivityDetection", because technically it doesn't have to be "voice" that activates/deactivates it.  It could be instrumental music or something else as well.  

@aboba There are two problems with using codec parameters in WebRTC 1.0:  1.  So far we have said codecs can't be changed, only reordered or removed, so we'd have to deal with that.  2.  It's awkward to say "for codec X, frob A, for codec Y, frob B".  It would be nice to have one frob for turning this on and off.

pthatcherg:
I thought about this a little more based on Cullen's comment.  Would it make sense to have 3 enum values and rename it this?

enum RTCTransmissionContinuity {
  "continuous",
  "discontinuous"
  // Discontinuous if CN not possible
  "discontinuousWithComfortNoise"
};

dictionary RTCRtpEncodingParameters {
  ...
  RTCTransmissionContinuity transmissionContinuity = "continuous";
}

aboba:
@pthatcherg Like the idea of consistency between codecs.  I understand settings to turn on DTX or turn off DTX. However I am unclear about the interaction with existing SDP negotiated CN. For some codecs (like Opus) couldn't you turn on DTX even if CN was not negotiated? How does discontinuousWithComfortNoise work if Opus is negotiated?

stefhak:
@aboba and @pthatcherg reading https://tools.ietf.org/html/draft-ietf-payload-rtp-opus-11 I get the impression that CN is not used with Opus, but comfort noise is instead generated by the decoder if no speech frames are received (like if DTX is enabled and no speech signal is present).

aboba:
@fluffy "if the SDP did not negotiate the ability to send CN/ DTX, then it will not be sent"
Is this still true in the case of Opus? Or would the attribute still affect use of DTX in that case even if CN were not negotiated?

stefhak:
Looking at codecs like AMR, AMR-WB and EVS it is unclear on if, and how, this API should be mapped. These codecs, to the best of my understanding, roughly work something like:

- DTX (Discontinuous Transmission) is built in

- for EVS it is possible to switch DTX off, but for AMR/AMR-WB default implementations run with DTX all the time

- whenever the VAD detects "no speech" DTX kicks in, and during DTX comfort noise data is sent by so called SID (SIlence Descriptor) frames. (There are two types, SID-first and SID-update but that does not matter that much.)

- when entering DTX a SID frame is generated followed by another one almost immediately after. From that point a new SID frame is generated every 8th frame (160ms)

- the decoder end uses the SID frames to generate comfort noise

- in absence of SID frames comfort noise would be generated by the decoder using the last speech frames received (I think this is similar to Opus) - there is a hangover period used meaning that the last few speech frames before DTX usually have no speech in them - for some time, then the decoder would mute its output signal.

- Turning DTX off would mean that speech frames (carrying the background noise) would be transmitted.

Assuming the use case is to have the decoder generate silence instead of comfort noise when there is no speech input signal this means: 

* Turning DTX off should be straightforward for EVS, and could, depending on implementation support in the codec lib, be possible for AMR/AMR-WB as well (but is not by default).

** *But* even if DTX is turned off the background noise would any way be carried to the receiver in speech frames.

* Suppressing the sending of the SID frames could be possible (e.g. by not forwarding them to the RTP stack), but then comfort noise would anyway be generated at the decoder side for some time (160ms - 1s are figures I have heard) before it mutes.

* The only way that seems possible is to replace SID frames for codec homing frames or modify the SID frames such that the decoder produce silence. However, that feels like a hack to me.

aboba:
@stefhak Thank you for the research. 

Does it make sense to:

a. Have a control to turn DTX on/off on an RtpSender.  On a codec with internal DTX this would turn off the internal DTX mechanism and on a codec without internal DTX it would turn off the sending of CN (and therefore could only be enabled if CN was negotiated).  

b. Have a control to turn comfort noise decoding on/off on an RtpReceiver. On a codec with internal DTX, this would substitute silence for internal comfort noise in the decoder and on a codec without internal DTX it would turn off the decoding of CN (if CN was negotiated).

stefhak:
@aboba I think that proposal makes sense.

One question is if you need a) if you have b) (assuming default is to use DTX if possible).

I worry a bit about error handling, would this knob exist on a video RtpSender, and what would happen if it was turned?

aboba:
@stefhak   a) is about the encoder, b) is about the decoder.  If internal DTX or CN is not being generated by the encoder, then the decoder will not decode CN or internal DX, no matter what the setting is. However, if internal DTX or CN is being generated by the encoder, then the decoder could choose not to decode it and substitute silence instead.  The question is whether this is useful or not.

These controls do not make sense on a video RtpSender/Receiver so the values would presumably be ignored.

webrtc.html
```diff
@@ -4741,6 +4741,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
              unsigned long       ssrc;
              RTCRtpRtxParameters rtx;
              RTCRtpFecParameters fec;
+             RTCDtxStatus        useDtx;
```
pthatcherg: Instead of "useDtx", can we just have "dtx"?

aboba:
@juberti @pthatcherg  Is it ok now?

webrtc.html
```diff
@@ -4830,11 +4844,39 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
         </section>
       </div>
       <div>
+        <pre class="idl">enum RTCDtxStatus {
+        "disabled",
+        "enabledIfNegotiated"
```
pthatcherg: I'm fine with this, but I thought Justin's suggestion was to just make this "enabled" so it was less verbose.  I'm fine either way.

pthatcherg:
LGTM

aboba:
@pthatcherg  Took Justin's suggestion.  Changed enum to "enabled" with the same description (only enabled if negotiated).

Issue #674, opened 2016-05-26
The doc should be updated to say that transceiver.stop() causes negotiation-needed to be set


alvestrand:
See discussion in #632

adam-be:
This would answer the question asked by #585

PR #676, opened 2016-05-26
transceiver.stop() causes negotiation-needed to be set

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/674


fluffy:
LGTM - I think this matches what we agreed to on the call

PR #681, opened 2016-05-31
Fix invalid id value that started by an '#'

vivienlacourba:
This is to fix this markup issue that prevents automatic publishing (has the markup validation fails).
```
"message": "Bad value “##widl-RTCPeerConnection-getStats-Promise-RTCStatsReport--MediaStreamTrack-selector” for attribute “href” on element “a”: Illegal character in fragment: “#” is not allowed."
```
https://lists.w3.org/Archives/Public/public-tr-notifications/2016May/0089.html


sideshowbarker:

PR #682, opened 2016-06-01
update JSEP link for reuse to include subsequent answers

burnburn:
In 5.1 addtrack() step 4, update JSEP reference to also include subsequent answers.

This is a work item for issue #337.


fluffy:
LGTM

Issue #692, opened 2016-06-06
Meaning of "Liveness checks have failed" for `disconnected` ICE state is not clear


taylor-b:
This definition leaves me with a couple question:

- What is a liveness check? Is it a [STUN consent check](https://tools.ietf.org/html/draft-ietf-rtcweb-stun-consent-freshness-16)?
- "Checks" is plural. So how many checks must fail before entering `disconnected`? Does the state return to `connected` when the next check succeeds?

I looked through the meeting minutes to try to find the answers, but [the minutes from when it was introduced](https://www.w3.org/2012/09/17-webrtc-minutes.html) barely reference it.

In my opinion it's important the the definition is unambiguous, because some developers use this state for ICE restart logic and even updating the call UI. So it's important that its behavior is consistent across browsers.

aboba:
@taylor-b Elsewhere in the description of the disconnected state, a reference is made to consent failure, so presumably "liveness checks have failed" does not represent consent failure (which cannot "resolve itself without action").

Does it really make sense to enter the disconnected state based on the failure of a few consent checks prior to consent failure? 30 seconds was chosen for the consent failure timer since this makes it unlikely that the failure would be due to a routing transient. Could we just delete the first two sentences in the disconnected state description?

taylor-b:
Well, suppose "liveness checks have failed" means simply "the last consent check failed". This would be very different than consent entirely failing.

I didn't consider that deleting the first two sentences was an option, since until recently these were the *only* sentences that defined the state. And as I mentioned earlier, developers use this for ICE restart logic. Other than using `getStats` to track binding request failures, there's no alternative I can see. That might actually be a reasonable alternative, if we can provide a shim for existing apps.

Also, for reference: Chrome currently enters "disconnected" if no STUN responses are received in the last 5 seconds. Chrome also sends them quite often; around twice a second.

aboba:
@stefhak Given that the PR has been merged, can we close this issue?

taylor-b:
I think so.

PR #695, opened 2016-06-08
Meaning of "Liveness checks have failed"

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/692

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5934,10 +5934,7 @@ sender.setParameters(params)
             </tr>
             <tr>
               <td><dfn><code>disconnected</code></dfn></td>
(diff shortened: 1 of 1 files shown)


stefhak:
@taylor-b do you have any time to look at this?

taylor-b:
Could this be a topic for the next interim? I feel there should be some discussion before removing this behavior, because it's used by developers currently to tell if a connection is temporarily disrupted, and also sometimes for ICE restart logic. If the only alternative is to call getStats periodically and detect when binding requests start failing, that's a bit of a burden compared to just listening to a state change event.

PR #701, opened 2016-06-12
addTrack: mention that MSID of the new track is added

mparis:


adam-be:
This is not a problem with this PR, but I think it's a bit unfortunate that there are pieces here and there about how createOffer/Answer should work. I think it's fine to have notes for clarity but we should try to define the behavior of functions at one place. That makes it much easier to implement them.

mparis:
+1

stefhak:
LGTM, but I'd like someone else to review (in general I find it problematic that we talk about MSID in this spec, ideally all that should be specced in JSEP IMO)

Issue #369, opened 2015-10-30
addTrack's streams parameter is unused.


jan-ivar:
We must either specify how this parameter is to be used, or remove it.

jan-ivar:
This applies to addTransceiver as well btw. It would be nice if the overall intent of this parameter was described. Also:

Is the parameter optional?
Should the API fail if a passed-in track is not in all or any of the passed-in streams?

I think the answers are yes and no respectively, because users can do this:

    var track = stream.getTracks()[0];
    pc.addTrack(track, stream);
    stream.removeTrack(track);

or even this:

    var transceiver = pc.addTransceiver ("video", { streams: [stream] });
    transceiver.sender.replaceTrack(track).catch(e => console.error(e));

where `replaceTrack` is defined to explicitly ignore stream associations.

So I think I'm good, but this could be a lot clearer.

jan-ivar:
> Is the parameter optional?

Sorry, this was resolved in https://github.com/w3c/webrtc-pc/issues/288 (it is), but the remaining question stands (validation), and I hope we agree that processing models should explain to implementers exactly what to do with input parameters.

alvestrand:
Copying over comment from #426: The expected result is that there are MSIDs for those streams listed in the SDP. This will probably take a JSEP reference (not sure where).

adam-be:
I agree with @jan-ivar that this processing model needs to be detailed out.

The addTrack() steps don't say much about how the track should be used either. Just that it's set on the sender object. This section needs some work. We should be able to define some steps that can be shared between addTrack() and addTransceiver().

adam-be:
Proposed fix: PR #702

PR #702, opened 2016-06-14
Define how to create RTCRtpSender/Receiver/Transceiver and let addTrack use them

adam-be:
Fixes: #369

This change adds steps to:
* create an RTCRtpSender with a track and streams
* create an RTCRtpReceiver with a kind (and id)
* create an RTCRtpTransceiver with a sender and a receiver

addTrack() uses these steps to create a new RTCRtpTransceiver (with a sender) if no existing sender could be matched. It also adds some text about configuring a matched sender (transceiver). The steps above can easily be extended to support the remote media case as well, for example by adding an mid "argument" to the 'create an RTCRtpTransceiver' steps.

Rendered preview: http://rawgit.com/adam-be/webrtc-pc/create-rtp-stuff/webrtc.html#dom-rtcpeerconnection-addtrack

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5231,22 +5231,22 @@ sender.setParameters(params)
           object [[!GETUSERMEDIA]].</p>
         </li>
         <li>
-          <p>Set <var>track.kind</var> to <var>kind</var>.
+          <p>Initialize <var>track.kind</var> to <var>kind</var>.
         </li>
         <li>
-          <p>If an id, <var>id</var>, was given as input to this algorithm, set
-          <var>track.id</var> to <var>id</var>. (Otherwise the value generated
-          when <var>track</var> was created will be used.)</p>
+          <p>If an id, <var>id</var>, was given as input to this algorithm,
+          initialize <var>track.id</var> to <var>id</var>. (Otherwise the value
+          generated when <var>track</var> was created will be used.)</p>
         </li>
         <li>
-          <p>Set <var>track.label</var> to the result of concatenating
+          <p>Initialize <var>track.label</var> to the result of concatenating
           the string <code>"remote "</code> with <var>kind</var>.
         </li>
         <li>
-          <p>Set <var>track.readyState</var> to <code>live</code>.</p>
+          <p>Initialize <var>track.readyState</var> to <code>live</code>.</p>
         </li>
         <li>
-          <p>Set <var>track.muted</var> to <code>true</code>. See the
+          <p>initialize <var>track.muted</var> to <code>true</code>. See the
           <a><code>MediaStreamTrack</code></a> section about how the
           <code>muted</code> attribute reflects if a <code>
           <a>MediaStreamTrack</a></code> is receiving media data or not.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -3840,6 +3840,22 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               <ol>
                 <li>
                   <p>Let <var>connection</var> be the
+                  <code><a>RTCPeerConnection</a></code> object on which this
+                  method was invoked.</p>
+                </li>
+                <li>
+                  <p>Let <var>track</var> be the <code>
+                  <a>MediaStreamTrack</a></code> object indicated by the
+                  method's first argument.</p>
+                </li>
+                <li>
+                  <p>Let <var>streams</var> be a list of <code>
+                  <a>MediaStream</a></code> objects constructed from the
+                  method's remaining arguments; if the method was called with a
+                  single argument the result is an empty list.</p>
+                </li>
+                <li>
+                  <p>Let <var>connection</var> be the
                   <code><a>RTCPeerConnection</a></code> object on which the
                   <code><a>MediaStreamTrack</a></code> <var>track</var>, is to
                   be added.</p>
@@ -3885,19 +3901,48 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                   </ul>
                 </li>
                 <li>
-                  <p>If <var>sender</var> was left unset by the previous step,
-                  create a new <code><a>RTCRtpTransceiver</a></code> with
-                  <code>sender.track</code> set to <var>track</var>, add it to
-                  <var>connection</var>'s set of transceivers.</p>
-                  <p>Let <var>sender</var> be the
-                  <code><a>RTCRtpSender</a></code> object associated with the
-                  newly created <code><a>RTCRtpTransceiver</a></code>.</p>
-                  <p>The initial value of an
-                  <code><a>RTCRtpTransceiver</a></code>'s <code><a data-for=
-                  "RTCRtpTransceiver">mid</a></code> attribute is null. Setting
-                  a new <code><a>RTCSessionDescription</a></code> may change it
-                  to a non-null value, as defined in <span data-jsep=
-                  "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>.</p>
+                  <p>If <var>sender</var> was set by the previous step, run the
+                  following steps to use that sender:</p>
+                  <ol>
+                    <li>
+                      <p>Set <var>sender.track</var> to <var>track</var>.</p>
+                    </li>
+                    <li>
+                      <p>Set <var>sender</var>'s
+                      [[<a>associated MediaStreams</a>]] to
+                      <var>streams</var>.</p>
+                    </li>
+                    <li>
+                      <p>Enable sending direction on the <code>
+                      <a>RTCRtpTransceiver</a></code> associated with
+                      <var>sender</var>.</p>
+                    </li>
+                  </ol>
+                </li>
+                <li>
+                  <p>If <var>sender</var> is still unset (no sender could be
+                  matched above), run the following steps:</p>
+                  <ol>
+                    <li>
+                      <p><a>Create an RTCRtpSender</a> with <var>track</var>
+                      and <var>streams</var> and let the result be
+                      <var>sender</var>.</p>
+                    </li>
+                    <li>
+                      <p><a>Create an RTCRtpReceiver</a> with
+                      <var>track.kind</var> as kind and let the result be
+                      <var>receiver</var>.</p>
+                    </li>
+                    <li>
+                      <p><a>Create an RTCRtpTransceiver</a> with
+                      <var>sender</var> and <var>receiver</var> and let the
+                      result be <var>transceiver</var>.</p>
+                    </li>
+                    <li>
+                      <p>Add <var>transceiver</var> to <var>connection</var>'s
+                      <a href="#transceivers-set">set of transceivers</a> </p>
+                    </li>
+                  </ol>
                 </li>
                 <li>
                   <p>A track could have contents that are inaccessible to the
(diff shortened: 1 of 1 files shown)


adam-be:
@jan-ivar, please review

webrtc.html
```diff
@@ -3840,6 +3840,22 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                <li>
+                  <p>Let <var>track</var> be the <code>
+                  <a>MediaStreamTrack</a></code> object indicated by the
+                  method's first argument.</p>
+                </li>
+                <li>
+                  <p>Let <var>streams</var> be a list of <code>
+                  <a>MediaStream</a></code> objects constructed from the
+                  method's remaining arguments; if the method was called with a
+                  single argument the result is an empty list.</p>
```
jan-ivar: What result? Maybe: *", or an empty list if the method was called with a single argument"* ?

webrtc.html
```diff
@@ -3840,6 +3840,22 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                </li>
+                <li>
+                  <p>Let <var>connection</var> be the
```
jan-ivar: `connection` appears twice.

webrtc.html
adam-be: Your proposal sounds better. I'll update.

webrtc.html
```diff
@@ -3885,19 +3901,48 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-                  <code><a>RTCRtpSender</a></code> object associated with the
-                  newly created <code><a>RTCRtpTransceiver</a></code>.</p>
-                  <p>The initial value of an
-                  <code><a>RTCRtpTransceiver</a></code>'s <code><a data-for=
-                  "RTCRtpTransceiver">mid</a></code> attribute is null. Setting
-                  a new <code><a>RTCSessionDescription</a></code> may change it
-                  to a non-null value, as defined in <span data-jsep=
-                  "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>.</p>
+                  <p>If <var>sender</var> was set by the previous step, run the
+                  following steps to use that sender:</p>
```
jan-ivar: `sender` is *optionally set* in the previous paragraph, which I found a bit hard to reason about. Rather than referencing previous logic, would it be simpler to bake it in, i.e. initialize it explicitly in it its own step?

  * *Let* sender *be* *`null`*.

And then here: *If* sender *is not `null`, run the following steps to use that sender:*

Also (github wont let me comment on lines directly above this patch):
* since you're in here, if you could change *"matches the following"* to *"matches all the following"*, I think it would be clearer (I tripped up on the "any" in the same sentence).
* The matching criteria references "sender", which is not `sender`, which is logically ok, but a bit confusing.

webrtc.html
```diff
@@ -3885,19 +3901,48 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                    <li>
+                      <p>Set <var>sender.track</var> to <var>track</var>.</p>
+                    </li>
+                    <li>
+                      <p>Set <var>sender</var>'s
+                      [[<a>associated MediaStreams</a>]] to
+                      <var>streams</var>.</p>
+                    </li>
+                    <li>
+                      <p>Enable sending direction on the <code>
```
jan-ivar: What does this mean?

webrtc.html
```diff
@@ -3885,19 +3901,48 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                      <a>RTCRtpTransceiver</a></code> associated with
+                      <var>sender</var>.</p>
+                    </li>
+                  </ol>
+                </li>
+                <li>
+                  <p>If <var>sender</var> is still unset (no sender could be
+                  matched above), run the following steps:</p>
```
jan-ivar: *If* sender *is `null`, run the following steps:*

webrtc.html
```diff
@@ -3885,19 +3901,48 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                  <ol>
+                    <li>
+                      <p><a>Create an RTCRtpSender</a> with <var>track</var>
+                      and <var>streams</var> and let the result be
+                      <var>sender</var>.</p>
```
jan-ivar: I'm no fan of inverse lets. I read *let x be y* as `let x = y;` not `let y = x;`

Maybe either: *"and let* sender *be the result."* or *"Let* sender *be the result of calling* Create an RTCRtpSender *..."* ?

webrtc.html
```diff
@@ -4371,6 +4416,33 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        </li>
+        <li>
+          <p>Let <var>sender</var> have an [[<dfn>associated
+          MediaStreams</dfn>]] internal slot, representing a list of <code>
+          <a>MediaStream</a></code> objects that the <code>
+          <a>MediaStreamTrack</a></code> object of this sender is associated
+          with.</p>
+        </li>
+        <li>
+          <p>Let <var>sender</var>'s [[<a>associated MediaStreams</a>]] slot be
```
jan-ivar: This should probably be *"Set* `sender`'s [[`associated MediaStreams`]] *to ..."* (you use that elsewhere).

webrtc.html
```diff
@@ -4371,6 +4416,33 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+      <a>MediaStreamTrack</a></code>, <var>track</var>, and a list of <code>
+      <a>MediaStream</a></code> objects, <var>streams</var>, run the following
+      steps:</p>
+      <ol>
+        <li>
+          <p>Let <var>sender</var> be a new <code><a>RTCRtpSender</a></code>
+          object.</p>
+        </li>
+        <li>
+          <p>Let <var>sender.track</var> be <var>track</var>.</p>
```
jan-ivar: Let or set?

webrtc.html
```diff
@@ -5152,6 +5224,45 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        <li>
+          <p>Let <var>sender</var> be a new <code><a>RTCRtpSender</a></code>
+          object.</p>
+        </li>
+        <li>
+          <p>Let <var>track</var> be a new <code><a>MediaStreamTrack</a></code>
+          object [[!GETUSERMEDIA]].</p>
+        </li>
+        <li>
+          <p>Initialize <var>track.kind</var> to <var>kind</var>.
```
jan-ivar: Yes, I think I like "Initialize" better than "set" on a newly created object, but we should be consistent.

webrtc.html
```diff
@@ -5152,6 +5224,45 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        </li>
+        <li>
+          <p>If an id, <var>id</var>, was given as input to this algorithm, let
+          <var>track.id</var> be <var>id</var>. (Otherwise the value generated
+          when <var>track</var> was created will be used.)</p>
```
jan-ivar: I guess *set* instead of *let* here? I guess not *initialize*, since we're referencing a generated value as the already-initialized value.

I'm a bit new to how we spec creation of public-facing objects without a public constructor, so I may be wrong here. Is there an established guideline for this?

jan-ivar: Update: I see http://rawgit.com/adam-be/webrtc-pc/create-rtp-stuff/webrtc.html#dispatch-receiver does this convincingly.

webrtc.html
```diff
@@ -5152,6 +5224,45 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+          <p>Initialize <var>track.readyState</var> to <code>live</code>.</p>
+        </li>
+        <li>
+          <p>Initialize <var>track.muted</var> to <code>true</code>. See the
+          <a><code>MediaStreamTrack</code></a> section about how the
+          <code>muted</code> attribute reflects if a <code>
+          <a>MediaStreamTrack</a></code> is receiving media data or not.</p>
+        </li>
+        <li>
+          <p>Let <var>sender.track</var> be <var>track</var>.</p>
```
jan-ivar: Initialize or set, I think.

webrtc.html
```diff
@@ -5347,7 +5458,28 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       <code><a>RTCRtpTransceiver</a></code> is defined by the kind of the
       associated <code><a>RTCRtpReceiver</a></code>'s
       <code><a>MediaStreamTrack</a></code> object.</p>
-      <div>
+      <p>To <dfn>create an RTCRtpTransceiver</dfn> with an <code>
+      <a>RTCRtpReceiver</a></code> object, <var>receiver</var>, an <code>
+      <a>RTCRtpSender</a></code> object, <var>sender</var>, run the following
```
jan-ivar: *",* **and** *a sender,"*

webrtc.html
```diff
@@ -5347,7 +5458,28 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+      steps:</p>
+      <ol>
+        <li>
+          <p>Let <var>transceiver</var> be a new <code>
+          <a>RTCRtpTransceiver</a></code> object.</p>
+        </li>
+        <li>
+          <p>Let <var>transceiver.sender</var> be <var>sender</var>.</p>
```
jan-ivar: Initialize

webrtc.html
```diff
@@ -5347,7 +5458,28 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        </li>
+        <li>
+          <p>Let <var>transceiver.receiver</var> be <var>receiver</var>.</p>
```
jan-ivar: Initialize

webrtc.html
```diff
@@ -5347,7 +5458,28 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        </li>
+        <li>
+          <p>Let <var>transceiver.stopped</var> be <code>false</code>.</p>
```
jan-ivar: Initialize

jan-ivar:
Lgtm with nits! Great improvement!

webrtc.html
adam-be: I'll update with the null-approach you suggest here. Perhaps we can clarify the "sender" vs ```sender``` in a separate change.

webrtc.html
adam-be: The intention is to go from recvonly to sendrecv and inactive to sendonly. Perhaps that needs to be defined (by JSEP)?

webrtc.html
adam-be: I'll update to your first option.

webrtc.html
adam-be: Perhaps set since it's an update.

webrtc.html
adam-be: I'm using "set" for everything that could be considered to have a value before to be consistent.

adam-be:
Thanks for reviewing @jan-ivar. I've tried to address all your comments (see second commit).

webrtc.html
```diff
@@ -5236,28 +5231,28 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           object [[!GETUSERMEDIA]].</p>
         </li>
         <li>
-          <p>Initialize <var>track.kind</var> to <var>kind</var>.
+          <p>Set <var>track.kind</var> to <var>kind</var>.
```
jan-ivar: Is there a default kind?

adam-be: No. So in this case it would be OK to say initialize.

webrtc.html
```diff
@@ -5236,28 +5231,28 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
         </li>
         <li>
-          <p>If an id, <var>id</var>, was given as input to this algorithm, let
-          <var>track.id</var> be <var>id</var>. (Otherwise the value generated
+          <p>If an id, <var>id</var>, was given as input to this algorithm, set
+          <var>track.id</var> to <var>id</var>. (Otherwise the value generated
```
jan-ivar: Definitely better, though since [id](https://w3c.github.io/mediacapture-main/getusermedia.html#widl-MediaStreamTrack-id) says: "Unless a *MediaStreamTrack* object is created as a part of a special purpose algorithm that specifies how the track id must be *initialized*, the User Agent MUST generate an identifier string and initialize the object's `id` attribute to that string."

I would think we could use "initialize" on all the track members here, including `id`, if we wanted.

jan-ivar:
Changes look great overall!

stefhak:
Everyone but Travis seem happy. Update and merge?

dontcallmedom:
(travis now happy again fwiw)

adam-be:
Thanks @dontcallmedom!

burnburn:
LGTM

PR #704, opened 2016-06-14
Use 'create an RTCRtpSender/Receiver/Transceiver' definitions (from #702) in addTransceiver()

adam-be:
Note that this PR is based on #702


adam-be:
This one should be easier to review now when #702 is merged and not included in the aggregated diff of this PR.

webrtc.html
```diff
@@ -4076,35 +4076,73 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-              <code>setRemoteDescription</code> is called with a corresponding
-              remote description that is able to receive multiple RTP encodings
-              as defined in [[!JSEP]], the <code><a>RTCRtpSender</a></code> may
-              send multiple RTP encodings and the parameters retrieved via the
-              transceiver&#8217;s <code>sender.getParameters()</code> will
-              reflect the encodings negotiated.</p>
-              <p>RID values passed into <code>init.sendEncodings</code> must be
-              composed only of case-sensitive alphanumeric characters (a-z,
-              A-Z, 0-9) up to a maximum of 16 characters.</p>
+              <p>When this method is invoked, the User Agent must run the
```
dontcallmedom: s/must/MUST/

webrtc.html
```diff
@@ -4076,35 +4076,73 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+              following steps:</p>
+              <ol>
+                <li>
+                  <p>If the dictionary argument is present, and it has a
+                  streams member, let <var>streams</var> be that list of <code>
```
dontcallmedom: I'd put the name of the members between `<code></code>` (here and for `sendEncodings` as well)

webrtc.html
```diff
@@ -4076,35 +4076,73 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                  <a>MediaStreamTrack</a></code>, let it be <var>track</var>
+                  and let <var>kind</var> be <var>track.kind</var>.</p>
+                </li>
+                <li>
+                  <p><a>Create an RTCRtpSender</a> with <var>track</var>,
+                  <var>streams</var> and <var>sendEncodings</var> and let
+                  <var>sender</var> be the result.</p>
+                  <p>If <var>sendEncodings</var> is set, then subsequent
+                  calls to <code>createOffer</code> will be configured to send
+                  with multiple RTP encodings as defined in <span data-jsep=
```
dontcallmedom: s/with// ?

webrtc.html
```diff
@@ -4076,35 +4076,73 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                  "subsequent-offers">[[!JSEP]]</span>. When
+                  <code>setRemoteDescription</code> is called with a
+                  corresponding remote description that is able to receive
+                  multiple RTP encodings as defined in [[!JSEP]], the <code>
+                  <a>RTCRtpSender</a></code> may send multiple RTP encodings
+                  and the parameters retrieved via the transceiver&#8217;s
```
dontcallmedom: I think we have settled on not using “fancy” apostrophes (to my dismay)

webrtc.html
```diff
@@ -4076,35 +4076,73 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                  <code>sender.getParameters()</code> will reflect the
+                  encodings negotiated.</p> <p>RID values passed into
+                  <code>init.sendEncodings</code> must be composed only of
```
dontcallmedom: I think that must is confusing; it would be better to describe what the UA should do if it is not the case (I realize that's a bug in the original text, not in this PR, but if fixing it is easy, we should probably do it; if not, I'll open an other issue)

adam-be:
Thanks for the review @dontcallmedom. I'll merge this when the bots turn green, but please file issues on the unclarities you found.

PR #716, opened 2016-06-27
Improve error handling for IdP proxy interactions

### LINKED ISSUE #562: What to do with an RTCIdentityProvider that returns rubbish
These might not return a promise.  They might not resolve a promise, they might reject.  I think that all needs to be more clearly described than it is.

martinthomson:
Closes #562.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8352,15 +8352,23 @@ function processStats() {
           provided.</p>
         </li>
         <li>
-          <p>The IdP returns a Promise to the <code>RTCPeerConnection</code>.
-          If the user has been authenticated by the IdP, and the IdP is willing
-          to generate an identity assertion, the IdP resolves the promise with
-          an identity assertion in the form of an
+          <p>The IdP proxy returns a Promise to the
+          <code>RTCPeerConnection</code> and generates an identity assertion
+          asynchronously.</p>
+          <p>If the user has been authenticated by the IdP, and the IdP is
+          willing to generate an identity assertion, the IdP resolves the
+          promise with an identity assertion in the form of an
           <code><a>RTCIdentityAssertionResult</a></code>.</p>
           <p>This step depends entirely on the IdP. The methods by which an IdP
           authenticates users or generates assertions is not specified, though
           they could involve interacting with the IdP server or other
           servers.</p>
+        </li>
+        <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -8352,17 +8352,25 @@ <h3 id="sec.identity-proxy-assertion-request">Requesting Identity
...
         <li>
-          <p>The IdP returns a Promise to the <code>RTCPeerConnection</code>.
-          If the user has been authenticated by the IdP, and the IdP is willing
-          to generate an identity assertion, the IdP resolves the promise with
-          an identity assertion in the form of an
+          <p>The IdP proxy returns a Promise to the
+          <code>RTCPeerConnection</code> and generates an identity assertion
+          asynchronously.</p>
+          <p>If the user has been authenticated by the IdP, and the IdP is
+          willing to generate an identity assertion, the IdP resolves the
```
stefhak: This part is a bit confusing. First it says ```IdP proxy ... generates an identity assertion``` and then ```if ... the IdP is willing to generate an identity assertion```. It looks like the IdP proxy has already generated that one.

webrtc.html
```diff
@@ -8466,7 +8466,13 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
...
           <p>The IdP proxy verifies the identity assertion using whatever means
           necessary. Depending on the authentication protocol this could
-          involve interacting with the IDP server.</p>
+          involve interacting with the IdP server.</p>
+        </li>
+        <li>
+          <p>If the IdP proxy produces an error or returns a promise that does
+          not resolve to a valid
+          <code><a>RTCIdentityValidationResult</a></code> (see <a href=
+          "#sec.idp-error-handling"></a>), then identity validation fails.</p>
```
stefhak: 8472 - 8475 is a copy of 8368 - 8371  - could it be moved to the error handling section?

stefhak:
As far as I can judge, this PR looks good (two minor comments above).

webrtc.html
martinthomson: Do you think that I should say "attempts to generate an identity assertion"?

stefhak:
Thanks for the update @martinthomson. To me (admitting I'm an IdP novice) this looks ready to merge now.

PR #719, opened 2016-06-27
The IdP environment can be spoofed

### LINKED ISSUE #253: Assurance that requests to IdP proxy originate from the user agent
In [section 9.1.2 Instantiating an IdP Proxy](http://w3c.github.io/webrtc-pc/#instantiating-an-idp-proxy), the spec claims that "A global property can only be set by the user agent or the IdP proxy itself. Therefore, the IdP proxy can be assured that requests it receives originate from the user agent. This ensures that an arbitrary origin is unable to instantiate an IdP proxy and impersonate this API in order obtain identity assertions".

It is true if the IdP JavaScript is loaded in the "isolated interpreted context" that the specification defines. However, what prevents an arbitrary origin from loading the IdP script in a regular window environment after having set "rtcIdentityProvider" to a fake value, thus impersonating the API?

There may be a way for the IdP proxy to validate the fact that it is running in an isolated interpreted context and/or that "rtcIdentityProvider" is indeed a platform object under the control of the user agent but I don't understand why the user agent can be the only one able to instantiate an IdP proxy. Did I miss something?

martinthomson:
This isn't a problem for validating assertions, presumably an
attacker would have an easier time asking RTCPeerConnection to
unpack an assertion if they wanted to learn the identity it
contains.

However, for generating an assertion it is important.  An IdP
therefore needs to draw on information that only it knows if it
is going to avoid being spoofed.  For any real IdP, that is
probably going to be automatic: they will look at what they have
stored (which is specific to their origin), or make requests
to servers.  Those requests to servers won't allow cross-origin
access unless something is seriously wrong.

Closes #253.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8057,11 +8057,6 @@ function processStats() {
         <code><a>RTCIdentityProviderRegistrar</a></code> named
         <var>rtcIdentityProvider</var> in the global scope of the <a>realm</a>.
         This object is used by the IdP to interact with the user agent.</p>
-        <p>A global property can only be set by the <a>user agent</a> or the
-        IdP proxy itself. Therefore, the IdP proxy can be assured that requests
-        it receives originate from the <a>user agent</a>. This ensures that an
-        arbitrary origin is unable to instantiate an IdP proxy and impersonate
-        this API in order obtain identity assertions.</p>
         <div>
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

PR #728, opened 2016-07-13
Fix mistakes in examples.

jan-ivar:
I was perusing the examples and found several mistakes, as well as some places where long lines made the code wrap around.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9153,13 +9153,13 @@ function start() {
     };
 
     // get a local stream, show it in a self-view and add it to be sent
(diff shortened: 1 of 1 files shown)


aboba:
@dontcallmedom  Why is the Travis CI build failing?  Doesn't appear to be anything wrong with this PR.

dontcallmedom:
@aboba this PR is based on a version of the tree where the travis.yml file doesn't work; it would need to be rebased to pass. That being said, I think we should feel free to ignore travis when we're confident it is wrong

Issue #566, opened 2016-04-05
Separate sender and receiver sets are unnecessary when we have a transceiver set


adam-be:
For example, getSenders() can be defined as a new sequence containing all senders in the transceiver set.

alvestrand:
Do you mean that the spec should define getSenders() in terms of transceivers, or that we should remove the getSenders() call? I'm in favor of the first, but not the last.

adam-be:
The first. I.e., we don't need to keep track of senders/receivers separately.

alvestrand:
"Add this sender to the sender set" is a hint that the spec hasn't been updated for transceivers; it can be a good exercise to clear this up.

burnburn:
After making the change you suggest to getSenders() and getReceivers() (see commit 503c9aa40e233bec07f8bbdbf515bbf4b9f1a170) I took a look at other uses of "set of senders" such as in addTrack().  I believe it may still be useful to either keep an explicit list of senders and receivers or create some internal function that generates the list on the fly from the transceivers.  Otherwise, changing the wording in addTrack, removeTrack, and close to talk about "all the senders in all of the transceivers in connection's set of transceivers" is both imprecise, wordy, and difficult to work into the text in those sections.

burnburn:
Decision on editor's call was to create internal functions for getSenders and getReceivers that we can reference in addTrack, removeTrack, etc.  I will do this.

burnburn:
PR 739 added algorithms and removed "set of senders" everywhere but not "set of receivers".  See the PR for details.

adam-be:
Remaining work will be tracked by issue #788.

Issue #563, opened 2016-03-24
Undefined 'transceiver' object used in 'dispatch a receiver' steps


adam-be:
Step 7 in https://w3c.github.io/webrtc-pc/archives/20160215/webrtc.html#dispatch-receiver

alvestrand:
@pthatcherg this is a place where the transceiver concept isn't integrated. Can you take a look?

mparis:
Hello,
I do not know if we should discuss this here or in #623.

My team and I decided to implement the next solution: 
if there is not any transceiver to manage a remote MediaStreamTrack,  an event is triggered to notify the developer if he wants to create a new transceiver to manage the incoming media.

This event is like an _asynchronous_ way of adding a new transceiver [1-addTransceiver].
For now, in our implementation the event only asks asks for a RTCRtpTransceiverInit:

- If it is NULL, the media will be rejected
- If not, it will be accepted and managed

It is working, but In next iterations we will take into account the info of RTCRtpTransceiverInit (RTCRtpTransceiverInit .receive], RTCRtpTransceiverInit .send, etc.) to reject the remote media, accept only for receiving, etc.

I hope this info can help us make a decision on this issue ;).

Best regards!!

Refs
[1-addTransceiver] https://w3c.github.io/webrtc-pc/archives/20160215/webrtc.html#widl-RTCPeerConnection-addTransceiver-RTCRtpTransceiver-MediaStreamTrack-DOMString-trackOrKind-RTCRtpTransceiverInit-init

alvestrand:
This seems like a wrong approach to me. The required owning transceivers should be created by the platform before the MediaStreamTrack is made visible to the application, and there definitely shouldn't be two ways to do it (when the transceivers and tracks are the result of setRemoteDescription, only one thing should happen).

This needs to be made clear. Tagging @pthatcherg @taylor-b .

stefhak:
I agree to Harald here. The transceivers should be created by the platform.

mparis:
Yes, in my proposal the transceiver is created by the platform (not by the developer), but it asks the developer if he wants to accept the media (so the transceiver is created) and also asks for the configuration (RTCRtpTransceiverInit) to create the transceiver, if not how knows the platform which configuration to apply?

taylor-b:
[...]

PR #739, opened 2016-08-11
issue 566: replace set of senders/receivers/transceivers with algorithms

burnburn:
I have (mostly) replaced "set of senders" and "set of receivers" with new algorithms based on the set of transceivers (issue #566), but I'm not clear on how to get access to the appropriate transceiver in addTrack, so it and the related track event have not been updated.  Interestingly, in 'dispatch a receiver' in 5.1.1, step 7, a transceiver variable is already being used but was never defined.  Clearly the magic happens in step 6, but again, not clear how to find the correct transceiver on which the new receiver should appear (unless I just defer to JSEP).

So we can apply this PR and then clean up the three remaining uses of "set of receivers" in a new PR, or someone can modify my PR.  Thoughts?

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1813,9 +1813,10 @@ interface RTCPeerConnection : EventTarget  {
                     </p>
                   </li>
                   <li>
-                    <p>
+                    <p>Let <var>senders</var> be the result of executing the
+                     <code><a>CollectSenders</a></code> algorithm.
                       For every <code><a>RTCRtpSender</a></code> <var>sender</var> in
-                      <var>connection</var>'s <a>set of senders</a>,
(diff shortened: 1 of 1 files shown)


alvestrand:
@adam-be @pthatcherg @taylor-b do you have time to look at this?

adam-be:
Regarding the alien transceiver in the 'dispatch a receiver' steps, we had a issue for that (#563) but it was closed (too quickly one might argue) with the resolution that we should look into an an 'ontransceiver' event. The creation of a transceiver initiated by setRemoteDescription should be integrated with the 'create an RTCRtpSender/Receiver/Transceiver' steps that were introduced recently, but that has not been done yet.

Issue #714, opened 2016-06-24
STUN/TURN OAuth token auth parameter passing


misi:
I think there is a confusion between the current PeerConnection W3C API and RFC7635 
In STUN/TURN auth crendtials/parameters handover...

https://tools.ietf.org/html/rfc7635#appendix-B

``` javascript
     HTTP/1.1 200 OK
        Content-Type: application/json
        Cache-Control: no-store

        {
          "access_token":
   "U2FsdGVkX18qJK/kkWmRcnfHglrVTJSpS6yU32kmHmOrfGyI3m1gQj1jRPsr0uBb
   HctuycAgsfRX7nJW2BdukGyKMXSiNGNnBzigkAofP6+Z3vkJ1Q5pWbfSRroOkWBn",
          "token_type":"pop",
          "expires_in":1800,
          "kid":"22BIjxU93h/IgwEb",
          "key":"v51N62OM65kyMvfTI08O"
          "alg":HMAC-SHA-256-128
        }
```

```
                        Figure 8: Response
```

Here below I have highlighted the three mandatory parameters that needed to pass to the ICE Agent
- "access_token":
- "kid":
- "key":

So we need to pass these 3 value at least to ICE Agent in the browser through PeerConnection iceServers configuration interface.

So according RFC 7635
- The key id (kid) (That is used like as username in Long Term Credential Auth),
  and has two other "credential" information pieces, that are needed to auth on remote TURN server. 
  -  the Session Key(Message Integrity, HMAC key). (That is used like as password in Long Term Credential Auth)
  -  the Access Token (SelfContaing token, e.g. Authenticated-Encryption with Associated-Data(AEAD) encrypted token.) In figure 8 it is the key:

See: https://w3c.github.io/webrtc-pc/#idl-def-rtciceserver

So in WebIDL I could find only one DOMString for Credential.

dictionary RTCIceServer {
    required (DOMString or sequence<DOMString>) urls;
             DOMString                          username;
             DOMString                          credential;
             RTCIceCredentialType               credentialType = "password";
};

And furthermore this credential field normally in case of "password" auth (Long Term Credential)  contains the Session Key(Message Integrity, HMAC key).

I am wondering what is the right way to pass the access token, the third value? 
How to pass the 3 information in 2 fields username/credential?
I propose to add a third field for the access_token, or add clarification in W3C PeerConnection.

The actual W3C webrtc-pc saying that the access_token need to be passed as credential
[...]

PR #740, opened 2016-08-15
STUN/TURN OAuth token auth parameter handover

misi:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/714


alvestrand:
Discussion at the Aug 16 interim did not resolve the issue, but indicated that this PR is not the preferred direction to go in. A new PR will come later.

misi:
@alvestrand I watched the last interim recording and my understanding was that no one understand OAuth+TURN. Everybody was confused => No one read through and understand the RFC. 
I mean it is complex, and no one had enough time on it. I am wondering how can you make decision on something that only one of you understand.. 
I guess no one have time to understand and dig in it, how OAuth + STUN/TURN works. 
So I started to work on AFAIU doc to present my understanding about it. May it help others too.
I will share it later..

The meeting participants trust in @juberti as the expert (who is a great expert and who I personally admire), but the point that he stated was wrong as I have pointed out in my comments on the issue 714. 

Yes you are right I wrote that I will update this PR, but after it I have changed my mind without a comment because, after counting all possibilities I still believe that this PR is the less worse solution. 
If we speak backward compatibility, minor change, etc. 

I still believe adding two or one field does not make so much confusion, then making a bigger change. The only change that I may would make, to cut back the change to one field, and not add the expiry timestamp, but only add the access_token. 

So junk dropped one plus field. Sounds Good?

I have double-checked this three info on the wire the kid=username,mac_key=password, access_token are enough for STUN/TURN client to Auth with the TURN server.

Again. 
1.I am sure that the mac_key could NOT be extracted from the access_key on client side.
2. And the mac_key if it is used as the password do not change the browsers internal workflow of credential/password handling. And it is not necessary to change if we move access_token to a new field.

Please consider to reopen this PR with the proposed small change if @juberti confirms that he changed his mind on WebRTC client side mac_key extraction from access_token.

PR #750, opened 2016-08-22
RTCIdentityProviderGlobalScope needs Exposed attributes

### LINKED ISSUE #748: Interface RTCIdentityProviderGlobalScope should be defined with [Global,Exposed]
As defined, the `RTCIdentityProviderGlobalScope` interface cannot be used as a [global object](https://heycam.github.io/webidl/#Global), which is what I think you're trying to create here, and its [exposure set](https://heycam.github.io/webidl/#dfn-exposure-set) is the primary global interface, meaning `Window` in practice and you probably do not want to expose that interface there.

The interface should rather be defined as:

```webidl
[Global, Exposed=RTCIdentityProviderGlobalScope]
interface RTCIdentityProviderGlobalScope { ... };
```

(This is similar to the way [SharedWorkerGlobalScope](https://html.spec.whatwg.org/multipage/workers.html#shared-workers-and-the-sharedworkerglobalscope-interface) is defined for instance).

Similarly, the `RTCIdentityProviderRegistrar` interface should be exposed on the `RTCIdentityProviderGlobalScope` global:

```webidl
[Exposed=RTCIdentityProviderGlobalScope]
interface RTCIdentityProviderRegistrar { ... }
```

By the way, the spec currently says: "The realm is populated with a global that implements WorkerGlobalScope" and never says that the global is actually an instance of the `RTCIdentityProviderGlobalScope` interface.

martinthomson:
Closes #748

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3834,7 +3834,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                   </ol>
                 </li>
                 <li>Let <var>senders</var> be a new sequence consisting of
-              all the <code><a>RTCRtpSender</a></code> objects in 
+              all the <code><a>RTCRtpSender</a></code> objects in
               <var>senderset</var>.
               The conversion from the senders set to the sequence
               is user agent defined and the order does not have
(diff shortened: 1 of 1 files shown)


dontcallmedom:
LGTM

aboba:
LGTM

PR #753, opened 2016-08-22
Update list of sections from JSEP

dontcallmedom:


fluffy:
LGTM - but if you want me to fix the sec- to sec. in JSEP, I can do that.

adam-be:
This works for my case so I'll go ahead and merge it.

Issue #755, opened 2016-08-23
setLocal/RemoteDescription should explicitly state which steps that are queued


adam-be:
Related to Issue #600

adam-be:
Closed by #756

PR #745, opened 2016-08-17
Add steps to createOffer and explicitly specify what is queued

adam-be:
Fixes #744 

This PR is currently based on PR #721 to make use of some changes to the generic queued operation section.

PR #721, opened 2016-06-28
Specify the synchronous and queued steps for addIceCandidate (ver 2) (issue #600)

adam-be:
I abandoned PR #601 after the move to the new WebIDL style and the reformatting of the document. Otherwise, this is the same change.

PR #756, opened 2016-08-23
Integrate queueing into the setLocal/RemoteDescription steps

adam-be:
Fixes issue #755 
Related to issue #600 

This change suggests that the entire 'set an RTCSessionDescription' algorithm should be enqueued instead of adding the queueing step to the algorithm and indenting all the steps.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -689,13 +689,9 @@
         </ol>
         <p>To <dfn id="set-description" data-lt="set the RTCSessionDescription">set an RTCSessionDescription</dfn>
         <var>description</var> on an <code><a>RTCPeerConnection</a></code>
-        object <var>connection</var>, run the following steps:</p>
+        object <var>connection</var>, <a href=
+        "#enqueue-an-operation">enqueue</a> the following steps:</p>
         <ol>
-          <li>
-            <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
-            <code>true</code>, the user agent MUST return a promise rejected
-            with an <code>InvalidStateError</code>.</p>
-          </li>
           <li>
             <p>Let <var>p</var> be a new promise.</p>
           </li>
@@ -1395,7 +1391,8 @@ interface RTCPeerConnection : EventTarget  {
                 which point the <code><a>RTCPeerConnection</a></code> can fully
                 adopt the pending local description, or rollback to the current
                 description if the remote side rejected the change.</p>
-                <p>When the method is invoked, the user agent must <a>set the
+                <p>When the method is invoked, the User Agent MUST return the
+                result of <a href="#set-description">setting the
                 RTCSessionDescription</a> indicated by the method's first
                 argument.</p>
                 <p><span data-jsep="configurable-sdp-paramete">[[!JSEP]]</span>
@@ -1435,9 +1432,11 @@ interface RTCPeerConnection : EventTarget  {
                 apply the supplied
                 <code><a>RTCSessionDescriptionInit</a></code> as the remote
                 offer or answer. This API changes the local media state.</p>
-                <p>When the method is invoked, the user agent must <a>set the
+                <p>When the method is invoked, the User Agent MUST return the
+                result of <a href="#set-description">setting the
                 RTCSessionDescription</a> indicated by the method's first
-                argument. In addition, a remote description is processed to
+                argument.</p>
+                <p>In addition, a remote description is processed to
                 determine and verify the identity of the peer.</p>
                 <p>If an <code>a=identity</code> attribute is present in the
                 session description, the browser <a href=


adam-be:
@jan-ivar, please have a look.

webrtc.html
```diff
@@ -689,14 +689,10 @@
         </ol>
         <p>To <dfn id="set-description" data-lt="set the RTCSessionDescription">set an RTCSessionDescription</dfn>
         <var>description</var> on an <code><a>RTCPeerConnection</a></code>
-        object <var>connection</var>, run the following steps:</p>
```
jan-ivar: (commenting on existing wording two lines above this one, which relates to changes later in this patch)

"RTCSessionDescription" is used as a descriptive name for the *description* input parameter here, as opposed to describing the target being set. This makes the meaning (and wording, "the" vs "an") subtly opposite of the text where we link from.

Would it be better to say "To **set the RTCSessionDescription** to a *description*" ?

webrtc.html
```diff
@@ -1395,7 +1391,8 @@
                 which point the <code><a>RTCPeerConnection</a></code> can fully
                 adopt the pending local description, or rollback to the current
                 description if the remote side rejected the change.</p>
-                <p>When the method is invoked, the user agent must <a>set the
+                <p>When the method is invoked, the User Agent MUST return the
+                result of <a href="#set-description">setting the
                 RTCSessionDescription</a> indicated by the method's first
```
jan-ivar: Hmm, I take that back. It seems to consistently refer to the input argument.

I guess I tripped over it expecting "set X" to mean: X is what's being modified, rather than X is the input argument. Shouldn't this link be calle something like "set the local description"? Are we avoiding this because of the "current" vs "pending" wording difficulty?

jan-ivar: Nm, we reuse this for remote and local. I see the problems now.

jan-ivar:
Looks correct, provided https://github.com/w3c/webrtc-pc/pull/745 makes it in with the `isClosed` check as part of **enqueuing**, so we don't lose it.

Do we want any input validation steps for `setLocalDescription` and `setRemoteDescription`? FWIW Firefox [checks a few things before enqueuing](https://hg.mozilla.org/mozilla-central/annotate/bd7645928990649c84609d3f531e803c2d41f269/dom/media/PeerConnection.js#l833) like that the *description.type* is valid and that *description.sdp* is not an empty string unless type is "rollback".

webrtc.html
jan-ivar: Nm, we reuse this for remote and local. I see the problems now.

jan-ivar:
I should have said "Do we want any *synchronous* input validation steps" (aka return an already rejected promise), like we do in mediacapture?

jan-ivar:
We [do some input validation in `addIceCandidate` too](https://dxr.mozilla.org/mozilla-central/rev/bd7645928990649c84609d3f531e803c2d41f269/dom/media/PeerConnection.js#1018). I forgot to mention that in https://github.com/w3c/webrtc-pc/pull/721.

adam-be:
> Looks correct, provided https://github.com/w3c/webrtc-pc/pull/745 makes it in with the `isClosed` check as part of **enqueuing**, so we don't lose it.

Yes. I forgot to mention that dependency. Hopefully #745 can be merged this afternoon at the editor's call.

> Do we want any input validation steps for `setLocalDescription` and `setRemoteDescription`? FWIW Firefox [checks a few things before enqueuing](https://hg.mozilla.org/mozilla-central/annotate/bd7645928990649c84609d3f531e803c2d41f269/dom/media/PeerConnection.js#l833) like that the *description.type* is valid and that *description.sdp* is not an empty string unless type is "rollback".

We can easily add and sync section, but I didn't have any content for it at the time of writing. Do we need to explicitly specify the `type` member check when we use an `RTCSessionDescriptionInit` dictionary (which uses an enum for that member). An empty `sdp` member will fail later in case the type is not rollback, so I don't know if it's needed to catch that early. There might be other things we can catch early as well, and it feels easier to just let the consumer of SDP fail later.

adam-be:
> We [do some input validation in `addIceCandidate` too](https://dxr.mozilla.org/mozilla-central/rev/bd7645928990649c84609d3f531e803c2d41f269/dom/media/PeerConnection.js#1018). I forgot to mention that in https://github.com/w3c/webrtc-pc/pull/721.

But there we have a sync section that checks input. Would you like to extend it?

jan-ivar:
Ah, you're right, the enum should suffice now that `type` is `required`. Cool, we can remove code.

As to catching an empty `sdp` member early, it's clearly not necessary, and this particular argument can be invalid in lots of more complicated ways that can only be determined asynchronously.

So I think it depends on how one feels about doing early input validation whenever possible. 
The benefit would be that developers could step over the method and instantly see that the returned promise is rejected (a now hopefully established pattern for learning that arguments to an asynchronous method were invalid), or in general, detect their mistake earlier and more precisely (especially since we do queuing, which could mean they'd otherwise see an error much later, with other subsequent actions piled up behind it, a state which may be harder to back out of).

The reason I lean toward checking it early is:
  1. that we can,
  2. that `null` is an acceptable input sometimes (rollback), but not others, and
  3. We should perhaps also check that description *is* `null` when `type` *is* rollback?
  4. Early validation it is known to make bugs easier to find.

jan-ivar:
I'll add a separate issue about adding input validation to addIceCandidate.

jan-ivar:
Nm, looks like addIceCandidate already does input validation, and we need to update our code instead.

PR #757, opened 2016-08-23
Support ufrag in IceCandidate and end of IceCandidate indicator 

### LINKED ISSUE #726: Add a ufrag attribute to the RTCIceCandidate structure to indicate which ICE context
At IETF 95, we agreed that we should have the ufrag included with ICE candidates so you could disambiguate candidates and end-of-candidates from different ICE generations (restarts). We should
1. Add ufrag to RTCIceCandidate
2. Add ufrag to one of the things that the serializer emits

We should also consider modifying the "done gathering marker" to be not RTCIceCandidate = null but rather RTCIceCandidate = { mid, null} or something.

fluffy:
Fixes #726

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -636,7 +636,10 @@
                 <p><a data-lt="update the ICE gathering state">Update
                 <var>connection</var>'s ICE gathering state</a> to
                 <code>completed</code> and let <var>newCandidate</var> be
-                null.</p>
+                <code><a>RTCIceCandidate</a></code> with the sdpMid,
+                 sdpMLineIndex, and ufrag set to the values used
+                 by this <a>ICE Agent</a>, and candiate and all
+                 other nullable members set to null.</p>
               </li>
             </ul>
           </li>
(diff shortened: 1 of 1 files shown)


fluffy:
In the current spec,  the end of candidates are indicated by a NULL, this changes this to return a map with the ufrag. It is not clear to me how to make this backwards compatible.

fluffy:
based on call today - still return the NULL at end of all candidates, but also send per mid end of candidates in a separate event.

fluffy:
Change PR to just do ufrag to ice candidate for now and worry about separate issue of what happened on end of candidates on separate bug to be filed.

alvestrand:
@fluffy are you intending to do the change indicated in the comment above, or are you waiting for someone else to do it?

alvestrand:
This was discussed at TPAC, but I don't think we ended up with a complete solution. Is this a partial solution we should merge anyway, or do we wait for a final one? @pthatcherg pinged.

stefhak:
Pinging @pthatcherg again. Any comment?

alvestrand:
Obsoleted by #968

Issue #732, opened 2016-08-04
replaceTrack with the previous one as ended


mparis:
Hello,
what should happen if we try to replaceTrack of an ended track?

If I am not wrong, [1] and [2] say that if a MediaStreamTrack is ended a renegotiation should be performed removing its info from the SDP

On the other hand, replaceTrack doc [2] says that a renegotiation must not be performed:
"Determine if negotiation is needed to transmit withTrack in place of the sender's existing track. Ignore which MediaStream the track resides in and the id attribute of the track in this determination. If negotiation is needed, then reject p with InvalidModificationError and abort these steps. "

So:
1.  if the previous track is ended should a renegotiation take place to notice the peer?
2. Or it cannot take place because the previous track must be in LIVE state?:
    "Attempts to replace the track **being sent** with another track provided, without renegotiation."

Refs
[1] http://w3c.github.io/webrtc-pc/#setting-negotiation-needed
[2] http://w3c.github.io/webrtc-pc/#clearing-negotiation-needed
[3] http://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack

alvestrand:
Good topic for next interim meeting, if we have time.

aboba:
Do we really need a negotiation when a MediaStreamTrack is ended?

jesup:
I think it should be up to the application to decide what to do in that case.  The track is still attached; we just won't get any more frames from it, and as benard said, the application could instead ReplaceTrack it  with another device, or with a canvas with "user disconnected their camera", etc - or just leave it.

adam-be:
I like @jesup's suggestion where the app can decide to send something else if a track ends for some reason. That would mean treating the track as muted until some action is taken (negotiation or replaceTrack).

One approach to make is usable would be to make sure the track "ended" event is fired before the "negotiationneeded" event. That way the app can do the replaceTrack call and have the negotiation needed flag cleared by that action (and prevent a negotiation that would stop the sender and make future replaceTrack calls throw).

alvestrand:
The critical point is whether the track on the receiving side is supposed to be ended or not. If it's ended, it can't resurrect (by the semantics of "ended") - and the information that it's ended and not merely paused has to be sent to the receiver somehow. (-msid gives 3 ways to do it - BYE, SSRC timeout and signalling).
If it's not supposed to be "ended", replacing it with a non-ended track is completely OK in terms of the model. If it's supposed to be "ended", it's not OK.

[...]

PR #758, opened 2016-08-23
replaceTrack with the previous one as ended

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/732


webrtc.html
```diff
@@ -2892,10 +2892,6 @@
         connection will be marked as needing negotiation. Examples of such
         operations include adding or stopping a track, or adding the first data
         channel.</p>
-        <p>Internal changes within the implementation can also result in the
-        connection being marked as needing negotiation. For example, if a
```
alvestrand: Suggest only removing the part after the dot (For example).

adam-be:
Do we need to add text that says that an `RTCRtpSender` sees a track as muted if it ends?

PR #776, opened 2016-08-26
Support for OAuth in TURN credentials (Issue 714 patch)

misi:
- update reference
- Just  add plus one attribute (accesstoken)
- Add an example


webrtc.html
```diff
@@ -214,8 +214,25 @@
...
                 <td><dfn><code>token</code></dfn></td>
-                <td>The credential is an access token, as described in
-                [[!TRAM-TURN-THIRD-PARTY-AUTHZ]], Section 6.2.</td>
+                <td>
+                  <p>The credential is an OAuth 2.0 Access Token.</p>
+                  <p>For authentication, at least these parameters are required
+                     , as described in [[!TRAM-TURN-THIRD-PARTY-AUTHZ]]</p>
+                  <ul>
+                    <li>
+                      The kid (the Access Token encrpytion key ID). Handover
```
alvestrand: Nit: "Handover it as" is not correct English. "This is passed in the API field called" might be.

webrtc.html
```diff
@@ -248,36 +266,66 @@
...
               </dd>
               <dt><dfn><code>credential</code></dfn> of type <span class=
               "idlMemberType"><a>DOMString</a></span></dt>
               <dd>
                 <p>If this <code><a>RTCIceServer</a></code> object represents a
-                TURN server, then this attribute specifies the credential to
-                use with that TURN server.</p>
+                TURN server, then this attribute, in case <code>
+                credentialType="password"</code> then it specifies the password
+                / credential that to use with that TURN server, or elseif in
```
alvestrand: "elseif" is not English. Sentence is hard to understand as written.

webrtc.html
```diff
@@ -248,36 +266,66 @@
...
+                case <code>credentialType="token"</code> then it is the
+                key / mac_key that to use with that TURN server. As it is
```
alvestrand: Remove word "As". It makes the sentence grammatically incorrect.

webrtc.html
```diff
@@ -248,36 +266,66 @@
...
+              <dt><dfn><code>accesstoken</code></dfn> of type <span class=
+              "idlMemberType"><a>DOMString</a></span></dt>
+              <dd>
+                <p>If this <code><a>RTCIceServer</a></code> object represents a
+                TURN server, then if <code>credentialType="token"</code> 
+                this attribute specifies the OAuth Access Token to use with 
+                that TURN server. The structure is described in
+                [[!TRAM-TURN-THIRD-PARTY-AUTHZ]] Section 6.2.
+                In case <code>credentialType="password"</code> then this field
+                is not used, and should have empty value.</p>
```
alvestrand: This is a dictionary, and the member doesn't have a "required" keyword. So the correct behavior is to omit it, not to give it the empty value.

alvestrand:
I am still worried about this solution, because it enshrines an idea that whatever new mechanism we add, we add fields to the dictionary rather than re-use the fields that are there.

So my comments are more "these things need fixing" than "this solution will be merged when those are fixed". I'd like @juberti @martinthomson @ekr to chime in on this.

misi:
@alvestrand THX for pointing out my mistakes, and for the suggested corrections. 
Let me know if theres is more mistake to correct.

> I am still worried about this solution, because it enshrines an idea that whatever new mechanism we add, we add fields to the dictionary rather than re-use the fields that are there.

There is a much cleaner another way, that avoids the new Auth type new attribute problem as I mentioned. 
Now I corrected it further and looks like this way: 

``` WebIDL
dictionary TokenCredential {
    required DOMString    kid;
    required DOMString    key;
    required DOMString    access_token;
             DOMString    alg;
             DOMTimeStamp          expiry;
};

dictionary LongTermCredential {
    required DOMString    username;
    required DOMString    password;
};

dictionary RTCIceCredential {
             required (LongTermCredential or TokenCredential) credential
             RTCIceCredentialType                             credentialType = "password";
}

dictionary RTCIceServer {
    required (DOMString or sequence<DOMString>) urls;
             sequence <RTCIceCredential>   credential;
};
```

It is more clean and future proof, and it solves if TURN server support both password and token based auth the same time (like coturn do), and pass both (token and password) to the ICE agent it may helps in case of transition phase from password to token.

But it needs bigger change in the API..
So the drawback is that it needs much bigger change in the API and implementations.

In my view it depends how often in the future a new Auth mechanism will be created to STUN/TURN, so I am not totally sure what is the the better solution to complicate the interface and make it future proof or add just one attribute. 

AFAIU webrtc-pc interface is in phase that you only merge small fix into it.
May adding one attribute is the smallest change.

Let me know, if you think I should submit an alternate PR with these dictionaries instead of the current add one attribute way.

alvestrand:
If we go down the path of adding more fields when we have more mechanisms, I think this is fine. Still waiting for @ekr @juberti @martinthomson to comment on the OAuth mechanics.

alvestrand:
At TPAC, the chairs took on the task to reach out to IETF's TRAM WG and ask for clarification.

alvestrand:
Linking to #714

misi:
PR updated.
(Dom confirmed that travis-ci check is broken.)

stefhak:
This PR was discussed at the November 9 2016 Virtual Interim, and we concluded that more work is needed on it.

aboba:
This has been obsoleted by PR https://github.com/w3c/webrtc-pc/pull/1033, correct?

misi:
Yes, this has been obsoleted, so I close it now.

PR #778, opened 2016-08-29
Remove void conformance requirement on interToneGap

### LINKED ISSUE #774: The sentence "It MUST be at least 30 ms" in the definition of the insertDTMF method seems to be content-free
Typically MUST is used for normative requirements, but I don't see how this sentence imposes a normative requirement on anyone.

The UA behavior is defined by the processing model and clamps the value to at least 30.  The caller behavior is not really something that you place a normative requirement on, in that there is no validation that happens of JS consumers of this API; you just define how it gets handled.

dontcallmedom:
No point in setting requirements for script authors
close #774


dontcallmedom:
wow, indeed, this was completely broken, not sure how I got there. Should be fixed now

Issue #764, opened 2016-08-25
Specify how an RTCRtpSender should treat an ended track


adam-be:
We have updated the replaceTrack steps; do we need to add something for the general case?

pthatcherg:
I think we could just treated muted/ended tracks the same as null tracks.

alvestrand:
Fixed in #786

PR #786, opened 2016-09-12
How an RTCRtpSender should treat an ended (or muted) track

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/764

Needs to reflect TPAC discussion (not ready to merge).

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4769,8 +4769,9 @@ interface RTCPeerConnectionIceErrorEvent : Event {
             <dd>
               <p>The <code>track</code> attribute is the track that is
               associated with this <code><a>RTCRtpSender</a></code> object.
-              If <code>track.stop()</code> is called, or if 
-              <var>track.muted</var> is set to <code>true</code>, the
+              If <code>track</code> is ended, or if
+              <code>track</code>.<var>muted</var> is set to <code>true</code>,
+              <code>track</code> is treated like a null track, and the
               <code>RTCRtpSender</code> does not send RTP.</p>
             </dd>
             <dt><dfn><code>transport</code></dfn> of type <span class=
--- a/webrtc.html
+++ b/webrtc.html
@@ -4771,8 +4771,8 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               associated with this <code><a>RTCRtpSender</a></code> object.
               If <code>track</code> is ended, or if
               <code>track</code>.<var>muted</var> is set to <code>true</code>,
-              <code>track</code> is treated like a null track, and the
-              <code>RTCRtpSender</code> does not send RTP.</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4768,7 +4768,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             nullable</dt>
             <dd>
               <p>The <code>track</code> attribute is the track that is
-              associated with this <code><a>RTCRtpSender</a></code> object.</p>
+              associated with this <code><a>RTCRtpSender</a></code> object.
+              If <code>track.stop()</code> is called, or if
```
adam-be: Perhaps we shouldn't limit ourselves to the track.stop() case here. I guess we want this behaviour regardless how the track was stopped. E.g. camera unplugged.

alvestrand: I think it should be "track.ended", not "track.stop() is called".

pthatcherg: What if the track is muted and VAD is off?  The purpose of turning VAD off is to make sure we keep sending RTP.

Perhaps instead we should say "an ended or muted is treated the same as a null track" (addTransceiver that has never had a track, for example).

alvestrand:
Discussed at TPAC. Will be updated accordingly.

webrtc.html
```diff
@@ -4768,7 +4768,11 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             nullable</dt>
             <dd>
               <p>The <code>track</code> attribute is the track that is
-              associated with this <code><a>RTCRtpSender</a></code> object.</p>
+              associated with this <code><a>RTCRtpSender</a></code> object.
+              If <code>track</code> is ended, or if
+              <code>track</code>.<var>muted</var> is set to <code>true</code>,
+              <code>track</code> is treated like a null track, and the
+              <code>RTCRtpSender</code> does not send RTP.</p>
```
mparis: In that case a negotiationneeded event is NOT triggered, isn't it?

webrtc.html
```diff
@@ -4768,7 +4768,11 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             nullable</dt>
             <dd>
               <p>The <code>track</code> attribute is the track that is
-              associated with this <code><a>RTCRtpSender</a></code> object.</p>
+              associated with this <code><a>RTCRtpSender</a></code> object.
+              If <code>track</code> is ended, or if
+              <code>track</code>.<var>muted</var> is set to <code>true</code>,
+              the <code>RTCRtpSender</code> sends silence (audio) or a black
```
mparis: I would accept sending only 1 black video frame, but what about stop sending audio directly instead of wasting bandwidth sending silence?

PR #791, opened 2016-09-14
Add text that fires the 'connectionstatechange' event

### LINKED ISSUE #554: We never fire the 'connectionstatechange' event
This event notifies listeners about RTCPeerConnectionState updates.

adam-be:
Fixes #554


stefhak:
LGTM

Issue #782, opened 2016-09-07
pc.createOffer(); pc.addTrack(track); should not include track


jan-ivar:
Now that [createOffer has procedural steps](https://rawgit.com/w3c/webrtc-pc/7f66138784d86bf3f944a7327386ca0b10e351a9/webrtc.html#dom-rtcpeerconnection-createoffer) thanks to @adam-be, we should add language to capture some state synchronously, so that in the infamous case of:
```js
    pc.addTrack(trackX, stream);
    pc.createOffer().then(offer => { ... })
    pc.addTrack(trackY, stream);
```

the `offer` does not include `trackY` in the typical (nothing on the queue) case, as we agreed over a year ago.

jesup:
And the same should apply to:
...
pc.createOffer().then(....);
pc.createDataChannel(...);
(assuming no createDataChannel before createOffer())

stefhak:
I agree. Willing to provide a PR?

alvestrand:
Does this mean that addTrack has a queued step (which doesn't return a promise, so it's unobservable when it finishes)?

stefhak:
@alvestrand I guess you are right in that it is not that easy to detect when a track added will start to result in RTP being sent (and received by the peer). In this case there would have to be another offer/answer exchange to make that happen. And observability is in stats!?

But I don't think addTrack in itself has a queued step, it's rather that createOffer samples the current tracks and if one is added after createOffer it won't be part of the produced offer.

jan-ivar:
Given that `addTrack` returns a sender, it would be hard to queue. I.e. I think one would expect `pc.getSenders()` to include this new sender immediately.

alvestrand:
So if createOffer has "gather info on all the tracks to be sent" in its synchronoous part, that works too. I guess that was what @jan-ivar said initially, I just didn't read it properly.

pthatcherg:
addTrack creates an RtpTransceiver (usually) and then returns the RtpSender for that transceiver.   So what we're talking about here is delaying when the transceiver shows up in createOffer.  Something like have a flag on each transceiver ready to go into createOffer and that doesn't happen immediately after creation.

taylor-b:
Apologies in advance if I have some fundamental misunderstanding about how PR #820 works, but to collect my thoughts on this while they're fresh:

I think the problem with this approach in general (taking a snapshot of the PC state) is that we're trying to fulfill two rules:

[...]

PR #792, opened 2016-09-15
Capture media config bundle in sync part of createOffer/Answer

adam-be:
Proposed fix for #782 

Perhaps it would be useful to include the example in the issue in the spec to highlight this as well.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1293,6 +1293,27 @@ interface RTCPeerConnection : EventTarget  {
                     method was invoked.</p>
                   </li>
                   <li>
+                    <p>Let <var>offerBundle</var> be an implementation specific
+                    collection of information that captures the changes made to
+                    the media configuration of <var>connection</var> since the
+                    last time an offer or answer was created, or since the
+                    creation of <var>connection</var> if no offer or answer has
+                    been created prior to this call. The following pieces of
+                    information MUST be captured:</p>
+                    <ul>
+                      <li>
+                        <p>If <code><a href=
+                        "#dom-rtcpeerconnection-createdatachannel">createDataChannel()</a></code>
+                        has been called for the first time.</p>
+                      </li>
+                      <li>
+                        <p>All <code><a>RTCRtpSender</a></code> objects that
+                        have been added to or removed from
+                        <var>connection</var>.</p>
+                      </li>
+                    </ul>
+                  </li>
+                  <li>
                     <p>Return the result of <a href=
                     "#enqueue-an-operation">enqueuing</a> the following
                     steps:</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1293,6 +1293,27 @@
...
+                    information MUST be captured:</p>
+                    <ul>
+                      <li>
+                        <p>If <code><a href=
+                        "#dom-rtcpeerconnection-createdatachannel">createDataChannel()</a></code>
+                        has been called for the first time.</p>
+                      </li>
+                      <li>
+                        <p>All <code><a>RTCRtpSender</a></code> objects that
+                        have been added to or removed from
```
pthatcherg: This should be in terms of RtpTransceivers, not RtpSenders.

webrtc.html
```diff
@@ -1293,6 +1293,27 @@
...
+                    collection of information that captures the changes made to
+                    the media configuration of <var>connection</var> since the
+                    last time an offer or answer was created, or since the
+                    creation of <var>connection</var> if no offer or answer has
+                    been created prior to this call. The following pieces of
+                    information MUST be captured:</p>
+                    <ul>
+                      <li>
+                        <p>If <code><a href=
+                        "#dom-rtcpeerconnection-createdatachannel">createDataChannel()</a></code>
```
pthatcherg: Perhaps this should be in terms of SctpTransport, which is like the data channel version of RtpTransceiver.

webrtc.html
```diff
@@ -1293,6 +1293,27 @@
                     method was invoked.</p>
                   </li>
                   <li>
+                    <p>Let <var>offerBundle</var> be an implementation specific
+                    collection of information that captures the changes made to
+                    the media configuration of <var>connection</var> since the
+                    last time an offer or answer was created, or since the
+                    creation of <var>connection</var> if no offer or answer has
+                    been created prior to this call. The following pieces of
+                    information MUST be captured:</p>
```
pthatcherg: This seems like a complex way of expressing this.

webrtc.html
```diff
@@ -1450,7 +1493,8 @@
                       </li>
                       <li>
                         <p>In parallel, start the process to generate an SDP
-                        answer, as described in <span data-jsep=
+                        answer that reflects changes in <var>answerBundle</var>,
+                        as described in <span data-jsep=
                         "generating-an-answer">[[!JSEP]]</span>.</p>
```
pthatcherg: This sounds like we have to rewrite JSEP in a major way to make this work.

pthatcherg: What happens with this?

setRemoteDescription(offer);
createAnswer();

Does that work?  

setRemoteDescription(offer);
createAnswer();
addTrack();

Does it affect setRemoteDescription?  Does it affect createAnswer?


I think this isn't going to work.

alvestrand:
Based on discussions at TPAC, this is not the way forward. Closing.

alvestrand:
#820 is the new attempt.

adam-be:
This PR was replaced by by #820.

PR #796, opened 2016-09-17
facingMode et al are non-nullable and cannot "return null".

jan-ivar:
Fix for language discovered in https://github.com/w3c/webrtc-pc/issues/305#issuecomment-247762455.


stefhak:
Thanks for fixing this. LGTM.

alvestrand:
LGTM. A function returns null; a dictionary member is missing.

stefhak:
BTW, it strikes me that a sentence similar to this one should probably be present in the -fromelement and -screen-share specs.

PR #804, opened 2016-09-20
Various small editorial changes.

taylor-b:
Most notably:
- Mark the RTCP mux and bundle policy enum descriptions as
  non-normative. JSEP is the source of truth here, and goes into much
  more detail.
- Stop using all-caps BUNDLE.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -123,7 +123,7 @@
               "idlMemberType">sequence&lt;<a>RTCIceServer</a>&gt;</span></dt>
               <dd>
                 <p>An array of objects describing servers available to be used
-                by ICE, such as STUN and TURN server.</p>
+                by ICE, such as STUN and TURN servers.</p>
               </dd>
               <dt><dfn><code>iceTransportPolicy</code></dfn> of type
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -336,25 +336,25 @@
           class="simple">
             <tbody>
               <tr>
-                <th colspan="2">Enumeration description</th>
+                <th colspan="2">Enumeration description (non-normative)</th>
```
alvestrand: Suggest replacing this with "Enumeration description, summarized from the normative description in (JSEP-reference)".

alvestrand:
Travis reacted to a duplicate definition. Please fix.

webrtc.html
taylor-b: I tried that, but the hyperlink ends up looking ugly on the blue background: [screenshot](http://i.imgur.com/5xbWaC9.png)

Also, the correct section of JSEP is already linked a couple lines above this, so it's a bit redundant anyway.

Any suggestions? My goal is to make it clear that the enum names are normative, but the meanings are not.

taylor-b:
Fixed.

aboba:
LGTM.

Issue #799, opened 2016-09-19
Unclear when a DTMFToneChangeEvent is fired with an empty string


dminor:
The steps for insertDTMF state "If toneBuffer is an empty string, fire an event named tonechange with an empty string at the RTCDTMFSender object and abort these steps." which seems to imply the event is only generated with an empty string when the toneBuffer is empty, so insertDTMF("AAB") would result in "A" "A" "B" "".

But the description of the 'tone' attribute of the RTCDTMFToneChangeEvent says that "The tone attribute contains the character for the tone that has just begun playout (see insertDTMF ). If the value is the empty string, it indicates that the previous tone has completed playback." That seems to imply that an empty string is generated after each tone, so that insertDTMF("AAB") would result in "A" "" "A" "" "B" "".

aboba:
@dminor Checking Section 7.2 of the latest draft: http://w3c.github.io/webrtc-pc/

Step 3 of the insertDTMF description now says now says:  "If toneBuffer is an empty string, return."

aboba:
Should the text in Section 7.3 be changed to say “the previous tones have completed playback”?"

alvestrand:
Fixed in #807

alvestrand:
This should probably say "when the buffer is empty", not "when the tones have played out".

jan-ivar:
@alvestrand when the buffer is empty AND the tones have played out, right?

alvestrand:
Fixed by #855

PR #807, opened 2016-09-22
When a DTMFToneChangeEvent is fired with an empty string

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/799


alvestrand:
Accepted at TPAC.

PR #819, opened 2016-09-22
Trickle ufrag in ICE candidate events

pthatcherg:


taylor-b:
Just to note what was discussed at TPAC: I believe the complaint against this approach was that you could get multiple "candidate = null" events, which is a backwards compatibility issue.

alvestrand:
Abandoned in favor of #968

PR #832, opened 2016-09-23
Webrtc NV Initial Work

pthatcherg:


aboba:
Should probably not be merged today :)

stefhak:
@pthatcherg I think this is not the time for NV stuff, we said we would wait for 1.0 to reach CR before. Also, when we start NV I think we will need more discussion before creating a first WG document (and additionally I expect that doc to have its own repo).

For you to maintain and develop a NV proposal in a repo of your own is of course fine.

Planning to close the PR unless we hear very good reasons for not doing it.

pthatcherg:
​I agree we shouldn't merge it now.  It isn't ready anyway (it has many
TODOs).  I just wanted to be prepared in case in came up at TPAC to have
something concrete to look at.  We didn't, and so it's not really needed
unless people are interested in seeing it between now and when we do talk
about it.

stefhak:
OK, I'm closing.

Issue #746, opened 2016-08-17
Need to specify what happens if `createDataChannel` is called with an invalid ID


taylor-b:
Currently, step 10 of `createDataChannel` instructs:

>  if the value of the id attribute is taken by an existing RTCDataChannel , throw a ResourceInUse exception and abort these steps.

However, there are situations where an id is unusable even though no existing `RTCDataChannel` with that ID exists. One unavoidable situation is when the requested ID is 65535 (the maximum value of a WebIDL `unsigned short`). This is also the maximum number of SCTP streams, which means the highest valid stream ID is 65534, since they're 0-based.

It's also possible for a WebRTC implementation to negotiate less than the maximum number of streams. Suppose 1000 streams are negotiated, and an application tries to create a data channel with stream ID 2000. When this happens, should an error be surfaced, or should the WebRTC implementation attempt to add 1001 additional streams using a RE-CONFIG message? If we prefer the latter, this behavior should probably be detailed in the rtcweb data channel spec.

adam-be:
In the first case I think we can throw an OperationError. In WebIDL there is an IndexSizeError ("The index is not in the allowed range") that we could use, but it might be safer to just go with OperationError.

The second issue should perhaps be taken to list to get the WG's view.

taylor-b:
I just listened to the discussion of this topic by the rtcweb working group at IETF 88. They definitely decided _against_ doing the RE-CONFIG, because of the added complexity. The conclusion was to just negotiate the maximum number of streams. There was some concern that current SCTP implementations allocate memory for each stream (this includes usrsctp, used by Chrome as of typing this). But people argued that implementations should be able to avoid this by using sparse data structures.

However, negotiating the maximum number of streams is still only a "SHOULD", not a "MUST", so the case still needs to be addressed.

My proposal would be to invoke the `onerror` event handler whenever this condition occurs. Why not throw an exception from `createDataChannel`? Because `createDataChannel` can be called before the PC knows how many streams will be negotiated, so the error _needs_ to occur asynchronously in that case. And it would be confusing for developers if the condition sometimes resulted in `onerror`, and sometimes resulted in a `DOMException`, depending on the timing of the SCTP handshake.

alvestrand:
The 65535 issue was addressed by #847, but "didn't negotiate them all" seems to be still open. @taylor-b , comments?

taylor-b:
I thought this paragraph would cover that case:

[...]

Issue #846, opened 2016-10-01
Not clear how data channel errors are reported, with ErrorEvent interface.


taylor-b:
The `RTCDataChannel` has an `EventHandler` of type `error`, which uses the `ErrorEvent` interface. However, this interface is used for reporting script errors associated with specific line/column numbers. So for an error that originates from background processing, it doesn't really make sense (to me). Do we just leave those attributes as the defaults?

In looking for an answer, the text that describes firing an error is somewhat lacking. It's simply:

> Fire a NetworkError event at _channel_.

Which I can extrapolate to mean:

> Fire an event named "error" which does not bubble, is not cancelable, and which uses the ErrorEvent interface with _message_ set to a user agent defined string (fingerprint warning?), with _filename_, _lineno_ and _colno_ left as the default values, and with _error_ set to a newly [created](https://heycam.github.io/webidl/#dfn-create-exception) NetworkError.

Is that what we want?

It looks like this topic also came up with mediacapture-main (https://github.com/w3c/mediacapture-main/issues/337), and the solution there was to use a simple event. But in this case we do want additional error information, or at least I assume.

If RTCDataChannels are meant to be similar to WebSockets, could we do the same thing they do, and put the additional information on the close event (in their case, a code and reason)?

gunnarhm:
I can just agree. It needs to be clear how various kinds of communication errors are reported, and how to find complementing information. 

draft-ietf-rtcweb-data-channel  gives some vague information.

Communication errors can cause the SCTP association to break so that all data channels using the association need to be closed. It would then be important to know that that had happened and which data was successfully transferred and not. I assume that this will happen e.g. when the connectivity watchdog acts on a failing association. 

If the channel is opened for reliable communication, then communication errors in the channel will cause the channel to be closed and indication on the error provided to applications. Information on how far the transmission was successful should also be provided to the application. 

If the channel is opened for partially reliable communication, then some general association errors will cause the association and channel to close, while transmission errors in the channel should cause an error indication without any channel closure. Instead, information on how far transmissions had succeeded should be available to the application.    

I assume that a lot of the above should be indicated through
https://www.w3.org/TR/webrtc/#rtcdatachannelevent
and that that specification needs to be more clear about how te different situations are signaled to te application.

taylor-b:
It sounds like this is a duplicate of issue #526 then. I'll close it.

PR #847, opened 2016-10-01
Specify how to handle invalid data channel IDs, or lack of IDs.

taylor-b:
Fixes issue #746.

For an ID of 65535, createDataChannel will throw an exception. If
it's determined that an ID is invalid for any other reason (such as
shortage of negotiated SCTP streams), or if no more valid IDs are
available when the user agent tries to pick an ID, the data channel will
fire an "error" event and be closed.

For now, I'm leaving the actual error as a "NetworkError", since I
think we need to separately decide how error events on data channels
work (see issue #846).


webrtc.html
```diff
@@ -7149,6 +7155,36 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
         </li>
       </ol>
+      <p>In some cases, the user agent may be unable to create an
+      <code><a>RTCDataChannel</a></code>'s <a>underlying data transport</a>.
+      For example, the data channel's <code><a
+      data-for="RTCDataChannel">id</a></code> may be outside the range
+      negotiated by the [[!RTCWEB-DATA]] implementations, or in the case where
+      the user agent generates an id according to [[!RTCWEB-DATA-PROTOCOL]], no
+      more IDs may be available. When the user agent determines that an
+      <code><a>RTCDataChannel</a></code>'s <a>underlying data transport</a>
```
pthatcherg: What's an "underlying data transport"?  Why not just say that an ID could not be allocated?

taylor-b: > What's an "underlying data transport"?

My assumption was that it refers to a DataChannel as defined by RTCWEB-DATA.

> Why not just say that an ID could not be allocated?

I wanted to be as specific as possible (by providing examples), but also use terminology consistent with the rest of this section. I think "the underlying data transport cannot be created" accomplishes that.

PR #848, opened 2016-10-04
Fix RTCDTMFSender attribute types in prose to match WebIDL.

jan-ivar:
The WebIDL says unsigned long, which matches insertDTMF inputs. Prose says long.

Fix prose to say unsigned long.


aboba:
LGTM.

aboba:
We have removed the interToneGap and duration attributes, so I believe this PR is no longer needed. Please reopen if that is incorrect.

Issue #843, opened 2016-09-29
DTMF: duration and interToneGap attributes: Do we need them?`


aboba:
Do we need the duration and interToneGap attributes?

fluffy:
I think we need to be able to do a "long pound" as that is commonly used in IVRs. So I see a use case for some way to indicate do DTMF that is a more than a few seconds long. I don't see much use for interToneGap

fluffy:
I can't see any reason we need the attributes (still need the at least the duration parameter in the function call)

alvestrand:
I think removing them makes the API simpler and no less powerful, so I'm happy to see them go.

alvestrand:
Closed by #852

smaug----:
Just a comment from someone who reviews often webidl changes in Gecko: why is the API being changed so often? Does any vendor have data whether the attributes here are used and whether it is safe to remove them?
It is unusual for web specs to be changed so often after there are implementations shipping.

PR #852, opened 2016-10-05
Remove duration and interToneGap attributes

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/843

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7833,8 +7833,6 @@ interface RTCDTMFSender : EventTarget {
     void insertDTMF (DOMString tones, optional unsigned long duration = 100, optional unsigned long interToneGap = 70);
                     attribute EventHandler ontonechange;
     readonly        attribute DOMString    toneBuffer;
-    readonly        attribute unsigned long         duration;
-    readonly        attribute unsigned long         interToneGap;
 };</pre>
         <section>
           <h2>Attributes</h2>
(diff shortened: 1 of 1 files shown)


rshpount:
Should toneBuffer be an array with DTMF tone, tone duration and subsequent gap? This way insertDTMF method would store desired tones with their duration and gap in toneBuffer and DTMF playback algorithm will play each tone based on the timing information stored for each tone. Otherwise it looks like you can adjust the duration of currently playing on un-played tones by calling insertDTMF.

aboba:
@rshpount Based on this PR, the duration and interToneGap arguments are used within the queued playout task, rather than the attributes.  So it appears to me that calling:

insertDTMF("bbbbb",  100, 70)
insertDTMF("111", 6000, 100)

could not cause "B" tones to be played with a duration of 6 seconds (e.g. the second insertDTMF call will queue a playout task that will execute after the first one).

rshpount:
@aboba This PR is ambiguous since in case

insertDTMF("bbbbb", 100, 70)
insertDTMF("111", 6000, 100)

It is unclear which value of duration and interToneGap parameters are referenced in the Playout task.

It would be cleaner to state that duration and interToneGap parameter values are stored for each tone by insertDTMF and these values are used by the Playout task. It means that internal state of the RTCDTMFSender is not toneBuffer string, but an array of tones with tone name, duration and interToneGap stored for each of them. If this is the internal state, then this is the data that should be exposed.

martinthomson:
If I call `insertDTMF("1", 100, 100)` then `insertDTMF("2", 200, 200)`, I agree with @rshpount that the "1" should play for 100ms and the "2" should play for 200ms, but what is the gap between the two?  (i.e., is the inter-tone gap a leading (`gapBefore2 = nextTone.gap`), trailing (`gapBefore = lastTone.gap`) or combinatorial property (`gapBefore2 = Math.max(lastTone.gap, nextTone.gap)`; see CSS margins for an example of this in practice).

alvestrand:
It's clear that people have problems reading the algorithm. Removing the attributes makes the text describing the algorithm simpler. So this change seems good.

rshpount:
The algorithm was clearer with the attributes, but it was saying the wrong thing (subsequent insertDTMF updates the duration and interToneGap attributes, which are used for the playback of queued tones, including tones queued by the previous insertDTMF). Without attributes it is unclear which duration and interToneGap values are used for tone playback.

I agree that attribute removal is a good thing, but should we create another issue about the playback algorithm? I think fixing the playback algorithm might affect the attribute exposed by RTCDTMFSender.

alvestrand:
@rshpount you missed the part where insertDTMF will clear the tone buffer.
It can never modify playback on previous tones, because the previous tones are either being played already, or they are cleared.

rshpount:
@alvestrand You are correct that I missed that insertDTMF resets or clears the buffer. The only thing to be might need to be clarified is that tone which is currently being played when new insertDTMF is called is followed up by the interToneGap from the insertDTMF call that queued the tone, not any of the subsequent ones.

Issue #813, opened 2016-09-22
getConfiguration doesn't include all of RTCConfiguration


pthatcherg:
getConfiguration is weird in that it makes a new RTCConfiguration object to return and explicitly fills in a few of the fields (ICE transports, ICE servers, certificates), but not others (bundlePolicy, rtcpMuxPolicy, iceCandidatePoolSize).  Should we just have a blanket "copy everything" statement, which would include any future things?  Or should we explicitly add the things that are missing?

aboba:
@pthatcherg  The method description says "Returns a RTCConfiguration object representing the current configuration of this RTCPeerConnection object."  Based on that, I'd expect it to "copy everything". Otherwise it would have said "representing some of the current configuration (in an inconsistent way, to keep you guessing)."

aboba:
Can we close this now?

PR #858, opened 2016-10-07
getConfiguration should include all of RTCConfiguration

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/813


aboba:
@pthatcherg Does this address Issue 813?

stefhak:
Alternative: store object in an internal slot.

adam-be:
Can't we spec this as storing a copy of the input to `setConfiguration` in an internal slot, and then return that in `getConfiguration` (instead of taking it apart and assembling it again)

aboba:
@stefhak  In addition to attributes set in setConfiguration or in the RTCPeerConnection constructor, there is the certificates attribute.  If this value is absent, then a set of certificates are generated for each RTCPeerConnection instance. As a result, getConfiguration().certificates may not return the same value as the RTCConfiguration passed in the RTCPeerConnection constructor (or in setConfiguration).

PR #860, opened 2016-10-07
Clarify when removeTrack should throw an exception.

taylor-b:
In PR #844, the intent was to throw an exception when the sender wasn't created
by the connection on which the method was called. And the assumption was that
senders are never removed, so the algorithm can simply check if it's in the
connection's set of senders.

But something I didn't consider is that rollbacks are able to remove
transceivers (and thus senders). So more clarity is needed.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4347,15 +4347,20 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                   exception and abort these steps.</p>
                 </li>
                 <li>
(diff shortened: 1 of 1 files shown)


taylor-b:
Note that the "check if sender is in the set of senders" steps may now be redundant, because a sender removed by a rollback is guaranteed to be stopped (I'd assume).

adam-be:
LGTM

Issue #814, opened 2016-09-22
What does setConfiguration change or not change?


pthatcherg:
I find the text of setConfiguration confusing:

"Let the value of configuration.iceTransportPolicy be the ICE Agent's ICE transports setting."  Does that mean set/change the ICE Agent's transport setting to configuration.iceTransportPolicy?  If so, it could be more clear.

But then there's a problem, I think, with "Let the value of configuration.bundlePolicy be connection's bundle policy."  What does it mean to change a bundle policy if the remote endpoint doesn't support BUNDLE?  Does that mean we potentially re-negotiate or undo negotiation of media sections?  It seems like a mess and I'd prefer not to allow changing the bundle policy.  This is similar to how we don't change the rtcpMuxPolicy.

aboba:
I agree that the setConfiguration text is confusing and incomplete.

aboba:
Can we close this now?

PR #861, opened 2016-10-13
What setConfiguration changes

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/814

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1893,8 +1893,8 @@ interface RTCPeerConnection : EventTarget  {
               <dd>
                 <p>The <code>setConfiguration</code> method updates the
                 configuration of this <code><a>RTCPeerConnection</a></code>
-                object.  This includes changing configuration of the <a>ICE
-                Agent</a>.</p>
+                object.  This includes changing the configuration of the
+                <a>ICE Agent</a>.</p>
                 <p>Since this call can change parameters relating to gathering
                 local candidates and pinging remote candidates, it may
                 result in a change to the state of the <a>ICE
--- a/webrtc.html
+++ b/webrtc.html
@@ -1894,12 +1894,10 @@ interface RTCPeerConnection : EventTarget  {
                 <p>The <code>setConfiguration</code> method updates the
                 configuration of this <code><a>RTCPeerConnection</a></code>
                 object.  This includes changing the configuration of the
-                <a>ICE Agent</a>.</p>
-                <p>Since this call can change parameters relating to gathering
-                local candidates and pinging remote candidates, it may
-                result in a change to the state of the <a>ICE
-                Agent</a>, and may result in a change to media state if it
-                results in connectivity being established.</p>
+                <a>ICE Agent</a>.  As noted in
+                <span data-jsep="ice-gathering-overview">[[!JSEP]]</span>,
+                when the ICE configuration changes in a way that requires a new
+                gathering phase, an ICE restart is required.</p>
                 <p>When the <dfn><code>setConfiguration</code></dfn> method is
                 invoked, the user agent MUST run the following steps:</p>
                 <ol>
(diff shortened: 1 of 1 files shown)


aboba:
@pthatcherg Can you review?

webrtc.html
```diff
@@ -1891,10 +1891,13 @@
               </dd>
               <dt><code>setConfiguration</code></dt>
               <dd>
-                <p>The <code>setConfiguration</code> method updates the <a>ICE
-                Agent</a> process of gathering local candidates and pinging
-                remote candidates.</p>
-                <p>This call may result in a change to the state of the <a>ICE
+                <p>The <code>setConfiguration</code> method updates the
+                configuration of this <code><a>RTCPeerConnection</a></code>
+                object.  This includes changing configuration of the <a>ICE
```
taylor-b: nit: "the configuration" for consistency

webrtc.html
```diff
@@ -1891,10 +1891,13 @@
...
+                Agent</a>.</p>
+                <p>Since this call can change parameters relating to gathering
+                local candidates and pinging remote candidates, it may
```
taylor-b: To be consistent with ICE terminology, should probably say "performing connectivity checks" instead of "pinging remote candidates".

webrtc.html
```diff
@@ -1891,10 +1891,13 @@
...
+                result in a change to the state of the <a>ICE
                 Agent</a>, and may result in a change to media state if it
                 results in connectivity being established.</p>
```
taylor-b: According to my understanding, pooled candidates still won't be used until an ICE restart occurs. So setConfiguration by itself won't change the ICE or media state.

webrtc.html
```diff
@@ -1891,10 +1891,13 @@
               </dd>
               <dt><code>setConfiguration</code></dt>
```
taylor-b: This whole section should probably point to JSEP somewhere since it goes into more detail.

webrtc.html
```diff
@@ -1932,26 +1935,29 @@
                   set and the set of certificates differs from the ones used
                   when <var>connection</var> was constructed, throw an
                   <code>InvalidModificationError</code>.</li>
```
taylor-b: nit: This paragraph regarding certificates could match the format of the new sections you added with InvalidModificationErrors

webrtc.html
```diff
@@ -1932,26 +1935,29 @@
...
+                  rtcpMux policy, throw an <code>InvalidModificationError</code>.
                   </li>
                   <li>
-                    <p>Let the value of
+                    <p>Set the <a>ICE Agent</a>'s <dfn id="ice-transports-setting">
+                    ICE transports setting</dfn> to the value of
                     <code><var>configuration</var>.<a data-for=
-                    "RTCConfiguration">bundlePolicy</a></code> be
-                    <var>connection</var>'s bundle policy.</p>
+                    "RTCConfiguration">iceTransportPolicy</a></code>.</p>
```
taylor-b: May want to clarify that this doesn't take effect immediately; only until the next ICE restart.

webrtc.html
```diff
@@ -1932,26 +1935,29 @@
...
                   </li>
                   <li>
-                    <p>Let the value of
+                    <p>Set the <a>ICE Agent</a>'s prefetched
+                    <dfn>ICE candidate poolsize</dfn> as defined
```
taylor-b: nit: should poolsize be one word?

webrtc.html
aboba: Yes, that is stated in JSEP Section 3.5.1, so we need to make that clear (as well as referencing that section of JSEP).

webrtc.html
aboba: I believe that the correct reference is to JSEP Section 3.5.1.

webrtc.html
aboba: Yes, that is stated in JSEP Section 3.5.1, so we need to make that clear (as well as referencing that section of JSEP).

webrtc.html
aboba: I'll insert a space.

Issue #841, opened 2016-09-28
transceiver.stop() effect on transceiver.receiver.track.onended EventHandler


aboba:
When transceiver.stop() is called, does transceiver.receiver.track's "onended" EventHandler fire? 

Related ORTC Issue: https://github.com/w3c/ortc/issues/596

PR #862, opened 2016-10-13
transceiver.stop() effect on transceiver.receiver.track "onended" event

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/841

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6042,8 +6042,8 @@ sender.setParameters(params)
               <p>The <dfn><code>stop</code></dfn> method stops the
               <a><code>RTCRtpTransceiver</code></a>. The sender of this
               transceiver will no longer send, the receiver will no longer
-              receive, <code>receiver.track</code>'s "onended" event will
-              fire, and the connection is marked as needing negotiation.</p>
+              receive, and the connection is marked as needing negotiation.
+              <code>receiver.track</code>'s "onended" event will not fire.</p>
               <div>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6042,7 +6042,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
               <p>The <dfn><code>stop</code></dfn> method stops the
               <a><code>RTCRtpTransceiver</code></a>. The sender of this
               transceiver will no longer send, the receiver will no longer
-              receive, and the <a>negotiation-needed flag</a> is set.</p>
+              receive, <code>receiver.track</code>'s "onended" event will
+              fire, and the <a>negotiation-needed flag</a> is set.</p>
```
adam-be: I suspect that we don't want to fire the "ended" event synchronously here. Since the script is the source of this action, I think we could just end the track directly (without any event).

Not introduced by this change, but should we replace the wording about setting the 'negotiation-neeed flag' and use "mark this connection as needing negotiation" instead?

stefhak: I guess the firing or not of the ended event can be discussed. While it's clear we should not fire if the app does ```stop()``` on the track, it is a secondary effect here coming from stopping the transeiver.

adam-be:
The RTCRtpTransceiver.stop method is underspecified and needs to detail out how the related tracks are ended. #878 is the new issue.

PR #864, opened 2016-10-13
Make channel member of RTCDataChannelEventInit required 

### LINKED ISSUE #817: The RTCDataChannelEventInit channel member needs to be required
https://w3c.github.io/webrtc-pc/#rtcdatachannelevent

If it isn't required, then `new RTCDataChannelEvent('type', {})` won't throw, and the resulting instance will have null for the `channel` attribute even though it's not nullable in the IDL. So:

```WebIDL
dictionary RTCDataChannelEventInit : EventInit {
    required RTCDataChannel channel;
};
```

adam-be:
...also add desccription

Fix #817


foolip:
Thanks!

PR #872, opened 2016-10-19
Pass peer identity to IdP

### LINKED ISSUE #678: Support assertions that identify the recipient
When imitating communications between two parties, some systems use assertions that only identify the initiator. However, these assertions have a security weakness in that they can be cut and pasted and sent to many different receiving parties because the intended recipient of the session is not identified in the assertion. Many identity systems, such as the STIR work at IETF, protect against that by including the intended recipient in the assertion. 

The current WebRTC spec can be used with the first type of identity systems but does not well suited for the the second type. To improve the security, the GenerateAssertionCallback  ( https://w3c.github.io/webrtc-pc/#idl-def-generateassertioncallback ) should be extended to allow and option recipientNameHint similar to the existing usernameHint ). 

This will make WebRTC fit into STIR as well as other identity assertion systems without loosing any existing functionally.

martinthomson:
Closes #678

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8734,7 +8734,7 @@ interface RTCIdentityProviderRegistrar {
               </dd>
               <dt><code>options</code> of type <span class=
               "idlMemberType"><a>RTCIdentityProviderOptions</a></span></dt>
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

PR #873, opened 2016-10-19
Tidy

martinthomson:
All I did was run `make tidy`.


alvestrand:
The result of "make tidy" is pushed on master.

Abandon all PRs, ye who enter here.

Issue #806, opened 2016-09-22
Need symmetry to #782 for createAnswer


fluffy:
In 
https://github.com/w3c/webrtc-pc/issues/782

we change how async work for createOffer. Need to make same change for createAnswer

alvestrand:
Merged.

PR #875, opened 2016-10-19
Rewrite the createOffer algorithm to eliminate race conditions (restructured).

taylor-b:
A revised version of PR #874. Turns the single bulky `createOffer` algorithm into three separate sub-algorithms. Some miscellaneous cleanup, but mostly just restructuring.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1316,6 +1316,11 @@ interface RTCPeerConnection : EventTarget  {
                     method was invoked.</p>
                   </li>
                   <li>
+                    <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+                    <code>true</code>, throw an <code>InvalidStateError</code>
+                    exception and abort these steps.</p>
+                  </li>
+                  <li>
                     <p>If <var>connection</var> is configured with an
                     identity provider, and an identity assertion has not
                     yet been generated using said identity provider, then
--- a/webrtc.html
+++ b/webrtc.html
@@ -970,6 +970,83 @@
             <p>Return <var>p</var>.</p>
           </li>
         </ol>
+        <p>The <dfn>steps to create an offer</dfn> given a promise
+        <var>p</var> are as follows:</p>
+        <ol>
+          <li>
+            <p>If the need for an identity assertion was identified when
+            <code>createOffer</code> was invoked, wait for <a href=
+            "#sec.identity-proxy-assertion-request">the identity assertion
+            request process</a> to complete.</p>
+          </li>
+          <li>
+            <p>If the identity provider was unable to produce an identity
+            assertion, reject <var>p</var> with a
+            <code>DOMException</code> object whose <code>name</code>
+            attribute has the value <code>NotReadableError</code>, and
+            abort these steps.</p>
+          </li>
+          <li>
+            <p>If <var>connection</var> was not constructed with a set of
+            certificates, and one has not yet been generated, wait for it
+            to be generated.</p>
+          </li>
+          <li>
+            <p>Inspect the system state to determine the currently
+            available resources as necessary for generating the offer, as
+            described in <span data-jsep="createoffer">[[!JSEP]]</span>.
+            </p>
+          </li>
+          <li>
+            <p>If this inspection failed for any reason, reject
+            <var>p</var> with a <code>DOMException</code> object whose
+            <code>name</code> attribute has the value
+            <code>OperationError</code>, and abort these steps.</p>
+          </li>
+          <li>
+            <p>Queue a task that runs the <a>final steps to create an
+            offer</a>, given <var>p</var>.</p>
+          </li>
+        </ol>
+        <p>The <dfn>final steps to create an offer</dfn> given a promise
+        <var>p</var> are as follows:</p>
+        <ol>
+          <li>
+            <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+            <code>true</code>, then abort these steps.</p>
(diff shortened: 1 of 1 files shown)


adam-be:
I like the structure of this new version.

adam-be:
Since `steps to create an offer` and `final steps to create an offer` won't be referenced by any other algorithms I think it's safe to put them below the description of `createOffer` to make the spec easier to read with less jumping.

webrtc.html
```diff
@@ -1316,6 +1393,15 @@
                     method was invoked.</p>
                   </li>
```
adam-be: Since we're doing work in the sync section here, we might want to squeeze in an additional [[isClosed]] check.

webrtc.html
```diff
@@ -970,6 +970,83 @@
...
+            <var>connection</var>, but while performing the <a>steps to
+            create an offer</a> in parallel, a video
+            <code><a>RTCRtpTransceiver</a></code> was added, requiring
+            additional inspection of video system resources.</div>
+          </li>
+          <li>
+            <p>Given the information that was obtained from previous
+            inspection, generate an SDP offer, <var>sdpString</var>, as
+            described in <span data-jsep="create-offer">[[!JSEP]]</span>.
+            </p>
```
adam-be: Nit:
Given the information that was obtained from previous inspection **and the current media configuration**, generate an SDP offer...

stefhak:
Assigning @jan-ivar for review.

webrtc.html
taylor-b: Done, though I said "the current state of *connection* and its `RTCRtpTransceiver`s" to be more specific.

webrtc.html
taylor-b: Good catch. Sidenote: it may be useful to define a **check that the connection is open** shorthand, since the [[isClosed]] check is used so often.

taylor-b:
Moved everything to the `createOffer` section and addressed comments.

webrtc.html
```diff
@@ -1316,6 +1316,20 @@
                     method was invoked.</p>
                   </li>
                   <li>
+                    <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+                    <code>true</code>, throw an <code>InvalidStateError</code>
+                    exception and abort these steps.</p>
+                  </li>
```
adam-be: We should return a rejected promise here instead of throwing.

taylor-b: Done.

webrtc.html
jan-ivar: @adam-be Why add that extra information? It adds confusion to me, because I thought the step directly above just went through a lot of trouble to ensure that the information obtained from previous inspection and the current state are *exactly the same*.

@taylor-b I think in practice, implementations will have generated the *sdpString* already at this point, in parallel (step 4) as part of the inspection (no reason not to). That's also what the spec says today (before this PR). Doing it here on the main thread is probably tricky.

That said, I like how this reads, and implementers can probably figure out that shortcut for themselves, given in all the assurances about state not having changed. Either way.

jan-ivar:
Lgtm!

webrtc.html
taylor-b: @jan-ivar The extra information is a little redundant, but I don't think the added clarity hurts.

As far as what implementations will do; they're still free to generate the SDP in parallel as long as the end result is equivalent. So to me, putting that step on the main thread just simplifies how the algorithm reads. It only has to reference JSEP and SDP generation in one place.

taylor-b:
Since it sounds like there's support for this approach, do you want me to do the same thing with `createAnswer` in this PR? Or leave it for a new one? It may be confusing if someone tries to read the spec while `createOffer` is revamped but `createAnswer` isn't.

webrtc.html
jan-ivar: @taylor-b I agree the readability is good, so the sdpString part works for me.

But on the extra information, I actually think it removes clarity, as it implies there may be more state to consider, when the whole point of this carousel back and forth is ensuring there isn't. Can we elaborate in the parallel step 4 instead, where the work is done?

There should be very little to do on the main thread, other than check that everything is the same, and return a dictionaried string.

jan-ivar:
@taylor-b Whichever you think is easier. I don't think we need to worry about readers for the subtle behavior change that this is. I'll try to respond faster next time.

webrtc.html
taylor-b: In my opinion, this text makes it explicit that when following the JSEP process for creating an offer, whenever JSEP refers to the set of transceivers (for example), the implementation should use the current set of transceivers, not the set at the point when `createOffer` was called.

I'm fine with changing it, but I'm not sure how it relates to parallel step 4. Parallel step 4 is "inspect the system state" (for example, see what hardware resources are available), but this step is referring to creating m= sections given the current set of `RTCRtpTransceivers`, with their current direction, codec preferences, etc. Can you write out what you had in mind?

taylor-b:
Alright, I'll do `createAnswer` in a separate PR then.

jan-ivar:
@taylor-b I believe there is one already: https://github.com/w3c/webrtc-pc/issues/806

webrtc.html
jan-ivar: @taylor-b I see, and in the parallel steps it is less clear which state should be used. Arguably, this is an optimization since the loop-around catches any changes from initial set anyhow, but I see how it might be harder to write in the parallel steps in a way that's unambiguous. Provided JSEP lets us split things up this way, then what you have WFM.

PR #880, opened 2016-10-24
Cleanup the definition of expires, use IDL

### LINKED ISSUE #879: Clarify the processing of expires setting in generateCertificate
When writing https://github.com/w3c/web-platform-tests/pull/4062 I tried this to get an expired certificate:

`RTCPeerConnection.generateCertificate({ name: "ECDSA", namedCurve: "P-256", expires: 0 }`

`DOMTimeStamp` is `unsigned long long`, and there's nothing in the spec that suggests that 0 shouldn't work. It works in Chrome, but not Firefox. This needs spec clarification or at least agreement on the right behavior since we now have two.

When writing tests for this, it would also be good to test -1, 0.1, NaN, Infinity, -Infinity and numbers around 2^31, 2^32, 2^63 and 2^64.

martinthomson:
The definition of expires was at best improvised.  This cleans it up to use WebIDL properly.

Closes #879

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3777,7 +3777,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
       SHA-256 hash.</p>
       <div>
         <pre class="idl">partial interface RTCPeerConnection {
-    static Promise&lt;RTCCertificate&gt; generateCertificate (AlgorithmIdentifier keygenAlgorithm);
+    static Promise&lt;RTCCertificate&gt; generateCertificate (RTCGenerateCertificateOptions keygenAlgorithm);
 };</pre>
         <section>
           <h2>Methods</h2>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3777,7 +3777,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       SHA-256 hash.</p>
       <div>
         <pre class="idl">partial interface RTCPeerConnection {
-    static Promise&lt;RTCCertificate&gt; generateCertificate (AlgorithmIdentifier keygenAlgorithm);
+    static Promise&lt;RTCCertificate&gt; generateCertificate (RTCGenerateCertificateOptions keygenAlgorithm);
```
foolip: Is it intentional that this will start rejecting with TypeError for `RTCCertificate.generateCertificate("string")`? The previous type was `object or DOMString`.

martinthomson: Yes, intentional.  Try finding a string that produces a sensible result.

henbos: It currently takes an AlgorithmIdentifier. Any string that would be accepted by WebCrypto's normalizeAlgorithm[1] is an AlgorithmIdentifier in Chromium. According to a comment in the code, a String is treated the same as a Dictionary with a "name" attribute and nothing else.

However, we reject any AlgorithmIdentifier that doesn't have expected non-name parameters such as modulusLength, publicExponent, hash and namedCurve, so that depends on if the default values for these align with parameters we accept in generateCertificate.

... And, I just checked, neither "RSA" or "ECDSA" are accepted Strings for generateCertificate. Perfect, that means this should be a safe change.

[1] https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/crypto/NormalizeAlgorithm.h?sq=package:chromium&dr&rcl=1477293892&l=66

Issue #803, opened 2016-09-20
Rules for negotiation-needed flag need to be updated for transceivers.


taylor-b:
I found a few problems with the current rules regarding the negotiation-needed flag:
- It's still described overall in terms of tracks, not transceivers (such as in section 4.8.1).
- It's only cleared by `setLocalDescription`, but `setRemoteDescription` should be able to clear it as well in some cases.
- `createAnswer` sets the flag? I don't see the reasoning behind this. I'd expect setting a description to possibly set the flag, but not generating one.
- The act of firing `negotiationneeded` after reaching the stable signaling state also clears the flag? This means if the next SDP exchange didn't finish negotiating something (such as if it was initiated by the remote peer), the event wouldn't be fired again.

taylor-b:
Another thing I realized: we need to specify what it means for all transceivers to be represented by the current negotiated description.

If you call `addTrack`, you get a "sendrecv" transceiver. But the remote endpoint may not have a track to send, resulting in the negotiated direction (on the offerer side) being "sendonly". Even though this doesn't match the transceiver's direction, it would be odd to get negotiation-needed in this case.

aboba:
Now that PR 913 has been merged, can we close this issue?

taylor-b:
Yes, the PR addressed all the problems I mentioned above.

PR #893, opened 2016-10-28
Rewrite createAnswer algorithm to eliminate race conditions.

taylor-b:
Fixes issue #806. This is the createAnswer equivalent of PR #875.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1623,7 +1623,6 @@ interface RTCPeerConnection : EventTarget  {
                     <p>Resolve <var>p</var> with <var>answer</var>.</p>
                   </li>
                 </ol>
-                </ol>
                 <table class="parameters">
                   <tbody>
                     <tr>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1502,12 +1502,25 @@ interface RTCPeerConnection : EventTarget  {
                     method was invoked.</p>
                   </li>
                   <li>
+                    <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+                    <code>true</code>, return a promise rejected with an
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1516,77 +1529,101 @@
...
+                    then in parallel begin the <a>steps to create an answer</a>
+                    again, given <var>p</var>, and abort these steps.</p>
+                    <div class="note">
+                      This may be necessary if, for example,
+                      <code>createAnswer</code> was called when an
+                      <code><a>RTCRtpTransceiver</a></code>'s direction was
+                      "recvonly", but while performing the <a>steps to create
+                      an answer</a> in parallel, the direction was changed to
+                      "sendrecv", requiring additional inspection of video
+                      encoding resources.
```
taylor-b: This is a much more niche situation than `createOffer`'s. But I figured it was worth including for symmetry, if nothing else.

webrtc.html
```diff
@@ -1516,77 +1529,101 @@
...
+                    <code><a>RTCSessionDescriptionInit</a></code> dictionary
+                    with its <code>type</code> member initialized to the string
+                    <code>"answer"</code> and its <code>sdp</code> member
+                    initialized to <var>sdpString</var>.</p>
+                  </li>
+                  <li>
+                    <p>Unless all non-stopped
+                    <code><a>RTCRtpTransceiver</a></code>s are represented in
+                    <var>answer</var>, mark <var>connection</var> as needing
+                    negotiation.</p>
```
taylor-b: I don't believe it's correct to do this here, but I'm leaving it in since it was there before. The overall issue of how negotiation-needed works is covered by issue #803.

PR #896, opened 2016-10-29
Use proper forms for setting RTCCertificate.expires

### LINKED ISSUE #891: Further generateCertificate changes required
A follow-up to https://github.com/w3c/webrtc-pc/issues/879 and https://github.com/w3c/webrtc-pc/pull/880:

A WebIDL dictionary doesn't allow for anything "extra", so using the `RTCGenerateCertificateOptions` dictionary, the only possible members would be `name` and `expires`. But the spec requires at least these things to work:
- `{ name: "RSASSA-PKCS1-v1_5", modulusLength: 2048, publicExponent: new Uint8Array([1, 0, 1]), hash: "SHA-256" }`
- `{ name: "ECDSA", namedCurve: "P-256" }`

The [algorithm normalization process](https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#algorithm-normalization) link is stale, but must mean that [normalize an algorithm](https://w3c.github.io/webcrypto/Overview.html#dfn-normalize-an-algorithm) should be used, which requires an `AlgorithmIdentifier` as its input. That's how it's implemented (@henbos) in Blink. (The prose for `generateCertificate` still refers to `AlgorithmIdentifier`, but that's an oversight I think.)

Paging @ericroman920 and @mwatson2 for advice on how to do this.

martinthomson:
The last patch that updated this was a complete disaster, as I discovered when trying to implement it.  After discussing with bz, this is what I arrived at.  Note that we need to use `AlgorithmIdentifier`.

I've also updated webcrypto links, which were all broken.

Closes #891.


webrtc.html
```diff
@@ -3819,7 +3820,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       SHA-256 hash.</p>
       <div>
         <pre class="idl">partial interface RTCPeerConnection {
-    static Promise&lt;RTCCertificate&gt; generateCertificate (RTCGenerateCertificateOptions keygenAlgorithm);
+    static Promise&lt;RTCCertificate&gt; generateCertificate (AlgorithmIdentifier keygenAlgorithm);
```
alvestrand: I don't get this. You're saying here that the argument is an AlgorithmIdentifier, and you're saying below that when generateCertificate is called with an Object, you will attempt to turn it into an RTCCertificateExpiration.
Is it an Object or an AlgorithmIdentifier?

Issue #898, opened 2016-11-01
Does RTCRtpReceiver.track.stop() reject the track?


Pehrsons:
In [**5.1.1 Processing Remote MediaStreamTracks**](https://w3c.github.io/webrtc-pc/archives/20160913/webrtc.html#processing-remote-mediastreamtracks) it says _Rejection of incoming MediaStreamTrack objects can be done by the application, after receiving the track, by stopping it._

It doesn't define _Rejection_, but I'll just assume for now it involves talking to the RTCPeerConnection.

However, [pull request 662](https://github.com/w3c/webrtc-pc/pull/662) added the text _Since `track.stop()` does not implicitly call `RTCRtpReceiver.stop()`, Receiver Reports continue to be sent._ which indicates that no rejection takes place on track.stop(), but rather on receiver.stop().

Though there is no `RTCRtpReceiver.stop()`, so I'll assume for now that it refers to `RTCRtpTransceiver.stop()`, but that again makes it even less clear.

I need clarification on what stopping a received MediaStreamTrack does, and how one rejects a received track.

These questions came up while implementing `MediaStreamTrack.stop()` for received tracks in Firefox. See [bug 1301675](https://bugzilla.mozilla.org/show_bug.cgi?id=1301675).

aboba:
@Pehrsons Removing an m-line can be accomplished by calling RTCRtpTransceiver.stop().  Calling RTCRtpTransceiver.setDirection() can be used to set the transceiver to "sendonly" or "inactive".  Both stop and setDirection set the negotiation-needed flag.

Pehrsons:
That clarifies it for me, thanks.

Shouldn't the reference to RTCRtpReceiver.stop() be dropped as well?

aboba:
@Pehrsons Reference to RTCRtpReceiver.stop() has been removed in the latest build: 
https://cdn.rawgit.com/w3c/webrtc-pc/master/webrtc.html

alvestrand:
Fixed in #900

jan-ivar:
So what does `track.stop()` do?

Issue #878, opened 2016-10-20
What happens when transceiver.stop() is called?


aboba:
The current text lacks detail.

PR #901, opened 2016-11-01
What happens when transceiver.stop() is called

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/878
Related to Issue https://github.com/w3c/webrtc-pc/issues/898

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6115,10 +6115,10 @@ sender.setParameters(params)
                   <p>Let <var>receiver</var> be <code><var>transceiver</var>.receiver</code>.</p>
                 </li>
                 <li> 
-                  <p><var>sender</var> will no longer send media.</p>
+                  <p>Stop sending media with <var>sender</var>.</p>
                 </li>
                 <li> 
-                  <p><var>receiver</var> will no longer receive media.</p>
+                  <p>Stop receiving media with <var>receiver</var>.</p>
                 </li>
                 <li> 
                   <p><code><var>receiver</var>.track</code> is ended.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -6116,6 +6116,9 @@ sender.setParameters(params)
                 <li> 
                   <p><var>receiver</var> is now considered <a>stopped</a>.</p>
                 </li>
+                <li> 
+                  <p><code><var>receiver</var>.track</code> is ended.</p>
+                </li>
                 <li>
                   <p>Set the value of <code><var>transceiver</var>.stopped</code>
                   to <code>true</code>.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -6095,6 +6095,10 @@ sender.setParameters(params)
                   be stopped.</p>
                 </li>
                 <li>
+                  <p>If <code><var>transceiver</var>.stopped</code> is <code>true</code>
+                  abort these steps.</p> 
+                </li>
+                <li>
                   <p>Let <var>connection</var> be the
                   <code><a>RTCPeerConnection</a></code> object on which
                   the <var>transceiver</var> is to be stopped.</p>
@@ -6111,10 +6115,10 @@ sender.setParameters(params)
                   <p>Let <var>receiver</var> be <code><var>transceiver</var>.receiver</code>.</p>
                 </li>
                 <li> 
-                  <p><var>sender</var> is now considered <a>stopped</a>.</p>
+                  <p><var>sender</var> will no longer send media.</p>
                 </li>
                 <li> 
-                  <p><var>receiver</var> is now considered <a>stopped</a>.</p>
+                  <p><var>receiver</var> will no longer receive media.</p>
                 </li>
                 <li> 
                   <p><code><var>receiver</var>.track</code> is ended.</p>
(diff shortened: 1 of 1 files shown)


aboba:
@taylor-b Can you review?

webrtc.html
```diff
@@ -6082,6 +6082,63 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                  <code><a>RTCPeerConnection</a></code> object on which
+                  the <var>transceiver</var> is to be stopped.</p>
+                </li>
+                <li>
+                  <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+                  <code>true</code>, throw an <code>InvalidStateError</code>
+                  exception and abort these steps.</p>
+                </li>
+                <li>
+                  <p>Let <var>sender</var> be the <code><var>transceiver</var>.sender</code>.</p>
```
taylor-b: extra "the"

webrtc.html
```diff
@@ -6082,6 +6082,63 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                </li>
+                <li>
+                  <p>Let <var>receiver</var> be <code><var>transceiver</var>.receiver</code>.</p>
+                </li>
+                <li> 
+                  <p><var>sender</var> is now considered <a>stopped</a>.</p>
```
taylor-b: It's not clear what it means to "consider a sender stopped". Could we just have a step that says: "immediately stop sending and receiving media on this transceiver."?

webrtc.html
```diff
@@ -6082,6 +6082,63 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                <li> 
+                  <p><code><var>receiver</var>.track</code> is ended.</p>
+                </li>
+                <li>
+                  <p>Set the value of <code><var>transceiver</var>.stopped</code>
+                  to <code>true</code>.</p>
+                </li>
+                <li>
+                  <p>Remove the <var>transceiver</var> from the
+                  <a>set of transceivers</a>.</p>
```
taylor-b: I always assumed a stopped transceiver would remain in the set. We'd have to rewrite parts of JSEP if we made this change, since the rules for createOffer/Answer deal with stopped transceivers.

webrtc.html
```diff
@@ -6082,6 +6082,63 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                </li>
+                <li>
+                  <p>Remove <var>sender</var> from the
+                  <a>set of senders</a>.</p>
```
taylor-b: "set of senders" isn't linked to anything. But I don't think this step (or the one below) is necessary, since "CollectSenders" just references the set of transceivers.

webrtc.html
```diff
@@ -6082,6 +6082,63 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                <li>
+                  <p>Let <var>connection</var> be the
+                  <code><a>RTCPeerConnection</a></code> object on which
+                  the <var>transceiver</var> is to be stopped.</p>
+                </li>
+                <li>
+                  <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+                  <code>true</code>, throw an <code>InvalidStateError</code>
+                  exception and abort these steps.</p>
+                </li>
```
taylor-b: Do we want an early return for if the transceiver is already stopped? At least, setting the negotiation-needed flag should be avoided in that case.

aboba: Done.

webrtc.html
```diff
@@ -6082,6 +6082,55 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                <li>
+                  <p>Let <var>sender</var> be <code><var>transceiver</var>.sender</code>.</p>
+                </li>
+                <li>
+                  <p>Let <var>receiver</var> be <code><var>transceiver</var>.receiver</code>.</p>
+                </li>
+                <li> 
+                  <p><var>sender</var> will no longer send media.</p>
```
taylor-b: May be just a matter of preference, but since this is in a list of steps it feels like it should be an instructive rather than declarative statement. So I'd prefer "Stop sending media with sender" (or "permanently stop", to be more definitive).

webrtc.html
```diff
@@ -6082,6 +6082,55 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                </li>
+                <li> 
+                  <p><var>receiver</var> will no longer receive media.</p>
+                </li>
+                <li> 
+                  <p><code><var>receiver</var>.track</code> is ended.</p>
```
taylor-b: Could this also be rephrased, as "Set receiver.track to ended", or just "End receiver.track"? Also, could link to https://www.w3.org/TR/mediacapture-streams/#track-ended, which has steps to perform when a track ends.

webrtc.html
```diff
@@ -6082,6 +6082,55 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                  <p>Let <var>sender</var> be <code><var>transceiver</var>.sender</code>.</p>
+                </li>
+                <li>
+                  <p>Let <var>receiver</var> be <code><var>transceiver</var>.receiver</code>.</p>
+                </li>
+                <li> 
+                  <p><var>sender</var> will no longer send media.</p>
+                </li>
+                <li> 
+                  <p><var>receiver</var> will no longer receive media.</p>
```
taylor-b: Conceptually, what object sends RTCP packets? The RtpSender or RtpReceiver? If RTCP is sent by an RTPReceiver and received by an RTPSender, then this should be worded such that a stopped transceiver stops sending RTCP.

Issue #903, opened 2016-11-02
Capitalization of RTCDTMFSender


fluffy:
DTMF may have been an acronym 50 years ago, but now it is more or less an word like ICE, or RTP. I'd suggest we change this to RTCDtmfSender. (and related changes for other uses of DTMF. ) Note it seems to me that the Initialisms principal applies to the RTC not the Dtmf.

martinthomson:
What is the effect on implementations that have already shipped the API?

fluffy:
I'm sure the clever implementors over there could figure out some way to make both work :-) And it would make this consistent with ORTC.

aboba:
Related issue: https://github.com/w3c/ortc/issues/601

martinthomson:
Yeah, I've done the aliasing thing before, it sucks but it's possible.

alvestrand:
Resolution with advice from the TAG: RTCDTMFSender is the correct casing.

PR #905, opened 2016-11-02
Capitalization of RTCDTMFSender

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/903

Related Issue: https://github.com/w3c/ortc/issues/601


alvestrand:
Suggest we don't do this, given strong pushback from Domenic/TAG on acronym casing even when they run together.

alvestrand:
Let's not.

PR #913, opened 2016-11-03
Revise the rules of the negotiation-needed flag for transceivers.

taylor-b:
Addresses issue #803.

Besides just revising the negotiation-needed sections to deal with
transceivers instead of tracks, this commit makes the following
improvements relevant to the negotiation-needed flag:

- Use an internal slot to represent the flag.
- Only update the flag when applying an answer, or modifying the
  PeerConnection in the "stable" signaling state. *Not* when creating an
  answer.
- Instead of directly firing the negotiationneeded event when applying an
  answer, queue a task to check the flag again and then fire it (which
  is how it's normally fired).
- More thoroughly specify the steps for updating the negotiation-needed
  flag, and the criteria for determining if negotiation is needed.
- Reference these steps in every method where the negotiation-needed
  flag may possibly be updated.
- Remove text talking about RtpReceivers allowing the application to
  control the receipt of media, requiring negotiation. There's no method
  on RtpReceiver that does anything but provide information to the
  application.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -519,6 +519,9 @@
             internal slot, initialized to <code>false</code>.</p>
           </li>
           <li>
+            <p>Let <var>connection</var> have a [[<dfn>needNegotiation</dfn>]]
+            internal slot, initialized to <code>false</code>.</p></li>
+          <li>
             <p>Let <var>connection</var> have an [[<dfn>operations</dfn>]]
             internal slot, representing an <a>operations queue</a>, initialized
             to an empty list.</p>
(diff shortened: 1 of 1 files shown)


taylor-b:
By the way, I called the slot "[[needNegotiation]]" because if I call it "negotiationNeeded" I get a ReSpec "duplicate definition" error due to the event with a matching name. I couldn't figure out a way around this; "data-dfn-for" only works for IDL definitions. If anyone has another idea, or a suggestion for a better name, I'll take it.

PR #919, opened 2016-11-04
Removing incorrect statement related to IP leaking issue.

taylor-b:
The incorrect statement is "Applications can avoid exposing IP
addresses... by not signalling non-relay ICE candidates."

However, simply not signaling candidates is not enough to prevent IP
addresses from leaking. These candidates will still be gathered and used
to form candidate pairs, which means connectivity checks will be sent
leaking the server reflexive address for every network interface.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3367,9 +3367,7 @@ interface RTCIceCandidate {
                   communicating party, either temporarily or permanently, by
                   forcing the <a>ICE Agent</a> to report only relay candidates
                   via the <code>iceTransportPolicy</code> member of
-                  <code><a>RTCConfiguration</a></code>, or by not signalling
-                  non-relay ICE candidates (e.g. until the user has accepted to
-                  share media).</p>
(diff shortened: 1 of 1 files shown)


aboba:
LGTM.

aboba:
@dontcallmedom  This tiny PR is failing due to a failure in travis-ci due to yarn not being installed: 
   cd node_modules/highlight.js/ && yarn && node ./tools/build.js -n xml javascript css http markdown xquery && cd ../../

dontcallmedom:
@aboba indeed, that's due to a recent change in respec :( I suggest ignoring Travis CI errors while we figure this out

@vivienlacourba could you look into this? I think the best option at this point is to force our CI infrastructure to the last version of ReSpec not using yarn (until https://github.com/travis-ci/apt-source-whitelist/pull/322 gets deployed)

vivienlacourba:
@aboba @dontcallmedom this issue has been fixed in #935

PR #928, opened 2016-11-08
Fixing typos in "create an RTCRtpReceiver"

youennf:


youennf:

Issue #897, opened 2016-10-31
Remove RTCSessionDescription constructor or throw error on missing sdp


jan-ivar:
The spec says of [sdp](http://w3c.github.io/webrtc-pc/#dom-rtcsessiondescriptioninit-sdp): *"if type is rollback, this member can be left undefined."* but no processing steps to say what happens if it's undefined for other types. E.g. What does this output?

    console.log(new RTCSessionDescription({type: "answer"}).sdp);

Alternatively, we could deprecate the constructor since it is no longer needed for anything.

Omitting sdp in the set description processing steps already yields `InvalidModificationError`, so I believe we are fine there.

bzbarsky:
> Alternatively, we could deprecate the constructor since it is no longer needed for anything.

Unless "deprecate" means "remove" here, the spec really needs to define its behavior.

foolip:
Don't any problems with this constructor also apply to any context that takes a RTCSessionDescriptionInit dictionary?

https://bigquery.cloud.google.com:443/savedquery/762219082167:4d698184e35549c2accc9fbe6ded90af has 1642 hits, so unless it's a maintenance burden it seems best to just keep it around. The interface itself is used after all, having a constructor is about as harmless (and useless) as many event constructors.

foolip:
(The BigQuery link just shows the query, the results aren't saved. If anyone is interested I can export them.)

jan-ivar:
> Don't any problems with this constructor also apply to any context that takes a RTCSessionDescriptionInit dictionary?

@foolip No because, http://w3c.github.io/webrtc-pc/#set-description already says *"If elements of the SDP were modified in an invalid way as specified in [JSEP] (section 6.), then reject p with an InvalidModificationError and abort these steps."*

foolip:
OK, looked a bit more. AFAICT, in spec and in Blink, a script-created RTCSessionDescription can't really be used for anything, except to pass it to a context expecting a RTCSessionDescriptionInit and depend on that error handling.

So, is it a problem if it's possible to create nonsensical RTCSessionDescription instances? Many event interfaces are like this, where scripts can create combinations of things that are otherwise impossible, but it doesn't actually do anything.

Of course, it might also be an option to throw in the constructor for certain combinations, if use counters or httparchive analysis show them to be very, very uncommon.

jan-ivar:
[...]

PR #929, opened 2016-11-09
Have RTCSessionDescription's sdp member default to ""

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/897.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3017,14 +3017,14 @@ interface RTCPeerConnection : EventTarget  {
         <h4>RTCSessionDescriptionCallback</h4>
         <div>
           <pre class="idl">
-          callback RTCSessionDescriptionCallback = void (RTCSessionDescription sdp);</pre>
+          callback RTCSessionDescriptionCallback = void (RTCSessionDescriptionInit description);</pre>
           <section>
             <h2>Callback <a class="idlType">RTCSessionDescriptionCallback</a>
             Parameters</h2>
             <dl data-link-for="RTCSessionDescriptionCallback" data-dfn-for=
             "RTCSessionDescriptionCallback" class="callback-members">
-              <dt><code>sdp</code> of type <span class=
-              "idlMemberType"><a>RTCSessionDescription</a></span></dt>
+              <dt><code>description</code> of type <span class=
+              "idlMemberType"><a>RTCSessionDescriptionInit</a></span></dt>
               <dd>The object containing the SDP [[!SDP]].</dd>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3017,14 +3017,14 @@
         <h4>RTCSessionDescriptionCallback</h4>
         <div>
           <pre class="idl">
-          callback RTCSessionDescriptionCallback = void (RTCSessionDescription sdp);</pre>
+          callback RTCSessionDescriptionCallback = void (RTCSessionDescriptionInit description);</pre>
```
foolip: Can you mention this change in the commit message?

foolip: Oh, there are two commits, I assumed the description was the commit message. Never mind then, as long as it's landed as two commits or one can otherwise figure out from the resulting commit why this was done.

jan-ivar:
Travis seems sad for unrelated reasons here.

webrtc.html
```diff
@@ -3017,14 +3017,14 @@
...
           <section>
             <h2>Callback <a class="idlType">RTCSessionDescriptionCallback</a>
             Parameters</h2>
             <dl data-link-for="RTCSessionDescriptionCallback" data-dfn-for=
             "RTCSessionDescriptionCallback" class="callback-members">
-              <dt><code>sdp</code> of type <span class=
-              "idlMemberType"><a>RTCSessionDescription</a></span></dt>
+              <dt><code>description</code> of type <span class=
+              "idlMemberType"><a>RTCSessionDescriptionInit</a></span></dt>
```
alvestrand: @foolip, is this a non-breaking change for user code?

foolip: It can certainly be observed, by `desc instanceof RTCSessionDescription` or trying to set `type` or `sdp` to something other than a string. The question is whether it's likely to break stuff. Unclear.

There seems to be a bunch of related changes here, and IMHO someone needs to have a look at them as a whole and see if they're likely to work out or not, adding use counters. What I did in https://codereview.chromium.org/2490543002 doesn't cover it all I think.

jan-ivar:
I've reverted the controversial part so we at least can merge the `sdp = ""` default.

The controversial part really belongs in (is an alternative fix to) https://github.com/w3c/webrtc-pc/issues/933 anyway.

jan-ivar:
Can we merge this now it's just `sdp = ""` again? I'm trying to land [this](https://bugzil.la/1313966) for Firefox.

Issue #933, opened 2016-11-09
RTCSessionDescription readonly members is not webcompat


jan-ivar:
Brought up by @foolip in https://github.com/w3c/webrtc-pc/issues/573#issuecomment-259103219:

> https://crbug.com/662898 is a bug originating from @philn with Igalia where they (I presume) they tried to follow the spec but ran into trouble.

There's some discussion on that pre-existing closed issue, but better to continue in a new issue.

I also have a PR proposal.

alvestrand:
I think this is solved. If not, please reopen.

Issue #573, opened 2016-04-11
RTCSessionDescription parameters should be readonly.


jan-ivar:
The [current language](http://w3c.github.io/webrtc-pc/#idl-def-RTCSessionDescription) implies that content can modify the local description like this:

    pc.localDescription.type = "answer";

which would be a travesty (an interface as an attribute is by reference).

In light of https://github.com/w3c/webrtc-pc/pull/302 I propose we solve this by changing `RTCSessionDescription`'s parameters to `readonly`, and change `createOffer` and `createAnswer` to resolve with `RTCSessionDescriptionInit` instead.

alvestrand:
"which would be a travesty (an interface as an attribute is by reference)."
I can't parse either the grammar of the sentence, what your opinion is, or why you have it.

The localDescription and remoteDescription parameters of the PC are already readonly, which makes the code you suggest illegal (unless I've misunderstood the properties of a readonly attribute).

4.7.2 says "Attributes on this interface are mutable for legacy reasons." I see no compelling reason to change it.

adam-be:
The type attribute was readonly to begin with, but we changed it to allow scripts to, for example, switch from an "answer" to a "pranswer" before setting.

jan-ivar:
I checked with our DOM guys, and `readonly` just means there's no setter for `.localDescription`. The returned interface is still mutable, which seems problematic.

If we do what I suggest then there's no reason to leave them for legacy reasons, since createOffer would return `RTCSessionDescriptionInit`, a dictionary, instead.

jan-ivar:
Also, dictionaries have copy semantics, whereas interfaces do not. So a conforming implementation of the current webidl MUST work like this:

```js
var pc = new RTCPeerConnection();
pc.createDataChannel("dummy");
pc.createOffer().then(offer => pc.setLocalDescription(offer))
.then(() => {
  console.log(pc.localDescription.type); // offer
  pc.localDescription.type = "answer";
  console.log(pc.localDescription.type); // answer
})
.catch(log);
```
Luckily, neither Firefox nor Chrome are compliant.

martinthomson:
You can't have an attribute with a dictionary type.  The right solution is to make `RTCSessionDescription` immutable itself, as @jan-ivar says originally.  That we are all current in flagrant defiance of expectations here isn't that bad, we just make the attributes `readonly` and move on.

jan-ivar:
[...]

PR #934, opened 2016-11-09
Add RTCSessionDescriptionReadOnly to make RTCSessionDescription membe…

jan-ivar:
…rs writable again for webcompat.

Fix for https://github.com/w3c/webrtc-pc/issues/933.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1021,13 +1021,13 @@ interface RTCPeerConnection : EventTarget  {
     Promise&lt;RTCSessionDescriptionInit&gt; createOffer (optional RTCOfferOptions options);
     Promise&lt;RTCSessionDescriptionInit&gt; createAnswer (optional RTCAnswerOptions options);
     Promise&lt;void&gt;                      setLocalDescription (RTCSessionDescriptionInit description);
-    readonly        attribute RTCReadonlySessionDescription? localDescription;
-    readonly        attribute RTCReadonlySessionDescription? currentLocalDescription;
-    readonly        attribute RTCReadonlySessionDescription? pendingLocalDescription;
+    readonly        attribute RTCSessionDescriptionReadOnly? localDescription;
+    readonly        attribute RTCSessionDescriptionReadOnly? currentLocalDescription;
+    readonly        attribute RTCSessionDescriptionReadOnly? pendingLocalDescription;
     Promise&lt;void&gt;                      setRemoteDescription (RTCSessionDescriptionInit description);
-    readonly        attribute RTCReadonlySessionDescription? remoteDescription;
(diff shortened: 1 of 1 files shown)


foolip:
The pattern I've seen in https://drafts.fxtf.org/geometry/ and some other places is InterfaceNameReadOnly.

How certain can we be that no content depends on being able to change type or sdp on localDescription, remoteDescription and similar? Per https://github.com/w3c/webrtc-pc/issues/573#issuecomment-259206448 it sounds like changing them does nothing in Edge, is that also true in Gecko?

jan-ivar:
Certain.

Firefox and Chrome return no-op copies too (`pc.localDescription == pc.localDescription` is false).

Setting these is what the methods `setLocalDescription` and `setRemoteDescription` are for.

alvestrand:
This is a breaking change for all code that uses the attribute names. We need to argue that the benefit outweights the cost.

foolip:
> Firefox and Chrome return no-op copies too (`pc.localDescription == pc.localDescription` is false).

Oh, that's too bad. That doesn't seem per spec, do you intent to change the spec or hope to match it? I assume match it, since otherwise it wouldn't be useful to make the return value readonly?

jan-ivar:
@foolip I have a patch for Firefox that fixes `pc.localDescription == pc.localDescription`.

@alvestrand What do you mean "code that uses the attribute names"? Can you give an example?

alvestrand:
the other one looks better.

jan-ivar:
@alvestrand Note that https://github.com/w3c/webrtc-pc/issues/949 only fixes the legacy callbacks, which was a regression, so thanks for merging it, just note it doesn't in itself address https://github.com/w3c/webrtc-pc/issues/933 which was about someone doing this:

```js
({type, offer}) => {
  var desc = new RTCSessionDescription({type});
  desc.sdp = sdp; // now silently ignored
  var desc = pc.setLocalDescription(desc);
}
```

rather than

```js
({type, offer}) => pc.setLocalDescription(new RTCSessionDescription({type, offer}))
```

Maybe no-one does that, but I could see it causing pain if someone modified sdp this way, and everything worked except the modification didn't take. But perhaps this breakage pales in comparison to the spec apparently deciding to no longer allowing sdp modifications of any kind, so I'm fine either way, just so we're on the same page.

alvestrand:
This particular example wouldn't work anyway, due to the "unchanged string" requirement, but it's relevant for "pc.setRemoteDescription".

Note - setLocalDescription() takes an RTCSessionDescriptionInit. I assume that the serializer in RTCSessionDescription means that it autoconverts to an RTCSessionDescriptionInit-compatible dictionary when passed to setLocalDescription - otherwise we'd already have broken the web.

"pc.setRemoteDescription({type, offer})" should Just Work, right?

Issue #908, opened 2016-11-03
Need to replace references to "stopped sender" with "recvonly/inactive transceiver"


taylor-b:
We used to have the concept of stopped/unstopped senders and receivers, which determined the direction of an m= section. But now we have the `RTCRtpTransceiver.setDirection` API for that. So I think the following changes are necessary:

- Remove the paragraph "An RTCRtpSender can be stopped, which indicates that it will no longer send any media.", or rewrite it in terms of transceiver direction.
- In the `removeTrack` steps:
  - Add a step for changing the transceiver's direction
  - Remove or rephrase the "Stop sender" step (which has a dead link to "Stop").

aboba:
@taylor-b  We can probably remove the paragraph starting with "An RTCRtpSender can be stopped" without losing anything.  

The description of transceiver.sender.removeTrack seems quite similar to that of transceiver.setDirection("inactive") with the exception that transceiver.setDirection would also affect transceiver.receiver.  Do you see any problem with removeTrack also affecting the receiver?

With respect to the now outmoded concept of stopped/unstopped senders and receivers, there is an  odd bit of text in the description of transceiver.stopped: 

stopped of type boolean, readonly
The stopped attribute ... is true if either stop has been called or if setting the local or remote description has caused the RTCRtpReceiver to be stopped.

Could we change this to "is true if stop has been called" and delete the rest?  I think this text was originally meant to describe how the effect of removeTrack would propagate to the remote peer, but reformulating things in terms of setDirection should take care of that.

taylor-b:
> Do you see any problem with removeTrack also affecting the receiver?

I don't think that should be possible, since it only modifies the "send" bit of the direction.

> Could we change this to "is true if stop has been called" and delete the rest?

No, because a transceiver being stopped means "transceiver will permanently stop sending/receiving RTP/RTCP packets". Even if the application didn't call "stop", if a remote offer or answer rejected the m= section, the transceiver will be stopped and there's nothing that can be done but create a new one.

Though this text *does* need to be changed to say RTCRtp**Trans**ceiver instead of RTCRtp**Re**ceiver...

alvestrand:
Note: I think a sender should be stoppable.
The chief role of the transceiver (to my mind) is to store the per-m-line information that came in on the previous SDP offer/answer, apply the needed changes to senders & receivers when that happens, and return the (possibly modified) information needed when createOffer/createAnswer is called.

[...]

PR #939, opened 2016-11-11
Remove "stopped" from removeTrack

aboba:
Partial fix for Issue  https://github.com/w3c/webrtc-pc/issues/908

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4551,14 +4551,14 @@ interface RTCPeerConnectionIceErrorEvent : Event {
             </dd>
             <dt><code>removeTrack</code></dt>
             <dd>
-              <p>Stops sending media from <var>sender</var>, pending
-              negotiation. The <code><a>RTCRtpSender</a></code>
-              will still appear in <code>getSenders</code>. Doing so
-              will cause future calls to <code>createOffer</code> to
-              mark the <a>media description</a> for the corresponding
-              transceiver as <code>recvonly</code> or
-              <code>inactive</code>, as defined in <span data-jsep=
-              "subsequent-offers">[[!JSEP]]</span>. <!-- onended --></p>
+              <p>Stops sending media from <var>sender</var>. The
+              <code><a>RTCRtpSender</a></code> will still appear
+              in <code>getSenders</code>. Doing so will cause future
+              calls to <code>createOffer</code> to mark the
+              <a>media description</a> for the corresponding transceiver
+              as <code>recvonly</code> or <code>inactive</code>,
+              as defined in <span data-jsep="subsequent-offers">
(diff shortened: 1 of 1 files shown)


aboba:
@taylor-b Can you review?

webrtc.html
```diff
@@ -4551,11 +4551,12 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             </dd>
             <dt><code>removeTrack</code></dt>
             <dd>
-              <p>Stops sending media from <var>sender</var>. The
-              <code><a>RTCRtpSender</a></code> will still appear in
-              <code>getSenders</code>. Doing so will cause future calls to
-              <code>createOffer</code> to mark the <a>media description</a> for
-              the corresponding transceiver as <code>recvonly</code> or
+              <p>Stops sending media from <var>sender</var>, pending
+              negotiation. The <code><a>RTCRtpSender</a></code>
```
taylor-b: At the virtual interim, was there consensus about the "pending negotiation" part? It seems like it would be surprising if an application calls "removeTrack" but the track is still being sent, and `sender.track` never changes.

Should this method set `sender.track` to `null` (and immediately stop sending it), in addition to what you already have?

webrtc.html
```diff
@@ -4592,11 +4596,24 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                  <code><a>RTCRtpTransceiver</a></code> object corresponding
+                  to <var>sender</var>.</p>
+                </li>
+                  <p>If <code><var>transceiver</var>.currentDirection</code>
+                  is <code>recvonly</code> or <code>inactive</code>, then
+                  abort these steps.</p>
+                <li>
+                  <p>If <code><var>transceiver</var>.currentDirection</code>
+                  is <code>sendrecv</code> set
+                  <code><var>transceiver</var>.direction</code> to
```
taylor-b: This should point to the `setDirection` steps you recently added,

aboba: Agreed.  I think we want to wait for the "current direction" PR though.

webrtc.html
taylor-b: I agree that `transceiver.stop`, `encodings[].active` and `setDirection` aren't solutions for this. But wouldn't disassociating the track from the sender work? If you call `addTransceiver`, you end up with a sender with a `null` track. So `removeTrack` could just return the sender to this state (of "can send, but have nothing to send"). Maybe this needs some list discussion. If we went this route it would also make sense to allow passing `null` into `replaceTrack`.

aboba: @taylor-b Worked on cleaning this up by setting track to null.  Can you check that it is fixed?

PR #949, opened 2016-11-17
Give legacy callbacks RTCSessionDescriptionInit so they can modify sdp

jan-ivar:
Alternative fix for https://github.com/w3c/webrtc-pc/issues/933.

If we don't take https://github.com/w3c/webrtc-pc/pull/934 then we need this IMHO.

Looking for swift resolution either way, as I'm trying to remove [the need for the constructors in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1263312#c17).


alvestrand:
Can you say if this wil require any changes in users' code, and if so what?

Offhand, I like this much better than the "access control in name of type" solution.

Issue #955, opened 2016-11-29
API for setting QoS priority of data channels


aboba:
On 28 Nov 2016, at 22:50, Peter Thatcher <pthatcher@google.com> wrote:

Looking at draft-ietf-tsvwg-rtcweb-qos, it's clear we want to apply the very-low/low/medium/high priority to data channels.  But we don't have any way in the WebRTC API, as far as I can tell, to set that priority. 

Is this just something we forgot to add?  If so, I propose that we add it as follows:

dictionary RTCDataChannelInit {
  ... 
  RTCPriorityType priority;
}

And it's meaning is defined by draft-ietf-tsvwg-rtcweb-qos.

PR #956, opened 2016-11-29
API for setting QoS priority of data channels

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/955

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7762,6 +7762,7 @@ interface RTCTrackEvent : Event {
              USVString      protocol = "";
              boolean        negotiated = false;
              unsigned short id;
+             RTCPriorityType priority;
 };</pre>
         <section>
           <h2>Dictionary <a class="idlType">RTCDataChannelInit</a> Members</h2>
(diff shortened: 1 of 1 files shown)


aboba:
@pthatcherg Can you review?

webrtc.html
```diff
@@ -7762,6 +7778,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
              USVString      protocol = "";
              boolean        negotiated = false;
              unsigned short id;
+             RTCPriorityType priority;
```
alvestrand: If we add a default value of "low" here, we don't have to special-case the missing attribute in the handling instructions above.

RTCPriorityType priority = "low";

(Note: I can't tell what the default for RTP channels is from the document. "low" sounds right, but consistency is important.)

alvestrand:
Looks good to me (modulo the issue of how to specify default priority).

aboba:
@alvestrand Added the default priority of "low".

Issue #965, opened 2016-12-07
How should sender.get/setParameters work on a stopped sender?


jan-ivar:
Should it throw `InvalidStateError` like sender.replaceTrack or be a no-op/return last set value?

aboba:
@jan-ivar  How about this? 

setParameters rejects the promise with InvalidStateError if transceiver.stopped is true.  
If sender.track is null or the value of transceiver.currentDirection is recvonly, the sender will not send, but I'd suggest this not cause the promise not be rejected.

alvestrand:
Since getParameters doesn't need to be special-cased, and pc.stop() stops all transceivers, the merge of #966 seems to resolve this issue.

PR #966, opened 2016-12-08
setParameters when transceiver.stopped is true

aboba:
Fix for https://github.com/w3c/webrtc-pc/issues/965


jan-ivar:
@aboba Thanks, looks good, except we typically fail with `InvalidStateError` on the main thread, not in parallel. Should we be consistent?

The difference is whether the promise can be immediately observed as rejected rather than pending after stepping over the method call in a JS debugger - a useful expectation about errors detectable synchronously. There's a micro-task ordering difference as well.

PR #967, opened 2016-12-08
change titles for getStats and identity samples

fippo:
changes the title of the getStats and identity samples section
to generate nicer link anchors than "#sample" and "#samples".


fippo:
hrm. Seems like @jan-ivar links to #sample in an error message about non-maplike getStats usage :-/

jan-ivar:
@fippo thanks for the heads up; [fix](https://bugzil.la/1323095) coming up.

PR #968, opened 2016-12-08
Add ufrag to IceCandidate, and use IceCandidate for end-of-candidates.

### LINKED ISSUE #726: Add a ufrag attribute to the RTCIceCandidate structure to indicate which ICE context

taylor-b:
Since RTCIceCandidate can be used for end-of-candidates, all the
attributes besides "ufrag" now become nullable.

Rebase of PR #757. Fixes #726.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1726,10 +1726,11 @@ interface RTCPeerConnection : EventTarget  {
                 <code>candidate</code>. The only members of the argument used
                 by this method are <code><a data-for=
                 "RTCIceCandidate">candidate</a></code>, <code><a data-for=
-                "RTCIceCandidate">sdpMid</a></code> and <code><a data-for=
-                "RTCIceCandidate">sdpMLineIndex</a></code>; the rest are
-                ignored. When the method is invoked, the User Agent MUST run
-                the following steps:</p>
+                "RTCIceCandidate">sdpMid</a></code>, <code><a data-for=
+                "RTCIceCandidate">sdpMLineIndex</a></code>, and
+                <code><a data-for="RTCIceCandidate">ufrag</a></code>; the rest
+                are ignored. When the method is invoked, the User Agent MUST
+                run the following steps:</p>
                 <ol>
                   <li>
                     <p>Let <var>candidate</var> be the methods argument.</p>
@@ -1758,28 +1759,6 @@ interface RTCPeerConnection : EventTarget  {
                       </li>
                       <li>
                         <p>Let <var>p</var> be a new promise.</p>
-                      </li>
-                      <li>
-                        <p>If <var>candidate</var> is <code>null</code>, the
-                        User Agent MUST queue a task that runs the following
-                        steps, and abort these steps:</p>
-                        <ol>
-                          <li>
-                            <p>If <var>connection</var>'s [[<a>isClosed</a>]]
-                            slot is <code>true</code>, abort these steps.</p>
-                          </li>
-                          <li>
-                            <p>For each <a>media description</a> in the last
-                            successfully applied remote description, perform
-                            the processing for an end-of-candidates indication
-                            for said <a>media description</a> as defined in
-                            [[TRICKLE-ICE]].</p>
-                          </li>
-                          <li>
-                            <p>Resolve <var>p</var> with
-                            <code>undefined</code>.</p>
-                          </li>
-                        </ol>
                       </li>
                       <li>
                         <p>If <var>candidate.sdpMid</var> is not null, run the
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1816,9 +1795,20 @@
                         </ol>
                       </li>
                       <li>
+                        <p>If <code><var>candidate</var>.ufrag</code> is not
+                        equal to any ufrag present in the corresponding
+                        <a>media description</a> of an applied remote
+                        description, reject <var>p</var> with an
```
alvestrand: Have to specify whether this is current description, pending description, or both. This matters in case of rollback (the issue Peter raised).

webrtc.html
```diff
@@ -6620,6 +6619,45 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+          <p>Fire an event named <code><a>icecandidate</a></code> with
+          <var>newCandidate</var> at <var>connection</var>.</p>
+        </li>
+      </ol>
+      <div class="issue">
+        ICEbis doesn't define what a "generation of candidates" is, though this
+        terminology is used in places like [[TRICKLE-ICE]]. Once a definition is
+        added we should reference it here.
+      </div>
+
```
alvestrand: This can be seen as triggered on transport.icegatheringstate change, right? I think we should describe it as such, rather than "when it's finished", since that decides which comes first - the event or the state change.

If we add a trigger on pc.icegatheringsummarystate that fires an empty event (no candidate at all), we don't even have to ask implementations to change (assuming they don't barf on the empty candidate).

webrtc.html
taylor-b: Hmm. It could definitely be either the pending/current description (rollback is the main reason we need to differentiate with the ufrag in the first place). But could it be anything more than the pending/current description? Is it expected that after applying an answer and the ICE restart is accepted, the old ICE generation immediately aborts? If so we should specify that in JSEP.

webrtc.html
alvestrand: JS doesn't admit to true simultaneous events. We may want to describe this as a handler for the internal "all candidates are gathered" event, which may or may not trigger a state change.
People will ask which event comes first, so let's just order them in one or the other order.

taylor-b: Alright. So we can structure this as two internal events: "gathering started for *ufrag*" (which may change the gathering state), and "gathering finished for *ufrag*" (which will fire this special candidate, and may change the gathering state, in a specified order).

taylor-b:
I replaced "null" with "empty string" and addressed Harald's concern about ordering between events (as currently written, the "end-of-candidates" indication occurs before the gathering state change).

webrtc.html
```diff
@@ -3454,51 +3447,56 @@
...
             </div>
           </section>
         </div>
         <div>
           <pre class="idl">dictionary RTCIceCandidateInit {
-    required DOMString       candidate;
+             DOMString       candidate;
              DOMString?      sdpMid = null;
              unsigned short? sdpMLineIndex = null;
+    required DOMString       ufrag;
```
foolip: This would make all existing uses of the `RTCIceCandidate` constructor or `addIceCandidate` throw exceptions. These changes could use review by a likely future implementer.

webrtc.html
```diff
@@ -3454,51 +3447,56 @@
...
               entries for the following attributes: candidate, sdpMid,
-              sdpMLineIndex.</p>
+              sdpMLineIndex, ufrag.</p>
             </div>
           </section>
         </div>
         <div>
           <pre class="idl">dictionary RTCIceCandidateInit {
-    required DOMString       candidate;
+             DOMString       candidate;
```
foolip: Should this have the default value ""? This way `addIceCandidate(null)` will be equivalent to `addIceCandidate({ candidate: "" })`. With no default value, some prose would have to say what a missing `candidate` member means,

webrtc.html
alvestrand: I think that saying that the ufrag MUST be present in at least one of the current and pending remote descriptions is fine.

taylor-b: Ok. Then the logical conclusion is that you can have at most 2 ICE generations at the same time (during the time between when the offer is sent and the answer accepts the ICE restart). Then when the answer is applied, the old ICE checklists can be discarded. Does this need to be detailed in draft-ietf-mmusic-ice-sip-sdp? I initially was thinking JSEP, but you're allowed to reject an offer even with normal SDP offer/answer semantics, so this seems like a general "ICE SDP" thing.

aboba:
LGTM.

PR #969, opened 2016-12-14
Make candidate argument in RTCPeerConnection#addIceCandidate non-nullable

LoonyBean:
It makes no sense to add a null object as an IceCandidate

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -965,7 +965,7 @@ interface RTCPeerConnection : EventTarget  {
     readonly        attribute RTCSessionDescription?    remoteDescription;
     readonly        attribute RTCSessionDescription?    currentRemoteDescription;
     readonly        attribute RTCSessionDescription?    pendingRemoteDescription;
-    Promise&lt;void&gt;                      addIceCandidate ((RTCIceCandidateInit or RTCIceCandidate)? candidate);
+    Promise&lt;void&gt;                      addIceCandidate ((RTCIceCandidateInit or RTCIceCandidate) candidate);
     readonly        attribute RTCSignalingState         signalingState;
     readonly        attribute RTCIceGatheringState      iceGatheringState;
     readonly        attribute RTCIceConnectionState     iceConnectionState;


alvestrand:
This option was added to mirror the "end of candidates = null candidate" from onicecandidate.
This allows doing "take whatever comes out of the icecandidate event and pass it into the other end's addIceCandidate, without checking if it's null or not", which apparently some people have been doing.

Apparently no browser treats the null end-of-candidate marker specially, and we've got a PR going in that doesn't use null candidates any more (instead it uses a candidate with a null candidate member), so the ability to null might not be relevant any more.

@taylor-b could comment further on what he thinks the status is; @pthatcherg is likely interested.

alvestrand:
#968 is the relevant change to candidates.

taylor-b:
Yes, PR #968 will make the candidate argument non-nullable again.

foolip:
The trouble here is not only that `pc.addIceCandidate(null)` seems like a silly thing, but that the `(RTCIceCandidateInit or RTCIceCandidate)` union type can't be nullable per WebIDL:
https://heycam.github.io/webidl/#idl-nullable-type

The things that can't be nullable include "a union type that itself has includes a nullable type or has a dictionary or record type as one of its flattened member types."

This in turn, is because for any dictionary, passing null or undefined behaves the same as `{}`, so making it nullable is in a sense redundant.

Since `RTCIceCandidateInit` has a required member, `pc.addIceCandidate({})` and thus `pc.addIceCandidate(null)` will throw TypeError.

Will https://github.com/w3c/webrtc-pc/pull/968 be merged soonish, or can we make this fix first?

alvestrand:
@foolip is what you are saying that you don't need to make the candidate nullable to make null legal?
In that case, the change seems completely logical. It's adding required members ot RTCIceCandidateInit that will be a breaking change for people using addIceCandidate(null).

Right?

alvestrand:
@LoonyBean ash-nazg doesn't like you (yet). Can you join the W3C WebRTC WG so that the chain of contribution agreements is complete? (Remember to use your google email for that)

foolip:
@alvestrand, right, because one of the union types is a dictionary, null already means something. It's just that it means `{}` which will throw an exception with the current definition of `RTCIceCandidateInit` so perhaps coupling it with the changes in https://github.com/w3c/webrtc-pc/pull/968 is better if that happens before long.

Is there no way to make PRs for WebRTC WG specs without joining? I did join for this purpose, but it got me subscribed to some lists too, and I wouldn't recommend it for one-off changes.

alvestrand:

alvestrand:
I (or you, I think) can mark the contribution as non-substantive. But I have to do that for every CL, which is why I don't like it much. One-off: marked it for this one.

So we'll merge this after #968.

foolip:
Thanks! Since #968 makes the same change, this'll probably end up just being closed, but I suppose leave it open in case further changes to #968 undoes this particular change.

dontcallmedom:
@foolip re not having to join the WG - the IPR checker now allows to bless contributions from people who are identified as working group for an organization who joined the group (without requiring individuals to join the group).

alvestrand:
@dontcallmedom is that done by clicking the "is a member" function in the ash-nazg API?
I couldn't find a way for me to list contributors on this one after I marked it "trivial", so I couldn't try clicking lunalu's entry to see if I could add her to the webrtc wg. Still a few details in ash-nazg's UI to resolve...

dontcallmedom:
@alvestrand the association between github username and affiliation is something that requires admin rights, so the best is to ask me at the moment. This hopefully will improve over time, but there are several things that need to happen before I can add it to ash-nazg.

(and yes, there are quite a few things that need improvement in ash-nazg :)

FWIW, it wasn't clear from @LoonyBean 's profile that she was working for Google, and no matter what, I can only fix the affiliation of contributors that have a W3C account (which Luna doesn't afaict).

aboba:
With the merger of PR https://github.com/w3c/webrtc-pc/pull/968 I believe this PR is obsolete.

PR #974, opened 2016-12-19
fix name of html2text in makefile

fluffy:


burnburn:
@fluffy before I merge this, just a quick confirmation that html2text is one of your tools, right?

fluffy:
Its a python package (https://pypi.python.org/pypi/html2text/2016.9.19) but I think I am probably the only person that uses it and the names changed from from html2text.py to html2text long ago.

burnburn:
Thanks.  I just meant that it is not something I get when I clone the repository, so it is pulled from somewhere else (by you).

Issue #983, opened 2017-01-02
RTCIceCandidateInit's required DOMString ufrag will throw for all existing content


foolip:
This issue was raised in https://github.com/w3c/webrtc-pc/pull/968#pullrequestreview-13093177 but not resolved before merging.

Perhaps @henbos or @jan-ivar can verify, but it looks to me like it's a bad idea to attempt implementing the current spec, as any existing use of the RTCIceCandidateInit dictionary will start throwing TypeError.

jan-ivar:
I agree with @foolip. Most scripts still use the explicit `RTCIceCandidate` constructor since most browsers still require it (few know adapter.js removes the need for it).

Even scripts that stringify whatever comes out of `onicecandidate` would throw in a connection between an old and a new browser, due to signaling. E.g. between Chrome and Firefox, unless we implemented ufrag at the same time and eradicate all old versions.

alvestrand:
@taylor-b can you write a PR that makes it not mandatory in IDL and instead specifies in text what to do if it's missing?
That seems the sensible solution.

taylor-b:
Sure. I still would find it helpful to know the general policy on backwards compatibility though (see my comment on PR #968).

alvestrand:
Solved by #987

foolip:
Thanks for fixing, @taylor-b!

> I still would find it helpful to know the general policy on backwards compatibility though

I can't speak for the WebRTC WG, but in all web standards work the approach I take is to ask "can this be implemented and shipped by everyone?" And to answer that question, one has to look into existing usage of an API, and whether some browsers might be constrained by web compat in ways that others are not, e.g. due to engine-specific code paths.

PR #987, opened 2017-01-05
Make the ufrag optional in RTCIceCandidateInit, for backwards compat.

taylor-b:
If it's missing, the candidate (or end-of-candidates indication) will be
applied for the most recent ICE generation/batch of candidates.

Fixes issue #983.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1807,16 +1807,20 @@ interface RTCPeerConnection : EventTarget  {
                         </ol>
                       </li>
                       <li>
-                        <p>If <code><var>candidate</var>.ufrag</code> is not
-                        equal to any ufrag present in the corresponding
-                        <a>media description</a> of an applied remote
-                        description, reject <var>p</var> with an
+                        <p>If <code><var>candidate</var>.ufrag</code> is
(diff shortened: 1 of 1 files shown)


alvestrand:
LGTM. Thanks!

webrtc.html
```diff
@@ -1807,16 +1807,20 @@
                         </ol>
                       </li>
                       <li>
-                        <p>If <code><var>candidate</var>.ufrag</code> is not
-                        equal to any ufrag present in the corresponding
-                        <a>media description</a> of an applied remote
-                        description, reject <var>p</var> with an
+                        <p>If <code><var>candidate</var>.ufrag</code> is
+                        non-null, and is not equal to any ufrag present in the
```
foolip: In IDL terms, the members of dictionaries are "missing" when they are not there. Since the `ufrag` dictionary member is of type DOMString and not nullable, must be a string if it is there at all.

foolip: @taylor-b, will you take care of it, or should I file an issue?

taylor-b: I will: https://github.com/w3c/webrtc-pc/pull/994

PR #989, opened 2017-01-08
Add dfn as anchor for interfaces, dictionaries, enums, ...

dontcallmedom:
Fix (most of) respec warnings

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -100,7 +100,7 @@
       <h3>Configuration</h3>
       <section>
         <h4>RTCConfiguration Dictionary</h4>
-        <p>The <code>RTCConfiguration</code> defines a set of parameters to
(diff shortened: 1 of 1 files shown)


alvestrand:
@dontcallmedom master is broken after merge :-(

dontcallmedom:
@alvestrand that was broken before I think

alvestrand:
@dontcallmedom maybe... but you find these things faster than me....

dontcallmedom:
looking at it quickly, this seems a respec bug - will look more into it

dontcallmedom:
so I have a work around in #992 - I think ReSpec doesn't expect enums to have upper case values (which indeed goes against what's recommended); I'm working on a patch for respec, but if #992 is acceptable, it's very likely to be much faster to fix Travis that way.

PR #990, opened 2017-01-11
Add an explicit stats selection algorithm.

alvestrand:
Fixes https://github.com/w3c/webrtc-stats/issues/116

This should match Edge's current implementation.

@ShijunS @jan-ivar @henbos please review and suggest additional objects to list.


webrtc.html
```diff
@@ -8852,6 +8854,34 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
     </section>
     <section>
+      <h3>The stats selection algorithm</h3>
+      <p>
+	The <dfn>stats selection algorithm</dfn> is as follows:
+	<ul>
+	  <li>If the selector is a MediaStreamTrack, return the following objects:
+	    <ul>
+	      <li>
+		The RTCMediaStreamStats corresponding to the selector.
```
jan-ivar: Shouldn't this be [RTCMediaStreamTrackStats](https://w3c.github.io/webrtc-stats/#mststats-dict*)?

webrtc.html
```diff
@@ -8852,6 +8854,34 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+	      </li>
+	      <li>
+		If the selector track is inbound, the RTCInboundRTPStreamStats corresponding to the selector.
```
jan-ivar: That's derived from [RTCRTPStreamStats](https://w3c.github.io/webrtc-stats/#dom-rtcrtpstreamstats) which has the following foreign-key references: `associateStatsId`, `mediaTrackId` and `transportId` today.

I think we need an algorithm that (recursively?) includes all of these stats objects as well. Do others feel differently?

webrtc.html
```diff
@@ -8852,6 +8854,34 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+	      </li>
+	      <li>
+		If the selector track is outbound, the RTCOutboundRTPStreamStats corresponding to the selector.
+	      </li>
+	      <li>
+		At the implementation's discretion, other stats that the
+		implementation considers relevant.
```
jan-ivar: This does not seem webcompat friendly. I'd replace this with the recursion rule above.

webrtc.html
henbos: I think so. But as long as we're returning more than track stats it makes sense to return the RTCMediaStreamStats too (stream stats whose trackIds contains the RTCMediaStreamTrackStats.id of the track).

webrtc.html
henbos: I don't think we should recursively explore the stats graph, that could end up pulling in a lot of stats. For performance and implementation simplicity I think the selector should filter aggressively, only pulling in specific things. It would be hard to find all relationships recursively without producing all the stats, and we want to apply the filter before collecting, not after (or else the caller might as well filter from a full getStats). If we want other things too we could think about other types of selectors and unions of selectors. If a large graph is desired, a full getStats should be made.

webrtc.html
```diff
@@ -8852,6 +8854,34 @@ <h2 id="sec.stats-model">Statistics Model</h2>
       [[!WEBRTC-STATS]].</p>
     </section>
     <section>
+      <h3>The stats selection algorithm</h3>
+      <p>
+	The <dfn>stats selection algorithm</dfn> is as follows:
+	<ul>
+	  <li>If the selector is a MediaStreamTrack, return the following objects:
```
henbos: Should this be "the RTCMediaStreamTrackStats whose trackIdentifier match the selector's id, and any RTCStats referencing that/those RTCMediaStreamTrackStats" or a list of specific stats (like you have now)?

The difference between that description and the list you described is you didn't include any RTCMediaStreamStats and there are stream stats with trackIds referencing RTCMediaStreamTrackStats. But these should be included?

alvestrand: I scratched my head about that. It seems that pulling in everything it references or that has references to (RTP, track) has two issues:
1) It gets a lot (pull in streams, you get references to the other tracks in the stream; should you pull those? pull in what RTP references, you get the transport, which gets the candidates..)
2) It isn't extension-friendly (add a new reference; do you pull what it points to or not?)

webrtc.html
alvestrand: Yes, (diffidently, if not differently). See comment below.

webrtc.html
alvestrand: Yes.

webrtc.html
henbos: Whatever you pull in you will risk (now, in the future or in an extension) getting stats objects with references to stats objects that are not in the returned set. I think we just have to accept that since recursively exploring the stats graph is IMO a mistake.

But you're right, if we make the rule about pulling in "any stats object referencing or referenced by the track stats" that might oblige us to pull in stuff we don't like if we update the spec with new references.

henbos: (By "any stats object referencing or referenced by the track stats" I meant direct references, not indirect (recursive) references.)

alvestrand: I checked - it seems that the only objects that have a reference to the MediaStreamTrack at the moment are the stream object and the RTPStream object. If we're OK with sending only those two objects, I think listing them explicitly is no great burden.

henbos: SGTM

webrtc.html
alvestrand: My reason for leaving it in is that we have a generic permission to add new stats (implementation specific or newly standardized), so preventing new stats from being added seems awkward.
If we want to prevent the addition of existing objects, such as candidates and transports, we should say so.
We could also replace this with something like:
NOTE: New versions of this specification may add more stats to this list, so the caller SHOULD be prepared to handle unexpected stats objects.

webrtc.html
jan-ivar: Actually no. RTCMediaStreamTrackStats have no back-links to RTCMediaStreamStats, so RTCMediaStreamStats should *not* be included, if we want to follow any rule of significance.

webrtc.html
jan-ivar: @henbos  I disagree. These foreign-keys are references, and leaving broken references leads to bugs. Had we had a different API where it was possible to look up these references with another call, then I might have agreed, but we don't.

jan-ivar: I've also heard the claim before that recursing ends up including everything, but this does not seem to be true from what I've found, because links tend to point "down" into the stack, not up. As long as we preclude stats for *other* tracks, I think there are ways to implement this that will be performant. The only real variable here is the number of tracks in a connection, which is why the track selector works imho.

webrtc.html
jan-ivar: @alvestrand The sentence should be narrowed to *new* stats then, because as it reads now it broadly allows browsers to differ in which stats they leave out among existing agreed-upon spec'ed stats.

That said, isn't it generally understood that new specs will iterate on old ones, unless it's somehow explicitly denied?

webrtc.html
jan-ivar: s/tracks/senders and receivers/ if we want to go that way.

henbos: (Unless you count a complete GetStats call as such API.) I don't think an API for looking up specific references works anyway, because a separate call occurs at a later time, at which these relationships could have changed.

I agree avoiding broken references would be nice to achieve if possible. Precluding broken references and other tracks would mean we can't say which stream a track belongs to, because other tracks may belong to that stream. On the other hand, including those tracks might be fine. Tracks don't reference things (other things reference tracks). We would have to be careful about not introducing the wrong references if we have selectors like this. This might be fine, but what if we want other types of selectors, or other references in the future? (Its possible I'm prematurely worrying.)

We could always not include references that we don't want to be followable, saying only relevant references are included, but that could be misleading.

webrtc.html
jan-ivar: Including objects "referencing track stats" is backwards, and does end up including everything. The goal here IMHO should be to return an object without broken references. So I vote for recursive *and* "referenced by", but not "referencing". Makes sense?

webrtc.html
alvestrand: RTCMediaStreamTrackStats doesn't have a reference to RTCRTPStreamStats either - the pointer goes the other way - the reason being that we might have to represent simulcast. See why I find an explicit listing simpler to understand?

webrtc.html
alvestrand: Which would leave RTCRtpStreamStats out in the cold.

webrtc.html
jan-ivar: @alvestrand Right, in my mind, we'd start with the RTCRTPStreamStats for the relevant track, which would work with a recursive "down" algorithm, as well as include at least the objects you wanted. Hmm, this tells me I was really thinking of the sender or receiver as the selector, so perhaps we should just make that explicit?

Btw, What does [`ssrcId`](https://w3c.github.io/webrtc-stats/#dom-rtcmediastreamtrackstats-ssrcids) point to? It's not documented.

webrtc.html
jan-ivar: Yeah we have to start with the sender/receiver of that track.

jan-ivar: Which probably means a track was always just a legacy selector for selecting a sender/receiver...

webrtc.html
alvestrand: A reason we chose DOMString for our references was that it was clear how to support broken references - you can still compare them  for equality, even when you can't access their content.

webrtc.html
alvestrand: Sender/receiver have no explicit representation in stats, and I haven't yet seen a number that I would want to hang off them.

webrtc.html
alvestrand: A spec is a contract between the two sides of the API. Version skew will lead to one side that expects the old version to see behavior from the other side that is only specified in the new version. That's why it's nice to say what new behaviors can be expected in places like this.

webrtc.html
alvestrand: The next version has added remote RTPStreamStats and MediaStreamStats to the list. Perhaps we can stop there?

alvestrand:
@jan-ivar @henbos PTAL.

webrtc.html
```diff
@@ -8857,6 +8859,58 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+            objects to <var>result</var>, if they exist:
+	    <ul>
+	      <li>
+		The RTCMediaStreamTrackStats corresponding to the selector.
+	      </li>
+	      <li>
+		If the selector track is inbound, the RTCInboundRTPStreamStats
+                corresponding to the selector that has "remote=false", and the
+                RTCOutboundRTPStreamStats corresponding to the selector that has
+                "remote=true".
```
henbos: It is my understanding that an inbound track has a corresponding RTCInboundRTPStreamStats (or multiple if receiving multicast is supported - can you clarify that it could be plural if so?), but never RTCOutboundTPStreamStats.

Is this comment incorrect or in what case would there be RTCOutboundRTPStreamStats corresponding to an inbound track?

webrtc.html
```diff
@@ -8857,6 +8859,58 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+	      </li>
+	      <li>
+		If the selector track is outbound, the RTCOutboundRTPStreamStats
+                corresponding to the selector that has "remote=false", and the
+                RTCInboundRTPStreamStats corresponding to the selector that has
+                "remote=true".
```
henbos: There may be one or more RTCOuboundRTPStreamStats for a single selector - can you clarify that there could be more than one RTCOutboundRTPStreamStats?

And again, it is my understanding that an outbound track does not have any corresponding RTCInboundRTPStreamStats. When could this happen?

webrtc.html
```diff
@@ -8857,6 +8859,58 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+	      </li>
+              <li>
+                The stats objects for all MediaStreams that have the
+                MediaStreamTrack as a member.
```
henbos: I would prefer if you refer to the stats object by type, i.e. "The RTCMediaStreamStats objects for all..." as opposed to "the stats objects", that is less ambiguous. nit: You could also say that have the RTCMediaStreamTrackStats.id as a member of trackIds to be clear what "as a member" means.

Note for reviewers: This will expose IDs of RTCMediaStreamTrackStats that we are not interested in, meaning broken references. I think we should include the stream the track belongs to. I don't think we should include other tracks.

webrtc.html
jan-ivar: The rtcp data for an inbound stream is stored in
```js
stats.get([...stats.values()].find(s => s.type == "inbound-rtp").associateStatsId);
````
and is of type `RTCOutboundRTPStreamStats`. I try to explain it [here](http://stackoverflow.com/a/29840886/918910).

jan-ivar: Example: https://jsfiddle.net/moz71L7h/

alvestrand: I'd phrase that as "the information we have about the other end's view of the connection is stored in the item you find through associateStatsId. Its values are derived from data received via RTCP."

@jan-ivar you left out the step of going from the MediaStreamTrackStats to the RTCRtpStreamStats.

alvestrand:
@jan-ivar this is awaiting your contributions after the previous interim.

Issue #723, opened 2016-06-29
Need JSEP reference for end-of-candidates indication


burnburn:
The WebRTC spec says in the description for RTCIceTransportState in section 5.6 (RTCIceTransportInterface):

"The failed and completed states require an indication that there are no additional remote candidates. This can be indicated either by canTrickleIceCandidates being set to false, or the processing of an end-of-candidates indication as described in [JSEP]."

I looked in JSEP but couldn't find an appropriate reference (anchor), so I filed issue https://github.com/rtcweb-wg/jsep/issues/293.

This is a work item for issue #337.

aboba:
Shouldn't it say: 

"This can be indicated by calling addIceCandidate with a null value for candidate, by canTrickleIceCandidates being set to false, or the processing of an end-of-candidates indication as described in [JSEP]."

burnburn:
Bernard:  yes.  I will include that change (or whatever we conclude) in the fix.
Status today:  JSEP issue #293 is still in progress.

PR #991, opened 2017-01-11
Update JSEP reference in RTCIceTransportState definition

burnburn:
This addresses issue #723.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7055,9 +7055,7 @@ sender.setParameters(params)
       a candidate value whose <code>candidate</code> property is set
       to an empty string, by <a data-link-for=
       "RTCPeerConnection">canTrickleIceCandidates</a> being set to
-      <code>false</code>, or the processing of an end-of-candidates
-      indication as described
-      in <span data-jsep="addicecandidate">[[!JSEP]]</span>.</p>
+      <code>false</code>.</p>
       <p>Some example transitions might be:</p>
       <ul>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7051,10 +7051,13 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-      indicated either by <a data-link-for=
+      indicated by calling <code><a>addIceCandidate</a></code> with
+      a candidate value whose <code>candidate</code> property is set
+      to an empty string, by <a data-link-for=
       "RTCPeerConnection">canTrickleIceCandidates</a> being set to
-      <code>false</code>, or the processing of an end-of-candidates indication
-      as described in [[!JSEP]].</p>
+      <code>false</code>, or the processing of an end-of-candidates
+      indication as described
+      in <span data-jsep="addicecandidate">[[!JSEP]]</span>.</p>
```
alvestrand: As I read JSEP at the moment, the end-of-candidates indication is the same as calling with a candidate value whose candidate property is the empty string, so this is saying the same thing twice.

burnburn: So the truth is that I still don't believe JSEP anywhere actually describes processing of an end-of-candidates indication but do believe that I'm never going to get a proper answer on that, so I almost deleted the reference to JSEP altogether.  I am happy to delete everything from ', or the processing . . .' onwards.

alvestrand: Deleting everythign after the comma would work for me.

PR #994, opened 2017-01-13
Changing "non-null" to "missing" to match IDL terminology.

taylor-b:


webrtc.html
```diff
@@ -1807,8 +1807,8 @@
                         </ol>
                       </li>
                       <li>
-                        <p>If <code><var>candidate</var>.ufrag</code> is
-                        non-null, and is not equal to any ufrag present in the
+                        <p>If <code><var>candidate</var>.ufrag</code> is not
+                        missing, and is not equal to any ufrag present in the
```
alvestrand: My hatred of double negatives is coming out. Can we say "present" instead of "not missing"?

stefhak:
I had exactly the same thought as @alvestrand.

On Thu, Jan 19, 2017 at 2:25 PM, Harald Alvestrand <notifications@github.com

webrtc.html
aboba: Elsewhere I have seen the term "set" used to mean "non-null".

alvestrand:
@jan-ivar @dontcallmedom do you have an opinion on what current terminology is?
@aboba suggests "set" and "not set" have been used elsewhere.

webrtc.html
jan-ivar: `ufrag` is not nullable so it can never be `null`.

I think we should be explicit, and say "If candidate.ufrag is not `undefined`,"

jan-ivar:
"If candidate.ufrag is not `undefined`"

webrtc.html
taylor-b: But, actually... `ufrag` _is_ nullable in the interface version of `RTCIceCandidate`. So if it's missing from the dictionary it will be `undefined`, and if it's missing from the interface it will be `null`, correct?

In that case, is "set" suitable, in order to avoid a double-negative, or is "not missing" most accurate?

jan-ivar: Good point, `candidate` can be either a dictionary or an object here, which is unusual! Even more reason to be explicit IMHO: "If candidate.ufrag is not `undefined` and not `null`", or maybe neither nor?

taylor-b:
Ok, changed to "not undefined or null".

jan-ivar:
@taylor-b Is that `!undefined || null` or `!(undefined || null)`? I realize one of those is redundant, meaning there's only one sensible interpretation, but there are still two interpretations. Didn't like nor?

taylor-b:
We tend to avoid "nor", but I guess that's technically the only way not to be ambiguous. Alright, changing.

PR #996, opened 2017-01-19
Describe when an RTCSctpTransport is created/set to null.

### LINKED ISSUE #979: When is an RTCSctpTransport created and destroyed?
This should be documented more clearly. Things that aren't obvious to me right now:

1. Is the SctpTransport created after applying an offer, or only after applying an answer?
2. If the data "m=" section is rejected, does the SctpTransport go away? I'd guess yes.
3. If the SCTP port ("a=sctp-port") is changed, and a new SCTP association is formed, is a new SctpTransport created to represent that association? I'd lean towards "yes".
4. Secondary question: Do we even want to support changing the SCTP port? It's not clear we have any use case for this.

taylor-b:
Fixes #979.

An SctpTransport represents a unique SCTP association. It's created from
applying a local description, and set to null when applying an answer
with a rejected data m= section.

SCTP-SDP reference will be added by this PR: https://github.com/tobie/specref/pull/327

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -899,18 +899,20 @@
                     <var>connection</var>.</p>
                   </li>
                   <li>
-                    <p>If <var>description</var> is set as a local description,
-                    and it creates a new SCTP association, as defined in
-                    [[!SCTP-SDP]], set the value of <var>connection</var>'s
-                    [[<code><a>sctpTransport</a></code>]] internal slot
-                    to a newly created <code><a>RTCSctpTransport</a></code>.</p>
+                    <p>If <var>description</var> is of type "answer", and it
+                    initiates the closure of an existing SCTP association, as
+                    defined in [[!SCTP-SDP]], Sections 10.3 and 10.4, set the
+                    value of <var>connection</var>'s
+                    [[<code><a>sctpTransport</a></code>]] internal slot to
+                    <code>null</code>.</p>
                   </li>
                   <li>
-                    <p>If <var>description</var> is of type "answer", and it
-                    closes an existing SCTP association, as defined in
-                    [[!SCTP-SDP]], set the value of <var>connection</var>'s
-                    [[<code><a>sctpTransport</a></code>]] internal slot
-                    to <code>null</code>.</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -895,6 +899,20 @@
                     <var>connection</var>.</p>
                   </li>
                   <li>
+                    <p>If <var>description</var> is set as a local description,
+                    and it creates a new SCTP association, as defined in
+                    [[!SCTP-SDP]], set the value of <var>connection</var>'s
+                    [[<code><a>sctpTransport</a></code>]] internal slot
+                    to a newly created <code><a>RTCSctpTransport</a></code>.</p>
```
aboba: I don't believe that a new SCTP association can be created until setRemoteDescription is called, because until then the remote port isn't known.  So you might say, "creates a new <code><a>RTCSctpTransport</a></code> object.".

taylor-b: Would "initiates a new SCTP association" be better wording? What I'm trying to capture is:

* data "m=" section negotiated for the first time
* data "m=" section going from rejected to non-rejected
* local port changed (which should never happen?)

aboba: Is it possible to initiate a new SCTP association until the remote port is known?

taylor-b: > Is it possible to initiate a new SCTP association before setRemoteDescription is called?

Yeah, "initiate" doesn't quite work either, especially given that SCTP uses an "INIT" message.

I'd like to not say something as vague as "If the description is associated with a new SCTP association"... How about "If the description *will lead to* a new SCTP association being established?"

Another option would be to only create a new object when applying an answer or pr-answer. The more I think about it, the more that seems like the more correct thing to do. For one, how is "maxMessageSize" set without any remote description?

aboba: If the SctpTransport is created when calling setLocalDescription then maxMessageSize needs to be nullable since it will be null prior to setting a remote description.  Perhaps a state attribute would clarify whether the SctpTransport was new, connecting, connected or closed.

aboba:
LGTM

Issue #961, opened 2016-12-01
Effect of a BYE on RtpReceiver.track


aboba:
If an RtpReceiver receives a BYE from an RtpSender, what happens? 

Does receiver.track.onmute fire?

aboba:
Some questions: 

1. When transceiver.stop() is called, does transceiver.sender send a BYE? 
2. When transceiver.sender.replaceTrack(null) is called does transceiver.sender send a BYE?
3. What about transceiver.sender.track.enabled = "false"?

alvestrand:
Clarification: it's the sender that sends the BYE when it's stopped, the transceiver is only inidirectly involved (because stopping the transceiver stops alll the senders).

aboba:
@alvestrand  @stefhank Correct.  Some thoughts: 

1. When transceiver.stopped is set to true, a BYE is sent on all transceiver.sender encodings[].ssrc.  When a BYE is received, the transceiver.receiver.track.onmute EventHandler is fired. 
2. Calling replaceTrack(null) or setting transceiver.sender.track.enabled to "false" does not causes a BYE to be sent (though the remote receiver might timeout the SSRC, as described in RFC 3550 Section 6.3.5).

juberti:
Seems strange that .enabled would send a BYE if replaceTrack(null) doesn't, so I think 3) above is 'no'.

alvestrand:
we want BYE to be sent if we're never going to use that SSRC again. So flipping .enabled shouldn't touch it. .ended is set from .stop, but we might want to resurrect it even then, so what do we do when

pc.addTrack(track1) => sender1
track1.stop()
sender1.replaceTrack(track2)

If the sequence

pc.addTrack(track1) => sender1
sender1.replaceTrack(track2)

continues sending on the same ssrc, it seems that the first sequence should also do so, and thus we shouldn't be sending BYE on ended.

after a transceiver.stop() we're not going to use that SSRC again, so sending a bye then seems OK.
similar logic brings me to think that a BYE says "I'm not going to be talking to you on this ssrc again", so doing track.stop and firing an ended event seems sensible.

If we have examples in the wild of people doing BYE & then resurrecting the SSRC, this logic falls flat.

aboba:
@alvestrand Agree with the principle that a BYE is only sent if we're never going to use the SSRC again, so flipping .enabled or replaceTrack(null) doesn't cause a BYE to be sent.  Also agree that sender.track.stop() shouldn't result in a BYE being sent.

PR #1001, opened 2017-01-26
Effect of a BYE on RtpReceiver.track

aboba:
Partial fix for Issue https://github.com/w3c/webrtc-pc/issues/961

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5843,15 +5843,14 @@ sender.setParameters(params)
             <dt><dfn><code>track</code></dfn> of type <span class=
             "idlAttrType"><a>MediaStreamTrack</a></span>, readonly</dt>
             <dd>
-              <p>The <code id="dom-rtpreceiver-track">track</code> attribute is
-              the track that is associated with this
-              <code><a>RTCRtpReceiver</a></code> object. When a BYE is
-              received on one of the SSRCs in
-              <code><var>receiver</var>.getParameters().encodings[<var>i</var>].ssrc</code>
-              (where <code><var>i</var></code> goes from 0 to
-              <code>encodings.length -1</code>) the <code>track.onmute</code>
-              EventHandler is fired.  When packets are received again, the
-              <code>track.onunmute</code> EventHandler is fired.</p>                
+              <p>The <code id="dom-rtpreceiver-track">track</code>
+              attribute is the track that is associated with this
+              <code><a>RTCRtpReceiver</a></code> object <var>receiver</var>.
+              When an RTCP BYE packet is received for one of the RTP
+              source streams received by <var>receiver</var>, the
+              <code>mute</code> event is fired at <code>track</code>.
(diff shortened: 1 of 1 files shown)


alvestrand:
Need review by @juberti to check that it says what we want it to say per WG discussion.

aboba:
@taylor-b  @pthatcherg Can you review?

webrtc.html
```diff
@@ -5845,11 +5845,18 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-              <code><a>RTCRtpReceiver</a></code> object. Note that
-              <code>track.stop()</code> is final, although clones are not
-              affected. Since <code>receiver.track.stop()</code> does not
-              implicitly stop <var>receiver</var>, Receiver Reports continue to
-              be sent.</p>
+              <code><a>RTCRtpReceiver</a></code> object. When a BYE is
+              received on one of the SSRCs in
+              <code><var>receiver</var>.getParameters().encodings[<var>i</var>].ssrc</code>
+              (where <code><var>i</var></code> goes from 0 to
+              <code>encodings.length -1</code>) the <code>track.onmute</code>
```
taylor-b: For consistency, should say the "mute" event is fired, rather than "onmute" EventHandler.

webrtc.html
```diff
@@ -5845,11 +5845,18 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+              EventHandler is fired.  When packets are received again, the
```
taylor-b: Change to "If and when packets are received again..." to avoid implication that packets will always be received again?

webrtc.html
```diff
@@ -5845,11 +5845,18 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
               <p>The <code id="dom-rtpreceiver-track">track</code> attribute is
               the track that is associated with this
-              <code><a>RTCRtpReceiver</a></code> object. Note that
-              <code>track.stop()</code> is final, although clones are not
-              affected. Since <code>receiver.track.stop()</code> does not
-              implicitly stop <var>receiver</var>, Receiver Reports continue to
-              be sent.</p>
+              <code><a>RTCRtpReceiver</a></code> object. When a BYE is
+              received on one of the SSRCs in
+              <code><var>receiver</var>.getParameters().encodings[<var>i</var>].ssrc</code>
```
taylor-b: This looks right, but I'm wondering if it could be simplified. Would it work to say "When a BYE is received for one of the RTP source streams received by <var>receiver</var>..."?

alvestrand:
Some timeout thing in JSEP needs to be finalized before merging this.

aboba:
@alvestrand Have modified the text to make it clear that the mute event depends on removal of the SSRC (due to BYE or timeout).  So I believe the delay is not implicitly taken into account.

stefhak:
LGTM.

Issue #962, opened 2016-12-01
Event when a transceiver is stopped via remote action


aboba:
If a re-Offer is received with an m-line having a zero port, when the Answerer calls setRemoteDescription, transceiver.stopped will be set to true.  Is an event also provided, or does the application have to check transceiver.stopped?

aboba:
This is related to Issue https://github.com/w3c/webrtc-pc/issues/961

If a BYE is sent by transceiver.sender as a result of transceiver.stop(), this could result in the remote transceiver.receiver.track.onmuted firing when setRemoteDescription is called.

alvestrand:
We don't have state change events on transceiver yet. Are you suggesting we add some?

alvestrand:
The track does have an .onmuted handle, which may be enough for the BYE case.  The zero case is different.

aboba:
@alvestrand We describe the steps taken when transceiver.stop() is called, but do not yet describe the steps taken when setRemoteDescription(zeromlineOffer) is called.  Some thoughts on what might happen:

1. Let transceiver be the RTCRtpTransceiver object corresponding to the zero'd m-line.

2. If transceiver.stopped is true, abort these steps.

3. Let connection be the RTCPeerConnection object corresponding to transceiver.

4. If connection's [[isClosed]] slot is true, abort these steps.

5. Let sender be transceiver.sender.

6. Let receiver be transceiver.receiver.

7. Stop sending media with sender. Send a BYE on all encodings. 

8. Stop receiving media with receiver.

9, Set receiver.track.readyState to ended.  Fire receiver.track.onended eventHandler.

10. Set transceiver.stopped to true.

PR #1002, opened 2017-01-26
Event when a transceiver is stopped via remote action

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/962


stefhak:
LGTM

Issue #950, opened 2016-11-22
Change SetLocalDescription to require unchanged offer/answer string


alvestrand:
As per discussions at IETF RTCWEB meeting in Seoul - it was decided some time ago to remove the ability to change the SDP string between createOffer / createAnswer and setLocalDescription. In Seoul, the decision was that a createOffer / answer invalidates all previous descriptions, and increments the version counter - only the last version number can be set with setLocalDescription.

This needs to be reflected in the webrtc-pc spec.

A proposal was also made that setLocalDescription's argument could omit the SDP string and that this would cause the last SDP created with createOffer/createAnswer to be used - since this is strictly an API matter, this ought to be a WEBRTC WG concern.

WG consensus on this needs to be established before doing it.

aboba:
So if description.sdp is the null string, then use description.type?

PR #1003, opened 2017-01-26
Change setLocalDescription to require unchanged offer/answer string

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/950

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1654,7 +1654,7 @@ interface RTCPeerConnection : EventTarget  {
                   last call to <code>createAnswer</code>.</li>
                   <li>If <code><var>description</var>.sdp</code> is
                   <code>null</code> and <code><var>description</var>.type</code>
-                  is <code>answer</code> or <code>pranswer</code>
+                  is <code>answer</code> or <code>pranswer</code>,
                   set <code><var>description</var>.sdp</code> to
                   <var>lastAnswer</var>.</li>
                   <li>If <code><var>description</var>.sdp</code> is
--- a/webrtc.html
+++ b/webrtc.html
@@ -1646,28 +1646,33 @@ interface RTCPeerConnection : EventTarget  {
                 a result, when the method is invoked, the User Agent MUST
                 run the following steps:
                 <ol>
-                  <li>Let <var>sdp</var> be the argument to
+                  <li>Let <var>description</var> be the argument to
                   <code>setLocalDescription</code>.</li>
                   <li>Let <var>lastOffer</var> be the result returned by the
                   last call to <code>createOffer</code>.</li>
                   <li>Let <var>lastAnswer</var> be the result returned by the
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1639,10 +1639,42 @@
...
+                  does not exactly match <var>lastOffer</var>,
+                  reject the promise with <code>InvalidParameters</code>
+                  and abort these steps.</li>
+                  <li>If <code><var>description</var>.type</code> is
+                  <code>answer</code> or <code>pranswer</code> and
+                  <code><var>description</var>.sdp</code> does not exactly
+                  match <var>lastAnswer</var>, reject the promise with
+                  <code>InvalidParameters</code> and abort these steps.</li>               
+                  <li>Return the result of <a href="#set-description">setting
+                  the RTCSessionDescription</a> indicated by the method's first
```
taylor-b: Should say either "the argument" or "the first argument" consistently

webrtc.html
```diff
@@ -1639,10 +1639,42 @@
...
+                  is <code>answer</code> or <code>pranswer</code>,
+                  set <code><var>description</var>.sdp</code> to
+                  <var>lastAnswer</var>.</li>
+                  <li>If <code><var>description</var>.sdp</code> is
+                  <code>null</code> and <code><var>description</var>.type</code>
+                  is <code>offer</code>, set <code><var>description</var>.sdp</code>
+                  to <var>lastOffer</var>.</li>
+                  <li>If <code><var>description</var>.type</code> is
+                  <code>offer</code> and <code><var>description</var>.sdp</code>
+                  does not exactly match <var>lastOffer</var>,
```
taylor-b: What went into JSEP isn't that the SDP must be *exactly* the same, but that the parsed representation must be the same. For example, "a=" lines appearing in a different order shouldn't matter.

Maybe we could simplify this by just referring to JSEP:

> If `description.sdp` is not `null`, and the browser determines that it was modified after the last call to `createOffer`/`createAnswer`, as described by JSEP Section 5.8, reject `p` with an `InvalidAccessError` and abort these steps.

webrtc.html
```diff
@@ -1639,10 +1639,42 @@
...
+                  reject the promise with <code>InvalidParameters</code>
```
taylor-b: We use `InvalidAccessError` for invalid parameters. Or maybe `InvalidModificationError` would be more appropriate? I'm not sure.

webrtc.html
aboba: I posted a question to the RTCWEB mailing list on this.

webrtc.html
aboba: Elsewhere, InvalidModificationError is used, so I changed it to that.

webrtc.html
aboba: Using "first argument" throughout.

PR #1005, opened 2017-01-26
Add offerToReceive* as legacy extensions

### LINKED ISSUE #709: offerToReceiveAudio/offerToReceiveVideo remain in implementations (likely needed for compat)
These were removed from the spec in e0c1dbee2ea722f5742610ec3cf1838f946cafa1 when `RTCRtpTransceiver` was added.

In https://codereview.chromium.org/2064593003/ it became apparent that support remains in the callback-based createOffer, and people expect them to work with the promise-based createOffer as well.

It's in Gecko and WebKit as well:
https://github.com/mozilla/gecko-dev/blob/b3ee63f6aa2af5788181d7bc42cd639bef08d4d9/dom/webidl/RTCPeerConnection.webidl#L62 (in `RTCOfferOptions`)
http://trac.webkit.org/browser/trunk/Source/WebCore/Modules/mediastream/RTCOfferAnswerOptions.cpp?rev=198492#L79 (also in `RTCOfferOptions`)

Lacking a concrete plan to get this removed from implementations with minimal compat risk, adding this back to the spec would probably be best. Perhaps it can be defined in terms of `RTCRtpTransceiver`?

Also, what to do about `OfferToReceiveAudio` and `OfferToReceiveVideo` with a capital O?

alvestrand:
Closes #709

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2181,6 +2181,8 @@ interface RTCPeerConnection : EventTarget  {
           These methods are kept on <code><a>RTCPeerConnection</a></code> for
           legacy purposes.
         </div>
+        <section>
+          <h4>Method extensions</h4>
         <div>
           <pre class="idl">partial interface RTCPeerConnection {
     Promise&lt;void&gt; createOffer (RTCSessionDescriptionCallback successCallback, RTCPeerConnectionErrorCallback failureCallback, optional RTCOfferOptions options);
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2604,6 +2606,40 @@
...
+          </pre>
+          <section>
+            <h2>Attributes</h2>
+            <dl data-link-for="RTCOfferOptions" data-dfn-for="RTCOfferOptions" class="dictionary-members">
+              <dt><dfn>offerToReceiveAudio</dfn> of type <span class="idlMemberType"><a>boolean</a></span></dt>
+              <dd>
+                <p>
+                  When this is given a non-false value, and no outgoing track of
+                  type "audio" is attached to the PeerConnection, createOffer()
+                  will behave as if addTransceiver("audio") had been called prior
```
stefhak: "called once prior"? We said the offer would contain only one receive-only m-line?

webrtc.html
```diff
@@ -2604,6 +2606,40 @@
...
+                  to the createOffer() call.
+                </p>
+              </dd>
+              <dt><dfn>offerToReceiveVideo</dfn> of type <span class="idlMemberType"><a>boolean</a></span></dt>
+              <dd>
+                <p>
+                  When this is given a non-false value, and no outgoing track of
+                  type "audio" is attached to the PeerConnection, createOffer()
+                  will behave as if addTransceiver("video") had been called prior
```
stefhak: Dito.

stefhak:
Had a minor comment, LGTM otherwise.

webrtc.html
aboba: Do you mean "no outgoing track of type audio" or "no outgoing track of type video"?

webrtc.html
aboba: What happens if there is an outgoing track of type "audio"?  Nothing?

alvestrand: Yes, if conditions are not satisfied, nothing happens.

alvestrand:
Editors meeting: Harald to fix typos and merge.

webrtc.html
alvestrand: Corrected to "video".

webrtc.html
alvestrand: I also added a precondition that no existing recv-able media section exists. Otherwise, doing a re-offer with offerToReceive=true would add an extra media section for each re-offer, and I don't think that was the intended behavior.

Issue #526, opened 2016-02-29
NetworkError event is not defined and might not be needed


ShijunS:
When discussing the following portion of Section 6.2 in issue #492

> 1. ...
> 2. Set channel's readyState attribute to closed.
> 3. If the transport was closed with an error, fire an NetworkError event at channel.
> 4. Fire a simple event named close at channel

@adam-be commented 

> This one is a bit strange. Is the "NetworkError event" even defined somewhere? I think we could do with one event here.

I'm opening a new issue to track this.

stefhak:
Seems like WebSocket uses a close event with reason rather than an error event.

adam-be:
Yes. But there's a simple error event as well [1]. We have a simple close event so it would take some work to align the behaviour.

[1] https://html.spec.whatwg.org/multipage/comms.html#feedback-from-the-protocol

adam-be:
How we want to communicate errors from DataChannel should perhaps be a TPAC or telco topic. It's also a question how current implementations work.

dontcallmedom:
#850 didn't fix this since NetworkError still appears in the draft.

#846 which was closed as a dup had this:
> webrtc-pc Issue #526 "NetworkError event is not defined and might not
be needed": RESOLUTION: we will change our close event to match with
WebSocket’s
> https://lists.w3.org/Archives/Public/public-webrtc/2016Oct/0018.html

aboba:
@adam-be  The Websocket onClose event handler provides the following information:

event.reason, eventWasClean, event.code:

1001    CLOSE_GOING_AWAY
1000    CLOSE_NORMAL
1005    CLOSE_NO_STATUS
1006    CLOSE_ABNORMAL

Note also that in order to provide this information the websocket.close method takes code and reason as arguments: 

void close(in optional unsigned long code, in optional DOMString reason);

So we have agreement that we want to add optional arguments to RTCDataChannel.close()?

aboba:
@stefhak Websockets has a mechanism to communicate the reason for the upper layer closure; there is no equivalent in the data channel. So I don't think it makes sense to add optional arguments to close() or to the onclose EventHandler.

aboba:
@dontcallmedom Section 13 says that the close Event (using the Event interface) is fired when "The RTCDataChannel object's underlying data transport has been closed."  So it does appear that there was an intent to use the close Event for this situation.

PR #1011, opened 2017-01-30
Eliminate NetworkError

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/526

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -12001,7 +12001,7 @@ interface RTCErrorEvent : Event {
       default values for attributes that would be left uninitialized without
       the init dictionary present.</li>
       <li>InvalidMediaStreamTrackError changed to InvalidParameter.</li>
-      <li>Fire the <code><a>RTCError</a></code> event
+      <li>Fire the <code><a>RTCErrorEvent</a></code>
       when the data transport is closed with an error.</li>
       <li>Add an exception for data channel with trying to use existing
       code.</li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -11160,9 +11160,9 @@ interface RTCErrorEvent : Event {
           "RTCDataChannel">bufferedAmountLowThreshold</a></code>.</td>
         </tr>
         <tr>
-          <td><dfn id="event-datachannel-error"><code>error</code></dfn></td>
-          <td><code><a>ErrorEvent</a></code></td>
-          <td>Any error occured from the data channel.</td>
+          <td><dfn id="event-datachannel-error"><code>RTCError</code></dfn></td>
+          <td><code><a>RTCErrorEvent</a></code></td>
+          <td>Any error occurred from the data channel.</td>
         </tr>
         <tr>
           <td><dfn id="event-datachannel-close"><code>close</code></dfn></td>
--- a/webrtc.html
+++ b/webrtc.html
@@ -12004,8 +12004,8 @@ interface RTCErrorEvent : Event {
       default values for attributes that would be left uninitialized without
       the init dictionary present.</li>
       <li>InvalidMediaStreamTrackError changed to InvalidParameter.</li>
-      <li>Fire the <code><a>RTCErrorEvent</a></code>
-      when the data transport is closed with an error.</li>
+      <li>Fire NetworkError when the data transport is closed with an
+      error.</li>
       <li>Add an exception for data channel with trying to use existing
       code.</li>
       <li>Change maxRetransmits to be an unsigned type.</li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -12001,7 +12001,7 @@ interface RTCErrorEvent : Event {
       default values for attributes that would be left uninitialized without
       the init dictionary present.</li>
       <li>InvalidMediaStreamTrackError changed to InvalidParameter.</li>
-      <li>Fire the <code>error</code> event
+      <li>Fire the <code><a>RTCError</a></code> event
       when the data transport is closed with an error.</li>
       <li>Add an exception for data channel with trying to use existing
       code.</li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -12001,8 +12001,8 @@ <h3 id="sec-constraints">Constraints</h3>
       default values for attributes that would be left uninitialized without
       the init dictionary present.</li>
       <li>InvalidMediaStreamTrackError changed to InvalidParameter.</li>
-      <li>Fire NetworkError when the data transport is closed with an
-      error.</li>
+      <li>Fire the <code>RTCDataChannel.onerror</code> EventHandler
```
alvestrand: This is not corrcect spec language - you fire errors, not event handlers.

aboba: Right.  Fixed the language but there is still the question about whether it is the RTCError event that should be fired, or Error.

alvestrand: I think you should "fire an event named Error at the DataChannel object, with "message" set to an appropriate error description".
If we need more data than "an error happened, write this to the debug log", we need to use a specialization of RTCError.

webrtc.html
```diff
@@ -12001,8 +12001,8 @@ <h3 id="sec-constraints">Constraints</h3>
       default values for attributes that would be left uninitialized without
       the init dictionary present.</li>
       <li>InvalidMediaStreamTrackError changed to InvalidParameter.</li>
-      <li>Fire NetworkError when the data transport is closed with an
-      error.</li>
+      <li>Fire the <code>error</code> event
+      when the data transport is closed with an error.</li>
```
stefhak: Would we need to say what the message of the event is?

webrtc.html
```diff
@@ -12001,8 +12001,8 @@ <h3 id="sec-constraints">Constraints</h3>
       default values for attributes that would be left uninitialized without
       the init dictionary present.</li>
       <li>InvalidMediaStreamTrackError changed to InvalidParameter.</li>
-      <li>Fire NetworkError when the data transport is closed with an
-      error.</li>
+      <li>Fire the <code><a>RTCError</a></code> event
+      when the data transport is closed with an error.</li>
```
taylor-b: It looks like this is in the changelog section of the spec; this shouldn't change, but a couple other places should.

Also, I'd recommend "an RTCErrorEvent" instead of "the RTCError event". And we may want to specify what errorDetail value(s) to use.

aboba: @taylor-b This relates to an error in the data channel transport,  such as might be communicated in an SCTP ERROR chunk.  The Cause Code values are listed here: http://www.iana.org/assignments/sctp-parameters/sctp-parameters.xhtml#sctp-parameters-24

The ERROR chunk also has a Cause-Specific-Information field.  So perhaps RTCError.errorDetail contains the Cause Code and RTCError.message contains the Cause-Specific-Information?

taylor-b: The data channel could also be closed with an error if there aren't enough stream IDs, which occurs above the SCTP level. For example: I call `createDataChannel` 1025 times, but the SCTP handshake (which occurs later) only negotiates 1024 streams. This was something I brought up at a past virtual interim.

So, this situation would need to be differentiated somehow from SCTP protocol errors. I'd propose using errorDetail to differentiate the two, and for SCTP errors, put the Cause Code in a new field (like is done for `httpRequestStatusCode`).

aboba: @taylor-b  It's probably better to add a new attribute for the SCTP Cause Code (e.g. sctpCauseCode) rather than using errorDetail. 

For data channel errors (e.g. above the transport layer), I am wondering if it makes sense to use the close event with a reason code, such as is done with WebSockets:
https://html.spec.whatwg.org/multipage/comms.html#handler-websocket-onclose

"When the WebSocket connection is closed, possibly cleanly, the user agent must queue a task to run the following substeps:
Change the readyState attribute's value to CLOSED (3).
If the user agent was required to fail the WebSocket connection, or if the the WebSocket connection was closed after being flagged as full, fire an event named error at the WebSocket object. [WSP]
Fire an event named close at the WebSocket object, using CloseEvent, with the wasClean attribute initialized to true if the connection closed cleanly and false otherwise, the code attribute initialized to the WebSocket connection close code, and the reason attribute initialized to the result of applying UTF-8 decode without BOM to the WebSocket connection close reason. [WSP]"

taylor-b: @aboba If I understand correctly, the `error` event for a WebSocket fires when the data channel is closed due to an error at the application level (closed before it finished being established, or buffer full), and the `close` event is used for *any* closure and provides additional protocol-level information.

In my opinion it does make sense for us to copy this pattern, since the working group already decided we should try to match WebSockets, and we have the same sort of situation here.

aboba: @taylor-b OK.  Think that implies that we need a reason code in the close event, and perhaps that the RTCError.errorDetail enum needs to cover both application layer and network-layer errors.

webrtc.html
```diff
@@ -7906,7 +7906,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             <dt><dfn><code>onerror</code></dfn> of type <span class=
             "idlAttrType"><a>EventHandler</a></span></dt>
             <dd>The event type of this event handler is <code><a data-for=
-            "RTCDataChannel">error</a></code>.</dd>
+            "RTCDataChannel">RTCErrorEvent</a></code>.
+            <code>RTCErrorEvent.errorDetail</code> contains the SCTP
+            Cause Code value, and <code>RTCErrorEvent.message</code>
+            contains the SCTP Cause-Specific-Information.</dd>
```
alvestrand: Should this be ", possibly with additional text", so that it's clear that the UA can add stuff here, and it's not machine parseable? (My generic fear about "message": That people will start parsing it.)

aboba:
Sentiment at the interim was to keep the RTCDataChannel close event simple.

PR #1016, opened 2017-02-07
Ensure that "track" event is only fired for "send" direction m-sections.

### LINKED ISSUE #1013: "track" event shouldn't fire for a "recvonly" description.
This is something I'm pretty confident we decided previously. Consider the legacy `addTrack` use case, where PeerConnection A adds a track, but PeerConnection B doesn't, and they do an offer/answer exchange. The "track" event should fire for B, but not for A. Later, if B calls `addTrack`, and they do another offer/answer, the direction will change and the "track" event will fire for A.

But currently, the "dispatch a receiver" steps will unconditionally fire the "track" event for every new "m=" section.

taylor-b:
Fixes #1013

Also does a bit of refactoring. The steps that used to be listed in the
"dispatch a receiver" section are now covered by "create a receiver" and
"stop a transceiver".

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -914,11 +914,31 @@
                     newly created <code><a>RTCSctpTransport</a></code>.</p>
                   </li>
                   <li>
-                    <p>If <var>description</var> is set as a remote description
-                    with new media descriptions <span data-jsep=
-                    "media-level-parse">[[!JSEP]]</span>, the User Agent MUST
-                    <a>dispatch a receiver</a> for all new media
-                    descriptions.</p>
+                    <p>If <var>description</var> is set as a remote description,
+                    then run the following steps for each media description in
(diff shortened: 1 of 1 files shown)


taylor-b:
@aboba That's what it seemed to me.

The "dispatch a receiver" steps for creating a track were identical to those in "create a receiver".

And the steps for "when a track is removed" are covered by the transceiver being stopped. This is the only time when a track *should* go to "ended", since that's a terminal state, and stopping a transceiver is the only terminal thing for its RtpReceiver. The differences I see are:

* The previous text mentioned that "an event" is fired when the track state goes to "ended". It looks like this is missing from the "stop a transceiver" steps; I'll add it.
* A task is queued to remove the receiver from the "set of receivers". This is no longer relevant here; there's a "set of transceivers" instead.

PR #1018, opened 2017-02-08
Mark negotitate in RTCRtcpMuxPolicy at risk

### LINKED ISSUE #997: RTCRtcpMuxPolicy of "negotiate" should be marked as an "at-risk" feature
JSEP allows implementations to not implement the "negotiate" RTCP mux policy (meaning they don't support non-muxed RTCP). Chrome and Firefox developers have expressed willingness to do this, so it's likely that the "negotiate" policy will not end up implemented in an interoperable way.

dontcallmedom:
close #997


stefhak:
LGTM.

PR #1019, opened 2017-02-09
Clarify that it is possible to send the same track in several copies.

### LINKED ISSUE #954: Receive a track multiple times
Hello,
while implementing we are dealing with the case of receiving the same track twice.
There are typically 2 cases:

### Receive a track previously removed
As explained in [1]:
> When the other peer stops sending a track in this manner, an ended event is fired at the MediaStreamTrack object.

The problem is that when the track is ended, it cannot be reused again [2].
So how should we manage the case of receiving again a track that has been set as ended?

### Receive the same track multiple times
The same track can be sent multiple times using addTransceiver[3].
How the receiver side should deal with this case?
A priori, we could have different MediaStreamTrack objects with the same ID and performing multiple onTrack events, but the ID should be unique, so this is not a proper solution.

## Solution proposal
As solution, I think that we must not allow sending a track more than once through addTransceiver in the same way addTrack is doing [4]:

>If an RTCRtpSender for track already exists in senders, throw an InvalidAccessError exception and abort these steps. 

And in the case that a receiver get this situation, it should reject the m-line associated.
In the case that the user wants send the same media multiple times, she should clone the track to generate a new ID.

What do you think about that?

Best!!

## Refs
[1] https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-addtransceiver
[2] https://www.w3.org/TR/mediacapture-streams/#idl-def-MediaStreamTrackState.ended
[3] https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-addtransceiver
[4] https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-addtrack

stefhak:
Close #954.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4285,7 +4285,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     noted that while <code>addTrack</code> checks if the 
     <code>MediaStreamTrack</code> given as argument is already being sent
     to avoid sending the same <code>MediaStreamTrack</code> twice, the other
-    ways to send does not and thus allows sending the same
+    ways to initiate sending do not and thus allow sending the same
     <code>MediaStreamTrack</code> (possibly using different
     <code><a>RTCRtpParameters</a></code> for different senders) in several
     copies simultaneously. Doing so can have the implication that at the
--- a/webrtc.html
+++ b/webrtc.html
@@ -4274,24 +4274,26 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     senders and receivers with some shared state. This set is initialized to
     the empty set when the <code><a>RTCPeerConnection</a></code> object is
     created.</p>
-    <p>There are several ways for a user to initiate
-    the sending of a <code>MediaStreamTrack</code> over a peer-to-peer
-    connection. One is to use the <code>addTrack</code> method on the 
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4274,6 +4274,24 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+    <code>MediaStreamTrack</code> argument) call, and then used, exist. It can be
+    noted that while <code>addTrack</code> checks if the 
+    <code>MediaStreamTrack</code> given as argument is already being sent
+    to avoid sending the same <code>MediaStreamTrack</code> twice, the other
+    ways to initiate sending do not and thus allow sending the same
+    <code>MediaStreamTrack</code> (possibly using different
+    <code><a>RTCRtpParameters</a></code> for different senders) in several
+    copies simultaneously. Doing so can have the implication that at the
+    receiving end of the connection there are several 
+    <code>MediaStreamTrack</code>'s that have identical
```
alvestrand: no apostrophe after "track" (plural not genitive)

aboba:
LGTM.

stefhak:
Bernard to clean up and merge.

stefhak:
@aboba none of the text made it into the spec(?). Was this by mistake or on purpose?

aboba:
@stefhak Not sure what happened, but it is merged now.

PR #1025, opened 2017-02-12
Mention that codecs can be reordered or removed but not modified.

### LINKED ISSUE #1024: Need to describe that codecs can be removed or reordered, but not modified
We definitely decided on this, but I don't see it spelled out anywhere.

taylor-b:
Fixes #1024

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5390,7 +5390,8 @@ sender.setParameters(params)
               be an entry in <code>codecs[]</code> with a <code>mimeType</code>
               attribute indicating retransmission via "audio/rtx" or
               "video/rtx", and an <code>sdpFmtpLine</code> attribute (providing
-              the "apt" and "rtx-time" parameters).</p>
+              the "apt" and "rtx-time" parameters). Codecs can be reordered or
+              removed, but otherwise this sequence can't be modified.</p>
             </dd>
             <dt><dfn><code>degradationPreference</code></dfn> of type
             <span class="idlMemberType"><a>RTCDegradationPreference</a></span>,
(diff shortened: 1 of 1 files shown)


taylor-b:
This is meant to mean that the codec dictionaries can't be modified, and codecs can't be added. Hopefully that's clear enough.

aboba:
@taylor-b How about this? 

Using the <code>setParameters</code> method, entries in the <code>codecs[]</code> sequence can be reordered or removed, but entries cannot be added, and <code><a>RTCRtpCodecParameters</a></code> dictionary attributes cannot be modified.

taylor-b:
You're right about `setCodecPreferences`. Might as well handle that now. I assume it's correct to say that codecs passed into `setCodecPreferences` can only contain codecs that are returned by `RTCRtpSender.getCapabilities()` or `RTCRtpReceiver.getCapabilities()`.

aboba:
@taylor-b Reading the PR, it seemed that setCodecPreferences has the same set of restrictions as setParameters() with respect to codecs.  Is that correct?  Is the only difference that setCodecPreferences can reorder or remove a codec from the SDP produced by createOffer/createAnswer (but not change the fmtp parameters), and sets negotiation-needed, whereas setParameters() never sets negotiation-needed and has no effect on the SDP produced by createOffer/createAnswer?

taylor-b:
@aboba That sounds correct. The other difference is that "setCodecPreferences" can include receive-only codecs.

aboba:
Based on the virtual discussion, this PR seems ready to merge.

PR #1030, opened 2017-02-14
Add stats selection algorithm based on sender or receiver of selector.

jan-ivar:
Fix for https://github.com/w3c/webrtc-stats/issues/116.

Also includes stats selection algorithm replacement for https://github.com/w3c/webrtc-pc/pull/990. See [here](https://github.com/w3c/webrtc-stats/issues/116#issuecomment-278860500) for what it ends up including.

@alvestrand PTAL.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9001,8 +9001,10 @@ interface RTCDTMFToneChangeEvent : Event {
                   <a>RTCRtpReceiver</a> on <var>connection</var> which
                   <code>track</code> member matches <var>selectorArg</var>.
                   If no such sender or receiver exists, or if more than one
-                  sender or receiver fit this criteria, throw an
-                  <code>InvalidAccessError</code>, and abort these steps.</p>
+                  sender or receiver fit this criteria, return a promise
+                  rejected with a newly
+                  <a data-link-for="exception" data-lt="create">created</a>
+                  <code>InvalidAccessError</code>.</p>
                 </li>
                 <li>
                   <p>Let <var>p</var> be a new promise.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -8951,11 +8951,9 @@ interface RTCDTMFToneChangeEvent : Event {
       object on which the stats request was issued. The calling Web application
       provides the selector to the <code><a data-for=
       "RTCPeerConnection">getStats()</a></code> method and the browser emits
-      (in the JavaScript) a set of statistics that it believes is relevant to
-      the selector.</p>
-      <div class="issue">
-        Evaluate the need for other selectors than MediaStreamTrack.
-      </div>
+      (in the JavaScript) a set of statistics that are relevant to the selector,
+      according to the <a>stats selection algorithm</a>. Note that that
+      algorithm takes the sender or receiver of a selector.</p>
       <p>The statistics returned are designed in such a way that repeated
       queries can be linked by the <code><a>RTCStats</a></code> <a data-for=
       "RTCStats">id</a> dictionary member. Thus, a Web application can make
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -8942,6 +8956,54 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+              <li>
+              All <code>RTCOutboundRTPStreamStats</code> objects corresponding
+              to <var>selector</var>.
+              </li>
+              <li>
+              All stats objects referenced directly or indirectly by the
+              <code>RTCOutboundRTPStreamStats</code> added.
+              </li>
+            </ul>
+          </li>
```
henbos: I believe this pulls in stats dictionaries of all types except for `RTCMediaStreamStats` (of which the track belongs), `RTCDataChannelStats` and `RTCPeerConnectionStats`. But only a subset of all object of each type need to be collected.

In terms of avoiding unnecessary collection of stats for performance, I'm still worried about the complexity of implementing this. We need to resolve references first, and collect stats afterwards, and only collect for relevant objects (otherwise there's no point to the selector). This should be doable, my concern stems with (Chromium's WebRTC codebase's) implementation details; lower layer objects not perfectly mapping with the spec, how references between objects are resolved, threads and collection order when it's not as pretty of a tree as I would like.

But based on my whiteboard sketching this should be doable.

webrtc.html
```diff
@@ -8777,13 +8780,27 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+                  track matches <var>selector</var>, and let <var>selector</var>
+                  be that sender or receiver. If no such sender or receiver
+                  exist, throw an <code>InvalidAccessError</code> exception and
+                  abort these steps.</p>
+                </li>
+                <li>
+                  <p>If <var>selector</var> was not created by
+                  <var>connection</var>, throw an
+                  <code>InvalidAccessError</code> exception and abort these
+                  steps.</p>
```
jan-ivar: The problem is *selector* may be a track then. And tracks are not created by *connection*.

webrtc.html
```diff
@@ -8792,15 +8809,12 @@ <h2 id="sec.stats-model">Statistics Model</h2>
                   <p>Run the following steps in parallel:</p>
```
jan-ivar: This follows [convention](https://www.w3.org/2001/tag/doc/promises-guide#explicit-async-steps). It means in parallel to JavaScript. We return *P as the last step.

webrtc.html
jan-ivar: Great! Compare to [mine](https://docs.google.com/presentation/d/1sfVD5VAHg1K7Hee37kooAbGEajfNIhaKUWmbIRZoioI/). The list is [here](https://github.com/w3c/webrtc-stats/issues/116#issuecomment-279084420) and is deterministic, so all the info should be known upfront about which stats to gather.

I count `RTCPeerConnectionStats`, `RTCDataChannelStats` and `RTCMediaStreamStats` as types not included. But more importantly is the multiple instances related to other senders/receivers a selector filters out.

jan-ivar:
Looks like an unrelated build-failure on Travis?

webrtc.html
```diff
@@ -8942,6 +8956,52 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
+            </li>
+          </ul>
+        </li>
+        <li>
+          Return <var>result</var>.
+        </li>
+      </ol>
+      <p>
+        A stats object is said to "correspond to" a selector if its "ssrc" stats
+        attribute matches an <code><a>ssrc</a></code> in the selector.
```
alvestrand: This needs a bit more language - SSRC needs to be one of the encodings on the sender, and needs to not match across transports.

jan-ivar: @alvestrand What do you mean when you say "needs to not match across transports"?

jan-ivar:
@alvestrand PTAL. Travis complains about unrelated build problem again.

webrtc.html
```diff
@@ -8777,13 +8775,25 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
-                  a valid <a>selector</a>, return a promise rejected with a
-                  <code>TypeError</code>.</p>
+                  <p>Let <var>connection</var> be the
+                  <code><a>RTCPeerConnection</a></code> object on which
+                  the method was invoked.</p>
+                </li>
+                <li>
+                 <p>If <var>selectorArg</var> is neither <code>null</code> nor
+                 a valid <a>MediaStreamTrack</a>, throw a <code>TypeError</code>
+                 and abort these steps.</p>
```
alvestrand: Isn't that what you're not supposed to do when you're returning a promise?

alvestrand:
Resolving the Travis issue may require a rebase.
Want clarification on the "throw error" vs "reject promise" question, then we can merge.

webrtc.html
jan-ivar: It's indistinguishable from a rejection [thanks to WebIDL](https://www.w3.org/2001/tag/doc/promises-guide#webidl-promise-return-values). That said, looking through this doc I see we mostly (*) use reject language, so I'll change it to be consistent.

*) [getIdentityAssertion](http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-getidentityassertion) returns a promise, and throws as part of its algorithm. This returns a rejected promise rather than throwing in the caller's JS.

jan-ivar:
Fixed and rebased.

Issue #1020, opened 2017-02-09
Don't fire events on a closed peer connection


jan-ivar:
To avoid this, whenever we *"Queue a task to run the following steps:"*, we immediately do:

 * If connection's [[isClosed]] slot is true, abort these steps.

We've adopted this approach, but we're not consistent. I found at least a couple of places where we don't do this:

 * The negotiationneeded event (we check [[needNegotiation]], but not [[isClosed]] [here](https://w3c.github.io/webrtc-pc/#dfn-update-the-negotiation-needed-flag)).
 * Resolve promise in [replaceTrack](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack).

There may be more places.

henbos:
Related: I found the [RTCPeerConnection::close()](https://rawgit.com/w3c/webrtc-pc/master/webrtc.html#dom-rtcpeerconnection-close) steps a bit unclear. It's doing a bunch of things, some which may or may not imply firing events, before reaching "_8. Set connection's [[isClosed]] slot to true._"

E.g. "_4. Set the state of all RTCIceTransports to closed._" Do you fire an RTCIceTransport statechange event or not?

jan-ivar:
> "4. Set the state of all RTCIceTransports to closed." Do you fire an RTCIceTransport statechange event or not?

@henbos Event firing is explicit, so no. The above merely sets the state.

`pc.close()` is always user-invoked now, and by design we don't fire events on user-invoked actions.

Unless some of the algorithms invoked include steps to fire events, then no events are fired. If you find one that fires an event, then that's probably a bug, as I'm pretty sure our design intent is to not cause events on close.

alvestrand:
There are more things  that need to be fixed even after #1031 is merged.

adam-be:
We also need to check the closed state when replaceTrack queues a task to do the actual swap of `track` to `withTrack` (replaceTrack step 9.3).

jhaenchen:
Am I understanding this properly? If I have an RTCPeerConnection with another client and an ongoing data channel connection with that same other client and the other client then force quits their browser, `oniceconnectionstatechange` will fire with `iceConnectionState = failed`. Then, if I call `.close` on the data channel, `dataChannel.readyState` goes to `closing` but never transitions to `closed`, I assume because the PeerConnection itself is broken but the data channel doesn't realize that.

Is this intended behavior?

jhaenchen:
[...]

PR #1031, opened 2017-02-15
Don't fire events on a closed peer connection

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1020

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -743,6 +743,11 @@
             <p>Let <var>p</var> be a new promise.</p>
           </li>
           <li>
+            <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+            <code>true</code>, reject <var>p</var> with
+            <code>InvalidStateError</code> and abort these steps.</p>
+          </li>
+          <li>
             <p>In parallel, start the process to apply <var>description</var>
             as described in <span data-jsep=
             "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>.</p>
@@ -752,10 +757,6 @@
                 reason, then user agent MUST queue a task runs the following
                 steps:</p>
                 <ol>
-                  <li>
-                    <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
-                    <code>true</code>, then abort these steps.</p>
-                  </li>
                   <li>
                     <p>If elements of the SDP were modified, then reject
                     <var>p</var> with an <code>InvalidModificationError</code>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5241,7 +5241,8 @@ sender.setParameters(params)
                 </li>
                 <li>
                   <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
-                  <code>true</code>, abort these steps.</p>
+                  <code>true</code>, return a promise rejected with an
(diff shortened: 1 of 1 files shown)


aboba:
@jan-ivar Can you review?  Are there other places where we need to check for a closed peer connection?

webrtc.html
```diff
@@ -5236,6 +5240,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   <var>sender</var>.</p>
                 </li>
                 <li>
+                  <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+                  <code>true</code>, abort these steps.</p>
+                </li>
+                <li>
```
jan-ivar: Great! I missed this one. I was actually thinking of step 10.3 below here:

*"Queue a task that sets sender's track attribute to withTrack, and resolves p with undefined."*

webrtc.html
```diff
@@ -3295,6 +3295,10 @@
         <var>connection</var>, run the following steps:</p>
         <ol>
           <li>
+            <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+            <code>true</code>, abort these steps.</p>
+          </li>
+          <li>
```
jan-ivar: This is great. Unfortunately, there's also this doozie of an exception in step 2.7 of the [set description algorithm](http://w3c.github.io/webrtc-pc/#set-description):

*"If connection's signaling state is now stable, update the negotiation-needed flag. If connection's [[ needNegotiation]] slot was true both before and after this update, **queue a task to check connection's [[needNegotiation]] slot and, if still true, fire a simple event named negotiationneeded at connection**."*

webrtc.html
```diff
@@ -743,6 +743,11 @@
             <p>Let <var>p</var> be a new promise.</p>
           </li>
           <li>
+            <p>If <var>connection</var>'s [[<a>isClosed</a>]] slot is
+            <code>true</code>, reject <var>p</var> with
+            <code>InvalidStateError</code> and abort these steps.</p>
+          </li>
+          <li>
```
adam-be: We don't need to check the closed state here since the 'enqueue' algorithm does the check before executing the operation.

webrtc.html
adam-be: I think we want to return a rejected promise here since it's an API call, not just abort.

And in the replaceTrack step 9.3, we need the queued task to do a closed state check before resolving the promise.

jan-ivar: @adam-be My understanding is we don't resolve asynchronous operations on a closed peer connection. At least that's the case for `setLocalDescription` and `setRemoteDescription`.

E.g. step 2.1.1. in http://w3c.github.io/webrtc-pc/#set-description
> If connection's [[isClosed]] slot is true, then abort these steps.

webrtc.html
jan-ivar: Good catch!

webrtc.html
alvestrand: I'm reading this as "whenever we do "queue a task to...", we need to say "queue a task that .... checks whether [[isClosed]] is true and then..."
Relegating fixing that to another PR.

adam-be:
We also need to check the closed state when replaceTrack queues a task to do the actual swap of `track` to `withTrack` (replaceTrack step 9.3.

PR #1037, opened 2017-02-23
Specify when random mid generation happens

### LINKED ISSUE #578: Need to specify precisely when MID generation happens
In #535 the algorithm is not updated to say exactly when MID generation happens. It's more implementable if it does.

adam-be:
Fixes #578

Note: Only review the top commit (based on non-master branch)


aboba:
LGTM.

PR #1038, opened 2017-02-24
Make RTCDataChannel.id nullable and describe when it's set.

### LINKED ISSUE #795: Specify when a data channel's ID is assigned, and what the `id` attribute returns when no ID is assigned.
This is an issue I noticed when working on a solution for issue #746.

The [WebRTC data channel establishment spec](https://tools.ietf.org/html/draft-ietf-rtcweb-data-protocol-08#section-6) requires that DTLS clients choose even stream IDs, and DTLS servers choose odd IDs.

However, the DTLS role may not be known at the time of creating a data channel. So, until the role is determined, what does the `id` attribute return? It's worth noting that both Chrome and Firefox currently return an ID of 65535, which is guaranteed to be out of range (the maximum usable ID is 65534). But it seems more "correct" to make the attribute nullable.

And then, when should a user script expect the ID to be populated? When the stable signaling state is reached with an SCTP media description negotiated? That would seem reasonable to me.

taylor-b:
Fixes #795.

When a data channel is initially created, and a ID isn't explicitly
provided, it will only be set after the DTLS role is negotiated by an
SDP answer. If a data channel is created after this answer is set, an
ID will be set instantly.

Until the id is set, the attribute will return null.

In the rare case that there are no IDs left for the given DTLS role
(for example, all even IDs are taken), a ResourceInUse exception
will be thrown (if this happens in `createDataChannel`) or fired in
an `error` event (if this happens in `setLocalDescription` or
`setRemoteDescription`).


aboba:
LGTM.

PR #1039, opened 2017-02-25
Clarify which timestamp RTCStats.timestamp represents.

### LINKED ISSUE #729: RTCStats timestamp source ambiguous
RTCStats has a timestamp member -- "has a timestamp value [...] associated with this object".  It's somewhat unclear what the source of this timestamp should be - perhaps the time the data returned for the child object of RTCStats was last updated.  Even that has some ambiguity in the case of remote RTP stats - is it the time they were measured (NTP time from the RTCP packet they came in on), or the time they were received locally?

Currently in Firefox we use the RTCP NTP time.  I haven't reviewed where we get the timestamp values for other stats.

We could expand this definition slightly, or push that down to the Stats spec to define for each derived dictionary what RTCStats.timestamp should be sourced from.

taylor-b:
Fixes #729.

"timestamp" always refers to the locally measured timestamp. In the case
of stats with a remote source, this means the timestamp at which the
RTCP packet was received.

The remote timestamp will be added in a separate webrtc-stats PR:
https://github.com/w3c/webrtc-stats/pull/164

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8929,7 +8929,12 @@ interface RTCDTMFToneChangeEvent : Event {
               <p>The <dfn><code>timestamp</code></dfn>, of type
               <code>DOMHighResTimeStamp</code> [[!HIGHRES-TIME]], associated
               with this object. The time is relative to the UNIX epoch (Jan 1,
-              1970, UTC).</p>
+              1970, UTC). For statistics that came from a remote source (e.g.,
+              from received RTCP packets), <code>timestamp</code> represents
+              the time at which the information arrived at the local endpoint.
+              The remote timestamp can be found in an additional field in an
+              <code><a>RTCStats</a></code>-derived dictionary, if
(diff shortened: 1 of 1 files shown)


alvestrand:
lgtm

webrtc.html
```diff
@@ -8929,7 +8929,12 @@ <h2 id="sec.stats-model">Statistics Model</h2>
               <p>The <dfn><code>timestamp</code></dfn>, of type
               <code>DOMHighResTimeStamp</code> [[!HIGHRES-TIME]], associated
               with this object. The time is relative to the UNIX epoch (Jan 1,
```
alvestrand: Related to https://github.com/w3c/webrtc-stats/issues/134 - can you replace "associated with this object" with "that indicates the time at which these values were measured"?

alvestrand:
Merging this and asking for a separate pull to cover the comment.

Issue #1040, opened 2017-03-01
Codecs supporting multiple clock rates/packetization-mode


aboba:
What happens if a codec supports multiple clock rates or encoding types?

Examples: 

- CN or SILK with clock rates of 8000 and 16000
- H.264/AVC implementation supporting packetization-mode of 0 or 1

Question: How do we differentiate these codecs in the following methods:
static RTCRtpCapabilities getCapabilities(DOMString kind);
void setCodecPreferences(sequence RTCRtpCodecCapability codecs);

Do we need to add attributes? If so, which can setCodecPreferences() set?

PR #1042, opened 2017-03-01
Add clockrate, channels, sdpFmtpLine to codec capability

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1040

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5837,9 +5837,16 @@ sender.setParameters(params)
           <h2>Dictionary <dfn>RTCRtpCodecCapability</dfn>
           Members</h2>
           <p>The <code>RTCRtpCodecCapability</code> dictionary provides
-          information about codec capabilities. Only capability combinations
-          utilizing a distinct payload type in the corresponding SDP
-          are provided.</p>
+          information about codec capabilities. Only capability
+          combinations that would utilize distinct payload types in a
+          generated SDP offer are provided.  For example:</p>
+          <ol>
+            <li>Exactly one Opus codec, with "stereo=1", indicating that
+              the browser supports either stereo or mono.</li>
+            <li>Two H.264/AVC codecs, one for each of two supported
+              packetization-mode values.</li>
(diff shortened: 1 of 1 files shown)


ibc:
Great. But, why `sdpFmtpLine` instead of a `parameters` object/dictionary`?

aboba:
Sentiment at March 1 WEBRTC WG Virtual Interim was to use DOMString sdpFmtpLine rather than a  parameters dictionary.   Since neither setParameters() nor setCodecPreferences() allows modification of the sdpFmtpLine string, the value of going with  a dictionary is reduced.  Also, there was some concern about proprietary strings that  might not fit the attribute/value pair paradigm.

webrtc.html
```diff
@@ -5829,10 +5829,17 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+             unsigned short channels = 1;
+             DOMString      sdpFmtpLine;
 };</pre>
         <section>
           <h2>Dictionary <dfn>RTCRtpCodecCapability</dfn>
           Members</h2>
+          <p>The <code>RTCRtpCodecCapability</code> dictionary provides
+          information about codec capabilities. Only capability combinations
+          utilizing a distinct payload type in the corresponding SDP
+          are provided.</p>
```
taylor-b: I think this wording could be improved, since if `getCapabilities` is called before any offer/answer is created, there isn't any corresponding SDP yet.

Maybe something like:

> Only capability combinations that would utilize distinct payload types in a generated SDP offer are provided.

Or, to use less SDP-specific terminology:

> Exactly one `RTCRtpCodecCapability` will be present for each supported combination of parameters that must be identical between the offer and answer.

Some examples may also be helpful. For example:

* Exactly one Opus codec, with "stereo=1", indicating that the browser supports either stereo or mono.
* Two H.264 codecs, one for each packetization mode.
* Two CN codecs with different clock rates

PR #1045, opened 2017-03-02
Clarifying exactly what "sdpFmtpLine" represents.

### LINKED ISSUE #1022: sdpFmtpLine isn't very convenient
As currently written, it sounds like this attribute should return the _entire_ "a=fmtp" line. For example:

"a=fmtp:96 foo=1;bar=2"

This seems unnecessary. The payload type (96) is already present in RTCRtpCodecParameters. I'd propose this attribute only returns the "format specific parameters" (the thing that comes after "a=fmtp:PT ").

Or, even better, if we can be confident that all codecs will use parameters in a "key=value" form, we could make this attribute a dictionary instead of an opaque string, as it is in ORTC.

taylor-b:
Fixes #1022.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5912,8 +5912,13 @@ sender.setParameters(params)
             <dt><dfn><code>sdpFmtpLine</code></dfn> of type <span class=
             "idlMemberType"><a>DOMString</a></span></dt>
             <dd>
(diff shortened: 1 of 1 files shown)


aboba:
LGTM.

henbos:
LGTM

PR #1047, opened 2017-03-02
Adding "[EnforceRange]" to RTCDataChannelInit.id.

### LINKED ISSUE #1046: RTCDataChannelInit.id should use [EnforceRange]
We already throw an exception if the ID is 65535 (valid SCTP stream ids are 0-65534), so we should definitely do so if the ID is outside the 0-65535 range.

taylor-b:
Fixes #1046.

Without doing this, values outside the 0-65535 range will be
converted to that range using modulo arithmetic, which isn't
something we want.


aboba:
LGTM.

PR #1054, opened 2017-03-03
Throw InvalidModificationError if changing pool size after SLD.

### LINKED ISSUE #1049: Description of handling iceCandidatePoolSize in setConfiguration is out of sync with JSEP
In JSEP, we decided that changing the candidate pool size is only allowed before setLocalDescription is called, and as a consequence, the pool will only be used for the first offer/answer exchange.

But webrtc-pc currently has a step in `setConfiguration` that says simply:

> Set the ICE Agent's prefetched ICE candidate pool size as defined in [JSEP] to the value of configuration.iceCandidatePoolSize. If the new ICE candidate pool size changes the existing setting, this may result in immediate gathering of new pooled candidates, or discarding of existing pooled candidates, as defined in [JSEP].

... Which doesn't say what to do if `setLocalDescription` has already been called.

I'd suggest throwing an InvalidModificationError like we do for everything else.

taylor-b:
Fixes #1049.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2143,6 +2143,13 @@ interface RTCPeerConnection : EventTarget  {
                   "RTCConfiguration">rtcpMuxPolicy</a></code> is set and its
                   value differs from the <var>connection</var>'s rtcpMux
                   policy, throw an <code>InvalidModificationError</code>.</li>
+                  <li>If <code><var>configuration</var>.<a data-for=
+                  "RTCConfiguration">iceCandidatePoolSize</a></code> is set and
+                  its value differs from the <var>connection</var>'s previously
+                  set <code>iceCandidatePoolSize</code>, and <code><a data-for=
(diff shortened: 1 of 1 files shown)


aboba:
@taylor-b Since we don't allow connection.peerIdentity, certificates, bundlePolicy or rtcpMuxPolicy to be changed (regardless of whether setLocalDescription has been called yet), it might be simpler just to prohibit changing the iceCandidatePoolSize.

taylor-b:
@aboba That was my initial thought, but there are some potential situations where changing the size before SLD is desirable. For example, an application may want to create a PeerConnection as early as possible for various reasons, but only set the candidate pool size when a "call" is imminent, since setting it too early would result in STUN/TURN candidates that need to be kept alive. So, this is what the JSEP editors ended up going with.

aboba:
Makes sense.

PR #1055, opened 2017-03-03
Changing iceCandidatePoolSize to an octet and adding EnforceRange.

### LINKED ISSUE #1048: Range checking needed for iceCandidatePoolSize
For one, `iceCandidatePoolSize` should be using `[EnforceRange]`, so that (for example) a value of -1 isn't converted to 65535 at the WebIDL level.

We also might want to restrict the 0-65535 range further. There shouldn't be any scenario where more than a handful of pooled candidates will be needed, and allowing WebRTC to take up every single available socket is probably not desirable.

taylor-b:
Fixes #1048.

Changing it to an octet limits the range to 0-255; there shouldn't be
any situation where an application wants to create thousands of pooled
candidates.

Adding [EnforceRange] means that if the application provides a negative
value or one greater than 255, a TypeError is thrown, rather than the
value being converted to the valid range by doing "mod 256".

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -111,7 +111,8 @@
              RTCRtcpMuxPolicy         rtcpMuxPolicy = "require";
              DOMString                peerIdentity;
              sequence&lt;RTCCertificate&gt; certificates;
-             unsigned short           iceCandidatePoolSize = 0;
+             [EnforceRange]
+             octet                    iceCandidatePoolSize = 0;
 };</pre>
           <section>
             <h2>Dictionary <a class="idlType">RTCConfiguration</a> Members</h2>


aboba:
@jan-ivar  Agree that it is silly to allow more than 255 candidates, but wondering if there are any backward compatibility issues with changing from unsigned short to octet.

jan-ivar:
@aboba Shouldn't be. No browser ([FF](https://bugzilla.mozilla.org/show_bug.cgi?id=1291894), [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=673395)) implements this yet anyway.

jan-ivar:
I didn't check Edge, but there shouldn't be any change in behavior from this other than numbers > 255 passed as arguments causing `TypeError`.

aboba:
LGTM.

PR #1056, opened 2017-03-06
Switching to new, consistent terminology when talking about exceptions.

### LINKED ISSUE #845: "Throw a FooError" steps not written in a consistent manner
I've seen the following:
- "throw a FooError."
- "[throw](http://heycam.github.io/webidl/#dfn-throw) a FooError." (which is the example wording suggested in WebIDL)
- "then throw a FooError exception and abort these steps."
- "throw a FooError and abort these, and all further steps."

And for promises, we have:
- "reject _p_ with FooError, and abort these steps"
- "reject _p_ with a DOMException object whose name attribute has the value FooError, and abort these steps"
- "reject _p_ with a DOMException object whose name attribute has the value FooError and stop processing any more steps"
- "reject _p_ with a DOMException object that has a name of FooError, and abort these steps"

Which formats do we want to use consistently?

For throwing an exception, I'd suggest:

> "[throw](http://heycam.github.io/webidl/#dfn-throw) a FooError."

The link makes it clear what this means, and "abort these steps" is redundant since it's implied that steps are aborted when an exception is thrown.

And for rejecting a promise, I'd suggest:

> "reject _p_ with a newly [created](https://heycam.github.io/webidl/#dfn-create-exception) FooError and abort these steps." 

This also uses wording suggested in WebIDL, and it's pretty short/sweet/to-the-point.

taylor-b:
Fixes #845.

Throwing exceptions will use the phrase "throw a FooError", which
implies that the current steps will be aborted, making it unnecessary
to clarify that.

Rejecting a promise with an exception will use the phrase "reject p
with a newly created FooError and abort these steps."

Each place where these phrases are used links to the "Terminology"
section, which links to where these terms are defined in WebIDL.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -84,6 +84,11 @@
     <p>The term <dfn>Blob</dfn> is defined in [[!FILEAPI]].</p>
     <p>The term <dfn>media description</dfn> is defined in [[!RFC4566]].</p>
     <p>The term <dfn>generation</dfn> is defined in [[!TRICKLE-ICE]] Section 2.</p>
+    <p>When referring to exceptions, the terms <dfn><a
+    href="https://www.w3.org/TR/WebIDL-1/#dfn-throw">throw</a></dfn> and
+    <dfn data-dfn-for="exception"><a href=
+    "https://www.w3.org/TR/WebIDL-1/#dfn-create-exception">create</a></dfn> are
+    defined in [[!WEBIDL-1]].</p>
   </section>
   <section>
     <h2>Peer-to-peer connections</h2>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -83,6 +83,11 @@
     <dfn>MediaStreamConstraints</dfn> are defined in [[!GETUSERMEDIA]].</p>
     <p>The term <dfn>Blob</dfn> is defined in [[!FILEAPI]].</p>
     <p>The term <dfn>media description</dfn> is defined in [[!RFC4566]].</p>
+    <p>When referring to exceptions, the terms <dfn><a
+    href="https://www.w3.org/TR/WebIDL-1/#dfn-throw">throw</a></dfn> and
```
taylor-b: Is this the right URL to link to?

webrtc.html
```diff
@@ -83,6 +83,11 @@
...
+    <dfn data-dfn-for="exception"><a href=
```
taylor-b: Is this an acceptable use of "data-dfn-for"? It seemed necessary to scope the "create" definition since it could apply to many things besides exceptions.

webrtc.html
aboba: I think so.

@adam-be Do you agree?

taylor-b: My concern was that this links to "latest published version", and there's no hard guarantee that a future published version won't remove the "dfn"s being linked to.

alvestrand:
@dontcallmedom can you check if the WebIDL references look right to you?

@taylor-b will rebase and merge.

dontcallmedom:
FWIW the WebIDL references look right to me :)

PR #1057, opened 2017-03-07
Add clockrate, channels, sdpFmtpLine to codec capability

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1040

Rebase of PR https://github.com/w3c/webrtc-pc/pull/1042

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5957,8 +5957,6 @@ sender.setParameters(params)
           combinations that would utilize distinct payload types in a
           generated SDP offer are provided.  For example:</p>
           <ol>
-            <li>Exactly one Opus codec, with "stereo=1", indicating that
-              the browser supports either stereo or mono.</li>
             <li>Two H.264/AVC codecs, one for each of two supported
               packetization-mode values.</li>
             <li>Two CN codecs with different clock rates.</li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5945,10 +5945,24 @@ sender.setParameters(params)
       <div>
         <pre class="idl">dictionary RTCRtpCodecCapability {
              DOMString mimeType;
+             unsigned long  clockRate;
+             unsigned short channels = 1;
+             DOMString      sdpFmtpLine;
 };</pre>
         <section>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5945,10 +5945,24 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       <div>
         <pre class="idl">dictionary RTCRtpCodecCapability {
              DOMString mimeType;
+             unsigned long  clockRate;
+             unsigned short channels = 1;
```
taylor-b: Not sure this needs a default value since it's always returned by the browser.

webrtc.html
```diff
@@ -5945,10 +5945,24 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
         <section>
           <h2>Dictionary <dfn>RTCRtpCodecCapability</dfn>
           Members</h2>
+          <p>The <code>RTCRtpCodecCapability</code> dictionary provides
+          information about codec capabilities. Only capability
+          combinations that would utilize distinct payload types in a
+          generated SDP offer are provided.  For example:</p>
+          <ol>
+            <li>Exactly one Opus codec, with "stereo=1", indicating that
+              the browser supports either stereo or mono.</li>
```
taylor-b: I think I was wrong about this example.

> Independent of the stereo parameter every receiver MUST be able to receive and decode stereo signals

So I suppose the "stereo" parameter in the capabilities would only represent the default preference of the browser. Maybe "max-fr" and "max-fs" with VP8 would be a better example.

webrtc.html
aboba: I put in a default value since setCodecPreferences takes a sequence of RTCRtpCodecCapability as an argument.

webrtc.html
aboba: I removed the Opus example.

webrtc.html
taylor-b: Oh, right. Not an issue if the application just takes capabilities and reorders them, but it might create the dictionaries from scratch.

Issue #1052, opened 2017-03-03
Section 10.3: Issue 5


aboba:
In Section 10.3 it says: 

"When a track comes from a remote peer and the remote peer has permanently stopped sending data the ended event must be fired on the track, as specified in [GETUSERMEDIA].

ISSUE 5
How do you know when it has stopped? This seems like an SDP question, not a media-level question. (Suggestion: when the track is ended, either through port 0, or removing the a=msid attrib)"

aboba:
I think we can remove the ISSUE 5 note, because the issue is clarified at various places in the specification.  For example:

removeTrack: 

"Stops sending media from sender. The RTCRtpSender will still appear in getSenders. Doing so will cause future calls to createOffer to mark the media description for the corresponding transceiver as recvonly or inactive, as defined in [JSEP] (section 5.2.2.). When the other peer stops sending a track in this manner, an ended event is fired at the MediaStreamTrack object."

transceiver.stop(): 

"When a remote description is applied with a zero port in the media description for the corresponding transceiver, as defined in [JSEP] (section 4.2.2), the user agent must run the above steps as if stop had been called. However in addition, the transceiver.receiver.track.onended Eventhandler is fired."

[BA] The last sentence should probably be: 

"In addition, the ended event is fired at transceiver.receiver.track."

taylor-b:
> When the other peer stops sending a track in this manner, an ended event is fired at the MediaStreamTrack object

This seems wrong. If the other peer sets a track again, and the direction changes to "sendrecv", what happens? The remote track can't be un-ended. So, I don't think remote tracks should be ended unless a transceiver is stopped.

aboba:
@taylor-b  Right.  Elsewhere we say that only transceiver.stop() results in the "ended" event. So both paragraphs should be deleted.

PR #1060, opened 2017-03-07
ended event fired on the track

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1052

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6486,9 +6486,8 @@ sender.setParameters(params)
               transceiver, as defined in <span data-jsep="stopped">
               [[!JSEP]]</span>, the user agent MUST run the
               above steps as if <code>stop</code> had been called.
-              However in addition, the 
-              <code><var>transceiver</var>.receiver.track.onended</code>
-              Eventhandler is fired.</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6486,9 +6486,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
               transceiver, as defined in <span data-jsep="stopped">
               [[!JSEP]]</span>, the user agent MUST run the
               above steps as if <code>stop</code> had been called.
-              However in addition, the 
-              <code><var>transceiver</var>.receiver.track.onended</code>
-              Eventhandler is fired.</p>
+              In addition, the <code>ended</code> event is fired on
+              <code><var>transceiver</var>.receiver.track</code>.</p>
```
taylor-b: Is there a way this could point to https://www.w3.org/TR/mediacapture-streams/#track-ended, to the steps that begin:

> When a MediaStreamTrack track ends for any reason other than the stop() method being invoked...

aboba: Yes, that would be better.

stefhak:
LGTM.

PR #1071, opened 2017-03-09
Specify behavior if browser doesn't implement "negotiate" rtcpMuxPolicy.

### LINKED ISSUE #1065: Need to specify what happens if the browser doesn't implement "negotiate" rtcpMuxPolicy
JSEP states:

> Implementations MAY choose to reject attempts by the application to set the multiplexing policy to "negotiate".

So, what does this mean at the API level? A "NotSupportedError" being thrown by the RTCPeerConnection constructor?

taylor-b:
Fixes #1065.

Also moved some steps around in the RTCPeerConnection constructor
steps, such that everything dealing with the RTCConfiguration is in one
place.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -513,7 +513,7 @@
           </li>
           <li>
             <p>If the <code>certificates</code> value in
-            <var>configuration</code> is non-empty, check that
+            <var>configuration</var> is non-empty, check that
             the <code>expires</code> on each value is in the future. If a
             certificate has expired, <a>throw</a> an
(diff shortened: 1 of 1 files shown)


aboba:
getConfiguration() obtains the configuration from the configuration internal slot, so I don't think you can delete storing the configuration in the configuration internal slot in lines 571 through 574.

taylor-b:
@aboba The "set the configuration" steps already do this though, so it was redundant.

stefhak:
LGTM.

taylor-b:
JSEP reference has been fixed. Ok to merge?

Issue #1079, opened 2017-03-15
Handling RTX in RTCRtpCodecCapabilities


aboba:
Now that we have added `clockRate` and `sdpFmtpLine` to `RTCRtpCapabilities` there is a question of what the values should be for RTX.  For example, is `sdpFmtpLine` set to the null string?

PR #1082, opened 2017-03-15
Handling RTX in RTCRtpCodecCapabilities

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1079

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5969,9 +5969,9 @@ sender.setParameters(params)
             "idlMemberType">sequence&lt;<a>RTCRtpCodecCapability</a>&gt;</span></dt>
             <dd>
               <p>Supported media codecs as well as entries for RTX, RED and FEC
-              mechanisms. For retransmission via RTX, there will be an entry
-              within <code>codecs[]</code> for each codec being retransmitted,
-              with <code>sdpFmtpLine</code> containing the parameters.</p>
+              mechanisms. There will only be a single entry in
+              <code>codecs[]</code> for retransmission via RTX, with
+              <code>sdpFmtpLine</code> set to "".</p>
             </dd>
             <dt><dfn><code>headerExtensions</code></dfn> of type <span class=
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5969,8 +5969,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             "idlMemberType">sequence&lt;<a>RTCRtpCodecCapability</a>&gt;</span></dt>
             <dd>
               <p>Supported media codecs as well as entries for RTX, RED and FEC
-              mechanisms. There will only be a single entry in
-              <code>codecs[]</code> for retransmission via RTX.</p>
+              mechanisms. For retransmission via RTX, there will be an entry
+              within <code>codecs[]</code> for each codec being retransmitted,
+              with <code>sdpFmtpLine</code> containing the parameters.</p>
```
taylor-b: What value goes in "apt" though? `RTCRtpCodecCapability` doesn't have a payload type that can be referred to.

I'd propose just making a special case for RTX, and saying "there's only one entry, and it doesn't have an 'apt' attribute because it represents general support for RTX with any codec".

aboba: @taylor-b Maybe the single "rtx" entry has sdpFmtpLine as the null string?  Not clear having more than one entry provides additional useful functionality.

taylor-b: @aboba As I recall, the working group discussed this at some point and no one came up with a use case for supporting RTX with some codecs and not others. So one entry with `sdpFmtpLine` undefined (or null, whatever is most consistent) sounds good to me.

aboba: @taylor-b Updated with `sdpFmtpLine` set to "".  Wondering if `channels` and `clockrate` should be undefined.

taylor-b: @aboba Is there a reason it should be an empty string rather than undefined? Not that it really matters, just wondering.

aboba:
Needs fixing: sdpFmtpLine should not be "".

PR #1088, opened 2017-03-17
Make "candidate" non-nullable in addIceCandidate parameter table.

### LINKED ISSUE #1077: Candidate from onicecandidate event and addIceCandidate are incompatible
The `addIceCandidate` method states:

> This method can also be used to indicate the end of remote candidates when called with an empty string for candidate.

First of all, do you mean the **member** `candidate` can be empty string or the **argument** `candidate` can be empty string? I'm guessing you mean the member but please clarify this in the description.
Please also state what should happen in case the argument `candidate` is `null`. Is this still an end of candidate indication but for all mids (as it was before)? I guess it would have to be to ensure backwards compatibility.

To the main issue: `RTCPeerConnectionIceEvent` contains an attribute `candidate` which states:

> This attribute is set to null when an event is generated to indicate the end of candidate gathering.

I guess this has just not been updated to the changes to `addIceCandidate`. I would expect the `candidate` attribute to always contain an `RTCIceCandidate` which cannot be `null` (instead, `candidate.candidate` would be an empty string).

taylor-b:
Fixes #1077.

It's non-nullable everywhere else, but we forgot to update this table.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2143,8 +2143,7 @@ interface RTCPeerConnection : EventTarget  {
                     method was invoked.</p>
                   </li>
                   <li>
-                    <p>If <var>candidate</var> is not <code>null</code> and
-                    both <var>sdpMid</var> and <var>sdpMLineIndex</var> are
+                    <p>If both <var>sdpMid</var> and <var>sdpMLineIndex</var> are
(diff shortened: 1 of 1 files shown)


taylor-b:
Fixed something else @lgrahl noticed. I guess I was a little sloppy.

PR #1094, opened 2017-03-23
Check RTCPeerConnection isClosed slot before running queued tasks

### LINKED ISSUE #1020: Don't fire events on a closed peer connection

adam-be:
Fixes  #1020

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1281,10 +1281,22 @@
                     <code>stable</code>, <a>update the negotiation-needed
                     flag</a>. If <var>connection</var>'s
                     [[<a>needNegotiation</a>]] slot was <code>true</code> both
-                    before and after this update, queue a task to check
-                    <var>connection</var>'s [[<a>needNegotiation</a>]] slot and,
-                    if still <code>true</code>, fire a simple event named
-                    <a>negotiationneeded</a> at <var>connection</var>.</p>
+                    before and after this update, queue a task that runs the
+                    following steps:</p>
+                    <ol>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5676,10 +5698,22 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                        <li>
+                          <p>Set <var>sender</var>'s <code><a data-for=
+                          "RTCRtpSender">track</a></code> attribute to
+                          <var>withTrack</var>.</p>
+                        </li>
+                        <li>
+                          <p>Resolve <var>p</var> with
+                          <code>undefined</code>.</p>
+                        </li>
+                      </ol>
```
adam-be: It's debatable if this check is needed.

taylor-b: We do this in other places before promises are resolved, so it seems consistent.

stefhak:
LGTM.

PR #1098, opened 2017-03-29
Attempt to update RTCRtpContributingSource objects at playout time.

### LINKED ISSUE #1085: RTCRtpContributingSource.audioLevel not guaranteed to be in sync with audio playout
My assumption is that this feature exists so that applications can show audio level UI indications for different participants of a call.

However, I don't see how this can be done in a robust manner, since the `RTCRtpContributingSource` objects are updated whenever a packet is received and not when audio is played out, with how the spec currently reads:

> Each time an RTP packet is received, the RTCRtpContributingSource objects are updated.

Consider these situations:

1. There is a noticeable delay between packets being received and audio playing out, due to poor network conditions, resulting in an audioLevel that's updated well in advance of audio playout; e.g., you see the volume indicator move before the speaker opens their mouth.
2. Traffic is bursty, resulting in the audio level jumping around when there's a burst of traffic, then remaining stagnant for a while.
3. Packets arrive out of order... and `timestamp` actually decreases?

How can these problems be mitigated? Could we change "Each time an RTP packet is received" to "each time a frame of media is delivered to the MediaStreamTrack" (or whatever the right terminology there is)?

Otherwise, what can an application do? Use `getStats` to figure out the playout delay, and then delay updating the audio level UI for that amount of time?

taylor-b:
Fixes #1085.

May not be the correct terminology, but the intention is that the
contributing source objects are updated at playout time, such that if
an application is using them to drive an audio level UI, that UI will be
in sync with the audio played out by the browser.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6596,11 +6596,13 @@ sender.setParameters(params)
         </section>
       </div>
       <p>The <dfn>RTCRtpContributingSource</dfn> objects contain information
-      about a given contributing source, including the time the most recent
-      time a packet was received from the source. The browser MUST keep
-      information from RTP packets received in the previous 10 seconds. Each
-      time an RTP packet is received, the
-      <code><a>RTCRtpContributingSource</a></code> objects are updated. The
+      about a given contributing source, including the most recent time a
+      packet that the source contributed to was played out. The browser MUST
+      keep information from RTP packets received in the previous 10 seconds.
+      When the first audio frame contained in an RTP packet is delivered to the
+      <code><a>RTCRtpReceiver</a></code>'s <code><a>MediaStreamTrack</a></code>
+      for playout, the relevant <code><a>RTCRtpContributingSource</a></code>
+      objects are updated. The
       <code><a>RTCRtpContributingSource</a></code> object corresponding to the
       SSRC is always updated, and if the RTP packet contains CSRCs, then the
       <code><a>RTCRtpContributingSource</a></code> objects corresponding to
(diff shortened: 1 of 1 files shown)


stefhak:
I would like someone that has been involved in the discussion about CSRC to review.

burnburn:
As I said on the call, I believe this is a slippery slope that we will regret, since in general we have no control over when any MST will play out; it may be duplicated, sent elsewhere, delayed via other APIs, etc.  I'm not going to officially object since I don't have a good alternative to propose, but I reserve the right to say 'I told you so' :)

taylor-b:
I understand the concern. Maybe calling it "playout time" is the mistake, but what we ultimately need is _some_ well-defined, post-jitter-buffer point of time, which the application has a way of correlating to playout time.

Even if the application does something with the MediaStreamTrack that introduces extra delay, the original problem is solved as long as the application has a way of estimating the additional delay and factoring it in when updating its audio level UI.

alvestrand:
We may get implementor feedback on this item, but it clearly reflects WG consensus.
Can be merged once conflicts are resolved.

taylor-b:
Merge conflicts resolved.

vr000m:
What is the motivation of accessing this on the getContributingSources (I suppose this is the value from the RtpHeader extension instead of the actual audio level).

Clarifying: are we adding audioLevel to both the webrtc-pc and webrtc-stats APIs? I feel in this case, having the audio levels accessible in two places is just confusing. 

See https://github.com/w3c/webrtc-stats/pull/194/

taylor-b:
> What is the motivation of accessing audioLevel on the getContributingSources (I suppose this is the value from the RtpHeader extension instead of the actual audio level).

For CSRCs, it's the value from the RFC6465 extension, yes. The motivation is that applications may want information about the speakers contributing to a mixed audio stream, which may be reflected in the application's UI.

The original reason to put the API on `RTCRtpReceiver` and not `getStats` was that we didn't want applications to be reliant on `getStats` for basic WebRTC functionality, since `getStats` calls may be expensive and time-consuming.

And I suggested adding the contributing sources to `getStats` as well, because we can put more granular information there, and it could provide value to services (like callstats.io?) that only have access to stats, and not individual `RTCRtpReceiver`s. An example of it being more granular: the working group decided that even if the `RTCRtpContributingSource`s are updated "post jitter buffer", the stats objects should be updated as soon as packets are received, which is consistent with other things in the stats.

PR #1099, opened 2017-03-29
Always update the RTCRtpContributingSource for SSRCs.

### LINKED ISSUE #1091: When exactly is an SSRC RTCRtpContributingSource object updated?
Current text:

> Each time an RTP packet is received, the RTCRtpContributingSource objects are updated. If the RTP packet contains CSRCs, then the RTCRtpContributingSource objects corresponding to those CSRCs are updated. If the RTP packet contains no CSRCs, then the RTCRtpContributingSource object corresponding to the SSRC is updated.

Is this meant to imply that the RTCRtpContributingSource object corresponding to the SSRC is *only* updated if a packet contains no CSRCs? I'd suggest having it always updated. There's no extra implementation burden and applications may want to know the SSRC audio level even when CSRCs are present.

taylor-b:
Fixes #1091.

As described in the issue, there may be cases where an application
wants to know the SSRC audio level, even when a CSRC is present in the
packet.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6565,12 +6565,11 @@ sender.setParameters(params)
       time a packet was received from the source. The browser MUST keep
       information from RTP packets received in the previous 10 seconds. Each
       time an RTP packet is received, the
-      <code><a>RTCRtpContributingSource</a></code> objects are updated. If the
-      RTP packet contains CSRCs, then the
-      <code><a>RTCRtpContributingSource</a></code> objects corresponding to
-      those CSRCs are updated. If the RTP packet contains no CSRCs, then the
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

fluffy:
LGTM

aboba:
@taylor-b Do we need to update the PR now that we're going to use getSources() to retrieve the SSRC info?

taylor-b:
@aboba I don't think it hurts to merge this now, and do the CSRC/SSRC split in a separate PR.

alvestrand:
Issue #1111 filed on the update strategy.

PR #1100, opened 2017-03-29
Clarify when RTCRtpContributingSource.audioLevel can be null.

### LINKED ISSUE #1090: When should RTCRtpContributingSource#audioLevel be null?
https://w3c.github.io/webrtc-pc/#dom-rtcrtpcontributingsource-audiolevel

The definition of the attribute does not mention a case where null should be returned. Is there such a case?

@taylor-b

taylor-b:
Fixes #1090.

audioLevel is only null if the RTCRtpContributingSource object
corresponds to a CSRC, and the RFC 6465 header extension wasn't used.


stefhak:
LGTM.

PR #1102, opened 2017-03-30
Don't link function names to legacy section (unless intended)

### LINKED ISSUE #1095: editorial: Some links to setLocalDescription points to the legacy extensions
Which they shouldn't

This applies to addIceCandidate as well.

adam-be:
Fixes #1095 

setLocalDescription and addIceCandidate has some bad links.


stefhak:
LGTM.

PR #1109, opened 2017-04-04
Adding configurable "ptime" member of RTCRtpEncodingParameters.

### LINKED ISSUE #1021: get/setParameters does not have a parameter for packetization interval
We have a few parameters for video and simulcast, would like to have one for packetization interval. Currently, I cannot find a way to change it without messing with the SDP.

taylor-b:
Fixes #1021.

Behaves just like the "ptime" attribute in a remote description, as
described in JSEP. If both are present, the one in the encoding
parameters takes precedence.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5883,6 +5883,7 @@ sender.setParameters(params)
              RTCDtxStatus        dtx;
              boolean             active;
              RTCPriorityType     priority;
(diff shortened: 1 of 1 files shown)


aboba:
You might say that ptime is only meaningful for an audio encoding.

taylor-b:
@aboba Good idea, done.

Issue #1092, opened 2017-03-23
DTLS failures


aboba:
Currently RTCDtlsTransportState has a "failed" value, but the RTCDtlsTransport does not have an onerror EventHandler.  Is there a way for the application to obtain  information on the reason why DTLS failed?

taylor-b:
Are there situations where a DTLS failure is actionable? Or is the failure reason only something that would be used for debugging, in which case it probably doesn't need its own error?

rshpount:
You probably do want to differentiate between the fingerprint match failure, protocol negotiation failure due to things like crypto capabilities or version mismatch,  and DTLS protocol negotiation failure due to network error.

aboba:
Now that browser DTLS implementations are moving to DTLS 1.2 with ECSDA support (and soon, 1.3), backward compatibility issues can arise with applications based on older code that only supports DTLS 1.0 and RSA certificates.  For example,  if an application running on a browser that defaults only to ECDSA certificates does not use the certificate management API to create an RSA certificate, it will fail to interoperate with a mobile application based on an older implementation that only supports RSA certificates.  So if the application can determine that this is what went wrong it could figure out what went wrong. 

I do not think that crypto capabilities or version mismatch negotiation failures would necessarily run afoul of the warning in the Websockets API spec, but fingerprint match failure might.  Here is the prohibition from that spec (https://html.spec.whatwg.org/multipage/comms.html#feedback-from-the-protocol):

User agents must not convey any failure information to scripts in a way that would allow a script to distinguish the following situations:

A server whose host name could not be resolved.
A server to which packets could not successfully be routed.
A server that refused the connection on the specified port.
A server that failed to correctly perform a TLS handshake (e.g., the server certificate can't be verified).
A server that did not complete the opening handshake (e.g. because it was not a WebSocket server).
A WebSocket server that sent a correct opening handshake, but that specified options that caused the client to drop the connection (e.g. the server specified a subprotocol that the client did not offer).
A WebSocket server that abruptly closed the connection after successfully completing the opening handshake.
In all of these cases, the the WebSocket connection close code would be 1006, as required by the WebSocket Protocol specification. [WSP]

Allowing a script to distinguish these cases would allow a script to probe the user's local network in preparation for an attack.

aboba:
Conclusion from the interim was that ICE consent removes much of the security risk. So can provide information on DTLS failures.

aboba:
Some ideas: 

[...]

PR #1115, opened 2017-04-07
DTLS failures

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1092

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -11647,9 +11647,9 @@ if (sender.dtmf) {
                 <td>The DTLS negotiation has failed or the connection
                 has been terminated with a fatal error. The
                 <code>message</code> contains information relating to
-                the nature of error.  If a DTLS alert was received,
+                the nature of error.  If a fatal DTLS alert was received,
                 the <code>receivedAlert</code> attribute is set to the
-                value of the DTLS alert received. If a DTLS alert was
+                value of the DTLS alert received. If a fatal DTLS alert was
                 sent, the <code>sentAlert</code> attribute is set to
                 the value of the DTLS alert sent.</td>
               </tr>
@@ -11660,7 +11660,7 @@ if (sender.dtmf) {
                 provided in the SDP. If the remote peer cannot match
                 the local certificate against the provided fingerprints,
                 this error is not generated. Instead a "bad_certificate"
-                DTLS alert might be received from the remote peer,
+                (42) DTLS alert might be received from the remote peer,
                 resulting in a "dtls-failure".</td>
               </tr>
               <tr>
--- a/webrtc.html
+++ b/webrtc.html
@@ -11795,7 +11795,7 @@ if (sender.dtmf) {
         </section>
         <section>
           <h2>RTCError.prototype.tlsAlert</h2>
-          <p>The initial value of the <code>dltsAlert</code> property of the prototype for 
+          <p>The initial value of the <code>tlsAlert</code> property of the prototype for 
           the <code><a>RTCError</a></code> constructor is 0.</p>
         </section>
         <section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -11648,9 +11648,9 @@ if (sender.dtmf) {
                 has been terminated with a fatal error. The
                 <code>message</code> contains information relating to
                 the nature of error.  If a fatal DTLS alert was received,
-                the <code>receivedAlert</code> attribute is set to the
+                the <code><a>receivedAlert</a></code> attribute is set to the
                 value of the DTLS alert received. If a fatal DTLS alert was
-                sent, the <code>sentAlert</code> attribute is set to
+                sent, the <code><a>sentAlert</a></code> attribute is set to
                 the value of the DTLS alert sent.</td>
               </tr>
               <tr>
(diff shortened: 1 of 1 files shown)


aboba:
Some questions: 

1. Do we need to distinguish between local and remote failures (e.g. by having separate "dtls-local-failure" and "dtls-remote-failure" errorDetail values)? 

2. Is it valuable to include the (D)TLS alert values? 
https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-6

Or does it make more sense to have tailored errorDetail values for actionable failure conditions (such as a ciphersuite negotiation problem)?

fluffy:
To try and answer questions .... on 1) my guess would be no 2) I like just including an exiting erro set like this instead of trying to define how each of theses errors maps to a new error we define. However, that is not enough by itself. There are some error that are only local and don't cause an alert. For example, if fingerprint did not match.

taylor-b:
> There are some error that are only local and don't cause an alert. For example, if fingerprint did not match.

I thought this causes a bad_certificate alert?

>    bad_certificate
>      A certificate was corrupt, contained signatures that did not
>      verify correctly, etc.

webrtc.html
```diff
@@ -11641,6 +11642,13 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
                 <td>The data channel has failed.</td>
               </tr>
               <tr>
+                <td><dfn><code>dtls-failure</code></dfn></td>
+                <td>The DTLS negotiation has failed or the connection
+                has been terminated with a fatal error. The 
+                <code>tlsAlert</code> attribute is set to the
+                DTLS alert value.</td>
+              </tr>
```
ekr: As discussed on the call:

1. I think we should have a "sentAlert" and "receivedAlert" value because both can happen.
2. Note that close_notify is alert code 0. Maybe "none"?

aboba: Added sentAlert and receivedAlert attributes, each with default values of null.

stefhak:
I think the comments made have been addressed. Mergeable?

martinthomson:
Where is the IDL for RTCError?  I notice that it is a "regular object".  I ask because you don't define the type of `sentAlert`/`receivedAlert` and that matters.  Otherwise, this is fine.

aboba:
@martinthomson  Filed separate issue https://github.com/w3c/webrtc-pc/issues/1158

adam-be:
I think we should have at least one usage of this error before we add it to the spec. That also works as a sanity check.

aboba:
@fluffy Added "fingerprint-failure" to RTCError, so that an RTCDtlsTransport can also provide this error in the onerror EventHandler.

taylor-b:
> The part of this that adds the DTLS error looks good but I don't think that can be used to for the fingerprint not matching error.

@fluffy Wouldn't the fingerprint not matching always result in a "bad_certificate" error as I mentioned previously? I'm basing that on Section 6.2 of RFC 8122.

aboba:
@fluffy Modified the PR to make it clear that inability to verify a remote certificate generates the "fingerprint-failure" error, but that if the remote peer cannot validate the local certificate and fingerprint, then it will an alert which will generate a "dtls-error" with receivedAlert of "bad_certificate".

webrtc.html
```diff
@@ -11641,6 +11643,27 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
                 <td>The data channel has failed.</td>
               </tr>
               <tr>
+                <td><dfn><code>dtls-failure</code></dfn></td>
+                <td>The DTLS negotiation has failed or the connection
+                has been terminated with a fatal error. The
+                <code>message</code> contains information relating to
+                the nature of error.  If a DTLS alert was received,
+                the <code>receivedAlert</code> attribute is set to the
```
taylor-b: Could link to receivedAlert/sentAlert descriptions.

webrtc.html
```diff
@@ -11786,6 +11809,16 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
           the <code><a>RTCError</a></code> constructor is 0.</p>
         </section>
         <section>
+          <h2>RTCError.prototype.receivedAlert</h2>
+          <p>The initial value of the <code>receivedAlert</code> property of the prototype for 
+          the <code><a>RTCError</a></code> constructor is null.</p>
```
taylor-b: What's the type of these properties? Are they strings (`"bad_certificate"`) or integers (`42`)? Might also want to point to an RFC where they're defined.

webrtc.html
```diff
@@ -11641,6 +11643,27 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
                 <td>The data channel has failed.</td>
               </tr>
               <tr>
+                <td><dfn><code>dtls-failure</code></dfn></td>
+                <td>The DTLS negotiation has failed or the connection
+                has been terminated with a fatal error. The
+                <code>message</code> contains information relating to
+                the nature of error.  If a DTLS alert was received,
```
taylor-b: May want to clarify that this only happens for fatal alerts: "If a fatal DTLS alert was received... If a fatal DTLS alert was sent..."

webrtc.html
aboba: Clarified that it is an integer (42).

webrtc.html
aboba: Fixed.

webrtc.html
aboba: Fixed.

PR #1119, opened 2017-04-11
Making legacy methods optional to implement.

### LINKED ISSUE #1086: Make legacy API optional to implement
The spec should drive users out of using the legacy API.
But it is not clear whether the legacy API is an optional part of the spec.
If it is not already, the legacy API section 4.3.3 should then be made optional.
Something like: 
Implementers MAY support these methods for backward compatibility.

stefhak:
Fixes #1086

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2622,10 +2622,9 @@ interface RTCPeerConnection : EventTarget  {
       </section>
       <section class="informative">
         <h3>Legacy Interface Extensions</h3>
-          <p>This section is non-normative; it specifies methods that can be
-          implemented for legacy purposes. However, if these methods are
-          implemented it is mandatory to implemented according to what
-          is specified here.</p>
+          <p>Supporting the methods in this section is optional. However,
+          if these methods are implemented it is mandatory to implement
+          according to what is specified here.</p>
         <div class="note">
           These methods are kept on <code><a>RTCPeerConnection</a></code> for
           legacy purposes.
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2620,8 +2620,12 @@
           </section>
         </div>
       </section>
-      <section>
+      <section class="informative">
         <h3>Legacy Interface Extensions</h3>
+          <p>This section is non-normative; it specifies methods that can be
```
foolip: Non-normative isn't appropriate here, sections that are non-normative should be possible to remove from the spec entirely and have no effect. I don't know of other specs that make whole sections optional, but when a single step of an algorithm is optional the wording can be "Optionally, run the following substeps" or similar. So, something like "Supporting the extensions in this section is optional" should suffice.

stefhak: Thanks @foolip (I was aware 'non-normative' is the the best fit here, I just wanted input on what I should say).
Is there another key word I could use instead of ```informative``` in 
```<section class="informative">``` ? I don't know what the valid ones are.

foolip: I don't think there is an appropriate existing class, but @tabatkins might know. Would you like the whole section to have a different background? You can always add your own CSS to achieve whatever effect effect is desired.

tabatkins: There's no special concept for "optional" in Bikeshed.

burnburn:
Sorry didn't see that you made this section informative.  If you remove that I'm okay with merging.

aboba:
@burnburn @stefhak  Removed the "informative" designation.  Is it ok now?

Issue #1120, opened 2017-04-11
RTCCertificate.getAlgorithm for implicitly generated local certificates


henbos:
RTCCertificate.getAlgorithm() is described as follows:

> Returns the value of keygenAlgorithm passed in the call to generateCertificate().

It is clear what to do for RTCCertificates generated by RTCPeerConnection.generateCertificate, but for certificates generated by the RTCPeerConnection contructor when no certificate was provided, and retrieved by RTCPeerConnection.getConfiguration, it is not clear what getAlgorithm() should return.

> If no certificates value was specified, one or more new RTCCertificate instances are generated for use with this RTCPeerConnection instance.

This could mean the PC called generateCertificate and we have a known AlgorithmIdentifier, or the PC could do something implementation-specific to generate certificates. Must an AlgorithmIdentifier have been used? Do we deduce which AlgorithmIdentifier could have been used? Do we return null if certificates was generated without explicitly calling generateCertificate with an AlgorithmIdentifier?

taylor-b:
The specification seems a bit ambiguous right now about whether generated certificates are even returned by `getConfiguration`. The next step is:

> Store the configuration, including the certificates, in a [[ configuration]] internal slot.

But I assumed this referred to only the certificates "included" in the configuration, not automatically generated certificates.

henbos:
I agree it's vague. I interpreted it as to include the generated certificates, thinking why else would you single out certificates out of all things included in a configuration? Otherwise you could just say "Store the configuration in a [[ configuration]] internal slot."

The same vagueness is here too (setConfiguration step 4): 

> If configuration.certificates is set and the set of certificates differs from the ones used when connection was constructed, throw an InvalidModificationError.

Does it refer to the set of certificates supplied in the constructor argument or the resulting set of certificates from the constructor steps? As long as we're consistent with [[ configuration]] internal slot.

aboba:
Since there are legacy native applications that only support DTLS 1.0 and RSA certificates, it may be useful to be able to determine if the browser is still generating an RSA certificate (and if not, to use `RTCPeerConnection.generateCertificate` to create one).  In that scenario, it would be useful for `RTCPeerConnection.getConfiguration` to return generated certificates, and then to be able to call `RTCCertificate.getAlgorithm()` for a generated certificate.

taylor-b:
If an application knows it only supports RSA certificates, it could just use `generateCertificate` 100% of the time.

My concern is that applications may be doing something like this:

```
[...]

PR #1122, opened 2017-04-12
RTCCertificate.getAlgorithm() to return a compatible AlgorithmIdentifier.

### LINKED ISSUE #1121: RTCCertificate.getAlgorithm() wording and serialization
RTCCertificate.getAlgorithm() is described as follows:

> Returns the value of _keygenAlgorithm_ passed in the call to generateCertificate().

Which refers to a `AlgorithmIdentifier keygenAlgorithm`.
The typedef is: `typedef (object or DOMString) AlgorithmIdentifier;`

As described by https://w3c.github.io/webcrypto/Overview.html#algorithm-normalization, the AlgorithmIdentifier is converted to one of the dictionaries inheriting from Algorithm by copying all the members that belong to that Algorithm from the input object (where a DOMString keygenAlgorithm is interpreted as object { name: keygenAlgorithm }).

Because keygenAlgorithm can be _any object_ that can be normalized, which could be any dictionary as long as it includes the same members as Algorithm & friends. Beyond those members, it could contain _anything_.
Because RTCCertificate is serializable, and getAlgorithm() should return the original keygenAlgorithm value, we are required by spec to be able to serialize almost _any_ dictionary. If the dictionary contains members of object types that are not serializable we are tasked with the impossible.

This is overkill. Can we change to wording to say:

> Returns an AlgorithmIdentifier corresponding to the one passed in the call to generateCertificate(), meaning an object that would produce the same result by the normalization process as the original _keygenAlgorithm_ value did.

As such, we are only tasked with serializing relevant members and don't have to remember if the original keygenAlgorithm was a string or an object.

henbos:
RTCCertificate.getAlgorithm() to return a compatible AlgorithmIdentifier to the one in generateCertificate() but not necessarily the exact value, so that we don't have to serialize any possible dictionary.

Fixes #1121.


henbos:
Sublime decided to create pseudo changes in various other lines, not sure why...

henbos:
PTAL @aboba

stefhak:
LGTM.

webrtc.html
```diff
@@ -4587,8 +4587,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             class="methods">
               <dt><dfn><code>getAlgorithm</code></dfn></dt>
               <dd>
-                <p>Returns the value of <var>keygenAlgorithm</var> passed in
-                the call to <code>generateCertificate()</code>.</p>
+                <p>Returns an AlgorithmIdentifier corresponding to the one passed in the call to
+                <code>generateCertificate()</code>, meaning an object that would produce the same
+                result by the normalization process as the original <var>keygenAlgorithm</var>
```
taylor-b: Why not just "the result of the normalization process", as opposed to "an object that would produce the same result", so that the return value is 100% predictable? Also, this could link to https://www.w3.org/TR/WebCryptoAPI/#dfn-normalize-an-algorithm

stefhak:
@henbos can you comment on @taylor-b 's comment, and perhaps update?

webrtc.html
henbos: Done, also added "or the equivalent if the certificate was generated by the RTCPeerConnection due to no certificates being specified in its constructor". This is only applicable if getConfiguration() returns such certificates though, otherwise all RTCCertificate objects will have been generated by generateCertificate().

henbos:
This also relates to #1120.

webrtc.html
henbos: Oh, nope, based on https://github.com/w3c/webrtc-pc/issues/1120#issuecomment-294765374 I removed the bit about what to do if the certificate was generated by the RTCPeerConnection.

henbos:
PTAL. Based on https://github.com/w3c/webrtc-pc/issues/1120#issuecomment-294765374 I assume getConfiguration() does not return certificates generated by the RTCPeerConnection (that were not generated by generateCertificate()) and updated the description to reflect that the only RTCCertificate objects that exists were generated by generateCertificate().

henbos:
Can someone with write access merge this pull request?

taylor-b:
@henbos The editors meet on Thursday, which is typically when PRs get merged.

Issue #1128, opened 2017-04-17
does MSID still work? (was: example 13: getReceivers semantics)


fippo:
http://w3c.github.io/webrtc-pc/#peer-to-peer-example-with-media-before-signaling
essentially does this:
```
    pc.addTrack(...)
    pc.addTrack(...)
    pc.getReceivers().forEach(...)
```
Firefox returns an empty array of receivers at this point. Chrome 59 (after addStream) returns an empty array as well.

Is the example wrong or the implementations?

@jan-ivar @henbos

jan-ivar:
That's post-transceiver behavior. Firefox is adding transceivers now in [bug 1290948](https://bugzilla.mozilla.org/show_bug.cgi?id=1290948).

Step 8.2 of [addTrack](http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtrack) creates a receiver, so the implementations are wrong (as of [last year](https://github.com/w3c/webrtc-pc/commit/2852f0b3515acacdf80a6b125f96e5d2fa3ea608)).

fippo:
Thanks. Now the next question, bear with me:
The receiver that is created and has a track. This track has a random id "randomA".
Media is received on that track (I assume this is possible if the MID header extension is used)
Now a remote answer arrives. The SDP contains two along the lines of 
    ```a=msid:remoterandomstreamid remoterandomaudio```

Does the receivers track id change?
Does ontrack fire with a track that has track.id=remoterandomaudio?
If yes: should the application replace the srcObject with the stream/track from that event?

taylor-b:
The reason Chrome doesn't return any receivers is that since it uses "Plan B" SDP, each m= section could be used to receive 0 to N streams. So there's no way to tell how many receivers are needed before the remote description is set.

To answer your questions:

> Does the receivers track id change?

No; the receiver already had a track before `setRemoteDescription`, so that would mean changing the ID of an existing track.

> Does ontrack fire with a track that has track.id=remoterandomaudio?

It would be with "randomA", but the event *would* be fired on setting the remote description. This issue is related: https://github.com/w3c/webrtc-pc/issues/1013

fippo:
so setRemoteDescription with an a=msid:streamid trackid only fires ontrack with a track whose id is 'trackid' only with an offer but not with an answer?! This doesn't make sense...

Not that track ids are useful but this seems to make the msid concept rather useless.

ibc:
[...]

PR #1129, opened 2017-04-18
examples: dont fiddle with srcObject if already set

fippo:
dont attempt to set srcObject if it is already set in the examples.

This covers two cases:
* ontrack with an audio and a video track
* early media (see discussion in #1128)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -10954,9 +10954,10 @@ function start() {
         .catch(logError);
     };
 
-    // once remote video track arrives, show it in the remote video element
+    // once remote track arrives, show it in the remote video element
     pc.ontrack = function (evt) {
-        if (evt.track.kind === "video")
+        // don't set srcObject again if it is already set.
+        if (!remoteView.srcObject)
           remoteView.srcObject = evt.streams[0];
     };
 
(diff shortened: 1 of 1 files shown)


taylor-b:
The early media case isn't relevant in these specific examples, since they don't set `srcObject` before the `track` event fires. But the point that it will fire twice, with an audio and video track, is valid, so LGTM.

taylor-b:
One issue may be that when the `track` event is first fired, as currently specified the stream will only include one track (the one in the event).

So will it be an issue if `srcObject` is set while the stream only has an audio track, then the video track is added in the next microtask? Or is this guaranteed to work? I've run into some issues with this kind of thing in Chrome, but was unsure if it was a bug or expected behavior.

taylor-b:
Created #1135 for the above issue.

Can we mention in a comment somewhere why the `!srcObject` check is needed? Which I'd assume is to avoid invoking the [media element load algorithm](https://www.w3.org/TR/2015/WD-html51-20150506/semantics.html#media-element-load-algorithm) twice.

fippo:
added a comment "don't set srcObject again if it is already set." to the samples. Which made me wonder if the check should be more specific and only avoid setting the same stream again but the example would blow up with multiple streams anyway.

We don't seem to have a description of what the consuming code is supposed to do with the track and stream from ontrack anywhere.

stefhak:
Comment to https://github.com/w3c/webrtc-pc/pull/1129#issuecomment-295758222: the intention is that once the MediaStream is attached to a video element (via srcObject), tracks can come and go and things should work. E.g. if a new audio track is added, the existing video and audio should continue to play with the new audio track being mixed in.

PR #1130, opened 2017-04-18
Clarify that configuration.certificates remains undefined in the RTCPeerConnection constructor.

### LINKED ISSUE #1120: RTCCertificate.getAlgorithm for implicitly generated local certificates

henbos:
Fixes #1120.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -729,7 +729,7 @@
             <code>InvalidAccessError</code>; otherwise, store the certificates.
             If no <code>certificates</code> value was specified, one or more
             new <code>RTCCertificate</code> instances are generated for use
-            with this <code>RTCPeerConnection</code> instance. This happens
+            with this <code>RTCPeerConnection</code> instance. This MAY happen
             asynchronously and the value of <code>certificates</code> remains
             undefined for the subsequent steps.</p>
           </li>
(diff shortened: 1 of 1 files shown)


henbos:
The phrase "Store the configuration, including the certificates, in a [[ configuration]] internal slot." has already been updated to "Store the configuration in a [[ configuration]] internal slot." so this PR does not have to clarify anything else.

webrtc.html
```diff
@@ -729,7 +729,9 @@
             <code>InvalidAccessError</code>; otherwise, store the certificates.
             If no <code>certificates</code> value was specified, one or more
             new <code>RTCCertificate</code> instances are generated for use
-            with this <code>RTCPeerConnection</code> instance.</p>
+            with this <code>RTCPeerConnection</code> instance. This happens
```
taylor-b: "This MAY happen asynchronously"?

henbos: Done.

PR #1133, opened 2017-04-20
Split getContributingSources into two methods, for CSRCs and SSRCs.

### LINKED ISSUE #1101: RTCRtpContributingSource naming
http://w3c.github.io/webrtc-pc/#dom-rtcrtpcontributingsource says
```
The RTCRtpContributingSource objects contain information
about a given contributing source
```
The text further down says the information is taken either from the RFC 6465 (csrc-audio-level) extension or the RFC 6464 (ssrc-audio-level) extension.

While I agree that a single API for both use-cases makes life much simpler, calling the latter a contributing source (a term which is defined in RFC 3550) is slightly confusing.

Ideally the "contributing" could be dropped from the name. If that ship has sailed,  
    about a given contributing source **or synchronization source**, including the
would clarify things at least.

taylor-b:
Fixes #1101.

There's now getContributingSources and getSynchronizationSources,
which return RTCRtpContributingSources and
RTCRtpSynchronizationSources, respectively. These interfaces are
slightly different, in that only RTCRtpSynchronizationSource has
voiceActivityFlag, and non-nullable audioLevel.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6450,7 +6450,8 @@ sender.setParameters(params)
     readonly        attribute RTCDtlsTransport? rtcpTransport; // Feature at risk
     static RTCRtpCapabilities          getCapabilities (DOMString kind);
     RTCRtpParameters                   getParameters ();
-    sequence&lt;RTCRtpContributingSource&gt; getContributingSources ();
+    sequence&lt;RTCRtpContributingSource&gt;    getContributingSources ();
+    sequence&lt;RTCRtpSynchronizationSource&gt; getSynchronizationSources ();
     Promise&lt;RTCStatsReport&gt;   getStats();
 };</pre>
         <section>
@@ -6557,14 +6558,27 @@ sender.setParameters(params)
             <dt><dfn><code>getContributingSources</code></dfn></dt>
             <dd>
               <p>Returns an <code><a>RTCRtpContributingSource</a></code> for
-              each unique CSRC or SSRC received by this RTCRtpReceiver in the
-              last 10 seconds.</p>
+              each unique CSRC identifier received by this RTCRtpReceiver in
+              the last 10 seconds.</p>
               <div>
                 <em>No parameters.</em>
               </div>
               <div>
                 <em>Return type:</em>
                 <code>sequence&lt;RTCRtpContributingSource&gt;</code>
+              </div>
+            </dd>
+            <dt><dfn><code>getSynchronizationSources</code></dfn></dt>
+            <dd>
+              <p>Returns an <code><a>RTCRtpSynchronizationSource</a></code> for
+              each unique SSRC identifier received by this RTCRtpReceiver in
+              the last 10 seconds.</p>
+              <div>
+                <em>No parameters.</em>
+              </div>
+              <div>
+                <em>Return type:</em>
+                <code>sequence&lt;RTCRtpSynchronizationSource&gt;</code>
               </div>
             </dd>
             <dt><code>getStats</code></dt>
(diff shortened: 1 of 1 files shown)


taylor-b:
I liked Harald's suggestion to just call it `getSsrcs`, but that seemed inconsistent with `getContributingSources` (which could just as easily be `getCsrcs`).

Another naming thing I don't like is that the CSRC or SSRC identifier is called `source`, as opposed to just `ssrc`/`csrc`, or even `identifier`. Is it too late to change this?

Naming gripes aside, I don't think there are any real issues with this.

webrtc.html
```diff
@@ -6611,24 +6625,25 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           </dl>
         </section>
       </div>
-      <p>The <dfn>RTCRtpContributingSource</dfn> objects contain information
-      about a given contributing source, including the most recent time a
+      <p>The <dfn>RTCRtpContributingSource</dfn> and
+      <dfn>RTCRtpSynchronizationSource</dfn> objects contain information about
+      a given contributing source (CSRC) or synchronization source (SSRC),
+      including the most recent time a
```
henbos: nit: weird line-break

webrtc.html
```diff
@@ -6611,24 +6625,25 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-      objects are updated. The
-      <code><a>RTCRtpContributingSource</a></code> object corresponding to the
-      SSRC is always updated, and if the RTP packet contains CSRCs, then the
-      <code><a>RTCRtpContributingSource</a></code> objects corresponding to
-      those CSRCs are also updated.</p>
+      and <code><a>RTCRtpSynchronizationSource</a></code> objects are updated. The
+      <code><a>RTCRtpSynchronizationSource</a></code> object corresponding to the
+      SSRC identifier is always updated, and if the RTP packet contains CSRC
+      identifiers, then the <code><a>RTCRtpContributingSource</a></code>
+      objects corresponding to those CSRC identifiers are also updated.</p>
```
henbos: SSRC and CSRC numbers are independent of each other? You could have one RTP packet (ssrc:1, csrc:[1,...]) and then another one (ssrc:2, csrc:[1,...])? Does it make sense to have a reference to the most recent RTCRtpSynchronizationSource  from an RTCRtpContributingSource?

Hmm actually that could be misleading because the SSRC could be updated and so a CSRC's SSRC could represent a different state and time than the CSRC. But what about a source id? Or is this not useful?

It would seem the SSRC's audioLevel would be enough (CSRC's audioLevel being a copy) if it wasn't for the reference representing a different state and time. Copy needed to have audioLevel for that state/time.

webrtc.html
```diff
@@ -6645,35 +6660,67 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-              otherwise <code>null</code>. RFC 6464 and 6465 define the level
-              as a integral value from 0 to 127 representing the audio level in
-              negative decibels relative to the loudest signal that the system
-              could possibly encode. Thus, 0 represents the loudest signal the
-              system could possibly encode, and 127 represents silence.</p>
+              this source. <code>audioLevel</code> will be the level value
+              defined in [[!RFC6464]], if the RFC 6464 header extension is
+              present. If the RFC 6464 extension header is not present, the
+              browser will compute a value for <code>audioLevel</code> as if it
+              had come from RFC 6464.
```
henbos: Is there a significant cost to making computing audioLevel non-optional per RTP packet?

henbos:
Chrome having shipped with names getContributingSources and RTCRtpContributingSource.source I don't think it is worth changing the name.

webrtc.html
taylor-b: Tried to do this to make the diff cleaner, but I guess it didn't make a difference.

webrtc.html
taylor-b: > You could have one RTP packet (ssrc:1, csrc:[1,...]) and then another one (ssrc:2, csrc:[1,...])?

I don't think this is a scenario we need to worry about. Technically the IDs might share the same space ("CSRC identifier" is just shorthand for "SSRC identifier of the CSRC") but I can't imagine how a CSRC could also be a primary SSRC in the same session.

The intention is basically that `RTCRtpContributingSource` represents "information from CSRC list and RFC 6465 extension" and `RTCRtpSynchronizationSource` represents "information from SSRC field and RFC 6464 extension".

webrtc.html
taylor-b: Hopefully not, since it's already computed for stats. Though I'm not changing that in this PR, just moving it around.

webrtc.html
henbos: My bad, I wasn't thinking about CSRC and SSRC namespaces colliding but rather if different SSRCs transport the same CSRCs at different points in time. I'll update the example to:

"You could have one RTP packet (ssrc:1, csrc:[10,...]) and then another one (ssrc:2, csrc:[10,...])?"

Would the CSRC information be copies of the latest SSRC packet information that contained that CSRC?

taylor-b: I understand your question now, sorry for my mixup.

Anyway: I was thinking of each `RTCRtpContributingSource` object as being tied to a particular `RTCRtpReceiver`. That seems more sensible given how the API is expected to be used.

So, if the same source is contributing to two different streams, I'd suggest having two different `RTCRtpContributingSource` objects, one for each stream it's contributing to.

If the application really _does_ need to know "which stream did this source most recently contribute to?," it still can do this by comparing timestamps.

henbos: Yes that sounds good to me.

stefhak:
Decided to merge at the May 2nd VI.

Issue #1116, opened 2017-04-07
"getParameters" and "setParameters" need more thorough specification


taylor-b:
The description of `RTCRtpSender.getParameters`, for example, is currently just:

> The getParameters method returns the RTCRtpSender object's current parameters for how track is encoded and transmitted to a remote RTCRtpReceiver.

But it's not obvious how and when these "current parameters" are populated. My understanding, from working group discussions back when the sender/receiver APIs were first being introduced, was that `getParameters` would return nothing until a remote description is set (or a local description, for an `RTCRtpReceiver`), after which the envelope of possible parameters is determined by SDP.

`setParameters` could also be more clear about all the things that are allowed to change. We already have "read-only parameter" markings where the parameter is defined, but a short summary at the top of the `setParameters` description wouldn't hurt.

aboba:
@taylor-b  One way to think about this is that there is a [[parameters]] internal slot that is distinct for the sender and receiver.  sender.setParameters() stores validated parameters in that slot and the getParameters() methods retrieve from it.  For the sender, setting the remote description stores into the [[parameters]] internal sender slot, and for the receiver, setting the local description stores into the [[parameters]] internal receiver slot. However, addTransceiver with sendEncodings or setCodecPreferences() does not store into the sender [[parameters]] internal sender slot (or the receiver slot).  Since sender.setParameters() cannot set the negotiation-needed flag (or the number of encodings), can the contents of the [[parameters]] internal slot affect createOffer() or createAnswer()?  I suspect the answer is "no".

aboba:
Subsets of the "current parameters" are affected by different methods: 

1. addTransceiver with sendEncodings can set rids, the number of encodings, and writeable encoding parameters on the sender.   This affects the SDP produced by createOffer/createAnswer. 

2. sender.setParameters() can reorder or remove codecs and can set writeable encoding parameters (but cannot change read-only parameters, the number of encodings or the sdpFmtpLine).  We say that setParameters cannot set the negotiation-needed flag, but do not say explicitly that setParameters does not affect the SDP produced by createOffer() and createAnswer().

3. setCodecPreferences() can reorder or remove codecs but cannot set the sdpFmtpLine.  This affects the SDP produced by createOffer/createAnswer. 

4. Calling setLocalDescription (for a receiver) or setRemoteDescription (for a sender) sets parameters from the SDP.

[...]

PR #1134, opened 2017-04-20
Add more detail about how getParameters and setParameters work.

### LINKED ISSUE #1073: Need to specify which members of the encodings in "sendEncodings" are actually used
The description of the `sendEncodings` member of `RTCRtpTransceiverInit` is currently:

> A sequence containing parameters for sending RTP encodings of media.

This somewhat implies that the full contents of the `RTCRtpEncodingParameters` are processed and used when generating an offer. But the only member that appears used by JSEP is `rid`. If this is correct, it should be clarified somewhere, either in the description of `sendEncodings` or in the `addTransceiver` steps.

taylor-b:
Fixes #1073 and #1116.

With regards to `sendEncodings`, all read-write parameters and `rid`
can be set, but the rest are ignored, as decided in the April virtual
interim.

Also adding some detail to the `getParameters` descriptions. For an
`RTCRtpSender`, `getParameters` should return the previously
provided `sendEncodings` even if `setRemoteDescription`
hasn't been called yet.

I'm not yet addressing the simulcast error point that arose in the
discussion of #1073. We can address that after we finish fleshing
out how simulcast errors will be handled in the first place.

There may still be some ambiguity remaining. But I think this is
a step in the right direction.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1255,7 +1255,9 @@
                         <p>Set <var>transceiver</var>'s <code><a data-for=
                         "RTCRtpTransceiver">mid</a></code> value to the mid of
                         the corresponding media description. If the media
-                        description has no mid, generate a random value as
+                        description has no MID, and <var>transceiver</var>'s
+                        <code><a data-for="RTCRtpTransceiver">mid</a></code>
+                        is unset, generate a random value as
                         described in <span data-jsep=
                         "applying-a-remote-desc">[[!JSEP]]</span>.</p>
                       </li>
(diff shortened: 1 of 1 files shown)


taylor-b:
I updated this PR to also address #1073. This PR already needed to update the `getParameters` description to talk about its interactions with `sendEncodings`, so I thought it may be easier to review the changes together like this.

taylor-b:
> An attempt to set a read-only parameter in setParameters() results in a rejection of the promise with an InvalidModificationError. Should addTransceiver behave the same way?

Good point. Now throwing errors.

> When will getParameters() return other attributes of RTCRtpParameters such as rtcp or degradationPreference?

Let's see. `degradationPreference` is only valid for a sender, and should return whatever it was last set to.

For the `RTCRtcpParameters`, it's a bit different between a sender and receiver.

**Sender**:

* `cname` should be there from creation time, since there should only be one CNAME per PeerConnection.
* `reducedSize` should be `true` if the remote description contains "a=rtcp-rsize" and the implementation supports it, and `false` otherwise.

**Receiver**:

* `cname` never set, since it's not signaled in SDP?
* `reducedSize` set to `true` if the receiver is prepared to received reduced-size RTCP. Which means either it was offered (in a local offer; still waiting for answer), or it's fully negotiated (in both local and remote description, `signalingState` is "stable").

Does that all sound reasonable?

aboba:
@taylor-b I'd assume degradationPreference is set to the default ("balanced") unless sender.setParameters() sets it to something else.  It also seems that it should be set to the default when retrieved via receiver.getParameters(), even though it only has meaning for a sender. 

Let me think more about cname on an RtpReceiver.  Otherwise rtcp logic seems reasonable.

taylor-b:
Updated this PR to go into more detail about all the members of `RTCRtpParameters`.

PR #1137, opened 2017-04-20
Adding more detail about RTCDataChannel.id's default value (null).

### LINKED ISSUE #1132: Define what the default RTCDataChannel id is
https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-createdatachannel

> If the id attribute is uninitialized (not set via the dictionary), and the DTLS role of the SCTP transport has already been negotiated, then initialize id to a value generated by the user agent, according to [RTCWEB-DATA-PROTOCOL], and skip to the next step. If no available id could be generated, or if the value of the id member of the dictionary is taken by an existing RTCDataChannel , throw a ResourceInUse exception.

However, Chrome and Firefox seem to be doing something much simpler, for `new RTCPeerConnection().createDataChannel('').id` both simply return 65535.

taylor-b:
Fixes #1132.

This is an editorial change; the behavior described here was already
decided when issue #795 was discussed.


webrtc.html
```diff
@@ -8099,7 +8101,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                 </li>
                 <li>
                   <p>If the <code><a data-for="RTCDataChannel">id</a></code>
-                  attribute is uninitialized (not set via the dictionary),
+                  attribute is <code>null</code> (due to no ID being
+                  passed into <code>createDataChannel</code>),
                   and the DTLS role of the SCTP transport has already been
                   negotiated, then initialize <code><a data-for=
                   "RTCDataChannel">id</a></code> to a value generated by the
```
foolip: Looking at https://tools.ietf.org/html/draft-ietf-rtcweb-data-protocol-09, the id is an "SCTP stream identifier (0-65534)". This step will either cause id to be a number, or an exception will be thrown, null is not a possible outcome AFACIT.

webrtc.html
```diff
@@ -8064,8 +8064,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   <code><a data-for="RTCDataChannel">negotiated</a></code>,
                   <code><a data-for="RTCDataChannel">id</a></code> and
                   <code><a data-for="RTCDataChannel">priority</a></code> attributes
-                  to the values of their corresponding dictionary members (if
-                  present in the dictionary).</p>
+                  to the values of their corresponding dictionary members, if
+                  present in the dictionary. For each member not present in the
+                  dictionary, initialize the corresponding attribute to its
+                  default value, as described in the attribute's description.
```
foolip: There are only 3 members in RTCDataChannelInit that don't have default values. It would be clearer, I think, to handle them explicitly in precisely the place that an implementation would. The attribute descriptions are an odd mix of non-normative description and normative requirements, and would be nice to reduce the amount of such mixing by extracting out the normative bits.

taylor-b: Good suggestion. I haven't cleaned up the attribute descriptions yet, but I at added default `null` values for the `RTCDataChannelInit` members and simplified this step.

webrtc.html
taylor-b: I'm not seeing this. Here's what I intended, in pseudo-code form:

```
dataChannel.id = init.id;
if (!dataChannel.id) {
  if (dtls_role_negotiated) {
    try {
      dataChannel.id = PickAvailableId();
    } catch (e) {
      throw ResourceInUse;
    }
  }
}

So it will end up null if no ID was provided and the DTLS role is not yet negotiated.

File: webrtc.html [hunk 3]
```diff
@@ -8748,13 +8760,13 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-             USVString      protocol = "";
-             boolean        negotiated = false;
+             boolean         ordered = true;
+             unsigned short? maxPacketLifeTime = null;
+             unsigned short? maxRetransmits = null;
+             USVString       protocol = "";
+             boolean         negotiated = false;
              [EnforceRange]
-             unsigned short id;
+             unsigned short? id = null;
```
foolip: https://heycam.github.io/webidl/#EnforceRange doesn't mention nullable types, but I think it'll work. Can you update https://github.com/w3c/web-platform-tests/blob/master/webrtc/RTCPeerConnection-createDataChannel.html in a companion PR to adjust the tests for the 3 nullable types? Test null and undefined for each.

webrtc.html
```diff
@@ -8055,17 +8055,20 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   the first argument.</p>
                 </li>
                 <li>
-                  <p>If the second (dictionary) argument is present, set
-                  <var>channel</var>'s <code><a data-for=
+                  <p>Let <var>init</var> be the second (dictionary) argument,
```
foolip: Filed https://github.com/heycam/webidl/issues/349 about this bit. In the meantime, I suggest writing prose that just assumes that the dictionary is there. https://wicg.github.io/mediasession/#dom-mediametadata-mediametadata is an example of this.

webrtc.html
foolip: I see, the "DTLS role of the SCTP transport has already been negotiated" bit is the important piece here.

foolip:
@jan-ivar @youennf, please check if you're happy with making the 3 dictionary members nullable.

webrtc.html
jan-ivar: Yeah, *init* is guaranteed to exist, thanks to webidl binding code.

webrtc.html
```diff
@@ -8748,13 +8760,13 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
       <div>
         <pre class="idl">dictionary RTCDataChannelInit {
-             boolean        ordered = true;
-             unsigned short maxPacketLifeTime;
-             unsigned short maxRetransmits;
-             USVString      protocol = "";
-             boolean        negotiated = false;
+             boolean         ordered = true;
+             unsigned short? maxPacketLifeTime = null;
+             unsigned short? maxRetransmits = null;
```
jan-ivar: Note that this makes `null` a valid input, i.e. users may specify `null`. Is that desirable? I think no.

The [advice](http://logs.glob.uno/?c=content#c437464) from our DOM team is *"'number or missing' is probably best represented as nullable, except in dictionaries/arguments where it should just be represented as optional"*

I think unless we want to allow users to pass in `null`, we're better off with non-nullable inputs, and some prose. Same for `id` below.

taylor-b: Ok, going back to non-nullable inputs and some prose.

webrtc.html
```diff
@@ -8748,13 +8758,13 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-             USVString      protocol = "";
-             boolean        negotiated = false;
+             boolean         ordered = true;
+             unsigned short  maxPacketLifeTime;
+             unsigned short  maxRetransmits;
+             USVString       protocol = "";
+             boolean         negotiated = false;
              [EnforceRange]
-             unsigned short id;
+             unsigned short  id;
```
taylor-b: Decided to keep this white space change since it makes things line up with "priority"

webrtc.html
```diff
@@ -8068,7 +8063,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   <code><a data-for="RTCDataChannel">negotiated</a></code>,
                   <code><a data-for="RTCDataChannel">id</a></code> and
                   <code><a data-for="RTCDataChannel">priority</a></code> attributes
-                  to the values of the corresponding members of <var>init</var>.
+                  to the values of the corresponding members of the
+                  <var>dataChannelDict</var> argument, using a value of
+                  <code>null</code> if the corresponding dictionary member is
```
foolip: A note here spelling out which 3 members can be missing would be good. Syntactically `{}` is missing all of them, but only those without a default value can be missing here, after the conversion to an IDL dictionary.

taylor-b: Done.

PR #1139, opened 2017-04-21
Call RTCDtlsFingerprint a dictionary, not an object

foolip:
RTCDtlsFingerprint is a dictionary. Web IDL says that "IDL dictionary type
values are represented by ECMAScript Object values", but they are not the same:
https://heycam.github.io/webidl/#es-dictionary

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7189,9 +7189,9 @@ sender.setParameters(params)
         </table>
       </div>
       <section id="rtcdtlsfingerprint*">
-        <h3>The RTCDtlsFingerprint Object</h3>
-        <p>The <dfn><code>RTCDtlsFingerprint</code></dfn> object includes the
-        hash function algorithm and certificate fingerprint as described in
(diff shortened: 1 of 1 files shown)


foolip:
See also https://w3c.github.io/webrtc-pc/#rtciceserver-dictionary

webrtc.html
```diff
@@ -7189,9 +7189,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
         </table>
       </div>
       <section id="rtcdtlsfingerprint*">
```
foolip: Any idea why this ID ends with a `*`? Compare with https://w3c.github.io/webrtc-pc/#rtciceserver-dictionary?

(Matters somewhat because we link to the sections from Blink's IDL files, and if this is a typo it'll presumably change.)

aboba: It's an artifact, due to the section being copied from the ORTC API spec.

foolip: Should I change it?

PR #1143, opened 2017-04-24
move legacy callback definitions to legacy section

### LINKED ISSUE #1142: "Callback definitions" section should go under "legacy interface extensions"
... Since these callbacks are only used for the legacy interface extensions.

Also, `RTCStatsCallback` is not referenced anywhere. Did we forget to add a legacy `getStats` method? Or can we just remove `RTCStatsCallback`?

fippo:
fixes #1142

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2676,26 +2676,6 @@ interface RTCPeerConnection : EventTarget  {
               </section>
             </div>
           </section>
-          <section>
-            <h4><dfn>RTCStatsCallback</dfn></h4>
-            <div>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2639,6 +2639,65 @@
           legacy purposes.
         </div>
         <section>
+          <h4>Callback Definitions</h4>
```
fippo: this is the only non-whitespace change

fippo:
Like this? Moved + removed one level of section tags so this does not show up in the TOC

PR #1145, opened 2017-04-25
FrozenArray, sequence and SameObject

### LINKED ISSUE #1138: Use [SameObject] for the FrozenArray<T> attributes that never change
These are the readonly attributes of type `FrozenArray<T>` in the spec:
* [RTCPeerConnection](https://w3c.github.io/webrtc-pc/#interface-definition)'s `static readonly attribute FrozenArray<RTCIceServer> defaultIceServers`
* [RTCCertificate](https://w3c.github.io/webrtc-pc/#rtccertificate-interface)'s `readonly attribute FrozenArray<RTCDtlsFingerprint> fingerprints`
* [RTCTrackEvent]'s `readonly attribute FrozenArray<MediaStream> streams`

These seem like they will always return the same value. Documenting this with [SameObject] makes this obvious an means that tests can `assert_equals(certificate.fingerprints, certificate.fingerprints)` and similar, something that cannot be assumed from the current spec.

@henbos

henbos:
Because dictionaries are returned by copy, returning FrozenArray\<DictionaryType\> must return a new array of new objects every time, and the dictionaries are mutable. In this case it makes more sense to use a getter method returning sequence\<DictionaryType\> than FrozenArray-attributes.

The one use of FrozenArray\<T\> where T is an interface gets [SameObject].

Fixes #1138.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1356,7 +1356,7 @@ interface RTCPeerConnection : EventTarget  {
     readonly        attribute RTCIceConnectionState     iceConnectionState;
     readonly        attribute RTCPeerConnectionState    connectionState;
     readonly        attribute boolean?                  canTrickleIceCandidates;
-    static readonly attribute FrozenArray&lt;RTCIceServer&gt; defaultIceServers;
+    static sequence&lt;RTCIceServer&gt;      getDefaultIceServers ();
     RTCConfiguration                   getConfiguration ();
     void                               setConfiguration (RTCConfiguration configuration);
     void                               close ();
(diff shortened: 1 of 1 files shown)


henbos:
Please take a look, @domenic and @foolip

domenic:
Generally on the platform we don't use the "get" prefix for methods that return values.

foolip:
There are lots and lots of exceptions to this, and for MediaStream and WebRTC it's common:
https://w3c.github.io/mediacapture-main/#media-stream-track-interface-definition
https://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface-extensions

I like the style of e.g. https://fetch.spec.whatwg.org/, but I think the best we could do is try to have some consistency for new APIs.

henbos:
It looks like all examples of webrtc-pc getter methods are called getFoo(). WebIDL says "sequences must not be used as the type of an attribute or constant" so getter should be the way to go.

stefhak:
@jan-ivar @adam-be can you take a look.

adam-be:
Lgtm

stefhak:
Agreed to merge at May 2nd VI.

PR #1149, opened 2017-05-01
Add paragraph about RtpContributingSources being updated simultaneously.

### LINKED ISSUE #1124: Related attributes should be updated at the same time (in the same microtask checkpoint?)
Issue #1111 deals with how attributes' values are updated; they shouldn't change within a single JS task execution.

But a similar thing I believe we should specify is that all related attributes, which the application might compare against each other, must be updated in the _same_ microtask checkpoint. For example, with `getContributingSources`, which brought up this topic: an application may compare `RTCRtpContributingSources` against each other to find out, for example, who is the most recent speaker. So, all the `RTCRtpContributingSource`s for a single `RTCRtpReceiver` should be updated at the same time. Otherwise, the application may think "this person stopped speaking", when really they're just looking at stale data.

It would be difficult to generalize this principle, so I suggest adding something to the description of `getContributingSources` itself, and any other place we think is relevant.

taylor-b:
Fixes #1124.

The intent of this PR is to ensure that if an application reads the
attributes of multiple `RTCRtpContributingSource`s, the values are all
taken from the same point in time, so they're comparable.

For example, an application may compare timestamps to see which
contributing source(s) are actively contributing (have the most recent
timestamp), or see which one is loudest. Without the guarantee this PR
attempts to add, a browser might return values from different points in
time for different `RTCRtpContributingSource`s, making such comparisons
unreliable.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6628,7 +6628,7 @@ sender.setParameters(params)
       <code><a>RTCRtpSynchronizationSource</a></code> objects based on the
       contents of the packet. The
       <code><a>RTCRtpSynchronizationSource</a></code> object corresponding to
-      the SSRC identifier is always updated, and if the RTP packet contains
+      the SSRC identifier is updated each time, and if the RTP packet contains
       CSRC identifiers, then the <code><a>RTCRtpContributingSource</a></code>
       objects corresponding to those CSRC identifiers are also updated.</p>
       <div class="note">As stated in the <a href="#conformance">conformance
--- a/webrtc.html
+++ b/webrtc.html
@@ -6623,19 +6623,23 @@ sender.setParameters(params)
       keep information from RTP packets received in the previous 10 seconds.
       When the first audio frame contained in an RTP packet is delivered to the
       <code><a>RTCRtpReceiver</a></code>'s <code><a>MediaStreamTrack</a></code>
-      for playout, the relevant <code><a>RTCRtpContributingSource</a></code>
-      and <code><a>RTCRtpSynchronizationSource</a></code> objects are updated. The
-      <code><a>RTCRtpSynchronizationSource</a></code> object corresponding to the
-      SSRC identifier is always updated, and if the RTP packet contains CSRC
-      identifiers, then the <code><a>RTCRtpContributingSource</a></code>
+      for playout, the user agent MUST queue a task to update the relevant
+      <code><a>RTCRtpContributingSource</a></code> and
+      <code><a>RTCRtpSynchronizationSource</a></code> objects based on the
+      contents of the packet. The
+      <code><a>RTCRtpSynchronizationSource</a></code> object corresponding to
+      the SSRC identifier is always updated, and if the RTP packet contains
+      CSRC identifiers, then the <code><a>RTCRtpContributingSource</a></code>
       objects corresponding to those CSRC identifiers are also updated.</p>
-      <p>All of the <code><a>RTCRtpContributingSource</a></code> objects that
-      correspond to the same <code><a>RTCRtpReceiver</a></code> MUST be
-      updated at the same time within the same microtask checkpoint, or updated
-      in a different but indistinguishable manner. This is necessary so that an
-      application can compare <code><a>RTCRtpContributingSource</a></code>
-      objects with each other, with the guarantee that they all return
-      information from the same point in time.</p>
+      <div class="note">As stated in the <a href="#conformance">conformance
+      section</a>, requirements phrased as algorithms may be implemented in
+      any manner so long as the end result is equivalent. So, an
+      implementaion does not need to literally queue a task for every
+      packet, as long as the end result is that within a single event loop task
+      execution, all <code><a>RTCRtpSynchronizationSource</a></code> and
+      <code><a>RTCRtpContributingSource</a></code> objects for a particular
+      <code><a>RTCRtpReceiver</a></code> return information from a single point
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6605,6 +6605,13 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       SSRC is always updated, and if the RTP packet contains CSRCs, then the
       <code><a>RTCRtpContributingSource</a></code> objects corresponding to
       those CSRCs are also updated.</p>
+      <p>All of the <code><a>RTCRtpContributingSource</a></code> objects that
+      correspond to the same <code><a>RTCRtpReceiver</a></code> MUST be
+      updated at the same time within the same microtask checkpoint, or updated
+      in a different but indistinguishable manner. This is necessary so that an
```
jan-ivar: This seems stricter than it needs to be, and there's an algorithm hiding here: Browsers must queue a task to update this information.

When that task runs, it runs to completion, including running and emptying everything on the micro-task queue, *which can only contain things put there by the algorithm itself*, before content JS runs again. So no need to dictate consistency at the microtask level.

I think we should specify that algorithm here.

webrtc.html
```diff
@@ -6605,6 +6605,13 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+      application can compare <code><a>RTCRtpContributingSource</a></code>
+      objects with each other, with the guarantee that they all return
+      information from the same point in time.</p>
```
jan-ivar: This second part is the important guarantee I think. This could perhaps be done by dictating that the algorithm MUST only be run on sets of values covering all sources, from the same point in time.

jan-ivar: Also, since packets arrive when they arrive, I'm not exactly sure what you mean with "same time".

taylor-b:
@jan-ivar Putting this in the form of an algorithm is probably a good idea since that's what we do in most places.

Would we be covered if I just replaced "when X happens, the relevant `RTCRtpContributingSource` objects are updated" with "when X happens, the user agent MUST queue a task to update the relevant `RTCRtpContributingSource` objects"? The actual implementation would still be allowed to do something different, due to this paragraph:

> Conformance requirements phrased as algorithms or specific steps may be implemented in any manner, so long as the end result is equivalent. (In particular, the algorithms defined in this specification are intended to be easy to follow, and not intended to be performant.)

Can you confirm that this is flexible enough to make Chrome's current implementation legal? Since queuing a task for each packet received is obviously bad for performance, it's doing the equivalent of this instead (code simplified for brevity):

```
RTCContributingSources RTCRtpReceiver::getContributingSources() {
  if (should_update_) {
    UpdateRTCContributingSources(internal_receiver_->GetContributingSources());
    should_update_ = false;
  }
  queueMicrotask([this] { should_update_ = true; });
  return contributing_sources_;
}

// Many, many layers of code deeper:
void InternalRtpReceiver::OnPacketReceived(Packet* p) {
  ScopedCriticalSection(cs_);
  UpdateContributingSources(p);
  // ...
}

ContributingSources InternalRtpReceiver::GetContributingSources() {
  ScopedCriticalSection(cs_);
  return contributing_sources_;
}
```

The full code can be seen [here](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/peerconnection/RTCRtpReceiver.cpp?type=cs&q=getcontributingsources&l=26).

Anyway, this ensures:

1. Within a single event loop task execution, the attributes of a given `RTCRtpContributingSource` will not change.
2. Within a single event loop task execution, all of the `RTCRtpContributingSource` objects for a given receiver will return information for the same point in the RTP stream.

Those are the requirements I really wanted to convey, but I was unsure how to phrase them in terminology that's consistent with the rest of the spec.

jan-ivar:
@taylor-b Ah, I had forgotten about the performance issue @jesup and I were supposed to look at here. In light of that, I'm fine either way.

Your implementation looks good to me. As for indistinguishable, you might consider queuing a task instead of a micro-task, otherwise I could observe a difference like this:
```js
let a = receiver.getContributingSources();
await Promise.resolve();
let b = receiver.getContributingSources();
log(JSON.stringify(a) == JSON.stringify(b)); // false
```
Not a problem per se, just different. I'm actually not sure what the TAG view is on state changes in the microtask queue.

taylor-b:
Fixed conflicts, and switched to using "queue a task" to enforce the normative requirement, with a `div class="note"` that explains further.

webrtc.html
```diff
@@ -6623,19 +6623,23 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-      and <code><a>RTCRtpSynchronizationSource</a></code> objects are updated. The
-      <code><a>RTCRtpSynchronizationSource</a></code> object corresponding to the
-      SSRC identifier is always updated, and if the RTP packet contains CSRC
-      identifiers, then the <code><a>RTCRtpContributingSource</a></code>
+      for playout, the user agent MUST queue a task to update the relevant
+      <code><a>RTCRtpContributingSource</a></code> and
+      <code><a>RTCRtpSynchronizationSource</a></code> objects based on the
+      contents of the packet. The
+      <code><a>RTCRtpSynchronizationSource</a></code> object corresponding to
+      the SSRC identifier is always updated, and if the RTP packet contains
```
jan-ivar: I had a bit of trouble parsing "always" here. Maybe s/always updated/updated each time/ to underscore we're talking about frequency and the packet in the previous sentence?

Issue #763, opened 2016-08-25
Handling of encoder/decoder errors


aboba:
There are some open questions about encoder/decoder errors, such as what happens if simulcast functionality is requested that the browser cannot carry out, such as specifying sendEncodings to send more encodings than the browser can handle.  setParameters is a Promise so an error will be returned, but it may not be obvious what exactly went wrong.  There are not even any capabilities to describe the limitations on simulcast (such as how many streams can be sent or received). 

Related Issue: https://github.com/w3c/webrtc-pc/issues/1323

aboba:
Some thoughts: 

Do we need RTCRtpCodecCapability.maxSimulcastStreams ?  (all we have today is mimeType)

try {
videoSender.setParameters(params); 
}
catch (e){
   console.log('params issue: ', e.message); 
}

Is there any information in e.message on what was wrong with params?  We may not even have called createOffer() by this point, so there isn't necessarily even any SDP to look at (or complain about).

pthatcherg:
Could we just choose to do the same thing we do for CPU and network adaptation: not send the layers we can't?  

How is this different than "poor man's simulcast", when someone calls addTrack N times for the same track?  Doesn't that have the same issue?

alvestrand:
Now that we're not permitting setParameters() to change number of layers, is this OBE?

alvestrand:
could check with @pthatcherg @taylor-b if there is an use case where a more specific error would be useful. If we don't see an obvious case (and one that can't be handled using error.name), we should close this.

taylor-b:
`setParameters()` can't change the number of layers, but it could activate layers that are inactive. If we assume a browser may have codec-specific restrictions, then consider this example:

1. `addTransceiver` called with 5 layers.
2. `createOffer` called. Suppose the offer contains one codec that supports all 5 layers, one that supports 3 layers, and another that doesn't support simulcast.
3. Offer/answer negotiation resulted in the last codec being chosen.
4. `sender.getParameters` should return 5 encodings, but 4 of them are not active (`active == false`). (Is this correct?)
5. If the application wants to switch to the codec that supports simulcast, it can set the `active` flag to true and reorder the codecs, then call `sender.setParameters`.
6. If more encodings were activated than are supported for the new codec, the promise needs to be rejected with an error.

[...]

PR #1150, opened 2017-05-01
Handling encoding parameter errors

aboba:
Work in progress (not ready to merge).  Need to incorporate feedback from the virtual interim.

Fix for Issue https://github.com/w3c/webrtc-pc/issues/763

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -11649,8 +11649,11 @@ if (sender.dtmf) {
               "idp-tls-failure",
               "idp-token-expired",
               "idp-token-invalid",
+              "invalid-codecs",
+              "invalid-encoding-parameters",
               "sctp-failure",
-              "sdp-syntax-error"
+              "sdp-syntax-error",
+              "too-many-encodings"
           };</pre>
           <table data-link-for="RTCErrorDetailType" data-dfn-for=
           "RTCErrorDetailType" class="simple">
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -11702,6 +11705,19 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
                 <td>The IdP token is invalid.</td>
               </tr>
               <tr>
+                <td><dfn><code>invalid-codecs</code></dfn></td>
+                <td>Invalid or unavailable <code>codecs[]</code>
```
taylor-b: May want to restrict this to just "unavailable", since "invalid" is already covered by InvalidModificationError.

fluffy:
This LGTM and like to see the stuff from call added in.

webrtc.html
aboba: What if setParameters() attempts to modify an RTCRtpCodecParameters dictionary entry in parameters.codecs[]?  Would this be covered by InvalidModificationError ?

taylor-b: I thought it would be. Or at least that was our intention.

stefhak:
Good enough to merge now?

stefhak:
@aboba what's the status of this one now?

aboba:
@stefhak I have closed the original issue, since we now have more specific Issues filed for each aspect of the problem. So this PR is now moot.

Issue #1156, opened 2017-05-03
Issue 1: Key shortening


aboba:
Issue 1 says: "[[!RFC7635]] doesn't explicitly define how keys are shortened; there is only a brief reference to this in Appendix B. This needs to be addressed in the IETF TRAM working group."

We will use this Issue to track progress (rather than putting an Issue in the specification itself, which causes Issue at Risk numbering to start with 2).

taylor-b:
FYI, I'm not sure this is really an issue. Copy of email I sent to @misi and @juberti:

> I don't think this is an issue after all... because RFC2104 already defines how HMAC keys are shortened:
> 
>> The authentication key K can be of any length up to B, the
>> block length of the hash function. Applications that use keys longer
>> than B bytes will first hash the key using H and then use the
>> resultant L byte string as the actual key to HMAC.
> 
> So if anything, the issue with RFC7635 is that it seems to assume the SHA1 key must be 20 bytes, when it can be up to 64 bytes and there's a defined procedure for shortening it if it's too large.
> 
> Or have I missed something? Is there a crypto expert in the working group we could check with?

misi:
@taylor-b I am also not a crypto expert, but I see the same way as you. 
RFC2104 defines how to short a longer key for HMAC algs in general.
I think You find the missing puzzle piece RFC, and so the algorithm definition that we missed earlier.

I don't see the issue, you have mentioned here:
> So if anything, the issue with RFC7635 is that it seems to assume the SHA1 key must be 20 bytes, when it can be up to 64 bytes and there's a defined procedure for shortening it if it's too large.

AFAIU RFC7635 is correct and describes according RFC2104, 
HMAC-SHA-256-128 use 256bit key https://tools.ietf.org/html/rfc4868#section-2.6
20byte is correct as the output of SHA1(256bit key) that will be used as input for the final SHA1.

After clarification of this I don't see any issue here, and I think we could close this issue.

taylor-b:
The issue I see with RFC7635 is this snippet:

> The key length of 160 bits MUST be supported (i.e., only the 160-bit key is used by HMAC-SHA-1 for message integrity of STUN messages).

This is incorrect, since any size key can be input into HMAC-SHA-1.

misi:
AFAIU RFC7635 allows any input key length. This sentence AFAIU is just declaring that the implementation should implement/support 160 bit input key length for sure (because of the key shortening SHA1 output is a fixed 160bit..). 
[...]

PR #1157, opened 2017-05-03
Move Issue 1 text to a note

aboba:
Related to Issue https://github.com/w3c/webrtc-pc/issues/1156 (this PR does not resolve the issue, just moves the text so that Issues at Risk can be numbered correctly).

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -291,13 +291,13 @@
                 Algorithm capabilities of the <a>ICE Agent</a>, and still
                 allows for hash agility as described by [[STUN-BIS]], Section
                 15.3.</p>
-                <div class="note">According to [[!RFC7635]] Section 4.1, the
-                HMAC key MUST be a symmetric key.
-                </div>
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

webrtc.html
```diff
@@ -291,14 +291,12 @@
                 Algorithm capabilities of the <a>ICE Agent</a>, and still
                 allows for hash agility as described by [[STUN-BIS]], Section
                 15.3.</p>
-                <div class="issue">[[!RFC7635]] doesn't explicitly define
+                <div class="note">According to [[!RFC7635]] Section 4.1, the
+                HMAC key MUST be a symmetric key. [[!RFC7635]] doesn't explicitly define
```
taylor-b: nit: The "key MUST be symmetric" sentence seems out of place here, since the text both above and below is talking about the key shortening. Could just change "issue" to "note" and have 2 notes here.

Issue #1159, opened 2017-05-04
Do we need getAlgorithm() ?


aboba:
Now that it is no longer possible to retrieve the pre-configured certificates via getConfiguration(), is the original reason for adding getAlgorithm (described in Issue https://github.com/w3c/webrtc-pc/issues/496) still valid?

Since the only certificates retrievable via getConfiguration() are those the application created itself, the application could presumably store the algorithms used to create those certificates on its own.

henbos:
Removing getAlgorithm() sounds good to me considering questions of having to serialize the AlgorithmIdentifier. It doesn't seem very useful and if you need it you can store it yourself.

aboba:
@martinthomson @taylor-b Does it make sense to remove getAlgorithm()?

taylor-b:
Sounds reasonable to me.

aboba:
Original Issue: https://github.com/w3c/webrtc-pc/issues/496

aboba:
@alvestrand (If you are not fishing ;)  Since you raised the issue originally, do you have an opinion on whether it is ok to remove getAlgorithm()?

henbos:
I'm guessing he is out fishing. :)

From the original issue by @alvestrand in #496,
> A question was raised on whether we could tell if a given cert was RSA or ECDSA - there doesn't seem to be one.

Default generated certificates are not be included in getConfiguration() (based on [this discussion](https://github.com/w3c/webrtc-pc/issues/1120#issuecomment-294987310)) and remote side certificates [are byte arrays](https://w3c.github.io/webrtc-pc/#dom-rtcdtlstransport-getremotecertificates). Since RTCCertificate only exist for certificates that _you_ generated with generateCertificate(), you would know if you care to keep track of that information.

PR #1160, opened 2017-05-04
Remove getAlgorithm()

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1159


aboba:
Dom has asked for a history of how getAlgorithms() got into the spec.  Will be discussed at the next virtual interim.

aboba:
At the virtual interim there was no opposition to removing getAlgorithm().

PR #1163, opened 2017-05-04
Update RTCCertificate ref to get serializer from WebIDL

fluffy:
We have several objects that in this spec that are serialized used the patters in WebIDL-1.  When the RTCCert got added, it used the pattern from what WG but our IDL does not even cover the syntax used. I made a few changes to terminology but no change to how it works to bring in line with rest of spec.


stefhak:
Obsoleted by #1171.

PR #1170, opened 2017-05-06
RTCIceCandidate: add component

### LINKED ISSUE #1167: RTCIceCandidate: component attribute missing?
https://tools.ietf.org/html/rfc5245#section-15.1 has a component-id which however is missing in
    https://w3c.github.io/webrtc-pc/#rtcicecandidate-interface
Given the lack of enthusiasm for non-mux was that intentional?

If this is missing and should be added should this be
```readonly attribute unsigned short?         component;```
or
```readonly attribute RTCIceComponent? component;```

### PR DESCRIPTION
**Pull request by fippo**
adds a component field to RTCIceCandidate. Fixes #1167.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3780,7 +3780,7 @@ interface RTCIceCandidate {
     readonly        attribute DOMString?              sdpMid;
     readonly        attribute unsigned short?         sdpMLineIndex;
     readonly        attribute DOMString?              foundation;
-    readonly        attribute octet?                  component;
+    readonly        attribute RTCIceComponent?        component;
     readonly        attribute unsigned long?          priority;
     readonly        attribute DOMString?              ip;
     readonly        attribute RTCIceProtocol?         protocol;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: webrtc.html [hunk 1]
```diff
@@ -3780,6 +3780,7 @@
     readonly        attribute DOMString?              sdpMid;
     readonly        attribute unsigned short?         sdpMLineIndex;
     readonly        attribute DOMString?              foundation;
+    readonly        attribute byte?                   component;
```
taylor-b: I forgot, `byte` is signed ([−128, 127]) while `octet` isn't, so I guess we want `octet` here.

webrtc.html
```diff
@@ -3855,6 +3856,10 @@
               <dd>A unique identifier that allows ICE to correlate candidates
               that appear on multiple
               <code><a>RTCIceTransport</a></code>s.</dd>
+              <dt><dfn><code>component</code></dfn> of type <span class=
+              "idlAttrType"><a>byte</a></span>, readonly, nullable</dt>
+              <dd>The assigned network component id of the candidate
+              (1 for <code>rtp</code>, 2 for <code>rtcp</code>).</dd>
```
taylor-b: nit: Why are "rtp" and "rtcp" in "code" blocks?

Also, this could reference ICEBis section 4.1.1.1 which defines the term "component id".

webrtc.html
stefhak: I agree, octet is better. But could it not be an enum instead?

webrtc.html
stefhak: what's the component if it is for a data channel?

webrtc.html
taylor-b: Sure; I guess it's impossible for it to be anything other than RTP or RTCP. Plus ORTC already has `RTCIceComponent`.

webrtc.html
fippo: it is 1 for data channels:
```sdpMid: data, sdpMLineIndex: 0, candidate: candidate:1108072699 1 udp 2122262783 8.8.8.8 38900 typ host generation 0 ufrag F9pm network-id 2 network-cost 10```

webrtc.html
fippo: right but so far all the other fields map 1:1 into the candidate attribute.
And this is supposed to be a convenience helper so maybe it shouldn't try to be too smart?

I have no strong opinion either way

webrtc.html
fippo: i'd rather reference ICE 15.1 at the beginning of the attributes. All but sdpMid, sdpMLineIndex and ufrag are defined there

stefhak:
I can live with this (ideally I think it should have been an enum, but this is good enough).

PR #1171, opened 2017-05-06
revert PR#1108

fluffy:
Revert PR 1108 and point at correct refs.


domenic:
On behalf of Chrome, I have to say that we can no longer implement this spec with this change, as it's now in conflict with our implementation of SharedArrayBuffer (which requires modern structured cloning).

I'd suggest referencing specs that implementers actually use, not [outdated forks](https://annevankesteren.nl/2016/01/film-at-11).

PR #1175, opened 2017-05-08
Expanding RTCPeerConnection introduction.

taylor-b:
The first sentence contained a grammatical error, so I fixed it and
expanded on it a bit. Also, changed "peer to peer" to "peer-to-peer"
since the latter is used more consistently.


stefhak:
LGTM.

PR #1176, opened 2017-05-08
Adding more detail to RTCIceTransportPolicy enum descriptions.

taylor-b:
The last sentence of the description for each enum value was somewhat
vague. And not really normative either. So I expanded each of them and
moved them into notes.


stefhak:
LGTM.

Issue #1186, opened 2017-05-09
editorial: terminology of Blob and "blob of SDP"


fippo:
https://w3c.github.io/webrtc-pc/#terminology
says
> The term Blob is defined in [FILEAPI]

While the uppercase term Blob is used in that sense there is the term "sdp blob" which is used like this
> The createOffer method generates a blob of SDP

It would probably be good to explain "blob of SDP" in the terminology to avoid confusion (yes, i am nitpicking here). I've found a fairly good description in JSEP:
> As detailed below, most applications should be able to treat the SessionDescriptions produced and consumed by these various API calls as opaque blobs; that is, the application will not need to read or change them.

but that was on page 2 of the search results

Another possible solution is to defer to JSEP on the first use of the term. It is used only in two places.

PR #1187, opened 2017-05-09
createAnswer: disambiguate blob

fippo:
editorial (while pondering on #1186)

this is a blob of SDP, similar to what is created in an offer and also
more similar to the description of create answer in JSEP


aboba:
While Section 3 (Terminology) says  "The term Blob is defined in [FILEAPI]." the term "blob" here does not appear to be covered by that. So "blob of SDP" is better.

PR #1208, opened 2017-05-12
Add bufferSize attribute to RTCDataChannel.

### LINKED ISSUE #1148: How do applications know a DataChannel's buffer capacity, so they can avoid filling it?
A well-behaving application that sends a large amount of data will use the "bufferedAmount" event and "bufferedAmountLowThreshold", to ensure it doesn't fill the DataChannel's buffer and cause the channel to be closed.

But how does the application know what to set "bufferedAmountLowThreshold" to? If the spec doesn't provide any guarantees about an implementation's buffer size, the application developer is forced to research how large different browsers' buffers are, pick a value low enough, and hope the browser buffers don't get smaller.

So I'd suggest either:

* Providing a `bufferSize` attribute (pretty straightforward)
* Guarantee something about the buffer size. For example, "it MUST be N MB or larger".
* Get rid of the "data channel closing on buffer filling" behavior, and just throw/return an error instead, like a non-blocking socket. This is a more radical change, but is anyone really attached to the current behavior? It seems to cause a fair amount of confusion. See discussion here: https://groups.google.com/forum/#!topic/discuss-webrtc/apmq9VLY9lc

taylor-b:
Fixes #1148.

This is option 1 for fixing the issue. An application can read this
bufferSize attribute, and check "messageSize + bufferedAmount <=
bufferSize" before sending data to ensure it won't abruptly close the
data channel.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8457,6 +8457,7 @@ interface RTCTrackEvent : Event {
     readonly        attribute unsigned short?     id;
     readonly        attribute RTCPriorityType     priority;
     readonly        attribute RTCDataChannelState readyState;
+    readonly        attribute unsigned long       bufferSize;
     readonly        attribute unsigned long       bufferedAmount;
                     attribute unsigned long       bufferedAmountLowThreshold;
                     attribute EventHandler        onopen;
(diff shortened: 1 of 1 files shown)


stefhak:
We should discuss if people think this is the right way to go. Interim topic?

taylor-b:
This is unneeded now since we decided to go with #1209.

PR #1209, opened 2017-05-12
Throw error if data channel's buffer is filled, rather than closing.

### LINKED ISSUE #1148: How do applications know a DataChannel's buffer capacity, so they can avoid filling it?

taylor-b:
Fixes #1148.

Option 2 for fixing the issue. The application doesn't need to know the
buffer's size ahead of time, because filling the buffer is no longer an
irrecoverable error. If it doesn't care about latency, an application
could send until an exception is thrown, then wait for the
"bufferedamountlow" event until it can send again, similar to a
non-blocking socket.

Note that this is deliberately different than how WebSockets work,
because the circumstances are slightly different: with WebSockets, it
was expected that applications had code to handle them being closed
unexpectedly. But this isn't true of WebRTC DataChannels, because before
an SCTP-level timeout can occur, the application would have done an ICE
restart and repaired the problem.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8941,11 +8941,16 @@ interface RTCTrackEvent : Event {
           started, then abort these steps.</p>
         </li>
         <li>
+          <p>If the size of <var>data</var> exceeds the value of <code><a
+          data-link-for="RTCSctpTransport">maxMessageSize</a></code> on
+          <var>channel</var>'s associated <code>RTCSctpTransport</code>,
+          <a>throw</a> an <code>InvalidAccessError</code>.</p>
+        </li>
+        <li>
           <p>Attempt to send <var>data</var> on <var>channel</var>'s
-          <a>underlying data transport</a>; if the data cannot be sent, e.g.
-          because it would need to be buffered but the buffer is full, the user
-          agent MUST abruptly <a>close</a> <var>channel</var>'s <a>underlying
-          data transport</a> <a>with an error</a>.</p>
+          <a>underlying data transport</a>. If the data cannot be sent, buffer
+          it, and if this is not possible because the buffer is full,
+          <a>throw</a> an <code>OperationError</code>.</p>
         </li>
       </ol>
       <div>
--- a/webrtc.html
+++ b/webrtc.html
@@ -8949,8 +8949,8 @@ interface RTCTrackEvent : Event {
         <li>
           <p>Queue <var>data</var> for transmission on <var>channel</var>'s
           <a>underlying data transport</a>. If queuing <var>data</var> is not
-          possible because no more buffer space is available, <a>throw</a> an
-          <code>OperationError</code>.</p>
+          possible because not enough buffer space is available, <a>throw</a>
+          an <code>OperationError</code>.</p>
           <div class="note">The actual transmission of data occurs in
           parallel. If sending data leads to an SCTP-level error, the
           application will be notified asynchronously through <code><a
--- a/webrtc.html
+++ b/webrtc.html
@@ -8894,8 +8894,8 @@ interface RTCTrackEvent : Event {
         </li>
         <li>
           <p>If <var>channel</var>'s <a data-for=
-          "RTCDataChannel"><code>readyState</code></a> attribute is
-          <code>connecting</code>, <a>throw</a> an
+          "RTCDataChannel"><code>readyState</code></a> attribute is not
+          <code>open</code>, <a>throw</a> an
           <code>InvalidStateError</code>.</p>
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -8941,11 +8941,16 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        </li>
+        <li>
           <p>Attempt to send <var>data</var> on <var>channel</var>'s
-          <a>underlying data transport</a>; if the data cannot be sent, e.g.
-          because it would need to be buffered but the buffer is full, the user
-          agent MUST abruptly <a>close</a> <var>channel</var>'s <a>underlying
-          data transport</a> <a>with an error</a>.</p>
+          <a>underlying data transport</a>. If the data cannot be sent, buffer
+          it, and if this is not possible because the buffer is full,
+          <a>throw</a> an <code>OperationError</code>.</p>
```
taylor-b: I don't know if these are the right errors; suggestions welcome.

stefhak:
Interim topic to decide between options? And are they exclusive?

taylor-b:
They're not necessarily mutually exclusive, no.

aboba:
Related to Issue https://github.com/w3c/webrtc-pc/issues/1287

soareschen:
I think there are few clarifications needed:

- When there are already buffers queued, it should attempt to queue the buffer without attempting to send to underlying data transport.
- If there is error sending to underlying transport, should there be error thrown? Or should the interaction with underlying transport be asynchronous and queued as a task?

taylor-b:
> - When there are already buffers queued, it should attempt to queue the buffer without attempting to send to underlying data transport.

Good catch... Reworded to make this clear.

> - If there is error sending to underlying transport, should there be error thrown? Or should the interaction with underlying transport be asynchronous and queued as a task?

It should always be asynchronous. That way the application only has one thing to listen to for SCTP protocol level errors: `onerror`.

> - If readyState is `closing` or `closed`, why is that error ignored while others are thrown?

I don't see how `send` could work if `readyState` is `"closing"` or `"closed"`. I changed the `readyState != "opening"` check to `readyState == "open"`.

soareschen:
> I don't see how send could work if readyState is "closing" or "closed". I changed the readyState != "opening" check to readyState == "open".

:clap: 

LGTM

webrtc.html
```diff
@@ -8941,11 +8941,20 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-          data transport</a> <a>with an error</a>.</p>
+          <p>If the size of <var>data</var> exceeds the value of <code><a
+          data-link-for="RTCSctpTransport">maxMessageSize</a></code> on
+          <var>channel</var>'s associated <code>RTCSctpTransport</code>,
+          <a>throw</a> an <code>InvalidAccessError</code>.</p>
+        </li>
+        <li>
+          <p>Queue <var>data</var> for transmission on <var>channel</var>'s
+          <a>underlying data transport</a>. If queuing <var>data</var> is not
+          possible because no more buffer space is available, <a>throw</a> an
```
stefhak: s/no/not enough/ ?

stefhak:
Made a comment, but I think we can merge now (regardless to whether or not we change based on that comment).

webrtc.html
```diff
@@ -8941,11 +8941,20 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
         <li>
-          <p>Attempt to send <var>data</var> on <var>channel</var>'s
-          <a>underlying data transport</a>; if the data cannot be sent, e.g.
-          because it would need to be buffered but the buffer is full, the user
-          agent MUST abruptly <a>close</a> <var>channel</var>'s <a>underlying
-          data transport</a> <a>with an error</a>.</p>
+          <p>If the size of <var>data</var> exceeds the value of <code><a
+          data-link-for="RTCSctpTransport">maxMessageSize</a></code> on
+          <var>channel</var>'s associated <code>RTCSctpTransport</code>,
+          <a>throw</a> an <code>InvalidAccessError</code>.</p>
```
adam-be: Perhaps we should go with an operation error here as well since InvalidAccessError is deprecated.

taylor-b: Hmm. What's the new "invalid parameter" error, then?

webrtc.html

lgrahl:
We have added this in Firefox 57. I'm currently writing a shim in adapter-js. Any other implementations that support this I should to be aware of (preferrably with version number)?

PR #1220, opened 2017-05-17
reorder createOffer/createAnswer paragraphs

### LINKED ISSUE #1193: editorial: createOffer description
http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-createoffer

has a weird structure which makes it hard to understand. It think it is easy to improve. Some nitpicking first:

> including descriptions of the local MediaStreamTracks attached to this RTCPeerConnection, the codec/RTP/RTCP options supported by this implementation, 

s/options supported by this implementation/capabilities of this implementation/
would make it clearer that there is little relationship between those options and the options argument.

> and any candidates that have been gathered by the ICE Agent.

Typically the candidates won't be included (because trickle). But more importantly, the offer contains the ICE username fragment and password.

The second paragraph seems to expand upon the codec/RTP/RTCP options while carefully avoiding to use any of those words.

The third paragraph makes a general statement about the lifetime. I think its better to move it up before the second paragraph.
> Calling this method is needed to generate the ICE username fragment and password.
This last sentence expands upon the ICE description earlier but does so in a "oh btw" which is rather unhelpful. Should be split into a new paragraph and expanded a bit to include the "candidate" stuff mentioned in the first paragraph.

The fourth paragraph should probably mention that the certificates are also used to create the dtls transports later.

fippo:
@aboba please take a look

fixes #1193


webrtc.html
```diff
@@ -1653,8 +1639,26 @@
...
+                <code>createOffer</code> is called after the session is
+                established, <code>createOffer</code> will generate an offer
+                that is compatible with the current session, incorporating any
+                changes that have been made to the session since the last
+                complete offer-answer exchange, such as addition or removal of
+                tracks. If no changes have been made, the offer will include
+                the capabilities of the current local description as well as
+                any additional capabilities that could be negotiated in an
+                updated offer.</p>
+                <p>The generated SDP will also contain the <a>ICE agent</a>s
```
taylor-b: Missing apostrophe?

webrtc.html
```diff
@@ -1653,8 +1639,26 @@
...
+                <code>usernameFragment</code>, <code>password</code> and
```
taylor-b: `usernameFragment` and `password` could link to the corresponding members of `RTCIceParameters`. `ice options` isn't code so I don't know if it makes sense in a code block.

fippo: done (in separate commit). The closest thing to a definition of ice-options I could find was in https://tools.ietf.org/html/rfc5245#section-14

webrtc.html
```diff
@@ -1840,8 +1840,23 @@
...
+                <a data-link-for="RTCIceParameters">usernameFragment</a>, 
+                <a data-link-for="RTCIceParameters">password</a> and
+                ICE options (as defined in [[!ICE]], Section 14) and may also contain
+                any local candidates that have been gathered by the agent.</p>
+                <p>The value for <code>certificates</code> in the
+                <code><a>RTCConfiguration</a></code> for the
+                <code>RTCPeerConnection</code> is used to produce a set of
+                certificate fingerprints. These certificate fingerprints are
+                used in the construction of SDP and as input to requests for
+                identity assertions.</p>
```
taylor-b: This isn't quite correct, since automatically generated certificates don't appear in the RTCConfiguration. But this isn't a new issue since this is just copying text from `createOffer`.

PR #1226, opened 2017-05-19
Removing WebIDL defaults for various RTP parameters.

### LINKED ISSUE #1173: degradationPreference shouldn't have a default at the WebIDL level
The fact that it has a default at the WebIDL level means that `receiver.getParameters` will always return `degradationPreference: "balanced"`, which doesn't make sense. Instead, we should specify that the `RTCRtpSender` has an internal degradation preference that defaults to "balanced", and `RTCRtpReceiver.getParameters` returns nothing for it.

taylor-b:
Fixes #1173.

`setParameters` is only expected to be called with a result of
`getParameters` (enforced by the `transactionId` field). So, these
WebIDL defaults serve no purpose. If a parameter has a default value,
`getParameters` will have already returned it.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5929,8 +5929,8 @@ sender.setParameters(params)
               preferred. If unset, the <code><a>RtpSender</a></code> defaults to
               the <code>balanced</code> policy.</p>
               <p>For an <code><a>RtpReceiver</a></code>,
-              <code>degradationPreference</code> is inapplicable will always be
-              <code>undefined</code>.</p>
+              <code>degradationPreference</code> is inapplicable and will
+              always be <code>undefined</code>.</p>
             </dd>
           </dl>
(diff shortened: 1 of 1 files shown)


taylor-b:
Addressed comments. Though I didn't add anything for `channels`; a codec's number of channels will be determined entirely from SDP, so there's no need to describe a default.

stefhak:
Looks overall good. However, if `channels` can't be changed, should it not be marked as 'Read-only parameter'?

taylor-b:
Added "Read-only parameter" to all codec parameter descriptions, since codecs can only be reordered. Something odd I noticed in the process is that the description of `payloadType` says "This value can be set to control which codec should be used to send a given encoding," and I don't see what this is supposed to mean so I removed it.

Issue #1224, opened 2017-05-18
RTCIceCandidate and candidate string parsing requirements


rwaldron:
In https://github.com/w3c/web-platform-tests/pull/5917#pullrequestreview-38431577, I included all of the properties of an RTCIceCandidate object for coverage. My understanding is that RTCIceCandidate constructor is expected to parse the contents of the provided candidate string and assign the resulting values to their corresponding instance properties, by correlating the named fields in https://tools.ietf.org/html/rfc5245#section-15.1 to their respective instance property. @soareschen [pointed out](https://github.com/w3c/web-platform-tests/pull/5917#issuecomment-302285678) that there is actually no explicit, normative prose that describes an expectation of parsing the string. 

Tangentially related, @fippo mentions writing a parser for candidate strings in this issue:  https://github.com/w3c/webrtc-pc/issues/1211 which supports the expectation I described above.

soareschen:
This is now fixed by #1229.

Issue #1165, opened 2017-05-05
Lack of validation in RTCIceCandidate constructor


soareschen:
There is no validation requirement for the arguments inside `RTCIceCandidate` constructor other than that `sdpMid` and `sdpMLineIndex` must not be both `null`. This seems to imply that the values in `RTCIceCandidateInit` can be anything as long as they are of valid type.

I suspect that the rationale for this is that the actual validation is only done when the candidate is passed to `addIceCandidate`. If so it is helpful to put a note in the `RTCIceCandidate` constructor section.

I also have difficulty understanding the use of the other fields in `RTCIceCandidate` that are not in `RTCIceCandidateInit`. Since these fields are read only there doesn't seem to have way to set their values at all. There is also no place in the spec where these fields are being read, including `addIceCandidate`.

foolip:
I also filed https://github.com/w3c/webrtc-pc/issues/1166

stefhak:
If adding a note is enough then this is editorial.

foolip:
"There is also no place in the spec where these fields are being read, including addIceCandidate" is not an editorial issue.

rwaldron:
It struck me as odd that: 

> sdpMid of type DOMString, readonly, nullable
> sdpMLineIndex of type unsigned short, readonly, nullable

That is: both are nullable, but then...

> When run, if both the sdpMid and sdpMLineIndex dictionary members are null, throw a TypeError.

foolip:
That's a case where Web IDL syntax can't express the constraints, and so it's done in prose instead. If a single dictionary member is required, the required keyword is used, but if it's more complicated, then things end up looking like this...

soareschen:
This is now fixed by #1229.

Issue #1211, opened 2017-05-13
RTCIceCandidate: how is extensibility handled?


fippo:
http://w3c.github.io/webrtc-pc/#rtcicecandidate-interface
defines a number of attributes of the RTCIceCandidate that should be pre-parsed from the candidate string. While implementing a shim for this I noticed that my parser was emitting two additional attributes, generation and [network-cost](https://tools.ietf.org/html/draft-thatcher-ice-network-cost-01)
```
Candidate:3805835391 1 udp 2113937151 192.168.0.107 57917 typ host generation 0 
ufrag 7POS network-cost 50
```

https://tools.ietf.org/html/rfc5245#section-15.1 allows extensibility (extension-att-name and extension-att-value) so this section should at the very least mention what happens with those extra attributes.

This might be somewhat tricky since e.g. when Firefox does not implement network-cost and therefore parsing a candidate from Chrome in the constructor will at best return a string network cost "50" while a browser that understands this will return an integer 50.
This might imply that this parsing is easier to be done in a javascript library which will give a consistent answer across browsers. So far nobody has implemented this even though I have a [PR for adapter.js to shim it](https://github.com/webrtc/adapter/pull/527).

taylor-b:
I don't think WebRTC 1.0 needs to handle this. If an application really needs to inspect non-standard attributes, it's already possible to parse the candidate string in JS; the `RTCIceCandidate` attributes are just there for convenience.

fippo:
But if you sometimes have to parse the candidate to find out what you need you can just as well unconditionally parse it myself and get consistent behaviour.

I really like this as a convenience function for onicecandidate. And if the effort to specify it is too high then i think it should be left to javascript.

rwaldron:
@taylor-b I need a little more clarification, I appreciate your patience. Can you explain how the values of these properties are populated? (from https://w3c.github.io/webrtc-pc/#rtcicecandidate-interface)

- [ ] `foundation`
- [ ] `priority`
- [ ] `ip`
- [ ] `protocol`
- [ ] `port`
- [ ] `type`
- [ ] `tcpType`
- [ ] `relatedAddress`
- [ ] `relatedPort`



@foolip 

> And if the effort to specify it is too high then i think it should be left to javascript.

I think it could be as simple as adding the following sentence (or something similar) to each property definition: 

[...]

PR #1229, opened 2017-05-19
Add detailed steps for constructing RTCIceCandidate

soareschen:
This PR address the issues in #1165 and #1224. It is also needed to have complete test coverage for `RTCIceCandidate()` in w3c/web-platform-tests#5917.

I am not very sure of the details for editing the spec, so this is a first cut draft that attempt to explain the changes to be made.

The main change is to add detailed steps for the `RTCIceCandidate` constructor to parse the candidate string, referencing the relevant specs, and populating the attributes with the parse result. 

To maintain backward compatibility, any parsing error is ignored and the derivative attributes are simply set with the default `null` value. This might not be the best approach, but I am not sure if we can make it throw an error since current browser implementations do not have the derivative attributes and have not done any parsing on the candidate string.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3851,18 +3851,49 @@ interface RTCIceCandidate {
     serializer = {candidate, sdpMid, sdpMLineIndex, ufrag};
 };</pre>
           <section>
-            <h2>Constructors</h2>
+            <h2>Constructor</h2>
             <dl data-link-for="RTCIceCandidate" data-dfn-for="RTCIceCandidate"
             class="constructors">
               <dt><code>RTCIceCandidate</code></dt>
               <dd>
-                The <dfn><code>RTCIceCandidate()</code></dfn> constructor takes
+                <p>The <dfn><code>RTCIceCandidate()</code></dfn> constructor takes
                 a dictionary argument, <var>candidateInitDict</var>, whose
                 content is used to initialize the new <a class=
-                "idlType"><code>RTCIceCandidate</code></a> object. When run, if
-                both the <code><a>sdpMid</a></code> and
-                <code><a>sdpMLineIndex</a></code> dictionary members are
-                <code>null</code>, <a>throw</a> a <code>TypeError</code>.
+                "idlType"><code>RTCIceCandidate</code></a> object.</p>
+                <p>When invoked, run the following steps:</p>
+                <ol>
+                  <li>If both the <code><a>sdpMid</a></code> and <code>
+                  <a>sdpMLineIndex</a></code> dictionary members in
+                  <var>candidateInitDict</var> are <code>null</code>, <a>throw</a>
+                  a <code>TypeError</code>.</li>
+                  <li>Let <var>iceCandidate</var> be a newly created
+                  <code>RTCIceCandidate</code> object.</li>
+                  <li>Set the corresponding attributes in <var>iceCandidate</var>
+                  with the dictionary member values of <var>candidateInitDict</var></li>
+                  <li>Let <var>candidate</var> be the <code><a>candidate</a></code>
+                  dictionary member of <var>candidateInitDict</var>. If
+                  <var>candidate</var> is not an empty string, run the following steps:
+                  <ol>
+                    <li>Parse <var>candidate</var> using the grammar for
+                    <code>candidate-attribute</code> as defined in section 15.1 of
+                    [[!ICE]], together with the grammar extension for ICE TCP as
+                    defined in section 4.5 of [[!RFC6544]].</li>
+                    <li>If parsing of <code>candidate-attribute</code> is successful,
+                    set the corresponding attributes in <var>iceCandidate</var> to
+                    the field values of the parse result.</li>
+                  </ol></li>
+                  <li>Return <var>iceCandidate</var></li>
+                </ol>
+                <div class="note">
+                  The validation for members in <var>candidateInitDict</var>
+                  is not done in the constructor for <code>RTCIceCandidate</code>.
+                  Instead, validation is done when calling <code><a data-for=
+                  "RTCPeerConnection">addIceCandidate()</a></code>. To maintain
+                  backward compatibility, if the <var>candidate</var> member do not
+                  match the grammar for <code>candidate-attribute</code>, the value
+                  remain set in the <code>candidate</code> attribute with the other
+                  derivative fields set to <code>null</code>
+                </div>
                 <table class="parameters">
                   <tbody>
                     <tr>
@@ -3925,7 +3956,9 @@ interface RTCIceCandidate {
               <dt><dfn><code>ip</code></dfn> of type <span class=
               "idlAttrType"><a>DOMString</a></span>, readonly, nullable</dt>
               <dd>
-                <p>The IP address of the candidate.</p>
+                <p>The IP address of the candidate. This corresponds to the
+                <code>connection-address</code> field in
+                <code>candidate-attribute</code></p>
                 <div class="note">
                   <p>The IP addresses exposed in candidates gathered via ICE
                   and made visibile to the application in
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3835,18 +3835,49 @@
...
+                <ol>
+                  <li>If both the <code><a>sdpMid</a></code> and <code>
+                  <a>sdpMLineIndex</a></code> dictionary members in
+                  <var>candidateInitDict</var> are <code>null</code>, <a>throw</a>
+                  a <code>TypeError</code>.</li>
+                  <li>Let <var>iceCandidate</var> be a newly created
+                  <code>RTCIceCandidate</code> object.</li>
+                  <li>Set the corresponding attributes in <var>iceCandidate</var>
+                  with the dictionary member values of <var>candidateInitDict</var></li>
+                  <li>Let <var>candidate</var> be the <code><a>candidate</a></code>
```
taylor-b: Does this `<a>` link to the right `candidate`?

webrtc.html
```diff
@@ -3835,18 +3835,49 @@
...
+                    the field values of the parse result.</li>
+                  </ol></li>
+                  <li>Return <var>iceCandidate</var></li>
+                </ol>
+                <div class="note">
+                  The validation for members in <var>candidateInitDict</var>
+                  is not done in the constructor for <code>RTCIceCandidate</code>.
+                  Instead, validation is done when calling <code><a data-for=
+                  "RTCPeerConnection">addIceCandidate()</a></code>. To maintain
+                  backward compatibility, if the <var>candidate</var> member do not
```
taylor-b: s/do/does

webrtc.html
```diff
@@ -3835,18 +3835,49 @@
...
+                  match the grammar for <code>candidate-attribute</code>, the value
+                  remain set in the <code>candidate</code> attribute with the other
+                  derivative fields set to <code>null</code>
```
taylor-b: Meaning, the DOMString `candidate` attribute will return the string passed in, but the parsed attributes will return `null`? I think this could be more clear.

webrtc.html
```diff
@@ -3835,18 +3835,49 @@
...
+                  dictionary member of <var>candidateInitDict</var>. If
+                  <var>candidate</var> is not an empty string, run the following steps:
+                  <ol>
+                    <li>Parse <var>candidate</var> using the grammar for
+                    <code>candidate-attribute</code> as defined in section 15.1 of
+                    [[!ICE]], together with the grammar extension for ICE TCP as
+                    defined in section 4.5 of [[!RFC6544]].</li>
+                    <li>If parsing of <code>candidate-attribute</code> is successful,
+                    set the corresponding attributes in <var>iceCandidate</var> to
+                    the field values of the parse result.</li>
```
taylor-b: s/parse/parsed?

webrtc.html
```diff
@@ -3835,18 +3835,49 @@
...
+                  </ol></li>
+                  <li>Return <var>iceCandidate</var></li>
+                </ol>
+                <div class="note">
+                  The validation for members in <var>candidateInitDict</var>
+                  is not done in the constructor for <code>RTCIceCandidate</code>.
```
taylor-b: This statement is somewhat misleading, since `sdpMid` and `sdpMLineIndex` (or at least, their presence) are validated above.

webrtc.html
```diff
@@ -3835,18 +3835,49 @@
...
+                <p>When invoked, run the following steps:</p>
+                <ol>
+                  <li>If both the <code><a>sdpMid</a></code> and <code>
+                  <a>sdpMLineIndex</a></code> dictionary members in
+                  <var>candidateInitDict</var> are <code>null</code>, <a>throw</a>
+                  a <code>TypeError</code>.</li>
+                  <li>Let <var>iceCandidate</var> be a newly created
+                  <code>RTCIceCandidate</code> object.</li>
+                  <li>Set the corresponding attributes in <var>iceCandidate</var>
+                  with the dictionary member values of <var>candidateInitDict</var></li>
```
taylor-b: Missing period

webrtc.html
```diff
@@ -3835,18 +3835,49 @@
...
+                  <ol>
+                    <li>Parse <var>candidate</var> using the grammar for
+                    <code>candidate-attribute</code> as defined in section 15.1 of
+                    [[!ICE]], together with the grammar extension for ICE TCP as
+                    defined in section 4.5 of [[!RFC6544]].</li>
+                    <li>If parsing of <code>candidate-attribute</code> is successful,
+                    set the corresponding attributes in <var>iceCandidate</var> to
+                    the field values of the parse result.</li>
+                  </ol></li>
+                  <li>Return <var>iceCandidate</var></li>
```
taylor-b: Missing period.

webrtc.html
```diff
@@ -3835,18 +3835,49 @@
...
+                  <li>Let <var>iceCandidate</var> be a newly created
+                  <code>RTCIceCandidate</code> object.</li>
+                  <li>Set the corresponding attributes in <var>iceCandidate</var>
+                  with the dictionary member values of <var>candidateInitDict</var></li>
+                  <li>Let <var>candidate</var> be the <code><a>candidate</a></code>
+                  dictionary member of <var>candidateInitDict</var>. If
+                  <var>candidate</var> is not an empty string, run the following steps:
+                  <ol>
+                    <li>Parse <var>candidate</var> using the grammar for
+                    <code>candidate-attribute</code> as defined in section 15.1 of
```
taylor-b: Can you use `<dfn>` for `candidate-attribute` and link to the definition where it's referenced below?

soareschen:
> Another question: what should happen if a field has a value that doesn't match any corresponding enum value. For example, a type that isn't "host/srflx/relay", but something else, possibly from an ICE extension the browser doesn't support.

This seems to related to #1211. We'd then have to relax the constraint and allow the `candidate-attribute` grammar to be extensible.

I have updated the paragraphs and attempt to clarify things based on feedback.

soareschen:
Rebased. I have added reference to `candidate-attribute` for the new `component` field in `RTCIceCandidate`. I also clarify the parsing step to require the parsing to either all pass or fail. e.g. If `component-id` has a value other than `1` or `2`, or if the IP address is malformed, the steps are aborted and all derivative attributes are set to null even if the other field values are valid. This is to prevent browsers from doing non strict parsing which could make it difficult to interpret the fields.

Please let me know if the proposal is sound. Frankly I feel that the simpler approach is to make all attributes non-nullable and simply throw an error when parsing failed. Because otherwise it would be hard for users to use these fields if their values are nullable depending on whether parsing is successful.

webrtc.html
```diff
@@ -3839,18 +3844,60 @@
...
+                    <li>If any field in the parse result represents an invalid value for the
+                    corresponding attribute in <var>iceCandidate</var>, abort these steps.</li>
+                    <li>Set the corresponding attributes in <var>iceCandidate</var> to
+                    the field values of the parsed result.</li>
+                  </ol></li>
+                  <li>Return <var>iceCandidate</var>.</li>
+                </ol>
+                <div class="note">
+                  <p>The constructor for <code>RTCIceCandidate</code> only does basic
+                  type checking for the dictionary members in <var>candidateInitDict</var>.
```
taylor-b: This sentence seems slightly misleading, since it doesn't just do basic type checking; it also does basic parsing.

webrtc.html
```diff
@@ -3839,18 +3844,60 @@
...
+                  <code>addIceCandidate()</code></a>.</p>
+
+                  <p>To maintain backward compatibility, any error on parsing the
+                  <var>candidate</var> attribute is ignored. In such case, the
+                  <a><code>candidate</code></a> attribute holds the raw
+                  <a data-dfn-for="idl-def-rtcicecandidateinit-candidate">
+                  <code>candidate</code></a> string given in <var>candidateInitDict</var>,
+                  but derivative attributes such as <a><code>foundation</code></a>,
+                  <a><code>priority</code></a>, etc are set to their default value,
+                  which is <code>null</code>.</p>
```
taylor-b: The "derivative attributes are set to `null` part" could be a bit more clear in the normative text. Maybe step 1 could be "initialize attribute values to `null`". Or is this implied?

webrtc.html
soareschen: I have updated the sentence to _"only does basic parsing and type checking"_.

webrtc.html
soareschen: I think the default value for nullable attributes is `null`. Anyway I clarified it in the constructor steps by explicitly listing the attributes to initialize.

stefhak:
Is there any conflict between this one and https://github.com/w3c/webrtc-pc/pull/1372/files? I don't think so, and I think this one can be merged!

PR #1230, opened 2017-05-20
Use data-cite for RTCStatsType

### LINKED ISSUE #1212: Can we directly link to WebRTC-Stats for RTCStatsType?
Current the spec [defines an empty RTCStatsType enum](http://www.w3.org/TR/webrtc/#rtcstats-dictionary) and then explains about the emptiness. Can we just directly link to http://www.w3.org/TR/webrtc-stats/#dom-rtcstatstype to prevent this empty IDL, as the spec already does on e.g. [EventHandler](http://www.w3.org/TR/webrtc/#rtcsctptransport-interface)?

SaschaNaz:
...which is explained [on ReSpec wiki](https://github.com/w3c/respec/wiki/data--cite) and [mailing list](https://lists.w3.org/Archives/Public/spec-prod/2017JanMar/0011.html). Should fix #1212.


SaschaNaz:

stefhak:
LGTM.

PR #1239, opened 2017-05-24
RTCIceConnectionEventInit: url is nullable

fippo:
describes the url as nullable as it is done in the idl definition.

Description is missing too. ORTC has "The URL of the server from which the candidate was obtained." shall I steal that (with @aboba's permission)?


stefhak:
LGTM, and your ortc steal sounds like something we should do.

fippo:
taken the definition from the previous section instead and squashed.

PR #1247, opened 2017-05-25
RTP/RTCP mux -> RTCP mux

fippo:
replace single use of RTP/RTCP mux with just RTCP mux for consitency with
other uses

We also don't define "RTCP mux" anywhere and don't have a reference to RFC 5763. Should we?

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7533,7 +7533,7 @@ sender.setParameters(params)
               <p>The <dfn id=
               "dom-icetransport-component"><code>component</code></dfn>
               attribute MUST return the ICE component of the transport. When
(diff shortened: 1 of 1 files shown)


taylor-b:
In the description of `RTCIceComponent`, there's also the phrase "RTP/RTCP-multiplexing". Can you replace that too?

fippo:
ah... I was looking for that but searching for "RTCP mux" didn't catch it. Updated + squashed

taylor-b:
Regarding the reference, RFC 5763 actually never uses the phrase "RTCP mux", but JSEP frequently does. So I'm not sure what to do about that.

But this change in itself LGTM.

PR #1253, opened 2017-05-29
clean up logic to support removal of RTCP mux "negotiated" value

fluffy:
This PR makes it easier to remove this at risk value and at the same times makes the text better for future extensions that define new values.


taylor-b:
I don't expect `RTCRtcpMuxPolicy` to be extended in the future; it's more likely it would be removed entirely. So, as long as there are only two possible values, and only one could possibly be unsupported, the existing text seems more clear than what this PR proposes.

stefhak:
Seems people think this is not needed. closing.

PR #1267, opened 2017-05-30
Remove "at-risk" marking from Identity

ekr:
As discussed on the May Virtual Interim:
https://lists.w3.org/Archives/Public/public-webrtc/2017May/0013.html


adam-be:
Already fixed by #1252

Issue #1285, opened 2017-05-31
Section 5.5.1: RFC 8122 consistency


aboba:
From EKR: https://lists.w3.org/Archives/Public/public-webrtc/2017May/0166.html

    One of the the hash function algorithms defined in the 'Hash function
    Textual Names' registry, initially specified in [ RFC4572] Section
    8. As noted in [JSEP] Section 5.2.1, the digest algorithm used for the
    fingerprint matches that used in the certificate signature.

I see that this was relaxed in [RFC 8122](https://tools.ietf.org/html/rfc8122). I wonder if we want to relax it here and in JSEP.

aboba:
My suggestion is to delete the second sentence, since the text has changed in [JSEP](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-20#section-5.2.1):

   For DTLS, all m= sections MUST use all the certificate(s) that have
   been specified for the PeerConnection; as a result, they MUST all
   have the same [I-D.ietf-mmusic-4572-update] fingerprint value(s), or
   these value(s) MUST be session-level attributes.

PR #1299, opened 2017-05-31
RFC 8122 consistency

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1285


stefhak:
@ekr can you review?

taylor-b:
JSEP still references this requirement:

>      An "a=fingerprint" line for each of the endpoint's certificates,
>      as specified in [RFC4572], Section 5; the digest algorithm used
>      for the fingerprint MUST match that used in the certificate
>      signature.

This seems like an issue with JSEP itself; it references RFC4572-update in some places and RFC4572 in other.

Issue #1276, opened 2017-05-31
Section 4.2.6: informative sentence


aboba:
From EKR: https://lists.w3.org/Archives/Public/public-webrtc/2017May/0166.html

   Gather ICE candidates for each media type in use (audio, video, and
   data). If the remote endpoint is not bundle-aware, negotiate only
   one audio and video track on separate transports.

The second sentence here (and below) are just information, they're
not instructions to the implementation, so they should be removed
or rewritten.

PR #1301, opened 2017-05-31
Section 4.2.6: remove informative sentences

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1276


taylor-b:
I don't think this is necessary since the entire enum description is meant to be non-normative (relying on JSEP for the normative definitions). The title of the table is "Enumeration description (non-normative)".

Issue #1204, opened 2017-05-11
setDirection needs an internal slot and an algorithm.


jan-ivar:
[direction](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-direction) says: *"The value of direction does not change except due to calls to setDirection."*

That seems wrong, as it appears to be changeable through addTrack and removeTrack as well.

[addTrack](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtrack) uses the phrase *"Enable sending direction ... "*.

[removeTrack](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-removetrack) says: *"set transceiver.direction to recvonly."*

Unfortunately, direction is a read-only attribute, so that doesn't work.

Even [setDirection](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setdirection) itself violates this: *"Set transceiver.direction to newDirection."*

Instead, I think transceiver needs a [[direction]] internal slot, and a set direction algorithm reused by the three places.

taylor-b:
The use of terminology like "set transceiver.direction to newDirection" is a more general problem, which I tried to summarize in issue #1112. @jan-ivar, could you weigh in on that issue?

adam-be:
#1321 adds internal two internal slots [[Direction]] and [[CurrentDirection]]

taylor-b:
I think this is resolved by #1321. Closing.

taylor-b:
Nevermind. The remaining work to be done is that `addTrack` and `removeTrack` should reference the same "set direction" algorithm.

PR #1321, opened 2017-06-01
Start integrating direction into 'create RTCRtpTransceiver' algorithm

### LINKED ISSUE #1206: transceiver.direction isn't initialized.
While the options dictionary [RTCRtpTransceiverInit](https://w3c.github.io/webrtc-pc/#idl-def-rtcrtptransceiverinit) has `direction = "sendrecv"`, it is unused.

Also, we need to initialize transceiver.direction not just in [addTransceiver](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtransceiver), but in the [Create a Transceiver](https://w3c.github.io/webrtc-pc/#dfn-create-an-rtcrtptransceiver) algorithm, used in several places.

adam-be:
Fixes #1206

Work in progress

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3815,7 +3815,7 @@ interface RTCSessionDescription {
                   <li>
                     <p>If <var>connection</var>'s <code><a data-for=
                     "RTCPeerConnection">currentLocalDescription</a></code>
-                    if of type "offer", and the direction of the associated m=
+                    is of type "offer", and the direction of the associated m=
                     section in neither the offer nor answer matches
                     <var>t</var>'s [[<a href="#direction-slot">Direction</a>]]
                     slot, return "true".</p>
@@ -3823,7 +3823,7 @@ interface RTCSessionDescription {
                   <li>
                     <p>If <var>connection</var>'s <code><a data-for=
                     "RTCPeerConnection">currentLocalDescription</a></code>
-                    if of type "answer", and the direction of the associated m=
+                    is of type "answer", and the direction of the associated m=
                     section in the answer does not match <var>t</var>'s
                     [[<a href="#direction-slot">Direction</a>]] slot intersected with
                     the offered direction (as described in <span data-jsep=
@@ -4987,7 +4987,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                       precisely, the [[<a href=
                       "#curr-direction-slot">CurrentDirection</a>]] slot of the
                       <code><a>RTCRtpTransceiver</a></code> associated with the
-                      sender has never had  a value of <code>sendrecv</code> or
+                      sender has never had a value of <code>sendrecv</code> or
                       <code>sendonly</code>.</p>
                     </li>
                   </ul>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1243,6 +1243,14 @@
                         "RTCRtpTransceiver">mid</a></code> value to the mid of
                         the corresponding media description.</p>
                       </li>
+                      <li>
+                        <p>If <var>description</var> is of type "answer", then
+                        set <var>transceiver</var>'s <code><a data-for=
+                        "RTCRtpTransceiver">currentDirection</a></code> value
+                        to an <code><a>RTCRtpTransceiverDirection</a></code>
+                        value representing the direction of the corresponding
+                        media description.</p>
+                      </li>
                     </ol>
                   </li>
                   <li>
@@ -1250,6 +1258,14 @@
                     then run the following steps for each media description in
                     <var>description</var>:</p>
                     <ol>
+                      <li>
+                        <p>Let <var>direction</var> be an <code>
+                        <a>RTCRtpTransceiverDirection</a></code> value
+                        representing the direction from the media
+                        description, but with the send and receive
+                        directions reversed to represent this peer's point
+                        of view.</p>
+                      </li>
                       <li>
                         <p>As described by <span data-jsep=
                         "applying-a-remote-desc">[[!JSEP]]</span>, attempt to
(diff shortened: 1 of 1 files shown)


stefhak:
@jan-ivar is this going in the right direction?

adam-be:
Patched 'create an RTCRtpTransceiver' algorithm. Updated addTrack, addTransceiver and 'set an RTCDescription' steps.

webrtc.html
```diff
@@ -1243,6 +1243,14 @@
                         "RTCRtpTransceiver">mid</a></code> value to the mid of
                         the corresponding media description.</p>
                       </li>
+                      <li>
+                        <p>If <var>description</var> is of type "answer", then
+                        set <var>transceiver</var>'s <code><a data-for=
+                        "RTCRtpTransceiver">currentDirection</a></code> value
```
jan-ivar: This is setting a read-only property. See https://github.com/w3c/webrtc-pc/issues/1204.

adam-be: True. I'll have a look at it.

adam-be:
I've modified this PR to define and use internal slots for direction and currentDirection. I've also updated the other text to use these slots so that fixes #1204. Please have a look @jan-ivar.

adam-be:
Ping @jan-ivar

webrtc.html
```diff
@@ -5208,6 +5238,12 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   empty list otherwise.</p>
                 </li>
                 <li>
+                  <p>If the dictionary argument is present, let
+                  <var>direction</var> be the value of the
+                  <code>direction</code> member. Otherwise let
+                  <var>direction</var> be <code>sendrecv</code>.</p>
```
jan-ivar: FWIW all other uses of "If the dictionary argument is present," are redundant as well.

webrtc.html
```diff
@@ -1243,6 +1243,14 @@
                         "RTCRtpTransceiver">mid</a></code> value to the mid of
                         the corresponding media description.</p>
                       </li>
+                      <li>
```
adam-be: > 7. If description is set as a local description, then run the following steps for each media description in description that is not yet associated with an RTCRtpTransceiver object:

This has to do with the mid. If a transceiver has beed assigned the mid of a media description, they are associated.

> 1. Let transceiver be the RTCRtpTransceiver used to create the media description.

The media description was created from data in the transceiver. They might not be associated though.

Clarifications to this part should be done in a separate PR I think.

webrtc.html
```diff
@@ -1243,6 +1243,14 @@
...
+                        <p>If <var>description</var> is of type "answer", then
```
adam-be: I don't know what to do for `"pranswer"` in this case.

webrtc.html
```diff
@@ -1243,6 +1243,14 @@
...
+                        set <var>transceiver</var>'s [[<a href=
+                        "#curr-direction-slot">CurrentDirection</a>]] slot
```
adam-be: Thanks for making an issue. This has been brought up at the editor's meeting as well.

webrtc.html
```diff
@@ -3782,8 +3805,8 @@
                 then perform the following checks:</p>
                 <ul>
                   <li>
-                    <p>If <var>t</var>'s <a data-for=
-                    "RTCRtpTransceiver">direction</a> is "sendrecv" or
+                    <p>If <var>t</var>'s [[<a href=
+                    "#direction-slot">Direction</a>]] slot is "sendrecv" or
```
adam-be: Let's do the "inactive" stuff in a separate issue/PR.

"sendrecv" is a string, but it's also an enum value. We should decide on syntax for enum values and use it everywhere.

webrtc.html
```diff
@@ -3794,15 +3817,15 @@
                     "RTCPeerConnection">currentLocalDescription</a></code>
                     if of type "offer", and the direction of the associated m=
                     section in neither the offer nor answer matches
-                    <var>t</var>'s <a data-for=
```
adam-be: yes

webrtc.html
```diff
@@ -4961,10 +4984,11 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
                     <li>
                       <p>The sender has never been used to send. More
-                      precisely, the <code><a>RTCRtpTransceiver</a></code>
-                      associated with the sender has never had a <code><a
-                      data-link-for="RTCRtpTransceiver">currentDirection</a></code>
-                      of <code>sendrecv</code> or <code>sendonly</code>.</p>
+                      precisely, the [[<a href=
+                      "#curr-direction-slot">CurrentDirection</a>]] slot of the
+                      <code><a>RTCRtpTransceiver</a></code> associated with the
+                      sender has never had  a value of <code>sendrecv</code> or
```
adam-be: fixed

webrtc.html
adam-be: Let's fix all these in #1369

webrtc.html
adam-be: Created issue #1370 to track the question about code-tags or strings for enum values.

stefhak:
I think we should merge now. @jan-ivar did you open all three issues talked about in https://github.com/w3c/webrtc-pc/pull/1321#pullrequestreview-43572624?

jan-ivar:
I haven't. @taylor-b would you mind opening them? I need to step away for a bit.

webrtc.html
taylor-b: `inactive` doesn't need explicit handling here. This step is just to make sure an "a=msid" line gets added. If you set the direction to "inactive" and it was previously something else, one of the next two bullets will return `true`:

>- If connection's currentLocalDescription if of type "offer", and the direction of the associated m= section in neither the offer nor answer matches t's direction, return "true".
>- If connection's currentLocalDescription if of type "answer", and the direction of the associated m= section in the answer does not match t's direction intersected with the offered direction (as described in [JSEP] (section 5.3.1.)), return "true".

taylor-b:
Filed issues, though I didn't file one for the "inactive" thing, since I'm fairly confident it's not an issue. See comment: https://github.com/w3c/webrtc-pc/pull/1321/files#r122254074

webrtc.html
jan-ivar: Makes sense, thanks! My only remaining issue here then is https://github.com/w3c/webrtc-pc/issues/1403.

PR #1326, opened 2017-06-02
Remove trailing whitespace

ekr:


fluffy:
This is only whitespace change in source so I have merged it.

Issue #1260, opened 2017-05-29
Does insertDTMF append to or replace the current tone buffer


fluffy:
Step 7 says that it replaces tone buffer it but the name implies otherwise so this might need some clarity to the introduction. If this changes, need to also update example 20.

aboba:
@fluffy Some of the examples appear incorrect.  Filed Issue https://github.com/w3c/webrtc-pc/issues/1336

PR #1338, opened 2017-06-05
Clarification: insertDTMF replaces the current tone buffer

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1260

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9325,7 +9325,7 @@ interface RTCDataChannelEvent : Event {
             <dd>
               <p>The <dfn>dtmf</dfn> attribute returns a RTCDTMFSender which
               can be used to send DTMF. A null value indicates that this
-              RTCRtpSender cannot send DTMF.</p>
+              <code><a>RTCRtpSender</a></code> cannot send DTMF.</p>
             </dd>
           </dl>
         </section>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -9453,7 +9453,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   </ol>
                 </li>
               </ol>
-              <p>Calling <code><a data-for=
+              <p>Since <code>insertDTMF</code> replaces the tone
+              buffer, in order to add to the DTMF tones to be played,
+              it is necessary to append additional tones to the
+              tone buffer. Calling <code><a data-for=
```
burnburn: Suggestion:  "in order to add to the DTMF tones being played, it is necessary to call <code>insertDTMF</code> with a string containing both the remaining tones (stored in <code>toneBuffer</code>) and the new tones appended together."

stefhak:
LGTM.

webrtc.html
```diff
@@ -9453,7 +9453,12 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   </ol>
                 </li>
               </ol>
-              <p>Calling <code><a data-for=
+              <p>Since <code>insertDTMF</code> replaces the tone
+              buffer, in order to add to the DTMF tones being played,
+              it is necessary to call <code>insertDTMF</code> with a
+              string containing both the remaining tones (stored in
+              <code>toneBuffer</code>) and the new tones appended
+              together. Calling <code><a data-for=
```
taylor-b: Is this actually safe? A tone could be sent in between `toneBuffer` being read and `insertDTMF` being called with the additional tones, resulting in one being sent twice.

stefhak: Since we have playout tasks and is on main loop this should be safe.

Issue #1272, opened 2017-05-31
Section 4.2.5/4.2.6: Enum Table Inconsistency


aboba:
From EKR: https://lists.w3.org/Archives/Public/public-webrtc/2017May/0166.html

There seems to be inconsistency about whether these tables
of enums are normative or non-normative. Compare S 4.2.6
to S 4.2.5. Both of these seem to be defined in JSEP.

   Gather ICE candidates for each media type in use (audio, video, and
   data). If the remote endpoint is not bundle-aware, negotiate only
   one audio and video track on separate transports.

The second sentence here (and below) are just information, they're
not instructions to the implementation, so they should be removed
or rewritten.

aboba:
@ekr Since neither Sections 4.2.5 nor 4.2.6 are denoted as non-normative, they both appear to be normative.  I do not recall any discussion about marking either RTCIceTransportPolicy or RTCBundlePolicy as "features at risk".  Is there something in this area that we should be talking about?

From what I can tell, the enum definitions are consistent with JSEP.  For example, RTCIceTransportPolicy (Section 4.2.5) is discussed in JSEP Sections 3.5.3 and 4.1.6. 
RTCBundlePolicy (Section 4.2.6) is discussed in Section JSEP 4.1.6.

PR #1340, opened 2017-06-05
Section 4.2.5/4.2.6: Enum Table Inconsistency

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1272

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -501,7 +501,7 @@
           "RTCIceTransportPolicy" class="simple">
             <tbody>
               <tr>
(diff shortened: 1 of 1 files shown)


aboba:
@ekr Is this what you were looking for?

taylor-b:
In the `RTCIceTransportPolicy` description above the table, should change "As noted in JSEP" to "As described in JSEP" if we want to rely on JSEP for the normative definitions. Otherwise this seems fine.

PR #1358, opened 2017-06-08
RTCDataChannel: Use internal slots and specify default values at one location

### LINKED ISSUE #1353: Verbose description for RTCDataChannel default values
In the attribute descriptions for RTCDataChannel, there are many verbose sentences that describe the same action as setting the attributes with the default/provided values in RTCDataChannelInit dictionary.

> The attribute must be initialized to ... by default and must return the value to which it was set when the RTCDataChannel was created.

Is there any reason why the sections require verbose description of default values? They are not really required since the sections that create data channel initialize the attributes with explicit values already.

> 6.1 createDataChannel
>   5. Set channel's ordered, maxPacketLifeTime, maxRetransmits, protocol, negotiated, id and priority attributes to the values of the corresponding members of the dataChannelDict argument, using a value of null if the corresponding dictionary member is missing.

> 6.2 When an underlying data transport is to be announce
>    4. Initialize channel's label, ordered, maxPacketLifeTime, maxRetransmits, protocol, negotiated and id attributes to their corresponding values in configuration.

For the attributes with default null value, the sentence can be omitted by making the respective fields in RTCDataChannelInit nullable with default value null.

adam-be:
fixes #1353

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8284,74 +8284,107 @@ interface RTCTrackEvent : Event {
                   <code><a>RTCDataChannel</a></code> object.</p>
                 </li>
                 <li>
-                  <p>Initialize <var>channel</var>'s <code><a data-for=
-                  "RTCDataChannel">label</a></code> attribute to the value of
-                  the first argument.</p>
+                  <p>Let <var>channel</var> have an [[<dfn id=
+                  "label-slot">Label</dfn>]] internal slot initialized to the
+                  value of the first argument.</p>
                 </li>
                 <li>
-                  <p>Set <var>channel</var>'s <code><a data-for=
-                  "RTCDataChannel">ordered</a></code>, <code><a data-for=
-                  "RTCDataChannel">maxPacketLifeTime</a></code>,
-                  <code><a data-for="RTCDataChannel">maxRetransmits</a></code>,
-                  <code><a data-for="RTCDataChannel">protocol</a></code>,
-                  <code><a data-for="RTCDataChannel">negotiated</a></code>,
-                  <code><a data-for="RTCDataChannel">id</a></code> and
-                  <code><a data-for="RTCDataChannel">priority</a></code> attributes
-                  to the values of the corresponding members of the
-                  <var>dataChannelDict</var> argument, using a value of
-                  <code>null</code> if the corresponding dictionary member is
-                  missing.</p>
-                  <div class="note">The members that do not have default
-                  values, and thus may be missing, are <code><a data-for=
-                  "RTCDataChannelInit">maxPacketLifeTime</a></code>, <code><a
-                  data-for="RTCDataChannelInit">maxRetransmits</a></code>, and
-                  <code><a data-for="RTCDataChannelInit">id</a></code>.</div>
+                  <p>Let <var>options</var> be the second argument, if present,
+                  otherwise leave it unset.</p>
                 </li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -8284,74 +8284,107 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-                  <code><a data-for="RTCDataChannelInit">id</a></code>.</div>
+                  <p>Let <var>options</var> be the second argument, if present,
+                  otherwise leave it unset.</p>
+                </li>
+                <li>
+                  <p>Let <var>channel</var> have an [[<dfn id=
+                  "maxPacketLifeTime-slot">MaxPacketLifeTime</dfn>]] internal
+                  slot initialized to <var>option</var>'s
+                  <code>maxPacketLifeTime</code> member, given that
+                  <var>options</var> is set and the member is present, otherwise
```
taylor-b: Should be able to remove "given that `options` is set" from these descriptions.

webrtc.html
```diff
@@ -8284,74 +8284,107 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                <li>
+                  <p>Let <var>channel</var> have an [[<dfn>DataChannelId</dfn>]]
+                  internal slot initialized to <var>option</var>'s
+                  <code>id</code> member, given that <var>options</var> is set
+                  and the member is present, otherwise <code>null</code>.</p>
+                </li>
+                <li>
+                  <p>If <var>options</var> is still unset, let it be a new
+                  <code><a>RTCDataChannelInit</a></code> dictionary instance
+                  with the specified default values.</p>
```
taylor-b: Again, WebIDL bindings already do this.

webrtc.html
```diff
@@ -8284,74 +8284,107 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   <code><a>RTCDataChannel</a></code> object.</p>
                 </li>
                 <li>
-                  <p>Initialize <var>channel</var>'s <code><a data-for=
-                  "RTCDataChannel">label</a></code> attribute to the value of
-                  the first argument.</p>
+                  <p>Let <var>channel</var> have an [[<dfn id=
```
adam-be: Will fix typo. I agree that we should try to be consistent on the naming and the "link names", if we need links to all slots.

webrtc.html
```diff
@@ -8284,74 +8284,107 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-                  to the values of the corresponding members of the
-                  <var>dataChannelDict</var> argument, using a value of
-                  <code>null</code> if the corresponding dictionary member is
-                  missing.</p>
-                  <div class="note">The members that do not have default
-                  values, and thus may be missing, are <code><a data-for=
-                  "RTCDataChannelInit">maxPacketLifeTime</a></code>, <code><a
-                  data-for="RTCDataChannelInit">maxRetransmits</a></code>, and
-                  <code><a data-for="RTCDataChannelInit">id</a></code>.</div>
+                  <p>Let <var>options</var> be the second argument, if present,
```
adam-be: Thanks for the pointer. The default value makes things a bit easier.

adam-be:
Thanks for reviewing @taylor-b. The next version of the patch should be simpler when we can assume a default value for the options dictionary.

stefhak:
LGTM.

Issue #1261, opened 2017-05-29
DTMF playout algorithm does not specify comma handling


fluffy:
Need to add a step for for the 2 second delay when playing a comma and make it clear that the event is fired for the comma just like any other tone - think of the , as the silence tone.

PR #1373, opened 2017-06-11
DTMF playout algorithm for comma

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1261

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9417,10 +9417,10 @@ interface RTCDTMFSender : EventTarget {
                     <li>Remove the first character from <code><a data-for=
                     "RTCDTMFSender">toneBuffer</a></code> and let that
                     character be <var>tone</var>.</li>
-                    <li>If <var>tone</var> is "," send silence for
+                    <li>If <var>tone</var> is "," delay sending tones for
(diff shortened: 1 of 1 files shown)


fluffy:
This looks good. Do we need anything to make it clear that the event is fired with the "," in it, or is that all clear.

aboba:
Added note about "," in the tonechange event.

stefhak:
LGTM.

webrtc.html
```diff
@@ -9419,10 +9417,17 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                     <li>Remove the first character from <code><a data-for=
                     "RTCDTMFSender">toneBuffer</a></code> and let that
                     character be <var>tone</var>.</li>
-                    <li>Start playout of <var>tone</var> for <code><a data-for=
+                    <li>If <var>tone</var> is "," send silence for
```
taylor-b: Should say "delay sending tones" instead of "send silence", since if you're sending audio you don't want to actually mute audio for 2 seconds. This should just insert a 2-second gap in between tones.

Issue #1181, opened 2017-05-08
End removed tracks remotely again; Make receiver.track nullable instead


jan-ivar:
Instead of https://github.com/w3c/webrtc-pc/issues/1161, let's fire `ended` on remote tracks again like browsers do today, to avoid [breakage](https://github.com/w3c/webrtc-pc/pull/1168#issuecomment-299589165).

Basically, when a `transceiver`'s direction loses the "send" bit as a result of setting a remote description, fire an `ended` event on `transceiver.receiver.track` and set `transceiver.receiver.track` to `null`.

This necessitates making `receiver.track` nullable.

Later, should a transceiver's direction regain its "send" bit as a result of setting a remote description, set `transceiver.receiver.track` to a new track of the same kind, and call [process the remote track](http://w3c.github.io/webrtc-pc/#process-remote-track) (which fires a new `track` event).

This would be backwards compatible.

taylor-b:
I was just talking about this idea with @pthatcherg today. I think it would work, though the receiver should also be recreated when a "sendrecv" _local_ description is set, so that media can be received before the answer.

stefhak:
I think it would be good to work through the events (in terms of `pc.removeTrack`, `transceiver.setDirection` and so on, assuming `onnegotiationneeded` is wired up) that would lead to "send" and "recv" bits being set/unset.

But overall this looks like something that could work. If the intent is to have the received track being constantly played or recorded, the sender can play with `replaceTrack` and use a null track for periods when there is nothing to send.

aboba:
@jan-ivar Section 5.1 says: "When the other peer stops sending a track in this manner, an ended event is fired at the MediaStreamTrack object." So according to the existing draft, this is already happening ;)

jan-ivar:
@aboba That's where this started, in https://github.com/w3c/webrtc-pc/pull/1168.

Sadly, the more contradictory language in a spec, the more doubt it casts on the consensus reached.

On the `muted` event, that seems to be true. Unfortunately, `mute` may fire at other times as well, from overconstrained conditions in mediacapture, or *"muted by the User Agent."* e.g. a browser implementing a privacy shade or similar feature, so it'd be hard to deduce much from it alone.

jan-ivar:
[...]

PR #1377, opened 2017-06-13
Add/remove remote tracks from msid streams based on direction.

jan-ivar:
Compromise fix for https://github.com/w3c/webrtc-pc/issues/1181 on idea from @pthatcherg. Note this PR relies on language in https://github.com/w3c/webrtc-pc/pull/1321.

@adam-be, @taylor-b PTAL.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5546,7 +5546,9 @@ interface RTCPeerConnectionIceErrorEvent : Event {
       <section>
         <h4>Processing Remote MediaStreamTracks</h4>
         <p>An application can reject incoming media descriptions by calling
-        <code><a>RTCRtpTransceiver</a>.stop()</code>.</p>
+        <code><a>RTCRtpTransceiver</a>.stop()</code> to stop both directions,
+        or set the transceiver's direction to "sendonly" to reject only the
+        incoming side.</p>
         <p>To <dfn id="process-remote-track-addition">
         process the addition of a remote track</dfn> for
         an incoming media description <span data-jsep=
--- a/webrtc.html
+++ b/webrtc.html
@@ -1594,9 +1594,19 @@
                       <li>
                         <p>If the direction of the media description is
                         <code>sendrecv</code> or <code>sendonly</code>, and
-                        <code>transceiver.receiver.track</code> has not yet
-                        been fired in a <code title="event-track"><a>track
-                        </a></code> event, <a>process the remote track</a> for
+                        <var>transceiver</var>'s
+                        [[<a href="#direction-slot">Direction</a>]] slot is
+                        neither "sendrecv" nor "sendonly",
+                        <a>process the addition of a remote track</a> for
+                        the media description, given <var>transceiver</var>.</p>
+                      </li>
+                      <li>
+                        <p>If the direction of the media description is
+                        <code>recvonly</code> or <code>inactive</code>, and
+                        <var>transceiver</var>'s
+                        [[<a href="#direction-slot">Direction</a>]] slot is
+                        neither "recvonly" nor "inactive",
+                        <a>process the removal of a remote track</a> for
                         the media description, given <var>transceiver</var>.</p>
                       </li>
                       <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1290,9 +1291,19 @@
                       <li>
                         <p>If the direction of the media description is
                         <code>sendrecv</code> or <code>sendonly</code>, and
-                        <code>transceiver.receiver.track</code> has not yet
-                        been fired in a <code title="event-track"><a>track
-                        </a></code> event, <a>process the remote track</a> for
+                        <var>transceiver</var>'s
+                        [[<a href="#direction-slot">Direction</a>]] slot is
```
taylor-b: This condition seems backwards. Instead of `[[Direction]] != "sendrecv" && [[Direction]] != "sendonly"`, shouldn't it be `[[CurrentDirection]] != "sendrecv" && [[CurrentDirection]] != "recvonly"`? In other words, "last time we negotiated, I didn't offer to receive a track from you"?

webrtc.html
```diff
@@ -1290,9 +1291,19 @@
...
+                        neither "sendrecv" nor "sendonly",
+                        <a>process the addition of a remote track</a> for
+                        the media description, given <var>transceiver</var>.</p>
```
taylor-b: Can link to the definition of "media description".

webrtc.html
```diff
@@ -1272,7 +1272,8 @@
                           </li>
                           <li>
                             <p><a>Create an RTCRtpTransceiver</a> with
-                            <var>sender</var> and <var>receiver</var>, and
+                            <var>sender</var>, <var>receiver</var> and
+                            "inactive" as direction, and
```
taylor-b: Should be with "recvonly" direction; we want to offer to receive things by default. See JSEP 5.9:

>          +  If no RtpTransceiver was found in the previous step, create
>             one with a recvonly direction.

webrtc.html
```diff
@@ -5059,9 +5070,6 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
               as <code>recvonly</code> or <code>inactive</code>,
               as defined in <span data-jsep="subsequent-offers">
               [[!JSEP]]</span>. <!-- onended --></p>
-              <p>When the other peer stops sending a track in this manner, an
-              <code title="event-MediaStreamTrack-ended">ended</code> event is
-              fired at the <code><a>MediaStreamTrack</a></code> object.</p>
```
taylor-b: Should this be replaced with "the track is removed from any remote streams it was added to"?

webrtc.html
jan-ivar: Ah yes, I initialized to "inactive" to make the subsequent steps below work where it diffs against this direction to take action. But as you point out below, I should be using `currentDirection` below instead. Will fix.

webrtc.html
jan-ivar: You'r right, and I can check [[CurrentDirection]] for `null` for the initial condition.

webrtc.html
jan-ivar: Will do.

webrtc.html
jan-ivar: I removed it as it seemed redundant, and I struggled with making it precise.

I guess it could be: *"When the other peer stops sending a track in this manner, the track is removed from any remote streams that were initially sent in the `track` event, and a *muted* event is fired at the track."* ?

webrtc.html
```diff
@@ -5390,29 +5406,29 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       </div>
       <section>
         <h4>Processing Remote MediaStreamTracks</h4>
-        <p>An application can reject incoming <code><a>MediaStreamTrack</a></code>
-        objects by calling <code><a>RTCRtpTransceiver</a>.stop()</code>.</p>
-        <p>To <dfn id="process-remote-track">process the remote track</dfn> for
+        <p>An application can reject incoming media descriptions by calling
```
stefhak: Isn't it tracks you can reject rather than the whole media description (SDP)? Or am I getting this wrong?

taylor-b: If you don't want to reject the whole media description, you can just set the transceiver's direction to "sendonly". That should probably be called out here.

jan-ivar:
Gah! That's what I get for pushing with --force-with-lease

jan-ivar:
Closed in favor of https://github.com/w3c/webrtc-pc/issues/1402.

jan-ivar:
(odd, I opened a new PR because github was displaying this issue as totally borked with way too much history and "diff too large to show", but github appears to have recovered, now displaying this PR correctly again. Oh well)

Issue #1112, opened 2017-04-06
Terminology around "setting" attributes may be incorrect


taylor-b:
In various places, we talk about "setting" attributes. This comes in three flavors:

1. Setting an internal slot, which is read by the attribute's getter. For example:

> Store the configuration in the [[ configuration]] internal slot.

2. Setting some abstract thing, which can be assumed to be an internal slot. For example:

> An RTCPeerConnection object has a **signaling state**, a **connection state**, an **ICE gathering state**, and an **ICE connection state**. These are initialized when the object is created.
> ...
> To update the **ICE gathering state** of an RTCPeerConnection instance _connection_, the user agent must queue a task that runs the following steps:
> ...
> 4. Set _connection_'s **ice gathering state** to newState.

This seems acceptable. Though, for consistency and to be more formal, I'd suggest just describing these things as internal slots as well.

3. Setting the attribute itself. For example:

> Set connection's pendingLocalDescription , currentLocalDescription , pendingRemoteDescription and currentRemoteDescription to null.

In WebIDL terminology, this seems as if it would be formally interpreted as "call the attribute's setter". But it's a readonly attribute, so it doesn't have a setter. Is this ok (since it's obvious enough what the intention was), or should we change these places to use an internal slot as well?

For example, say in the description of `pendingLocalDescription`:

> Returns the value of the `[[pendingLocalDescription]]` internal slot.

And change the above quote to:

> Set connection's `[[pendingLocalDescription]]` , `[[currentLocalDescription]]`, `[[pendingRemoteDescription]]` and `[[currentRemoteDescription]]` internal slots to null.

jan-ivar:
Number 3 is definitely wrong. Some care seems to have gone into number 2, but consistency wins for me as well. Internal slots for everything.

adam-be:
I think we have started moving towards internal slots, but there are still a lot of old spec text using "number 3" above.

taylor-b:
@jan-ivar A question that's coming up as we're trying to fix this: what naming convention should we use for internal slots that map 1:1 to attributes? If we use the same name, we'll get duplicate definition respec errors.

jan-ivar:
From looking at other specs, internal slots seem to use [[FooBar]] vs. attribute fooBar.

taylor-b:
What's an example? I ran into problems even with that, since definitions aren't case sensitive.

jan-ivar:
[...]

Issue #1360, opened 2017-06-08
Editorial: consistent Internal slot names e.g. [[IsClosed]] vs [[isClosed]]?


jan-ivar:
I can't find a TAG link for this, just cargo-culting from the html and webidl specs. Seems less likely to name clash with API names as well.

I'm filing this issue rather than asking for internal consistency in https://github.com/w3c/webrtc-pc/pull/1321 review.

jan-ivar:
[Apparently](https://github.com/w3c/webrtc-pc/issues/1112#issuecomment-308846383) slot names are fairly ad hoc atm. but we should at least make them consistent within the same spec.

aboba:
Note:  [Dom suggests Capitalized Case. ](https://github.com/w3c/webrtc-pc/pull/1395)

PR #1395, opened 2017-06-15
Standardize how internal slots are named and linked to.

taylor-b:
Related to #1112.

This replaces any instance of `[[<dfn>fooBar</dfn>]]` with
`<dfn>[[\FooBar]]</dfn>`, which means we don't have to worry about
definitions conflicting, and can have a slot whose name matches
an attribute without `<dfn>`s conflicting.

Also changed all internal slot names to be UpperCamelCase, for consistency and because this matches EcmaScript.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/internal_slots.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/4e42d60...taylor-b:58909ca.html)


burnburn:
@dontcallmedom can you give an opinion on casing, a la issue #1360 ?

dontcallmedom:
since [EcmaScript](https://tc39.github.io/ecma262/) uses CapitalizedCase for internal slots, and since that's where that notion is grounded, I would suggest we adopt this.

adam-be:
Nice that you found a solution to the attribute and slot name collision issue @taylor-b. Should we go for initial upper case before merging this?

One other thing I was thinking about is when slots have very short and generic names; like [[\Id]] for example. It might make sense to add some more information to that name to avoid future name collisions and give a hint about its origin. Like [[\DataChannelId]] for example.

webrtc.html
```diff
@@ -8419,98 +8403,90 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
                 <li>
-                  <p>Let <var>channel</var> have an [[<dfn id=
-                  "maxRetransmits-slot">MaxRetransmits</dfn>]] internal slot
+                  <p>Let <var>channel</var> have a <dfn>[[\maxRetransmits]]</dfn> internal slot
                   initialized to <var>option</var>'s <code>maxRetransmits</code>
                   member, if present, otherwise <code>null</code>.</p>
                 </li>
                 <li>
-                  <p>Let <var>channel</var> have an [[<dfn>DataChannelId</dfn>]]
+                  <p>Let <var>channel</var> have an <dfn>[[\id]]</dfn>
```
adam-be: Slot name too generic?

taylor-b:
> Should we go for initial upper case before merging this?

I don't see why not.

> It might make sense to add some more information to that name to avoid future name collisions and give a hint about its origin. Like [[\DataChannelId]] for example.

Sounds good to me. Are there any others that are too ambiguous (like "protocol")?

adam-be:
Maybe "handle" and "label".

taylor-b:
Made some names more descriptive and switched to UpperCamelCase naming.

Issue #1277, opened 2017-05-31
Section 4.3.1: operations queue and setDescription


aboba:
From EKR: https://lists.w3.org/Archives/Public/public-webrtc/2017May/0166.html

(operations queue)
This algorith seems to imply that operation is run synchronously
if the queue depth is one (step 6) whereas if it's not, it's run
asynchronously. This definitely isn't invisible in the callback
version (reentrancy) and isn't actually invisible in the promises
version because these operations have side effects. Is it really
the expectation that we should (for instance) start sending
media before the promise resolves? That seems odd.

(setDescription)
The says:
    To set an RTCSessionDescription description on an RTCPeerConnection
    object connection, enqueue the following steps:

    1. Let p be a new promise.

I may be confused, but I don't think that the promise creation and
resolution is in fact queued.

    2. If elements of the SDP were modified, then reject p with a newly created InvalidModificationError and abort these steps.

This only applies to local decriptions.

This algorithm is extremely prescriptive about which error checks
you need to perform and in which order. This is externally observable
because I throw different errors. For instance, why is it the
spec's business whether I check for being in the right state before
or after I check for the SDP being modified.

    5. If the content of description is invalid, then reject p with a
    newly created InvalidAccessError and abort these steps.

What would invalid be here? Anything else?

    2. If description is applied successfully, the user agent must
    queue a task that ru ns the following steps:

Didn't we already queue above?

soareschen:
Note: the first part of this is related to #1218.

stefhak:
Re-assigning to @jan-ivar

jan-ivar:
I think all issues reported here have been addressed by now, so we can close this issue.

> (operations queue)
This algorithm seems to imply that operation is run synchronously
if the queue depth is one (step 6) whereas if it's not, it's run
asynchronously.

This was discussed in https://github.com/w3c/webrtc-pc/pull/1397 and closed. Only an operation's chained preamble ever runs synchronously—matching how `async` functions work—so reentrancy shouldn't be an issue.

> `1. Let p be a new promise.`

The latest version of the [chain an operation](https://w3c.github.io/webrtc-pc/#dfn-chain-an-operation) algorithm returns a promise, which the outer steps return appropriately, so this should be fixed.

> `2. If elements of the SDP were modified, then reject p with InvalidModificationError`
> 
> This only applies to local decriptions.

[...]

PR #1397, opened 2017-06-16
Queue enqueued operation as task if length of operations is 1

### LINKED ISSUE #1218: Race condition in enqueue an operation
In step 6 of enqueue an operation:

> 6. If the length of [[operations]] is exactly 1, execute operation.

This result in race condition with synchronous code following call to RTCPeerConnection methods. That is, if length of the operation queue is 1, the operation is executed _before_ the next line of synchronous code; But if the length of operation queue is > 1, the operation is executed _after_ the current microtask return to the event loop. This can result in code having unexpected behavior depending on whether the internal operation queue is empty or not when the operation is enqueued.

For example, if the next line of code is `peerconnection.close()`, then the operation may or may not be executed depending on the operation queue length.

To fix this, the operation should be enqueued to the next tick if the length of operation queue is 1.

### LINKED ISSUE #1347: Enqueue an operation vs queue a task
Section 3 imports the concept of queue a task from HTML5.

Section 4.3.1.2 defines _"Enqueue an operation"_ which has an operation queue of running one task at a time.

I tried searching through the spec and there are two phrases being used:

1. queue a task that runs the following steps
2. enqueue the following steps

I'd guess that (1) refers to the HTML5 task and (2) refers to the operation queue. If so, the only section that make use of enqueue operation is 4.3.1.6 _"Set the RTCSessionSessionDescription"_.

If that is correct, I think the sections should be better organized to document the use. In particular, place the _"Enqueue an operation"_ side by side with _"Set the RTCSessionSessionDescription"_, and add a note to clarify that the remaining sections refer to the HTML5 task.

Because _"Enqueue an operation"_ is introduced so early and prominently in the spec, I have been wrongly assuming that all tasks are enqueued as operation instead of HTML5 task. :(

soareschen:
Fixes #1218, ~#1347~, and first part of #1277. (Update: remove patch for #1347 and move to #1421)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1158,6 +1158,11 @@
             <p>Return <var>p</var>.</p>
           </li>
         </ol>
+        <div class="note">
+          Enqueing an operation is different from queuing a task, which is
+          defined in [[HTML5]]. The operation queue is connection-specific,
+          and runs sequentially in first in, first out (FIFO) manner.
+        </div>
         </section>
 
         <section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1114,8 +1114,8 @@
             <p>Append <var>operation</var> to [[<a>operations</a>]].</p>
           </li>
           <li>
-            <p>If the length of [[<a>operations</a>]] is exactly 1, execute
-            <var>operation</var>.</p>
+            <p>If the length of [[<a>operations</a>]] is exactly 1, queue a task
+            to execute <var>operation</var>.</p>
           </li>
           <li>
             <p>Upon fulfillment or rejection of the promise returned by the


burnburn:
@jan-ivar can you please review this since you wrote most of the original text?

taylor-b:
Also, @soareschen, it would be helpful to split this up into different PRs for the different issues being fixed.

webrtc.html
```diff
@@ -1158,6 +1159,11 @@
             <p>Return <var>p</var>.</p>
           </li>
         </ol>
+        <div class="note">
+          Enqueing an operation is different from queuing a task, which is
+          defined in [[HTML5]]. The operation queue is connection-specific,
+          and runs sequentially in first in, first out (FIFO) manner.
```
jan-ivar: Would it be useful to append *...whenever more than one operation is enqueued*?

webrtc.html
```diff
@@ -1114,8 +1115,8 @@
             <p>Append <var>operation</var> to [[<a>operations</a>]].</p>
           </li>
           <li>
-            <p>If the length of [[<a>operations</a>]] is exactly 1, execute
-            <var>operation</var>.</p>
+            <p>If the length of [[<a>operations</a>]] is exactly 1, queue a task
```
jan-ivar: It's not OK to queue here, because each operation has a synchronous part that needs to run synchronously in the typical case, to ensure errors are thrown correctly.

The reason this matters is because we follow a pattern of doing all input validation synchronously, such that a JavaScript developer can observe the that the returned promise is in rejected state on return when observed in a JavaScript debugger.

soareschen:
> Also, @soareschen, it would be helpful to split this up into different PRs for the different issues being fixed.

Sorry, was thinking that the fixes are short and trivial but turns out they are not. I have made another PR #1421 and moved the clarification patches to there.

soareschen:
Closing this as per discussion in #1218.

PR #1398, opened 2017-06-16
Remove condition "If channel's underlying data transport is not established yet, then abort these steps" from send()

### LINKED ISSUE #1396: In data channel send(), remove the condition "If channel's underlying data transport is not established yet, then abort these steps"
Missed this when reviewing #1209. The data can still be queued while channel's underlying data transport is being established. Ignoring it would cause initial data being sent to be silently dropped.

It is also possible that the condition is unreachable, since the new step 2 already guard against non-open `readyState`.

soareschen:
Fixes #1396.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9270,10 +9270,10 @@ interface RTCTrackEvent : Event {
             </li>
           </ul>
         </li>
+
         <li>
-          <p>If <var>channel</var>'s <a>underlying data transport</a> is not
-          established yet, or if the <code><a>closing procedure</a></code> has
-          started, then abort these steps.</p>
+          <p>If the <code><a>closing procedure</a></code> for
+          <var>channel</var> has started, then abort these steps.</p>
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


aboba:
@soareschen Since Step 2 throws when readyState != "open", why does step 4 still need to check whether the closing procedure has been started?

nils-ohlmeier:
I think @aboba is right here. Since the first thing close() does is to change the readyState I think it is safe to remove the whole step 4 as by PR #1404 .

soareschen:
My thinking was that the closing procedure does not necessarily get invoked by closed, e.g. when closed by remote peer. In that case the changing from `open` to `closing` could is queued as a task and could result in race condition.

But you are right that the whole step can be removed entirely. Because even if we guard against that, the remote peer can still close while data being sent and result in data loss. It is simpler to just queue it to the buffer regardless and let the closing procedure handle discarding the queue.

Closing this in favor of #1404.

Issue #1217, opened 2017-05-17
Missing References to JSEP spec


soareschen:
There have been a few times when I failed to find any normative definition for some behavior in webrtc-pc, only to found it later in JSEP. Some examples:

- No normative definition of valid state transitions for `RTCSignalingState`.
- `setLocalDescription` triggers ICE candidate gathering process.

If it is alright I will submit PRs to add reference to JSEP whenever I find one missing.

burnburn:
Yes, @soareschen that would be great.  Thanks!

soareschen:
The first part has been fixed in master branch. The second part has been evolved to #1422. Closing this issue.

PR #1399, opened 2017-06-16
Add reference to JSEP that setLocalDescription triggers ICE gathering

soareschen:
Fixes second part of #1217.

***
[Preview](https://s3.amazonaws.com/pr-preview/soareschen/webrtc-pc/setlocaldescription-icegathering.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/0b0db47...soareschen:5a6bd95.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2363,9 +2363,6 @@ interface RTCPeerConnection : EventTarget  {
                 which point the <code><a>RTCPeerConnection</a></code> can fully
                 adopt the pending local description, or rollback to the current
                 description if the remote side rejected the change.</p>
-                <p>As noted in <span data-jsep="ice-candidate-pool">[[!JSEP]]</span>,
-                calling this method triggers the ICE candidate gathering process
-                by the <a>ICE Agent</a>.</p>
                 <p>As noted in <span data-jsep="modifying-sdp">[[!JSEP]]</span>
                 the SDP returned from <code>createOffer</code> or
                 <code>createAnswer</code> MUST NOT be changed
@@ -2389,9 +2386,14 @@ interface RTCPeerConnection : EventTarget  {
                   is <code>offer</code>, set <code><var>description</var>.sdp</code>
                   to <var>lastOffer</var>.</li>
                   <li>Return the result of <a href="#set-description">setting
-                  the RTCSessionDescription</a> indicated by 
+                  the RTCSessionDescription</a> indicated by
                   <code><var>description</var></code>.</li>
                 </ol>
+                <div class="note">
+                  <p>As noted in <span data-jsep="applying-a-local-desc">[[!JSEP]]</span>,
+                  calling this method may trigger the ICE candidate gathering process
+                  by the <a>ICE Agent</a>.</p>
+                </div>
                 <table class="parameters">
                   <tbody>
                     <tr>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2363,11 +2363,14 @@
                 which point the <code><a>RTCPeerConnection</a></code> can fully
                 adopt the pending local description, or rollback to the current
                 description if the remote side rejected the change.</p>
+                <p>As noted in <span data-jsep="ice-candidate-pool">[[!JSEP]]</span>,
```
taylor-b: This should reference section 5.8, "applying a local description", which defines this behavior most precisely.

webrtc.html
```diff
@@ -2363,11 +2363,14 @@
...
+                calling this method triggers the ICE candidate gathering process
```
taylor-b: "triggers" => "may trigger"

webrtc.html
```diff
@@ -2363,11 +2363,14 @@
...
+                by the <a>ICE Agent</a>.</p>
                 <p>As noted in <span data-jsep="modifying-sdp">[[!JSEP]]</span>
                 the SDP returned from <code>createOffer</code> or
                 <code>createAnswer</code> MUST NOT be changed
-                before passing it to <code>setLocalDescription</code>. As
```
taylor-b: I don't see why the "as a result" is being removed; it adds clarity, since it segues to how the "MUST" is enforced.

webrtc.html
soareschen: Unfortunately section 5.8 doesn't mention about triggering ICE gathering. The only place it is hinted that ICE gathering is triggered by setLocalDescription in section 3.5.4 for ICE candidate pool:

> When setLocalDescription is eventually called, and the JSEP implementation goes to gather the needed ICE candidates, it SHOULD start by checking if any candidates are available in the pool.

webrtc.html
soareschen: There are now two notes referencing JSEP, and I wanted to split it into multiple paragraphs to improve readability. I see the steps are defined the same way as other methods, so it's not like there are steps in `setLocalDescription` because of the JSEP requirement.

Anyway I have added back the "as a result", since it doesn't matter much either way.

webrtc.html
soareschen: Should this be a MUST or SHOULD requirement?

soareschen:
It seems that this PR may not be sufficient to clarify when ICE gathering should be triggered. I have filed #1422 to seek clarification on how this can be fixed.

webrtc.html
taylor-b: From section 5.8:

>    o  If this m= section is new, begin gathering candidates for it, as
      defined in [RFC5245], Section 4.1.1, unless it has been marked as
      bundle-only.

>   o  Or, if the ICE ufrag and password values have changed, and it has
      not been marked as bundle-only, trigger the ICE agent to start an
      ICE restart, and begin gathering new candidates for the m= section
      as described in [RFC5245], Section 9.1.1.1.  If this description
      is an answer, also start checks on that media section as defined
      in [RFC5245], Section 9.3.1.1.

webrtc.html
taylor-b: The normative requirement is in JSEP, so if anything this should be in a `<div class="note">`. We've been pretty lax about mixing normative and non-normative text...

webrtc.html
soareschen: Thanks for pointing! I was searching for the wrong keyword when reading the spec.

soareschen:
Thanks for the feedback. I have made the suggested changes and move it to a new note section at the bottom of `setLocalDescription`.

PR #1400, opened 2017-06-16
Add RTCDataChannelPriority enum and step to initialize data channel priority

### LINKED ISSUE #1368: Missing initialization of priority attribute when firing datachannel event
In step 4 of section 6.2. _When an underlying data transport is to be announced_, it missed initializing `priority` from the configuration.

> Initialize channel's label, ordered, maxPacketLifeTime, maxRetransmits, protocol, negotiated and id attributes to their corresponding values in configuration.

It is also not clear what the value of `priority` should be, if the byte value in configuration do not match one of the corresponding enum value of `RTCPriorityType`. The valid values specified is only a SHOULD requirement.

soareschen:
Fixes #1368.

The priority values defined in rtcweb-data-channel is different from the priority values in rtcweb-transports. As a result the enum `RTCPriorityType` does not match the usage of priority in data channels.

I have defined a new `RTCDataChannelPriority` enum based on rtcweb-data-channel and update the attribute definition. An additional step is added to the _"underlying data transport is to be announced"_ section to initialize the `priority` attribute in the data channel.

This should be safe to change as current browsers have not implemented the priority attribute for data channels.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8713,6 +8713,15 @@ interface RTCTrackEvent : Event {
           values in <var>configuration</var>.</p>
         </li>
         <li>
+          <p>Initialize <var>channel</var>'s
+          <a data-for="RTCDataChannel"><code>priority</code></a> attribute to
+          the <a>RTCDataChannelPriority</a> enum value that corresponds to the
+          integer priority value in <var>configuration</var>, as defined in
+          section 6.4 of [[!RTCWEB-DATA]]. If the integer priority value does
+          not have a matching enum value, match it to the nearest corresponding
+          enum value.</p>
+        </li>
+        <li>
           <p>Set <var>channel</var>'s <code><a data-for=
           "RTCDataChannel">readyState</a></code> attribute to
           <code>connecting</code>.</p>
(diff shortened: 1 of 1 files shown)


aboba:
@pthatcherg The priority values defined in rtcweb-data-channel are indeed different from the priority values in rtcweb-transports.  Was this intentional?

webrtc.html
```diff
@@ -8713,6 +8713,15 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           values in <var>configuration</var>.</p>
         </li>
         <li>
+          <p>Initialize <var>channel</var>'s
+          <a data-for="RTCDataChannel"><code>priority</code></a> attribute to
+          the <a>RTCDataChannelPriority</a> enum value that corresponds to the
+          integer priority value in <var>configuration</var>, as defined in
```
taylor-b: The value in *configuration* is an enum though, not an integer. It's an enum that gets turned into an integer, for the "underlying data transport", not the other way around.

webrtc.html
```diff
@@ -9303,7 +9348,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
         "RTCDataChannelState" class="simple">
           <tbody>
             <tr>
-              <th colspan="2"><dfn>RTCDataChannelState</dfn> Enumeration description</th>
```
taylor-b: Why remove this dfn?

soareschen: I put the IDL definitions `RTCDataChannelPriority` and `RTCDataChannelState` in new sub-sections. So the new dfn is defined in the new section header few lines above.

webrtc.html
soareschen: I'm not sure of that. _configuration_ is not exactly a JavaScript object and there is no type definition of what fields and types it has. Step 3 only says:

> Let configuration be an information bundle received from the other peer as a part of the process to establish the underlying data transport described by the WebRTC DataChannel Protocol specification [RTCWEB-DATA-PROTOCOL].

If _configuration.priority_ already have the `RTCDataChannelPriority` type, we'd be able to just set the value directly, and there wouldn't need to have so much explanation on how to construct `channel.priority` from _configuration_.

taylor-b: Oh, my mistake. Hmm. This is a problem then, because the implementation is allowed to use values other than 128/256/512/1024. What should the attribute return if the data channel was opened with a priority of 200? Maybe the `priority` attribute should be an integer, while the `RTCDataChannelInit` member remains an enum. Seems like a good thing to bring up in the upcoming virtual interim.

taylor-b:
Filed https://github.com/w3c/webrtc-pc/issues/1423 for the issue I pointed out in the comment.

taylor-b:
I talked to Peter, and realized that the same enum is intended to affect both the DSCP markings used for the SCTP transport, and the priority used by the SCTP scheduler (see: https://datatracker.ietf.org/doc/draft-ietf-tsvwg-rtcweb-qos/?include_text=1). Which seems odd to me, since there's also the requirement that all packets over the same SCTP transport use the same DSCP marking. It would have made more sense to me to have one priority enum for the transport level (priority of data relative to audio/video), and one for the SCTP scheduling level (priority of data channels relative to each other).

But unless the IETF side changes, we should have a single enum, and acknowledge somewhere that the names in the two IETF specs don't match up. There's still [this issue](https://github.com/w3c/webrtc-pc/issues/1423) though, about what you do with the priority value in the "open" message, which I added as a virtual interim topic.

aboba:
Given discussion at the virtual interim, we will close this PR.

Issue #1381, opened 2017-06-13
pc.close() method should close sctp transport


soareschen:
Picking out from #1306.

In the `close()` method, it iterates through the `DtlsTransport` for each sender/receiver for each transceiver and close them. However the `SctpTransport` at `sctp` attribute is missed so its underlying `DtlsTransport` is not closed.

aboba:
Related Issue: https://github.com/w3c/webrtc-pc/issues/1288

PR #1401, opened 2017-06-16
close all DtlsTransports in pc.close() (including sctp.transport)

aboba:
Work-in-progress, do not merge.

Fix for Issue https://github.com/w3c/webrtc-pc/issues/1381

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/issue-1381-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/0b0db47...5322655.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2903,36 +2903,11 @@ interface RTCPeerConnection : EventTarget  {
                     "RTCIceTransportState">closed</a></code>.</p>
                   </li>
                   <li>
-                    <p>Set the <var>connection</var>'s
-                    <code>sctp.transport.state</code>
+                    <p>Set the <code><a data-link-for=
+                    "RTCDtlsTransport">state</a></code> of each of
+                    <var>connection</var>'s <code><a>RTCDtlsTransport</a></code>s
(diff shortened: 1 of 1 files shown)


aboba:
It seems odd that RTCSctpTransport does not have a state attribute to set to "closed".

webrtc.html
```diff
@@ -2903,6 +2903,12 @@
                     "RTCIceTransportState">closed</a></code>.</p>
                   </li>
                   <li>
+                    <p>Set the <var>connection</var>'s
+                    <code>sctp.transport.state</code>
+                    to <code><a data-link-for=
+                    "RTCDtlsTransportState">closed</a></code>.</p>
+                  </li>
+                  <li>
```
taylor-b: We could define a "collect set of `RTCDtlsTransport`s" algorithm, as we did for "set of `RTCRtpSender`s". Though it looks like that was removed at some point? Anyway, if we did this, then we could just have "for each transport in the set of transports", and would avoid this algorithm touching the same transport multiple times.

PR #1414, opened 2017-06-21
RTCPeerConnection.close: release transceiver before ICE transport

mparis:
Right now the steps described for [RTCPeerConnection.close](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-close) make to release ICE transport before transceivers.
I think this is wrong and transceivers should be released before in order to ensure that RTCP BYES are properly sent (the transport should be active).
From my point of view the order should be the proposed in this PR.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2890,6 +2890,44 @@ interface RTCPeerConnection : EventTarget  {
                     <code>true</code>, abort these steps.</p>
                   </li>
                   <li>
+                    <p>Let <var>transceivers</var> be the result of executing the
+                    <code><a>CollectTransceivers</a></code> algorithm. For every
+                    <code><a>RTCRtpTransceiver</a></code> <var>transceiver</var> in
+                    <var>transceivers</var>, run the following steps:</p>
+                    <ol>
+                      <li>
+                        <p>If <code><var>transceiver</var>.stopped</code> is
+                          <code>true</code>, abort these steps.</p>
+                      </li>
+                      <li>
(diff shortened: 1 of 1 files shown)


fippo:
Good catch.

I think this should be done slightly different since currently we repeat a lot of the steps from http://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-stop (search for RTCP BYE).
Instead we should
1) run the steps to stop the transceiver
2) close the dtls transport (so a close_notify is sent)
3) close the ice transport

Not from your PR but a few lines below:
> set
>                     <code><var>sender.transport.state</var></code> and
>                     <code><var>sender.transport.transport.state</var></code> to
>                     "closed".

This does not make sense even in the current spec since the ice transport was closed before.

mparis:
Hello @fippo,
I see your point, the problem is that [RTCRtpTransceiver.sop](http://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-stop) implies [updating the negotiation-needed flag](http://w3c.github.io/webrtc-pc/#dfn-update-the-negotiation-needed-flag), which should not be done if PeerConnection is being close, should it?

fippo:
@mparis  you're right... I would have expected that pc.close sets IsClosed way earlier (not as the last step) which would stop any events from being emitted...

aboba:
@fippo Can you review PR https://github.com/w3c/webrtc-pc/pull/1565 which touches this same section?

PR #1426, opened 2017-06-26
getstats: improve selector definiton

### LINKED ISSUE #1374: Section 8.2: Null selector properly handled by getStats?
From fippo:

we might have an issue in the text @vr000m cites:
http://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface-extensions-2
That step (4) does not seem to cover the case that selector is null.

fippo:
fixes #1374

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9835,12 +9835,18 @@ interface RTCDTMFToneChangeEvent : Event {
                 </li>
                 <li>
                   <p>If <var>selectorArg</var> is neither <code>null</code> nor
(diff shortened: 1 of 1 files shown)


aboba:
At line 9838, why was the word "valid" removed?

fippo:
see [this comment](https://github.com/w3c/webrtc-pc/issues/1374#issuecomment-307890217) -- essentially if we only allow valid tracks in that step we can not throw InvalidAccessError later.

Issue #1198, opened 2017-05-10
RTCIceServer "represents a TURN server"


fippo:
http://w3c.github.io/webrtc-pc/#rtciceserver-dictionary
says this about credential:
> If this RTCIceServer object represents a TURN server, then this attribute specifies the credential to use with that TURN server.

I did some digging and this seems to have been touched after moving from url to urls. For a single URL it is pretty clear what "represents a TURN server" means. For multiple urls there is the (theoretical) case of
```[stun:foo, turn:foo, turns:foo]```
mixing different classes of urls should not be allowed imo but I think changing the definition to say that the username and credential will only be used with TURN servers is easier.

fippo:
Related, http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-setconfiguration step 11:
> 3. For each url in server.urls parse url and obtain scheme name. If the scheme name is not implemented by the browser, or if parsing based on the syntax defined in [ RFC7064] and [RFC7065] fails, throw a SyntaxError.

> 4. If scheme name is turn or turns, and either of server.username or server.credential are omitted, then throw an InvalidAccessError.

This seems to implicitly assume a single scheme name. Steps 4, 5 and 6 need to be done for each url in server.urls.

PR #1427, opened 2017-06-26
rtciceserver: fix represents a turn server

fippo:
fixes https://github.com/w3c/webrtc-pc/issues/1198

seemed simpler than what I had in mind initially.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -421,11 +421,11 @@
               <dt><dfn><code>username</code></dfn> of type <span class=
               "idlMemberType"><a>DOMString</a></span></dt>
               <dd>
-                <p>If this <code><a>RTCIceServer</a></code> object represents a
+                <p>If any of the <code><a>RTCIceServer</a>.urls</code> represents a
                 TURN server, and <code>credentialType</code> is
                 <code>"password"</code>, then this attribute specifies the
                 username to use with that TURN server.</p>
-                <p>If this <code><a>RTCIceServer</a></code> object represents a
+                <p>If any of the <code><a>RTCIceServer</a>.urls</code> represents a
                 TURN server, and <code>credentialType</code> is
                 <code>"oauth"</code>, then this attribute specifies the Key ID
                 (<code>kid</code>) of the shared symmetric key, which is shared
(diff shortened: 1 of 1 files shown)


taylor-b:
Previously, it was implied that each `RTCIceServer` represents a single "server" in the abstract sense, where the list of URLs represents a list of ways of connecting to that server.

If we're loosening that definition, and allowing a single `RTCIceServer` to include STUN/TURN URLs, then the question becomes: "What's the difference between putting multiple URLs in one `RTCIceServer` instead of multiple?" It looks like step 11 of `setConfiguration` flattens the structure (or at least, is intended to)? If so, this should be called out in a note. If not, what *is* the difference?

fippo:
My interpretation is that if the application specifies a group of urls like
```[turn:foo?transport=udp, turn:foo?transport=tcp, turns:foo?transport=tcp]```
then the ICE engine may interpret this as "just get me the best candidate from _any_ of these servers (which would at least be the same physical entity). So if you know you have a relay candidate over turn/udp that is all you need and you could stop gathering turn/tcp and turn/tls candidates.
This optimization reduces the number of candidates and therefore checking times etc.

But we haven't written that down anywhere what this "represent" means.

nils-ohlmeier:
To be honest I never really understood the difference between the list of servers and the list of URLs.
It make to some degree sense if the list of servers means "try to get responses from all of these" and the list of URL means "a response from just of these is enough". But as @fippo pointed out I think that expected behavior needs to be better documented.

stefhak:
@pthatcherg could you comment on this one?

taylor-b:
I asked Peter, and he said he doesn't recall there being any intention to treat a list of servers with one URL differently from one server with multiple URLs. In other words, they all get flattened to the same thing. He says "if there's logic to discard a TCP candidate if a UDP candidate could be allocated, that should happen regardless of how the URLs are grouped".

So I'd propose closing this, and just making the `RTCIceServer`-flattening more clear, if it isn't already;
 after @fippo's [PR](https://github.com/w3c/webrtc-pc/pull/1457) to change the indentation of some steps, we may be fine.

stefhak:
@taylor-b do you think we can close this one now?

taylor-b:
#1457 was merged, so that's fine by me.

Issue #1242, opened 2017-05-24
rtcsessiondescription: attributes are not mutable


fippo:
http://w3c.github.io/webrtc-pc/#rtcsessiondescription-class
says
> Attributes on this interface are mutable for legacy reasons.

The IDL specifies them as readonly. I'm a bit confused by the back and forth but my reading of #573 is that they are not mutable in the spec anymore even if they still are in the wild. If that is the case this sentence should be removed.

aboba:
@fippo Can you submit a PR?

PR #1428, opened 2017-06-26
rtcsessiondescription: attributes are not mutable

fippo:
removes text which claims the attributes were mutable for legacy reasons.

Fixes https://github.com/w3c/webrtc-pc/issues/1242


fippo:
@jan-ivar if you could double-check too that would be great

Issue #1238, opened 2017-05-24
which ice-tcp types will a browser emit?


fippo:
http://w3c.github.io/webrtc-pc/#rtcicetcpcandidatetype-enum

describes the different types of ICE-TCP candidates that exist. IIRC nobody really wants to have browsers talking ice-tcp to other browsers (for example Chrome sets the port on the tcp candidates it emits to 9), this might be a good place to set the expectations about what browsers are required to implement

taylor-b:
I agree that the expectations/recommendations for what candidates a browser gathers should be described somewhere.

> for example Chrome sets the port on the tcp candidates it emits to 9

This just means it's an "active" TCP candidate, and will only be used to make an outgoing connection if a "passive" remote candidate is added. Which *is* something that should be supported, so that the browser can make a TCP connection to an Internet-accessible server when that's the only option available.

aboba:
@fippo Can you submit a PR?

PR #1429, opened 2017-06-26
ice-tcp: add note about ice-tcp types a UA will gather

fippo:
fixes https://github.com/w3c/webrtc-pc/issues/1238

I didn't see any good language for "but expect to receive this from non-browsers" unfortunately :-/

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4218,6 +4218,8 @@ interface RTCIceCandidate {
               </tbody>
             </table>
           </div>
+          <p>Note: the user agent will typically only gather <code>active</code> 
+          ICE TCP candidates.</p>
         </section>
         <section>
           <h4>RTCIceCandidateType Enum</h4>
--- a/webrtc.html
+++ b/webrtc.html
@@ -4218,7 +4218,7 @@ interface RTCIceCandidate {
               </tbody>
             </table>
           </div>
-          <p>Note: the user agent will typically only gather <code>active</code> 
+          <p class="note">The user agent will typically only gather <code>active</code> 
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4218,6 +4218,8 @@
               </tbody>
             </table>
           </div>
+          <p>Note: the user agent will typically only gather <code>active</code> 
+          ICE TCP candidates.</p>
```
fluffy: FWIW ... I don't really know what that means. Could we point at some other docs to help explain

taylor-b: The definition and RFC link are right above.

webrtc.html
```diff
@@ -4218,6 +4218,8 @@
               </tbody>
             </table>
           </div>
+          <p>Note: the user agent will typically only gather <code>active</code>
```
taylor-b: Should use `<div class="note">`

webrtc.html
taylor-b: May want to include some reasoning, like: "because `passive` candidates typically only work if both peers are on the same LAN, and `so` candidates may not be supported by the operating system." I assume that's the reason. Or have we just not got around to implementing it? Maybe @juberti or @pthatcherg know.

taylor-b: The definition and RFC link are right above.

webrtc.html
fippo: <p class="note"> worked too.

webrtc.html
fippo: I vaguely that nobody really wants browsers to connect via tcp.

Issue #1383, opened 2017-06-13
offerToReceive* legacy behaviour spec does not match the behaviour of legacy implementations


fippo:
Run this snippet in Chrome or Firefox:
```
var pc = new RTCPeerConnection();
pc.createOffer({offerToReceiveAudio: true})
.then((offer) => {
  console.log('offer 1', offer.sdp)
  return pc.createOffer({offerToReceiveVideo: true})
 })
 .then((offer) => {
  console.log('offer 2', offer.sdp)
 });
```
The first offer will have an audio m-line. The second offer will have a video m-line.

Now according to the [specification of the legacy offerToReceive*](http://w3c.github.io/webrtc-pc/#configuration-data-extensions) this is not valid since offerToReceive internally creates a transceiver. In the addStream-model createOffer was stateless -- I think the same applies to addTrack.
This is quite a surprising side effect, looks like we managed to specify legacy behaviour that does not match the behaviour of legacy clients.

I do not think the spec should be changed but there should at least be a note explaining this.
I actually like the new behaviour more than having to remember what arguments I used previously.

fippo:
Note that the actual impact, i.e. people calling createOffer without setLocalDescription is very low.

ibc:
So, in the new transceiver based spec, isn't createOffer stateless? If I properly understand, it's not stateless because it creates a transceiver, so the app can set transceiver's parameters and then... createOffer + setRemoteDescription?
Everything around transceivers is crazy. Death to SDP.

jan-ivar:
Whether createOffer internally creates a transceiver is not very clear from the [spec language](http://w3c.github.io/webrtc-pc/#dom-rtcofferoptions-offertoreceiveaudio) IMHO:

 *"When this is given a non-false value, no outgoing track of type "audio" is attached to the PeerConnection, and the existing localDescription (if any) doesn't contain any sendrecv or recv audio media sections, createOffer() will behave as if addTransceiver("audio") had been called once prior to the createOffer() call. "*

In the narrowest interpretation, this statement only modifies createOffer's behavior: to act as it would have acted had addTransceiver() been previously called. This to me does not say it acts as addTransceiver.

jan-ivar:
That said, it raises questions about the state of things after createOffer(). Without a transceiver, there's no sender nor receiver, and no tracks.

jan-ivar:
[...]

Issue #1361, opened 2017-06-09
Clarify offerToReceiveAudio and offerToReceiveVideo in renegotiation


fluffy:
Not clear if setting  RTCOfferOptions changes what happens in renegotation  or can this only be used in an initial offer.

stefhak:
There is indeed no such info. However, it is said that 
```When this is given a non-false value, no outgoing track of type "audio" is attached to the PeerConnection, and the existing localDescription (if any) doesn't contain any sendrecv or recv audio media sections, createOffer() will behave as if addTransceiver("audio") had been called once prior to the createOffer() call. ``` 
which I read as this would happen every time createOffer() is called.
Is this the way it should work in your opinion?

stefhak:
Related to #1383

ekr:
My question is: what existing software depends on the creation of the transceiver at this time as opposed to (say) at setLocal()

taylor-b:
It's not that software depends on it, it's that _JSEP_ does. JSEP says "here's how you create an offer given a set of transceivers". So without actually adding a transceiver, how can we say "create an offer according to JSEP"? We'd need to say "create an offer according to JSEP, acting as if these extra transceivers exist if `offerToReceiveX` was true". And we'd need to specify how `setLocalDescription` creates transceivers.

This PR is just much simpler than any other alternative, and solves most of the backwards compatibility use cases, which is what `offerToReceive` exists for in the first place. I don't think we should introduce extra complexity just so that the legacy API matches the object model perfectly; the legacy API only exists for practical reasons, so it makes sense to design it according to practical reasons.

aboba:
[...]

PR #1430, opened 2017-06-28
Have createOffer call addTransceiver() on offerToReceive.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1383 and https://github.com/w3c/webrtc-pc/issues/1361

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3464,7 +3464,7 @@ interface RTCPeerConnection : EventTarget  {
         </div>
         </section>
       <section>
-        <h4>Configuration data extensions</h4>
+        <h4>Legacy configuration extensions</h4>
         <div>
           <pre class="idl">partial dictionary RTCOfferOptions {
             boolean offerToReceiveAudio;
@@ -3477,11 +3477,15 @@ interface RTCPeerConnection : EventTarget  {
               <dt><dfn>offerToReceiveAudio</dfn> of type <span class="idlMemberType"><a>boolean</a></span></dt>
               <dd>
                 <p>
-                  When this is given a non-false value, no outgoing track of
-                  type "audio" is attached to the PeerConnection, and the existing
-                  localDescription (if any) doesn't contain any sendrecv or recv audio media sections,
-                  createOffer() will behave as if addTransceiver("audio") had been called once
-                  prior to the createOffer() call.
+                  Whenever this is given a non-false value, and the
+                  <code><a>RTCPeerConnection</a></code> has no non-stopped
+                  "sendrecv" or "recvonly" audio transceivers, createOffer()
+                  MUST as its first step invoke the equivalent of
+                  <code>addTransceiver("audio")</code> on the
+                  <code><a>RTCPeerConnection</a></code> object, except that this
+                  MUST NOT <a>Update the negotiation-needed flag</a>,  and,
+                  provided this does not fail, proceed with createOffer()'s
+                  regular steps.
                 </p>
                 <p>
                   In all other situations, it will be disregarded.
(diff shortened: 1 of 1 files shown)


fippo:
LGTM, thanks!

aboba:
@ekr Can you review?

aboba:
From https://www.w3.org/2017/06/28-webrtc-minutes

EKR to comment on the PR.

aboba:
@ekr Can you review this?

ekr:
None of this seems right. JSEP seems pretty clear that createOffer() doesn't modify the state of the PC. @juberti?

stefhak:
I don't see this as createOffer() modifying state, but rather that when a user uses the legacy api ```offerToReceive``` they should really do ```addTrancseiver```, and the UA does it for the user (sort of a shim really).

This is is for legacy compatibility, and JSEP has removed everything about ```offerToReceive``` so I'm personally not overly concerned about aligning to the JSEP "spirit". I think this PR is a pragmatic way to add legacy compatibility we decided to do in a VI.

aboba:
@stefhak Agree that on a browser that supports `addTransceiver` but not `offerToReceive`, `offerToReceive` could be implemented as a shim that called `addTransceiver`. 

This is also an argument for why `offerToReceive` is more of a convenience than a necessity.  Since there isn't an alternative way to support `offerToReceive` that creates `RtpSender`/`RtpReceiver` objects and all the methods that depend on them, if native support for `offerToReceive` is considered problematic, then the best alternative is probably to remove it from the spec and support it via shim if needed.


In fact, without an implicit call to `addTransceiver` it is not clear to me how to create RtpSender/RtpReceiver objects and enable methods referencing those objects to function as expected.

stefhak:
@aboba I agree fully to what is said in https://github.com/w3c/webrtc-pc/pull/1430#issuecomment-327587494. However, to remove ```offerToReceove``` we would have to revert the decision to bring it back into the spec (remember it was gone for some time).

jan-ivar:
@aboba I'm not seeing how the availability or ease of writing a shim makes a difference. Some sites will stop working (break), and would need manual intervention to resume working. AFAIK the triviality of intervention does not nullify web compat problems (or we wouldn't have nearly as many legacy APIs). Also, once we have a developer's attention, I'd rather they fix their code to call the new API directly than install some shim that perpetuates old modalities.

stefhak:
Just connecting some (seemingly related) comments. I think https://github.com/w3c/webrtc-pc/issues/1383#issuecomment-327944759 is a response to https://github.com/w3c/webrtc-pc/pull/1430#issuecomment-327004905

stefhak:
@jan-ivar I think you're perfectly right. But I was under the impression that many sites use the same shim, so that could have been a way to handle at least those. Anyway, I agree to that if we have the developer's attention we should get the to use transceivers instead.

jan-ivar:
@stefhak If you're thinking of [adapter.js](https://github.com/webrtc/adapter), it generally shims new APIs on older browsers, not old APIs on new ones. Much like babel.js lets sites write es7+ without worrying about older browsers.
 
That's in some sense the opposite of a legacy shim that perpetuates old APIs: Move to adapter.js to write against the latest API, not the oldest API. It solves the legacy burden for people who keep up with the spec, while *browsers* handle the other group. The doc language on this being a "shim to insulate apps from spec changes" is a bit misleading. It generally insulates apps from browser differences, *if* you follow the spec.

For this reason, production sites are encouraged not to use [adapter-latest.js](https://webrtc.github.io/adapter/adapter-latest.js), which is generally the least stable version, not the most stable (nor do they like such runtime dependencies anyway, which works out well). Adapter.js has minted breaking major version updates at least a couple of times in the last couple of years to keep up with spec changes.

So it requires "developer attention" to update a production site's copy of adapter.js, which means it's of limited use to mitigate web compat problems.

**There are two exceptions**: Legacy APIs on Safari, and `createObjectURL(stream)`. The thinking here was that by transitioning support from browsers to adapter.js in a *minor* adapter release, it would move this breakage off individual browser release calendars, and onto the adapter.js major version update calendar. Kudos to the Safari team here.

I suppose we could attempt a similar exception for `offerToReceiveX`, *provided browsers commit to ending support*. Basically we need to follow through on it, and yank adapter support eventually in a major adapter version, after all browsers have yanked support.

Accumulating exceptions in adapter.js hurt its reputation as the spec-reference module. We've heard confusion already from people who think it's a "bloated legacy library" when it is anything but. We want sites to use adapter, which means keeping it lean and forward-thinking.

stefhak:
@jan-ivar good points.

To re-iterate, the current decision is to have ```offerToReceive``` in the spec, and I personally think that with this PR merged we have it specced well enough. I hope we can get consensus to merge this PR and then be done with ```offerToReceive```.

Issue #1390, opened 2017-06-15
removeTrack: does not always cause createOffer to mark media description


soareschen:
In the last few steps for `removeTrack`, `transceiver.direction` is altered based on the value of its `currentDirection`, not `direction`. However `currentDirection` is nullable and only changes after session descriptions.

This means if `removeTrack` is called when `currentDirection` is null, such as before setting session descriptions, the transceiver's direction is not changed. Is this intentional?

aboba:
The text for removeTrack says: 

"Stops sending media from sender. The `RTCRtpSender` will still appear in `getSenders`. Doing so will cause future calls to `createOffer` to mark the media description for the corresponding transceiver as `recvonly` or `inactive`, as defined in [JSEP] (section 5.2.2.)."

taylor-b:
This is an issue in JSEP as well. Filed: https://github.com/rtcweb-wg/jsep/issues/757

PR #1434, opened 2017-06-28
Don't proceed w/removeTrack() if sender.track is already null.

jan-ivar:
Optimization.

***
[Preview](https://s3.amazonaws.com/pr-preview/jan-ivar/webrtc-pc/removetrack.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/c602d52...jan-ivar:ec20c9d.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5223,6 +5223,10 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                   type "rollback"), then abort these steps.</p>
                 </li>
                 <li>
+                  <p>If <code><var>sender</var>.track</code> is null, then
+                  abort these steps.
+                </li>
+                <li>
                   <p>Set <code><var>sender</var>.track</code> to null.
                 </li>
                 <li>


fippo:
without this
```
pc.replaceTrack(someSender);
pc.replaceTrack(someSender);
```
might have some unintended consequences (like firing ONN)

taylor-b:
Fundamentally this isn't about ONN. It's more "don't change transceiver direction in `removeTrack` if sender's track is already `null`". So I'd suggest changing the PR description.

aboba:
@jan-ivar Is there are issue associated with this PR?

stefhak:
I think we could merge this one. There is some more discussion to be had (relation to #1390), but this PR in itself looks good.

jan-ivar:
Hi, I'm back from a week off. This is unrelated to https://github.com/w3c/webrtc-pc/issues/1390. Sorry for the delay in answering and not filing a separate issue, as I thought it trivial (I was wrong)! 

First, I believe @fippo meant to write:
```js
pc.removeTrack(someSender);
pc.removeTrack(someSender);
```
I think he and I were confused by [current behavior](https://jsfiddle.net/jib1/ucvgcaox/) where the above fires ONN 3 (instead of 2!) times in Chrome (and fails for unrelated reasons in Firefox).

But @taylor-b is right, the [[[needNegotiation](http://w3c.github.io/webrtc-pc/#dfn-update-the-negotiation-needed-flag)]] internal slot already catches this; doesn't fire twice, so the spec was already fine here wrt ONN.

The remaining issue would be the implicit setting of direction. What should the following output?
```js
pc.removeTrack(transceiver.sender);
transceiver.setDirection("sendrecv");
pc.removeTrack(transceiver.sender);
console.log(transceiver.direction); // "recvonly" or "sendrecv" ?
```

jan-ivar:
@fippo To clarify, with this PR merged, the 2nd `removeTrack` in my example skips remaining steps that used to (re)set `transceiver.direction`. It's now a no-op, leaving `"sendrecv"`.

taylor-b:
> The remaining issue would be the implicit setting of direction. What should the following output?

That's the question. I can imagine arguments for either, but since this is a corner case and you can always just use `replaceTrack`/`setDirection` directly to get what you want, I think we just need to decide on _something_. Given the updated description, I'd be fine going with this.

PR #1441, opened 2017-06-30
update current*Description and pending*Description

### LINKED ISSUE #1192: editorial: currentLocalDescription getting updates with candidates
http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-currentlocaldescription
says (and a similar thing happens with pendlingLocalDescription and the corresponding remote descriptions)

> any local candidates that have been generated by the ICE Agent since the offer or answer was created.

and then repeats "plus the candidates" immediately

> complete with any local candidates that have been generated by the ICE Agent since the offer or answer was created.

I think removing the first occurence will not harm the technical requirements.

> Note that the value of currentLocalDescription.sdp need not match the value of localDescription.sdp.

This should be moved to the first paragraph since it is not a statement about the return value.

fippo:
updates the description of the current*Description and pending*Description
attributes. Fixes #1192

***
[Preview](https://s3.amazonaws.com/pr-preview/fippo/webrtc-pc/fix-1192.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/c232b6e...fippo:2839ffb.html)


webrtc.html
```diff
@@ -1767,18 +1767,17 @@
                 attribute represents the local
                 <code><a>RTCSessionDescription</a></code> that was successfully
                 negotiated the last time the <code>RTCPeerConnection</code>
-                transitioned into the stable state plus any local candidates
-                that have been generated by the <a>ICE Agent</a> since the offer
-                or answer was created.</p>
+                transitioned into the stable state.</p>
+                <p>Note that the value of
+                <code>currentLocalDescription.sdp</code> need not match the
+                value of <code>localDescription.sdp</code>.</p>
```
taylor-b: `localDescription` will always be either `currentLocalDescription` or `pendingLocalDescription`, right? If so, it would be better if the note was phrased in those terms. And also used `class="note"`.

fippo:
I like that plan and will update this PR!

stefhak:
@fippo ping!

fippo:
the problem I hit whenever I try to do what @taylor-b proposed is that it is a much bigger change that is no longer "local". These slots would need to be updated both in the icecandidate event as well.
And that seems a bit too effort much for something I still consider a cosmetic change :-/

Issue #1438, opened 2017-06-29
Should refer to enums in a consistent way


taylor-b:
Should we refer to enums:

1. In a code block? `connected`
2. In quotes? "connected"
3. Both? `"connected"`

Since it's both code and a string, I'd prefer the latter.

burnburn:
@aboba , you say that PR #1485 is a partial fix.  What remains?

aboba:
@burnburn AFAIK it is fixed.  If there are any remaining issues we can reopen it.

PR #1442, opened 2017-06-30
Editorial: Add 'Style Conventions' section to contributing guidelines

adam-be:
This is a first stab at a set of 'Style Conventions'. We need to work out the details. My guess is that there might be some controversy around strings and enums.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,3 +28,20 @@ Section aliases, to be used as data-jsep values, are found in [jsep-mapping/map.
 
 ### Refactoring: Moving Text
 If your PR needs to both move a section of text and update it, please do these operations in two separate commits (or even PRs). That will make the review process simpler.
+
+### Style Conventions
+
+**Strings**. Use quotes and enclose in code tags (to distinguish strings literals from quoted prose).
+Example: `<code>"String literal"</code>`
+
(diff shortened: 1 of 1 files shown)


adam-be:
Related: #1438, #1395, #1409

CONTRIBUTING.md
```diff
@@ -28,3 +28,20 @@ Section aliases, to be used as data-jsep values, are found in [jsep-mapping/map.
...
+
+**Booleans**. Enclose booleans in code tags (without quotes).\
+Example: `<code>true</code>`
+
+**Enum Values**. See **Strings**.
+
+**Notes**. Use div tags for notes.\
+Example: `<div class='note'>FYI</div>`
+
+**Internal Slots**. Use upper camel case (initial upper case letter) with a leading backslash and enclosed by two brackets. The leading backslash prevents your slot name from colliding with other definitions such as attribute names. If a slot name is very general, like "Id", it might be useful to add more information to it to avoid name collisions and make it easier to identify its owning entity. For example, "Id" could be extended to "ComponentId".\
```
taylor-b: Can this be split into multiple lines? Also, the leading backslash isn't to prevent names from colliding, it's to prevent the slot name from being interpreted as a bibliography link (I think).

PR #1443, opened 2017-06-30
SLD/SRD: Check if transceiver is stopped before setting currentDirection

adam-be:
This change also adds a [[Stopped]] internal slot to RTCRtpTransceiver.

We might want to prevent more things from happening in this scenario as well.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/set-desc-check-transceiver.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/2d424aa...927e896.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1516,6 +1516,10 @@
                         the corresponding media description.</p>
                       </li>
                       <li>
+                        <p>If <var>transceiver</var>'s <a>[[\Stopped]]</a> slot
+                        is <code>true</code>, abort these sub steps.</p>
+                      </li>
+                      <li>
                         <p>If <var>description</var> is of type
                         <code>"answer"</code> or <code>"pranswer"</code>, then
                         set <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1525,6 +1525,10 @@
                         the corresponding media description.</p>
                       </li>
                       <li>
+                        <p>If <var>transceiver</var>'s <a>[[\Stopped]]</a> slot
+                        is <code>true</code>, abort these sub steps.</p>
```
taylor-b: Is the meaning of "abort these sub steps" clear enough? Maybe it's time to break the giant list of "set a description" steps into different algorithms. Also, this is where `[[CurrentDirection]]` is set for a local description, but it's also set for a remote description further down.

aboba:
@adam-be  How does `transceiver.stopped` get set?  I think we need language that states that the `stopped` attribute is retrieved from the internal slot.

adam-be:
@aboba, it's described in the attribute description that on "getting", the attribute must return the value of the slot.

webrtc.html
adam-be: Yes, it might be a good idea to split it up. We have a lot of conditions regarding local/remote and offer/answer. It would be nice to have separate algorithms for setLocal/setRemote and perhaps have a set of shared substeps that could be referenced.

adam-be:
Had conflicts; rebased.

PR #1450, opened 2017-07-05
Remove duplicate info on method parameters

dontcallmedom:
These were inherited from the previous respec WebIDL format, but they now represent a maintenance footgun

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/noparamtables.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/cdf9e75...b0ac6f1.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1876,26 +1876,6 @@ interface RTCPeerConnection : EventTarget  {
               <dd>
                 See the <a data-lt="RTCPeerConnection()">RTCPeerConnection
                 constructor algorithm</a>.
-                <table class="parameters">
-                  <tbody>
-                    <tr>
-                      <th>Parameter</th>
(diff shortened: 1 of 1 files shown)


burnburn:
@dontcallmedom we should be consistent with gUM.  What do you think would be best to do that would give us consistency?

dontcallmedom:
I think we should aim at this consistency, but that requires someone doing the work (not only for gum, but all the other specs who were converted); I'm happy to be assigned relevant issues in these other specs if no one else is available.

But in any case, I don't think we need to land these patches simultaneously - consistency over time is fine from my perspective.

aboba:
@dontcallmedom Can you rebase this PR?

dontcallmedom:
rebased

Issue #1125, opened 2017-04-13
Should the spec describe addStream/onaddstream as legacy API?


youennf:
Related discussion in https://github.com/w3c/web-platform-tests/issues/5533

addStream and onaddstream are widely used but not specified in the WebRTC specification.
This API is more convenient to use from a web developer point of view for simple A/V calls then addTrack/ontrack.
This API is available in Edge, Chrome and Firefox while addTrack is only supported in Firefox.

From a pragmatic point of view, it seems like the webrtc spec should describe this legacy API, in terms of existing API like addTrack for instance.

nils-ohlmeier:
> This API is more convenient to use from a web developer point of view for simple A/V calls then addTrack/ontrack.
Then write a library or some code which simplifies the API usage, but don't bloat the native API to support a simple and a more complex version.

> This API is available in Edge, Chrome and Firefox while addTrack is only supported in Firefox.
Then get the implementations which are lacking behind the spec to finally implement what the spec is and don't give them excuses to implement what ever they want.

> From a pragmatic point of view, it seems like the webrtc spec should describe this legacy API, in terms of existing API like addTrack for instance.
I think the WebRTC specs are way to complex already. Adding sections about legacy stuff will only bloat them and cause confusion on what has to be implemented and what is available. How about you create a separate document/spec which describes the old, legacy APIs.

youennf:
>> This API is more convenient to use from a web developer point of view for simple A/V calls then addTrack/ontrack.
> Then write a library or some code which simplifies the API usage, but don't bloat the native API to support a simple and a more complex version.

The point I was trying to make was just that web developers will not have a huge incentive to move towards addTrack given addStream is often more convenient. Contrary to callback/promise APIs.
This might mean that addStream might remain in browsers for a long time.

>> This API is available in Edge, Chrome and Firefox while addTrack is only supported in Firefox.
> Then get the implementations which are lacking behind the spec to finally implement what the spec is and don't give them excuses to implement what ever they want.

As a browser implementor, not shipping addStream might be a problem (currently under assessment in WebKit).
As a browser implementor, implementing addStream without a spec or a path towards convergence is a problem.

> I think the WebRTC specs are way to complex already. Adding sections about legacy stuff will only bloat them and cause confusion on what has to be implemented and what is available. How about you create a separate document/spec which describes the old, legacy APIs.

Agreed on the complexity of the WebRTC spec.
[...]

PR #1451, opened 2017-07-05
Add legacy note about addStream.

jan-ivar:
Potentially for https://github.com/w3c/webrtc-pc/issues/1125.

***
[Preview](https://s3.amazonaws.com/pr-preview/jan-ivar/webrtc-pc/legacynote.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/3345dad...jan-ivar:0e23326.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3003,9 +3003,11 @@ interface RTCPeerConnection : EventTarget  {
           if these methods are supported it is mandatory to implement
           according to what is specified here.</p>
         <div class="note">
-          The addStream(stream) method that used to exist on
-          <code><a>RTCPeerConnection</a></code> is easy to polyfill as:<br>
-          stream.getTracks().forEach(track => this.addTrack(track, stream));
+          The addStream method that used to exist on
+          <a>RTCPeerConnection</a></code> is easy to polyfill as:
+          <pre>RTCPeerConnection.prototype.addStream = function(stream) {
+  stream.getTracks().forEach(track => this.addTrack(track, stream));
+};</pre>
         </div>
         <section>
           <h4>Method extensions</h4>
--- a/webrtc.html
+++ b/webrtc.html
@@ -3003,8 +3003,9 @@ interface RTCPeerConnection : EventTarget  {
           if these methods are supported it is mandatory to implement
           according to what is specified here.</p>
         <div class="note">
-          These methods are kept on <code><a>RTCPeerConnection</a></code> for
-          legacy purposes.
+          The addStream(stream) method that used to exist on
+          <code><a>RTCPeerConnection</a></code> is easy to polyfill as:<br>
+          stream.getTracks().forEach(track => this.addTrack(track, stream));
         </div>
         <section>
           <h4>Method extensions</h4>
--- a/webrtc.html
+++ b/webrtc.html
@@ -3004,7 +3004,7 @@ interface RTCPeerConnection : EventTarget  {
           according to what is specified here.</p>
         <div class="note">
           The addStream method that used to exist on
-          <a>RTCPeerConnection</a></code> is easy to polyfill as:
+          <code><a>RTCPeerConnection</a></code> is easy to polyfill as:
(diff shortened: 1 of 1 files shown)


fippo:
if we do it then write it like this
```
RTCPeerConnection.prototype.addStream = function(stream) {
  var pc = this;
  stream.getTracks.forEach(function(track) {
    pc.addTrack(track, stream);
  });
};
```
as a pre class="example highlight"?

jan-ivar:
@fippo `class="example highlight"` didn't work inside a note, so just using `pre`.

taylor-b:
Looks good, but should we also acknowledge the behavior described [here](https://bugs.chromium.org/p/webrtc/issues/detail?id=7815)? This makes the polyfill slightly more complicated:

```
RTCPeerConnection.prototype.addStream = function(stream) {
  var pc = this;
  stream.getTracks().forEach(function(track) {
    pc.addTrack(track, stream);
  });
  // Listen to future changes to the stream, and add/remove tracks automatically.
  oldAddTrack = stream.addTrack;
  oldRemoveTrack = stream.removeTrack;
  stream.addTrack = function(track) {
    pc.addTrack(track, stream);
    oldAddTrack.call(stream, track);
  }
  stream.removeTrack = function(track) {
    pc.removeTrack(track, stream);
    oldRemoveTrack.call(stream, track);
  }
};
```

jan-ivar:
@taylor-b AFAIK that behavior was never in the spec, and isn't web compatible, so I don't think it belongs there. Maybe MDN?

For me the goal of this note is someone searching the spec for "addStream" finding a vestige of it.

taylor-b:
This is how Chrome has always behaved, and given that Chrome doesn't support `addTrack`/`removeTrack` (until very soon!), this is the only way people have had to modify a stream (besides removing it and re-adding it, which has consequences). I can imagine there's a lot of code that looks like this:

```
stream.addTrack(track);
if (pc.addTrack) {
  pc.addTrack(track, stream);
}
```

Or some variant thereof.

Also, this behavior *was* in the standard (or at least was heavily implied) at one point:

> If something in the browser changes that causes the RTCPeerConnection object to need to initiate a new session description negotiation, a negotiationneeded event is fired at the RTCPeerConnection object.
> 
> In particular, if an RTCPeerConnection object is consuming a MediaStream on which a track is added, by, e.g., the addTrack() method being invoked, the RTCPeerConnection object must fire the "negotiationneeded" event. Removal of media components must also trigger "negotiationneeded".

jan-ivar:
> I can imagine there's a lot of code that looks like this:
> ```
> stream.addTrack(track);
> if (pc.addTrack) {
>   pc.addTrack(track, stream);
> }
>  ```

@taylor-b then we should ***not*** polyfill automatic addition, or `addTrack` throws `InvalidAccessError`.

taylor-b:
@jan-ivar Oh, you mean this code will start failing once Chrome does implement `addTrack`? Then replace it with:

```
if (chrome) {
  stream.addTrack(track);
} else {
  pc.addTrack(track, stream);
}
```

I'm not an application developer myself, I'm not sure what people do most often.

But my point is, it's behavior that's been around for a long time, which we _will_ need to help people migrate off of. If you don't think it should be mentioned in the note here, that's fine; I just wanted to call it out in case people were unaware of it.

jan-ivar:
@taylor-b I'm not an application developer eiter, but FWIW it's my understanding that feature-testing (your first example) is considered a good pattern, wheras user-agent testing (the latter) is not.

jan-ivar:
@taylor-b Lastly, we try to encourage polyfills of new APIs, not legacy ones. 

I've added a [Migration](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addStream#Migration) section to the addStream MDN entry. PTAL.

taylor-b:
@jan-ivar The migration section you added looks good, thanks.

PR #1453, opened 2017-07-05
Removing text talking about key shortening that was incorrect.

### LINKED ISSUE #1156: Issue 1: Key shortening

taylor-b:
Fixes #1156.

The HMAC algorithm itself accepts *any* key size. So all we have to say
here is that WebRTC allows any key size 20 bytes or greater (even if the
implementation has transitioned to HMAC-SHA-256, and would prefer a
32-byte key for the added security).

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1156_key_shortening.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/3345dad...taylor-b:52ca510.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -278,21 +278,10 @@
                 <p>However, this specification uses a simplified <a>alg</a>
                 approach. The length of the HMAC key
                 (<code>RTCOAuthCredential.macKey</code>) MAY be any multiple of
-                bytes greater than 20 (160 bits). When the <a>ICE Agent</a>
-                computes the message integrity attributes it supports, it will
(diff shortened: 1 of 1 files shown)


aboba:
Do you really want to say "MAY be any multiple of bytes greater than 20"?

Issue #1447, opened 2017-07-05
setConfiguration throwing InvalidModificationError with default values


soareschen:
In set configuration, the fields `bundlePolicy`, `rtcpMuxPolicy`, and `iceCandidatePoolSize` are checked against the original values and if they are not the same, `InvalidModificationError` is thrown. e.g.:

> If configuration.bundlePolicy is set and its value differs from the connection's bundle policy, throw an InvalidModificationError.

But there is a problem with the wording _"if ___ is set"_. Because these fields all have default values as defined in `RTCConfiguration`, this means they are always set. So the following example would throw error:

```javascript
const pc = new RTCPeerConnection({ bundlePolicy: 'max-bundle' });
// throws InvalidModificationError because provided default bundlePolicy is 'balanced'
pc.setConfiguration({});
```

taylor-b:
Do we just need to change the wording to be more clear, or do you think the error being thrown is a problem? If an application does this, then things should work fine:

```
var config = { bundlePolicy: 'max-bundle' };
const pc = new RTCPeerConnection(config);

// Later on, fiddle with some bits of |config|.
config.iceServers = [...]
pc.setConfiguration(config);
```

Which seems like more maintainable code anyway.

aboba:
While I would suggest removing the "is set" statements for attributes with defaults, the errors are being thrown correctly (a dictionary entry is being changed and that is forbidden).

soareschen:
> Do we just need to change the wording to be more clear, or do you think the error being thrown is a problem?

In general I find the behavior of default values in `setConfiguration` to be unintuitive and may bring surprise to unfamiliar users. Other than the issues mentioned above and #1454, the default value for `iceTransportPolicy` can also cause unexpected bug:

```javascript
const pc = new RTCPeerConnection({ iceTransportPolicy: 'relay' });
pc.setConfiguration({});

pc.getConfiguration().iceTransportPolicy; // now become 'all'
```

The coding convention @taylor-b mentioned above can workaround the issue. But that means users would need to follow the convention to avoid unexpected bugs, and there should be notes some where to recommend the coding convention.

taylor-b:
> there should be notes some where to recommend the coding convention.

Assuming people don't want to change the behavior, that seems like the best resolution.

Issue #1454, opened 2017-07-06
Using setConfiguration() to add application certificates to an RTCPeerConnection post-construction?


soareschen:
This issue is similar to #1447. The last step of setConfiguration saves the provided configuration object to the `[[Configuration]]` internal slot, overriding the previous values. #1447 covers the fields with default values, so they cannot be unset. But fields like `iceServers`, `peerIdentity` and `certificates` can seemingly be "unset" from the user-visible configuration. For example:

```javascript
const pc = new RTCPeerConnection({ 
  iceServers: [...], 
  certificates: [...], 
  peerIdentity: ... 
});

pc.setConfiguration({});
const config = pc.getConfiguration();

config.iceServers // undefined
config.certificates // undefined
config.peerIdentity //undefined
```

Note that this only affects the user-visible configuration. Since the spec lets `RTCPeerConnection` save the actual configuration internally elsewhere, the "unset" never affects the actual connection.

soareschen:
A way to fix this is to update the individual fields in the `[[Configuration]]` internal slot, if the field is set in the provided configuration and update is allowed. This way the other fields cannot get "unset" from the result of `getConfiguration()`.

aboba:
getConfiguration() says "When this method is called, the user agent must return the RTCConfiguration object stored in the [[ configuration]] internal slot."  So why would iceServers, certficates and peerIdentity be undefined?

soareschen:
The last step of `setConfiguration()` says:

> Store the configuration in the [[configuration]] internal slot.

Which means the object `[[configuration]]` gets replaced with the new configuration object passed to `setConfiguration()`.

aboba:
In the example, `setConfiguration()` will throw an exception since it is attempting to change parameters (such as certificates) that cannot be modified post-construction.  So `getConfiguration()` will return the original configuration.

taylor-b:
Unsetting the certs will fail with `InvalidModificationError. But if I did:

```
pc = new RTCPeerConnection( { iceServers: [...] } );
pc.setConfiguration({});
```

The ICE servers *would* get unset. And I don't see a problem with this; maybe it's a little confusing but it's how the API was designed, and is consistent with how `getParameters`/`setParameters` work. You have to pass in the full dictionary every time, not just the things you want to change.

soareschen:
[...]

PR #1460, opened 2017-07-06
Remove "is set" text

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1447

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/issue-1447-patchv3.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/7ba8467...bd2797a.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1683,18 +1683,18 @@
             set and the set of certificates differs from the ones used
             when <var>connection</var> was constructed, <a>throw</a> an
             <code>InvalidModificationError</code>.</li>
-            <li>If <code><var>configuration</var>.<a data-for=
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1683,18 +1683,18 @@
             set and the set of certificates differs from the ones used
```
taylor-b: Don't we also want to remove "is set" from here? And add a default value (`[]`)? I may be getting my `setConfiguration` issues confused.

aboba: I believe that relates to Issue https://github.com/w3c/webrtc-pc/issues/1454 .

Issue #1456, opened 2017-07-06
setConfiguration({ peerIdentity }) when there is no initial peerIdentity


soareschen:
`setConfiguration` says:

>  If configuration.peerIdentity is set and its value differs from the target peer identity, throw an InvalidModificationError.

`setRemoteDescription` says

> The target peer identity cannot be changed once set.

If a peer connection is initially constructed without a `peerIdentity`, there is no target peer identity established, so it is not set. So when `setConfiguration()` is called with a new `peerIdentity`, is there error thrown?

aboba:
I would assume not, because `peerIdentity` was not set when `setConfiguration()` was called.

taylor-b:
So basically, we want to say "If `configuration.peerIdentity` is set, and `[[Configuration]].peerIdentity` is not set..."?

aboba:
@jan-ivar If `peerIdentity` is not set on construction, can it be added later?

jan-ivar:
That's my best reading of: The (presence of the) [peerIdentity](http://w3c.github.io/webrtc-pc/#dom-rtcconfiguration-peeridentity) dictionary member *"Sets the target peer identity for the RTCPeerConnection."*- and *"The target peer identity cannot be changed once set."*

Check with @martinthomson though if this was the intent, in case there's some security reason to not allow this.

martinthomson:
To @aboba's question, yes, you could set it later, but only if it agrees with the current target peer identity.

So the challenge with `setConfiguration()` is that it can be called after the session is established.  If the current peer is unidentified, then we have a problem because the session now disagrees with the configuration.

It seems like the right solution here is to define an internal slot `[[TargetPeerIdentity]]`, which can three states: unspecified, null, and a string value.  This is initially unspecified.  Specifying a value using `setConfiguration()` sets the value to null/string.  Connecting sets the value to null/string.  If `setConfiguration()` attempts to set a value and it isn't either the same as the current value, or the current value is unspecified, then `setConfiguration()` fails.  The same process applies when connecting.

PR #1464, opened 2017-07-07
setConfiguration({ peerIdentity }) when there is no initial peerIdentity

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1456

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/issue-1456-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/7ba8467...0b011a6.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1679,6 +1679,9 @@
             set and <a>[[\Configuration]]</a>.<code>peerIdentity</code> is
             not set, <a>throw</a> an <code>InvalidModificationError</code>.
             </li>
+            <li>If <code><var>configuration</var>.peerIdentity</code> is
+            set and its value differs from the <a>target peer
+            identity</a>, <a>throw</a> an <code>InvalidModificationError</code>.
             <li>If <code><var>configuration</var>.certificates</code> is
             set and the set of certificates differs from the ones used
             when <var>connection</var> was constructed, <a>throw</a> an
--- a/webrtc.html
+++ b/webrtc.html
@@ -1679,9 +1679,6 @@
             set and <a>[[\Configuration]]</a>.<code>peerIdentity</code> is
             not set, <a>throw</a> an <code>InvalidModificationError</code>.
             </li>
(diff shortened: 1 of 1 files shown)


soareschen:
This removes the original condition. Does this means peerIdentity can be changed if it is initially set?

aboba:
@soareschen Restored the original condition. `peerIdentity` should not be changeable.

martinthomson:
This would be a fine answer to the question.  Note that you could take the approach I stated earlier, but that's more complicated.

webrtc.html
```diff
@@ -1679,9 +1679,6 @@
             set and <a>[[\Configuration]]</a>.<code>peerIdentity</code> is
             not set, <a>throw</a> an <code>InvalidModificationError</code>.
             </li>
-            <li>If <code><var>configuration</var>.peerIdentity</code> is
```
soareschen: Is the original condition removed again?

webrtc.html
```diff
@@ -1676,8 +1676,8 @@
             <li>Let <var>connection</var> be the target
             <code><a>RTCPeerConnection</a></code> object.</li>
             <li>If <code><var>configuration</var>.peerIdentity</code> is
-            set and its value differs from the <a>target peer
-            identity</a>, <a>throw</a> an <code>InvalidModificationError</code>.
+            set and <a>[[\Configuration]]</a>.<code>peerIdentity</code> is
+            not set, <a>throw</a> an <code>InvalidModificationError</code>.
```
taylor-b: This sounds like it allows the peer identity to change, just not go from being unset to being set. I don't think that's what we want. I thought the plan was to give everything a default value (`null` for `peerIdentity`, `[]` for `certificates`, etc.), and just say "if the value is different, throw an `InvalidModificationError`".

aboba: Yes, that's better.

Issue #1461, opened 2017-07-07
We forgot about {offerToReceiveAudio: false, offerToReceiveVideo: false}


jan-ivar:
The legacy [offerToReceive](http://w3c.github.io/webrtc-pc/#dom-rtcofferoptions-offertoreceiveaudio) section only covers how we used to set `"recvonly"`:
```js
pc.createOffer({offerToReceiveAudio: true});
```
but not how we used to set `"sendonly"`:
```js
pc.addStream(stream);
pc.createOffer({offerToReceiveAudio: false});
```

Try it with this [fiddle](https://jsfiddle.net/jib1/1h8mr8fw/). Was this omitted on purpose?

Having it implicitly call `setDirection("sendonly")` might suffice.

jan-ivar:
Closing as not needed. See discussion in PR.

PR #1466, opened 2017-07-07
Cover legacy sendonly using {offerToReceiveAudio/Video: false}.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1461.

***
[Preview](https://s3.amazonaws.com/pr-preview/jan-ivar/webrtc-pc/offertoreceivefalse.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/cc8d80f...jan-ivar:eef2557.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3484,7 +3484,13 @@ interface RTCPeerConnection : EventTarget  {
                   prior to the createOffer() call.
                 </p>
                 <p>
-                  In all other situations, it will be disregarded.
+                  Whenever this is given a false value, and the
+                  <code><a>RTCPeerConnection</a></code> has a non-stopped
+                  "sendrecv" audio transceiver, createOffer() MUST as its first
+                  step set that transceiver's
+                  [[<a href="#direction-slot">Direction</a>]] slot
+                  to "sendonly", and then proceed with createOffer()'s regular
+                  steps.
                 </p>
               </dd>
               <dt><dfn>offerToReceiveVideo</dfn> of type <span class="idlMemberType"><a>boolean</a></span></dt>
@@ -3498,7 +3504,13 @@ interface RTCPeerConnection : EventTarget  {
                   to the createOffer() call.
                 </p>
                 <p>
-                  In all other situations, it will be disregarded.
+                  Whenever this is given a false value, and the
+                  <code><a>RTCPeerConnection</a></code> has a non-stopped
+                  "sendrecv" video transceiver, createOffer() MUST as its first
+                  step set that transceiver's
+                  [[<a href="#direction-slot">Direction</a>]] slot
+                  to "sendonly", and then proceed with createOffer()'s regular
+                  steps.
                 </p>
               </dd>
             </dl>
--- a/webrtc.html
+++ b/webrtc.html
@@ -3485,11 +3485,11 @@ interface RTCPeerConnection : EventTarget  {
                 </p>
                 <p>
                   Whenever this is given a false value, and the
-                  <code><a>RTCPeerConnection</a></code> has a non-stopped
-                  "sendrecv" audio transceiver, createOffer() MUST as its first
-                  step set that transceiver's
-                  [[<a href="#direction-slot">Direction</a>]] slot
-                  to "sendonly", and then proceed with createOffer()'s regular
+                  <code><a>RTCPeerConnection</a></code> has one or more
+                  non-stopped "sendrecv" or "recvonly" audio transceivers,
+                  createOffer() MUST as its first step set those transceivers'
+                  <a>[[\Direction]]</a> slot to "sendonly" or "inactive"
+                  respectively, and then proceed with createOffer()'s normal
                   steps.
                 </p>
               </dd>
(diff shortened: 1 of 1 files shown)


fippo:
Are people using this to put someone on hold and we need to cover the recvonly->inactive case too?

webrtc.html
```diff
@@ -3484,7 +3484,13 @@
                   prior to the createOffer() call.
                 </p>
                 <p>
-                  In all other situations, it will be disregarded.
+                  Whenever this is given a false value, and the
+                  <code><a>RTCPeerConnection</a></code> has a non-stopped
+                  "sendrecv" audio transceiver, createOffer() MUST as its first
+                  step set that transceiver's
+                  [[<a href="#direction-slot">Direction</a>]] slot
```
taylor-b: This slot link can be updated.

webrtc.html
```diff
@@ -3484,7 +3484,13 @@
                   prior to the createOffer() call.
                 </p>
                 <p>
-                  In all other situations, it will be disregarded.
+                  Whenever this is given a false value, and the
+                  <code><a>RTCPeerConnection</a></code> has a non-stopped
+                  "sendrecv" audio transceiver, createOffer() MUST as its first
```
taylor-b: Sorry to be pedantic, but what if there are multiple such transceivers? Does this affect all of them, just the first, just the last?

jan-ivar:
@fippo hmm that [actually works](https://jsfiddle.net/jib1/c48q11jd/), though is anyone is using it? It'd only work on one side.

webrtc.html
jan-ivar: Firefox [appears](https://jsfiddle.net/jib1/kvzj12sb/) to apply it to all.

jan-ivar:
Went with all. @taylor-b PTAL.

jan-ivar:
I also added @fippo's bit since it matches two implementations.

burnburn:
Perhaps offerToReceiveX, as legacy behavior, really only needs to be supported for initial configurations and not as a control for later behavior.

jan-ivar:
> applications are likely to do a normal createOffer() later to take the call "off of hold"

@taylor-b That [doesn't actually work today](https://jsfiddle.net/jib1/c48q11jd/3/), but `createOffer({{offerToReceiveAudio: true})` [does](https://jsfiddle.net/jib1/c48q11jd/4/).

jan-ivar:
@aboba What problem requires list discussion? Is this not good to merge?

taylor-b:
> @taylor-b That doesn't actually work today, but createOffer({{offerToReceiveAudio: true}) does.

Ok, then the problem will be that `createOffer({{offerToReceiveAudio: true})` doesn't change the direction to `sendrecv`; same point.

The thing that needs list discussion is whether we want the `offerToReceive` functionality to cover just the most common usage (generating an "m=" section that otherwise wouldn't be present), or if we want to cover the direction-changing use cases as well.

jan-ivar:
@taylor-b I'm happy to cover only basic usage. What text changes would you like to see?

jan-ivar:
To clarify, I consider *basic usage* to include `true` and `false` for initial configurations, as I think @burnburn suggests.

taylor-b:
Let me clarify. There are two classes of use cases I see for `offerToReceive`:

1. The endpoint wants to offer to receive media, but doesn't have media to send. `offerToReceive` is the only way to force audio/video m= sections to be generated in this case.
2. The endpoint wants to manipulate the direction attribute of the "m=" section, because it's dealing with a SIP endpoint.

1 is something I'd expect almost _every_ sufficiently robust application to be using; if the endpoint doesn't have a webcam, it probably still wants to be able to receive video. In the call where we decided to add `offerToReceive` to the legacy section, this was the motivation.

2 is more corner casey, and we could tell these applications "use the transceiver API". Supporting use cases that involve SIP endpoints is why we introduced the transceiver API in the first place...

jan-ivar:
I guess I imagined:

  3. The endpoint has media to send but never wants to receive media. `offerToReceiveAudio/Video: false` is the only way to force "sendonly" as opposed to "sendrecv".

Are there no sites using `{offerToReceiveAudio/Video: false}` to set "sendonly" on initial connection?

taylor-b:
I'm sure those sites exist, but my assumption was that they'd be much less common. What's an example of how an application might use this?

jan-ivar:
> 1. The endpoint wants to receive media, but doesn't have media to send.

> 3. The endpoint has media to send but doesn't want to receive media.

Isn't this two sides of the same coin, the difference being who the offerer is?

I assumed they had the same use-cases.

taylor-b:
> Isn't this two sides of the same coin, the difference being who the offerer is?

Not necessarily. If "A" has media to send but "B" doesn't, the typical offer/answer from A->B will be "sendrecv"/"recvonly", and from B->A will be "recvonly"/"sendonly". In neither case would there be a "sendonly" offer.

jan-ivar:
OK I'm all for less legacy to support!

PR #1479, opened 2017-07-14
rename ufrag to usernameFragment

### LINKED ISSUE #1210: ufrag vs usernameFragment
http://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-ufrag
uses ufrag.

http://w3c.github.io/webrtc-pc/#dom-rtciceparameters-usernamefragment
uses usernameFragment.

They're the same. Can we use just one name?

fippo:
renames the ICE ufrag to usernameFragment for consistency.
Fixes #1210

***
[Preview](https://s3.amazonaws.com/pr-preview/fippo/webrtc-pc/ufrag.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/a12bf47...fippo:0060df4.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2637,7 +2637,7 @@ interface RTCPeerConnection : EventTarget  {
                 "RTCIceCandidate">candidate</a></code>, <code><a data-for=
                 "RTCIceCandidate">sdpMid</a></code>, <code><a data-for=
                 "RTCIceCandidate">sdpMLineIndex</a></code>, and
-                <code><a data-for="RTCIceCandidate">ufrag</a></code>; the rest
+                <code><a data-for="RTCIceCandidate">usernameFragment</a></code>; the rest
                 are ignored. When the method is invoked, the user agent MUST
                 run the following steps:</p>
                 <ol>
@@ -2703,9 +2703,9 @@ interface RTCPeerConnection : EventTarget  {
                         </ol>
                       </li>
                       <li>
-                        <p>If <code><var>candidate</var>.ufrag</code> is neither
+                        <p>If <code><var>candidate</var>.usernameFragment</code> is neither
                         <code>undefined</code> nor <code>null</code>, and is not
-                        equal to any ufrag present in the corresponding
+                        equal to any usernameFragment present in the corresponding
                         <a>media description</a> of an applied remote
                         description, reject <var>p</var> with a newly
                         <a data-link-for="exception" data-lt="create">created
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4106,7 +4106,7 @@
               the candidate that it is derived from. For host candidates, the
               <code>relatedPort</code> is <code>null</code>. This corresponds to
               the <code>rel-port</code> field in <a><code>candidate-attribute</code></a>.</dd>
-              <dt><code><dfn>ufrag</dfn></code> of type <span class=
+              <dt><code><dfn>usernameFragment</dfn></code> of type <span class=
               "idlAttrType"><a>DOMString</a></span>, readonly, nullable</dt>
               <dd>This carries the <code>ufrag</code> as defined in section
```
fippo: here ufrag is the right thing and doesn't need to be replaced.
But technically https://tools.ietf.org/html/rfc5245#section-15.4 talks about ice-ufrag, not ufrag...

taylor-b: "ice-ufrag" is the attribute, "ufrag" is the field inside that attribute. So I think we're ok.

webrtc.html
```diff
@@ -2703,9 +2703,9 @@
                         </ol>
                       </li>
                       <li>
-                        <p>If <code><var>candidate</var>.ufrag</code> is neither
+                        <p>If <code><var>candidate</var>.usernameFragment</code> is neither
                         <code>undefined</code> nor <code>null</code>, and is not
-                        equal to any ufrag present in the corresponding
+                        equal to any usernameFragment present in the corresponding
```
taylor-b: Since this isn't code, should just say "username fragment" here like ICE itself does.

webrtc.html
```diff
@@ -2716,8 +2716,8 @@
                         <p>In parallel, add the ICE candidate
                         <var>candidate</var> as described in <span data-jsep=
                         "addicecandidate">[[!JSEP]]</span>. Use
-                        <code><var>candidate</var>.ufrag</code> to identify the
-                        ICE <a>generation</a>; if the ufrag is null, process the
+                        <code><var>candidate</var>.usernameFragment</code> to identify the
+                        ICE <a>generation</a>; if the usernameFragment is null, process the
```
taylor-b: Here we're talking about the member, not the username fragment in the abstract sense, so should be `<code>usernameFragment</code>`.

webrtc.html
```diff
@@ -7687,7 +7687,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           <code><a data-for="RTCIceCandidate">sdpMLineIndex</a></code>
           set to the values associated with this
           <code><a>RTCIceTransport</a></code>, with
-          <code><a data-for="RTCIceCandidate">ufrag</a></code> set to the ufrag
+          <code><a data-for="RTCIceCandidate">usernameFragment</a></code> set to the usernameFragment
```
taylor-b: Here's another place.

taylor-b:
I went ahead and addressed my comments in the online editor FYI.

stefhak:
LGTM.

webrtc.html
aboba: Is line 2125 intentional or is that a typo?

Issue #1470, opened 2017-07-10
Performing rollback via the other set*Description


soareschen:
JSEP 4.1.8.2. says:

> A rollback is performed by supplying a session description of type "rollback" with empty contents to either setLocalDescription or setRemoteDescription, depending on which was most recently used (i.e. if the new offer was supplied to setLocalDescription, the rollback should be done using setLocalDescription as well).

The current spec doesn't seem to impose such restriction. So say if one calls `setLocalDescription(offer)` then `setRemoteDescription(rollback)`, should there be error thrown?

aboba:
Looks like steps should be added to 4.4.1.6 Set the RTCSessionSessionDescription to check for invalid rollback. Also, it may make sense to file an issue against JSEP.

fluffy:
Given rollback is one of the way to recover from an some errors, my preference is pretty much no matter what state you are in, and no matter which set* got called, exactly the same things happens which is you roll back to last stable state. If you are in a stable state, then it just stays at that same stable state. What I am trying to avoid is needing error handling for my error handling code. It would be really nice if doing a rollback did not cause an error. That would make is easy to use for when a error happened and it would just move you back to basically last known good  state. From an implementation point of view in the browsers, I don't think this changes very much what needs to be implemented.

fluffy:
As side note, JSEP has bugs with this same conversation going on. I suspect that JSEP is the right place to figure out what we want to have happen then figure out how to reflect any call or errors needed into the API.

PR #1487, opened 2017-07-15
Check for invalid rollback

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1470

Related JSEP PR:  https://rawgit.com/w3c/webrtc-pc/master/webrtc.html

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/issue-1470-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/a12bf47...e70dc31.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1377,7 +1377,7 @@
                         to null. Finally set <var>connection</var>'s
                         <a>signaling state</a> to <code>"stable"</code></p>
                       </li>
-                       <!-- B) transition rollback to haveLocalOffer to stable -->
+                      <!-- B) transition rollback to haveLocalOffer to stable -->
                       <li>
                         <p>If <var>description</var> is of type <code>"rollback"</code>,
                         then this is a rollback. Set
--- a/webrtc.html
+++ b/webrtc.html
@@ -1375,24 +1375,24 @@
                         and <code><a data-for=
                         "RTCPeerConnection">pendingLocalDescription</a></code>
                         to null. Finally set <var>connection</var>'s
-                        <a>signaling state</a> to <code>"stable"</code></p>
+                        <a>signaling state</a> to <code>stable</code></p>
                       </li>
                       <!-- B) transition rollback to haveLocalOffer to stable -->
                       <li>
-                        <p>If <var>description</var> is of type <code>"rollback"</code>,
+                        <p>If <var>description</var> is of type <code>rollback</code>,
                         then this is a rollback. Set
                         <code><var>connection</var>.<a data-for=
                         "RTCPeerConnection">pendingLocalDescription</a></code>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1301,6 +1301,14 @@
                     <code>InvalidModificationError</code> and abort these steps.
                   </li>
                   <li>
+                    <p>If <var>description</var> is set as a local description,
+                    if <code><var>description</var>.type</code> is <code>"rollback"</code>
+                    and <a>signaling state</a> is not <code>"have-local-offer"</code>
```
jan-ivar: Maybe *"is neither "have-local-offer" nor "have-local-pranswer""* ?

I'm not a JSEP expert, but [section 4.1.8.2.](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-20#section-4.1.8.2) seems to imply rollback is allowed from any state but stable.

webrtc.html
```diff
@@ -1392,6 +1400,14 @@
                     <p>Otherwise, if <var>description</var> is set as a remote
                     description, then run one of the following steps:</p>
                     <ul>
+                      <li>
+                        <p>If <var>description</var> is set as a remote description,
+                        if <code><var>description</var>.type</code> is <code>"rollback"</code>
+                        and <a>signaling state</a> is not <code>"have-remote-offer"</code>
```
jan-ivar: *"is neither "have-remote-offer" nor "have-remote-pranswer""* ?

webrtc.html
```diff
@@ -1392,6 +1400,14 @@
                     <p>Otherwise, if <var>description</var> is set as a remote
                     description, then run one of the following steps:</p>
                     <ul>
+                      <li>
+                        <p>If <var>description</var> is set as a remote description,
```
soareschen: This should also be in step 2.2 where it covers _"If the process to apply description fails for any reason..."_.

taylor-b:
I agree with Jan-Ivar. This should be covered by "If the description's type is invalid for the current signaling state of connection, then reject p". If the issue is that JSEP isn't clear enough about what's invalid, then we should fix JSEP. Here's an issue (I made a PR too): https://github.com/rtcweb-wg/jsep/issues/760

fluffy:
See my comments in bug ... I'm not sure this is the behavior we want - I think we need to resolve that in JSEP - but if this is the behavior we decide on there, then of source we need something like this.

taylor-b:
The JSEP decision was "you can roll back PR-answers, and it makes no difference whether you call SLD or SRD". So this PR should be updated. The only invalid rollback is a rollback from a "stable" state.

stefhak:
LGTM.

PR #1402, opened 2017-06-16
Add/remove remote tracks from msid streams based on direction (rebased)

jan-ivar:
Rebase of https://github.com/w3c/webrtc-pc/pull/1377.
Fix for https://github.com/w3c/webrtc-pc/issues/1181.

***
[Preview](https://s3.amazonaws.com/pr-preview/jan-ivar/webrtc-pc/removeremote2.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/a12bf47...jan-ivar:f13a05e.html)

PR #1519, opened 2017-08-09
Ensure that "track" events fire before the SRD promise resolves.

### LINKED ISSUE #1508: setRemoteDescription and ontrack order of events: SRD before ontrack?
TL;DR: Should `setRemoteDescription` be resolved before the resulting `ontrack` fires for remote tracks or after in the same task? @jan-ivar @taylor-b 

`setRemoteDescription` does, in parallel to returning a promise, a number of things. On success this includes:
```
1. Process remote tracks.
   1a. Queue a task to fire an event named track.
2. Resolve the promise.
```
As described, the promise would be resolved before `ontrack` fires. Invoking `then()` of a resolved promise means the resolver is not executed immediately, but [at the end of the current run of the JavaScript event loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#Timing), which is still before any tasks queued for the next JavaScript event loop. We get this order of events:
```
1. Synchronous code after our setRemoteDescription() call is executed as part of
   the current loop.
2. resolve is called (setRemoteDescription(...).then(resolve)) at the end of
   the current loop.
3. ontrack fires in the next loop.
```

Question: Is this on purpose? setRemoteDescription happens before tracks are fired.

If this isn't on purpose it's not obvious how to make the ontrack events fire first. If we don't queue them but fire them immediately, they're fired before we've completed all steps of setRemoteDescription, so we may be in an incomplete state. If we queue a task to resolve the promise then it would be possible to run a task in-between calling setRemoteDescription having had its effect and before the promise being resolved, which might yield unexpected behavior when examining the pc.

taylor-b:
Fixes #1508.

This was the behavior before, but I accidentally broke it in PR #1136.

This ensures that if, for example, a description is applied with
audio/video tracks in the same stream, the order of events is:

1. "track" event with audio track, and stream that contains both tracks.
2. "track" event with video track, and stream that contains both tracks.
3. SRD promise resolves.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1508_srd_ontrack_order.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/4f0ef95...taylor-b:1a96e82.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5124,9 +5124,9 @@ interface RTCPeerConnectionIceErrorEvent : Event {
             list.</p>
           </li>
           <li>
-            <p>Queue a task, that if <var>track.muted</var> is
-            <code>false</code>, will <a>update the muted state</a>
-            of <var>track</var> with the value <code>true</code>.</p>
+            <p>If <var>track.muted</var> is <code>false</code>,
+            <a>update the muted state</a> of <var>track</var> with the value
+            <code>true</code>.</p>
           </li>
         </ol>
         <p>To <dfn id="set-associated-remote-streams">set the associated remote streams</dfn> given
--- a/webrtc.html
+++ b/webrtc.html
@@ -1542,83 +1542,99 @@
                   </li>
                   <li>
                     <p>If <var>description</var> is set as a remote description,
-                    then run the following steps for each media description in
-                    <var>description</var>:</p>
+                    then run the following steps:</p>
                     <ol>
                       <li>
-                        <p>Let <var>direction</var> be an <code>
-                        <a>RTCRtpTransceiverDirection</a></code> value
-                        representing the direction from the media
-                        description, but with the send and receive
-                        directions reversed to represent this peer's point
-                        of view.</p>
+                        <p>Let <var>trackEvents</var> be an empty list.</p>
                       </li>
                       <li>
-                        <p>As described by <span data-jsep=
-                        "applying-a-remote-desc">[[!JSEP]]</span>, attempt to
-                        find an existing <code><a>RTCRtpTransceiver</a></code>
-                        object, <var>transceiver</var>, to represent the media
-                        description.</p>
-                      </li>
-                      <li>
-                        <p> If no suitable transceiver is found
-                        (<var>transceiver</var> is unset), run the following
-                        steps:</p>
+                        <p>Run the following steps for each media description
+                        in <var>description</var>:</p>
                         <ol>
                           <li>
-                            <p><a>Create an RTCRtpSender</a>, <var>sender</var>,
-                            from the media description.</p>
+                            <p>Let <var>direction</var> be an <code>
+                            <a>RTCRtpTransceiverDirection</a></code> value
+                            representing the direction from the media
+                            description, but with the send and receive
+                            directions reversed to represent this peer's point
+                            of view.</p>
                           </li>
                           <li>
-                            <p><a>Create an RTCRtpReceiver</a>,
-                            <var>receiver</var>, from the media description.</p>
+                            <p>As described by <span data-jsep=
+                            "applying-a-remote-desc">[[!JSEP]]</span>, attempt to
+                            find an existing <code><a>RTCRtpTransceiver</a></code>
+                            object, <var>transceiver</var>, to represent the media
+                            description.</p>
                           </li>
                           <li>
-                            <p><a>Create an RTCRtpTransceiver</a> with
-                            <var>sender</var>, <var>receiver</var> and
-                            <var>direction</var> set to <code>"recvonly"</code>,
-                            and let <var>transceiver</var> be the result.</p>
+                            <p> If no suitable transceiver is found
+                            (<var>transceiver</var> is unset), run the following
+                            steps:</p>
+                            <ol>
+                              <li>
+                                <p><a>Create an RTCRtpSender</a>, <var>sender</var>,
+                                from the media description.</p>
+                              </li>
+                              <li>
+                                <p><a>Create an RTCRtpReceiver</a>,
+                                <var>receiver</var>, from the media description.</p>
+                              </li>
+                              <li>
+                                <p><a>Create an RTCRtpTransceiver</a> with
+                                <var>sender</var>, <var>receiver</var> and
+                                <var>direction</var> set to <code>"recvonly"</code>,
+                                and let <var>transceiver</var> be the result.</p>
+                              </li>
+                            </ol>
+                          </li>
+                          <li>
+                            <p>Set <var>transceiver</var>'s <code><a data-for=
+                            "RTCRtpTransceiver">mid</a></code> value to the mid of
+                            the corresponding media description. If the media
+                            description has no MID, and <var>transceiver</var>'s
+                            <code><a data-for="RTCRtpTransceiver">mid</a></code>
+                            is unset, generate a random value as
+                            described in <span data-jsep=
+                            "applying-a-remote-desc">[[!JSEP]]</span>.</p>
+                          </li>
+                          <li>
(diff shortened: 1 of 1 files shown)


aboba:
@soareschen Can you take a look?

soareschen:
Will look at this beginning of next week.

webrtc.html
```diff
@@ -1526,83 +1526,99 @@
...
+                            and <var>trackEvents</var>.</p>
+                          </li>
+                          <li>
+                            <p>If <var>direction</var> is
+                            <code>sendonly</code> or <code>inactive</code>, and
+                            <var>transceiver</var>'s
+                            <a>[[\CurrentDirection]]</a> slot
+                            is neither "sendonly" nor "inactive",
+                            <a>process the removal of a remote track</a> for
+                            the media description, given <var>transceiver</var>.</p>
```
henbos: Here or in a separate issue/PR, do we want to do something similar for the removal of tracks and updating their muted state (which fires mute on track)? Currently it still says to "_Queue a task_, that if track.muted is false, will update the muted state of track with the value true." I suspect it should be "Add the track to tracksMuted" and before/after "for each trackEvents" we do "for each tracksMuted, if track.muted is false, update the muted state to true".

webrtc.html
```diff
@@ -1526,83 +1526,99 @@
...
+                          </li>
+                          <li>
+                            <p>If <var>direction</var> is <code>sendrecv</code> or
+                            <code>recvonly</code>, and
+                            <var>transceiver</var>'s
+                            <a>[[\CurrentDirection]]</a> slot
+                            is neither "sendrecv" nor "recvonly",
+                            <a>process the addition of a remote track</a> for
+                            the media description, given <var>transceiver</var>
+                            and <var>trackEvents</var>.</p>
```
soareschen: In the 20170605 editor's draft, there was a guard for _"if transceiver.receiver.track has not yet been fired in a track event"_. Here it seems to suggest that if `CurrentDirection` is `null`, `sendonly`, or `inactive`, and not changed, the same track event can be fired multiple times. e.g. calling `setRemoteDescription(offer)` twice.

webrtc.html
taylor-b: The reason for adding the track events to a list and processing it later was so the streams would be updated with the full list of tracks by the time the event fires. But the "muted" event doesn't include the stream, so this isn't necessary. So I think we could just remove the "queue a task" part.

@jan-ivar, any thoughts? When do you think the track "muted" event should fire relative to the SRD promise resolving, and other events?

webrtc.html
taylor-b: I think this was a consequence of https://github.com/w3c/webrtc-pc/pull/1402 and was intentional. If a track is added, removed, and re-added, you'll get a "track" event, a "muted" event (and "removetrack" event on the stream), then another "track" event, so every action has a corresponding event.

soareschen: I think the problem here is that each call to `setRemoteDescription` always fire track event for every media description that is sendrecv or recvonly, regardless of whether the track is existing one, new, or readded. Is that the intended new behavior?

Anyway I see that the issue already exist before the PR, so we can merge this first and raise a separate issue for that.

taylor-b: That won't happen, because there's a check for whether `[[CurrentDirection]]` is `"sendrecv"` or "`recvonly`". Meaning the event will only fire if the remove peer is offering to send something, and nothing is currently being received.

soareschen: Ah I see! I guess I missed it because the "sendrecv" and "recvonly" for `[[CurrentDirection]]` are not formatted as `<code>`. Can you fix the formatting?

webrtc.html
```diff
@@ -1526,83 +1526,99 @@
...
+                          </li>
+                          <li>
+                            <p>If <var>direction</var> is
+                            <code>sendonly</code> or <code>inactive</code>, and
+                            <var>transceiver</var>'s
+                            <a>[[\CurrentDirection]]</a> slot
+                            is neither "sendonly" nor "inactive",
```
soareschen: Here the "sendonly" nor "inactive" should be formatted as `<code>` as well.

webrtc.html

webrtc.html
taylor-b: Also done

webrtc.html
```diff
@@ -5624,10 +5641,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
-            <p>Queue a task to fire an event named <code title=
-            "event-track"><a>track</a></code> with <var>transceiver</var>,
-            <var>track</var>, and <var>streams</var> at the
-            <var>connection</var> object.</p>
+            <p>Add a new <code><a>RTCTrackEvent</a></code> with
+            <var>transceiver</var>, <var>track</var>, and
+            <var>streams</var> to <var>trackEvents</var>.</p>
           </li>
         </ol>
         <p>To <dfn id="process-remote-track-removal">
```
jan-ivar: In reply to @taylor-b, yes just removing the "queue a task" part from the [step below here](http://w3c.github.io/webrtc-pc/#process-remote-track-removal) that [updates the muted state](https://w3c.github.io/mediacapture-main/getusermedia.html#update-track-muted), seems reasonable to me. Especially since that step in media-capture queues another task itself! That looks like a mistake.

It would even have been nice for consistency perhaps if all `muted` events fired right before all new `track` events, if any, and on the same task as those, but I can't think of a real reason, so that may or may not warrant a change to mediacapture.

That said, the webrtc version we'd remove seems more careful about not firing spurious events, checking in the queued task itself if the state is already in the new state to void firing twice. Perhaps media-capture should do the same at least? 

As a general comment on media-capture, the event-firing algorithms it exposes seem to assume they're being called off-main-thread, or that queuing a separate task for each event firing is desirable. Here it seems we are instead preferring to do multiple events on the same task, to convey a larger set of changes happening as part of the same cause.

Would it have been more flexible if media-capture exposed algorithms that didn't queue a task as their first step, and instead left queuing up to callers of those algorithms? @domenic do you know what common practice is here, and if there is any general spec design advice on this?

domenic: I think it's generally the same as most other software development. If you have a use case for the non-queueing versions, then yeah, keep that outside. Otherwise, to deduplicate, feel free to move it inside the algorithm.

To be explicit, if you have

```js
function caller1() {
  // ...
  wrapperOperation(doStuff);
  // ...
}

function caller2() {
  // ...
  wrapperOperation(doStuff);
  // ...
}
```

then maybe you should refactor to a single function `doWrapperOperationAndDueStuff()` and call that instead. But if you have a `caller3` that directly calls `doStuff()`, then don't refactor.

taylor-b: This algorithm isn't actually invoked by anything in mediacapture-main other than this prose:

> There can be several reasons for a MediaStreamTrack to be muted: the user pushing a physical mute button on the microphone, the user toggling a control in the operating system, the user clicking a mute button in the browser chrome, the User Agent (on behalf of the user) mutes, etc.

Is there any precedent for saying something like "run the following steps, queuing a task to do so if not already running from the relevant event loop"?

jan-ivar: That, or maybe something like:

"If either of these occur, the User Agent MUST queue a task to **update a track's muted state** to *newState*, using the [user interaction task source](http://w3c.github.io/html/webappapis.html#user-interaction-task-source).

To **update a track's muted state** to *newState*, run the following steps:" ?

This split would let webrtc specify its own task source.

taylor-b: Filed a new issue for the "mute" ordering specifically: https://github.com/w3c/webrtc-pc/issues/1568

taylor-b:
Rebased and ended up with this gem:

> For each RTCTrackEvent trackEvent in trackEvents, fire a track event named track with trackEvent...

PR #1522, opened 2017-08-09
Formalize how createOffer interacts with identity providers.

### LINKED ISSUE #1184: Race condition between createOffer and setIdentityProvider
Suppose an application does:

```
pc.setIdentityProvider(a);
pc.createOffer().then(...);
// Some time later, but before createOffer's promise has resolved:
pc.setIdentityProvider(b);
```

Is there a guarantee which identity provider will be used to generate "a=identity"?

Note that I already tried to fix the `createOffer` race conditions here: https://github.com/w3c/webrtc-pc/pull/875

But this one may still exist.

taylor-b:
Fixes #1184.

This cleans things up slightly, and fixes a corner case race condition,
where in the following sequence:

```
pc.setIdentityProvider(a);
pc.createOffer().then(...);
pc.setIdentityProvider(b);
pc.createOffer().then(...);
```

It wasn't completely clear if the first `createOffer` would use the
assertion from `a` or `b`. The answer now is "it will use `a`".

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1184_setidentityprovider_race.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/48532fb...taylor-b:1e6ed2a.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2178,13 +2178,9 @@ interface RTCPeerConnection : EventTarget  {
                     <code>InvalidStateError</code>.</p>
                   </li>
                   <li>
-                    <p>Let <var>provider</var> be <var>connection</var>'s
-                    currently configured identity provider if one has been
-                    configured, or <code>null</code> otherwise.</p>
-                  </li>
-                  <li>
-                    <p>If <var>provider</var> is non-null, then begin <a href=
-                    "#sec.identity-proxy-assertion-request"> the identity
+                    <p>If <var>connection</var> is configured with an identity
+                    provider, then begin <a href=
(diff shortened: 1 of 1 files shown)


jan-ivar:
> if the first createOffer would use the assertion from a or b. The answer now is "it will use a".

@taylor-b How come you chose the opposite behavior from `addTrack` and everything else?

taylor-b:
Good point. I can fix that, for consistency if for no other reason.

taylor-b:
@aboba, can you take one more look and merge if there aren't any issues?

PR #1530, opened 2017-08-09
Adding defaults for RTCRtpEncodingParameters.active and priority

### LINKED ISSUE #1491: Default values for RTCRtpEncodingParameters
The current definition seems to imply that all fields are optional and have the value `undefined` if not set. However it seems that at least some of the fields should have default values, e.g. `active` and `priority`.

For example, the following examples have different `active` values:

```javascript
// sender.encodings[0].active == true
pc.addTransceiver('audio');

// sender.encodings[0].active == true
pc.addTransceiver('audio', { sendEncodings: [] });

// sender.encodings[0].active == undefined
pc.addTransceiver('audio', { sendEncodings: [{}] });
```

taylor-b:
Fixes #1491.

The rest of the encoding parameters (`ptime`, `maxBitrate`,
`maxFramerate`, etc.) are optional, but `active` and `priority` need
default values in case they aren't specified in the `sendEncodings`
passed into `addTransceiver`.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1491_rtpencodingparameters_defaults.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/226d9af...taylor-b:453ba4f.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6289,8 +6289,8 @@ sender.setParameters(params)
              RTCRtpRtxParameters rtx;
              RTCRtpFecParameters fec;
              RTCDtxStatus        dtx;
-             boolean             active;
-             RTCPriorityType     priority;
+             boolean             active = true;
+             RTCPriorityType     priority = "low";
              unsigned long       ptime;
              unsigned long       maxBitrate;
              double              maxFramerate;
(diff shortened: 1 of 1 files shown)


taylor-b:
I used "low" as the default priority, since this is the default for data channels.

Issue #1499, opened 2017-07-20
Processing Remote MediaStreamTracks should pass on receiver when constructing track event


soareschen:
In the last step of 5.1.1 Processing Remote MediaStreamTracks:

> Queue a task to fire an event named track with transceiver, track, and streams at the connection object.

It missed out the receiver attribute in ` RTCTrackEvent`, which we know is simply getting from `transceiver.receiver`.

stefhak:
I think this is on purpose, as you say the app can easily get hold of the receiver via the Transceiver.

jan-ivar:
```js
pc.ontrack = e => e.tranceiver.receiver;
```

soareschen:
What I mean is the WebIDL definition for `RTCTrackEvent` has a `receiver` field. So that attribute should either be removed or filled with `transceiver.receiver`, no?

jan-ivar:
@soareschen Yeah that seems redundant to me. We should remove both the `receiver` readonly attribute and the corresponding member in `RTCTrackEventInit` IMHO.

jan-ivar:
Since Firefox already supports `e.receiver`, we'd continue to for a while for legacy reasons.

stefhak:
@soareschen got it (https://github.com/w3c/webrtc-pc/issues/1499#issuecomment-319687209). Could you provide a PR?

jan-ivar:
PR up.

soareschen:
Thanks @jan-ivar!

@stefhak sorry I am rushing on another project, didn't get the time to work on it.

jan-ivar:
Doh, I misread @stefhak's message as coming from @soareschen 🙂

jan-ivar:
I've closed https://github.com/w3c/webrtc-pc/pull/1532.

@soareschen can you add a PR to fill in `event.receiver` with `transceiver.receiver` instead?

stefhak:
Ping @soareschen !

stefhak:
This has been a while now. @jan-ivar, what would be the gain of 
> fill in `event.receiver` with `transceiver.receiver` instead
?

It seems to me we can leave things as they are

PR #1532, opened 2017-08-10
Remove event.receiver.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1499.

***
[Preview](https://s3.amazonaws.com/pr-preview/jan-ivar/webrtc-pc/noreceiver.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/cdf9e75...jan-ivar:f29a0b1.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8279,14 +8279,14 @@ sender.setParameters(params)
       <code><a>RTCTrackEvent</a></code> interface.</p>
       <p><dfn id="fire-track-event" data-lt="fire track event">Firing an
       RTCTrackEvent event named <var>e</var></dfn> with an
-      <code><a>RTCRtpReceiver</a></code> <var>receiver</var>, a
+      <code><a>RTCRtpTransceiver</a></code> <var>transceiver</var>, a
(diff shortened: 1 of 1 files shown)


soareschen:
So does this mean that the compat issue is resolved and we can safely remove event.receiver?

jan-ivar:
I don't know about safely. There are definitely implementations of `event.receiver` today. I've added a polyfill to adapter in https://github.com/webrtc/adapter/pull/647 FWIW.

We should probably ask for consensus before merging this.

jan-ivar:
I think in hindsight I'm withdrawing this PR. More browsers than I realized already support this, i.e. this works in Safari:
```js
pc.ontrack = event => {
  console.log(event.track);
  console.log(event.receiver.track);
  console.log(event.transceiver.receiver.track);
  video.srcObject = event.streams[0];
};
```
The redundancy extends to `track` as well, a can of worms best left unopened, and the value of removing any of these seem marginal, whereas the compat hit substantial.

Issue #1293, opened 2017-05-31
Section 9.6: protocol type


aboba:
From EKR: https://lists.w3.org/Archives/Public/public-webrtc/2017May/0166.html

Is there any valid reason to allow the protocol to be a DOMString as
opposed to ASCII or LDH? It's about to be shoved into a URL...

aboba:
In terms of WebIDL types, we only have DOMString, ByteString and USVString.

aboba:
[RFC 3986](https://tools.ietf.org/html/rfc3986) defines the syntax.  As noted in Section 2.3, unreserved characters include ALPHA / DIGIT / "-" / "." / "_" / "~"

Issue #1500, opened 2017-07-21
setIdentityProvider should throw error if protocol contains invalid characters


soareschen:
Description for `protocol` field in `RTCIdentityProviderOptions`:

> The name of the protocol that is used by the identity provider. This MUST NOT include '/' (U+002F) or '\' (U+005C) characters. This value defaults to "default" if not provided.

Right now there is no step for validating `protocol` in `setIdentityProvider()`. It should instead throw specific kind of error. (`InvalidAccessError` perhaps?)

martinthomson:
Not a question, just a bug.

aboba:
@soareschen Can you submit a PR?

PR #1538, opened 2017-08-17
protocol can only contain characters legal in a URI

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1293

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/issue-1293-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/945b618...d863e68.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9388,7 +9388,8 @@ interface RTCIdentityProviderRegistrar {
               "idlMemberType"><a>DOMString</a></span>, defaulting to
               <code>"default"</code></dt>
               <dd>
-                <p>The protocol parameter used for the IdP.</p>
+                <p>The protocol parameter used for the IdP. This attribute MUST
+                contain only ASCII characters.</p>
               </dd>
             </dl>
           </section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -9389,7 +9389,7 @@ interface RTCIdentityProviderRegistrar {
               <code>"default"</code></dt>
               <dd>
                 <p>The protocol parameter used for the IdP. This attribute MUST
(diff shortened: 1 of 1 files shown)


soareschen:
Related to #1500: Not just any ASCII character is allowed.

aboba:
@soareschen According to [RFC 3986 Section 2.3](https://tools.ietf.org/html/rfc3986#section-2.3), the unreserved characters are: ALPHA / DIGIT / "-" / "." / "_" / "~"

soareschen:
In ietf-security-arch:

> The path, starting with "/.well-known/idp-proxy/" and appended
with the IdP protocol.  Note that the separator characters '/'
(%2F) and '\' (%5C) MUST NOT be permitted in the protocol field,
lest an attacker be able to direct requests outside of the
controlled "/.well-known/" prefix.  Query and fragment values MAY
be used by including '?' or '#' characters.

This makes me realize that this PR is dealing with `protocol` in `RTCIdentityProviderDetails`, while #1500 mentions `protocol` in `RTCIdentityProviderOptions`.

`RTCIdentityProviderDetails` is returned from the IdP proxy, which means there is an issue of whether we want to validate both `domain` and `protocol` returned are well formed, and make methods that invoke "Requesting Identity Assertions" fail if they are not. 

That being said, an attacker can serialize arbitrary assertion string into the SDP and pass it to a remote peer. In such case it should result in failure in lower level libraries and cause `setRemoteDescription` to fail. So it probably doesn't matter much if the local peer doesn't catch the error early.

PR #1544, opened 2017-08-19
s/outboundrtp/outbound-rtp

fippo:
stat type names are using hyphens now.

@jan-ivar is the example still correct with your recent "remote" change to -stats?

***
[Preview](https://s3.amazonaws.com/pr-preview/fippo/webrtc-pc/outbound-fix.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/945b618...fippo:1eb447f.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9081,7 +9081,7 @@ pc.getStats(selector).then(function (report) {
 function processStats() {
     // compare the elements from the current report with the baseline
     currentReport.forEach (now =&gt; {
-        if (now.type != "outboundrtp")
+        if (now.type != "outbound-rtp")
             return;
 
         // get the corresponding stats from the baseline report


jan-ivar:
@fippo It is ***only*** correct with my recent "remote" changes 🙂.

But even with your fix (thanks) this example still lags [the one in stats](https://w3c.github.io/webrtc-stats/#example-of-a-stats-application).

Which begs the question, why maintain two? I suggest we nuke this, and link to the example in stats.

@aboba Thoughts?

jan-ivar:
See also https://github.com/w3c/webrtc-stats/issues/236.

vr000m:
AFAICT, the original decision for stats (at TPAC15) was to keep the definition, example, and a MTI list in webrtc-pc. IMO, implementers will need to read webrtc-stats anyway, hence keeping a very simple example in webrtc-pc. And webrtc-stats can provide more examples, simple and complex, etc.

jan-ivar:
@vr000m Ok. I vaguely remember that now, thanks.

stefhak:
lgtm.

PR #1548, opened 2017-08-23
Add note that IdP must treat contents as opaque

### LINKED ISSUE #1505: Is contents validation based on exact string matching or fingerprint list matching when verifying identity assertion?
There are a few places that mention that `contents` returned by `validateAssertion` must be exact string match of `contents` passed to `generateAssertion`:

> Callback GenerateAssertionCallback Parameters
> contents of type DOMString
>   The contents parameter includes the information that the user agent wants covered by the identity assertion. A successful validation of the provided assertion MUST produce this string.

> Dictionary RTCIdentityValidationResult Members
> contents of type DOMString, required
>    The payload of the identity assertion. An IdP that validates an identity assertion MUST return the same string that was provided to the original IdP that generated the assertion.
>    The user agent uses the contents string to determine if the identity assertion matches the session description.

And in rtcweb-security-arch:
> 5.7.  Verifying Assertions
> contents:  The original unmodified string provided by the AP as input to the assertion generation process.

However in section 9.4 Verifying Identity Assertions, `contents` is validated by only decoding it and match against all a=fingerprint attributes.

> 9.4 Verifying Identity Assertions
> 8. The RTCPeerConnection decodes the contents and validates that it contains a fingerprint value for every a=fingerprint attribute in the session description. This ensures that the certificate used by the remote peer for communications is covered by the identity assertion.

This means it doesn't need to be exact, e.g. there can be extra fingerprints in `contents` or fingerprints can be reordered.

Other than that, there are issues on enforcing either of the requirements.

For exact string matching, this is not possible because the original `contents` is only known by the remote peer, and there is no additional metadata in the SDP that mention what the original `content` must be. We cannot rely on the a=identity line because the assertion string returned from the IdP is opaque, and may not contain the original `contents` string, e.g. it can be stored on server side.

~For fingerprint list matching, `contents` is supposed to be an opaque string and rtcweb-security-arch doesn't specify what should be the format of `contents`. In other words the format could be implementer specific and thus could introduce interoperability issues.~ (Update: format for `contents is specified in 5.6.4.)

soareschen:
Fixes #1505.

This adds note that although `RTCPeerConnection` do not perform string based comparison with the original `contents`, the IdP MUST treat `contents` as opaque and return the same string to conform with rtcweb-security-arch.

***
[Preview](https://s3.amazonaws.com/pr-preview/soareschen/webrtc-pc/issue-1505-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/f798246...soareschen:89aeed6.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9296,9 +9296,10 @@ interface RTCIdentityProviderRegistrar {
               <dt><code>contents</code> of type <span class=
               "idlMemberType"><a>DOMString</a></span></dt>
               <dd>The <var>contents</var> parameter includes the information
-              that the user agent wants covered by the identity assertion. A
-              successful validation of the provided assertion MUST produce this
-              string.</dd>
+              that the user agent wants covered by the identity assertion. The
+              IdP MUST treat <code>contents</code> as opaque string. A
+              successful validation of the provided assertion MUST produce the
+              same string.</dd>
               <dt><code>origin</code> of type <span class=
               "idlMemberType"><a>DOMString</a></span></dt>
               <dd>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -9605,6 +9606,14 @@ <h3 id="sec.identity-verify-assertion">Verifying Identity Assertions</h3>
           communicate with peers that offer a certificate that doesn't match an
           <code>a=fingerprint</code> line in the negotiated session
           description.</p>
+          <p class="note">The user agent decodes <a data-for=
+          "RTCIdentityValidationResult"></code>contents</code></a> using
+          the format described in [[!RTCWEB-SECURITY-ARCH]] and do not perform
+          string-based comparison with the original <a
```
martinthomson: I don't understand why you need to say "do not perform string-based comparison with the original".  Isn't it sufficient to say that the contents are opaque and should be treated as such?  We're just asking the IdP not to touch this nicely, which should be sufficient.

That the IdP could tamper with this because we have no inbuilt mechanism to detect that isn't that relevant to a specification.  The consequence of tampering is that it later breaks something if the browser-browser protocol changes.

soareschen: Makes sense. I have removed that part of the sentence.

burnburn:
@martinthomson can you review again and approve if okay?

PR #1549, opened 2017-08-23
Allow createAnswer to be called only in valid signaling state

### LINKED ISSUE #1183: createAnswer does not guard against invalid remoteDescription or signaling state
The current spec for `createAnswer` have the statement:

> If connection's remoteDescription is null return a promise rejected with a newly created InvalidStateError.

However `remoteDescription` can have the type `answer` on the offerer side and this statement allows it to pass through.

The statement also allows `createAnswer` to still be called on the anwerer side after `setLocalDescription` is called and the signaling enters `stable` state. Unlike `createOffer`, repeated call to `createAnswer` cannot be used again with `setLocalDescription` because doing so is invalid in `stable` state.

The statement should be updated to check for the type and signaling state as well.

soareschen:
Fixes #1183.

This makes createAnswer throw `InvalidStateError` if the signaling state is not `have-remote-offer` or `have-local-pranswer`. Checking signaling state also implicitly assert that `pendingRemoteDescription` is set, so that checking is removed.

There is a breaking change here that existing implementations allow createAnswer to succeed after a call to `setLocalDescription(answer)`. Nevertheless the resulting new answer cannot be applied because the connection have transition  back to stable state. If anyone think it is important not to break this behavior, I am happy to modify the patch to only throw error if `pendingRemoteDescription` is not of type `offer`.

***
[Preview](https://s3.amazonaws.com/pr-preview/soareschen/webrtc-pc/issue-1183-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/f798246...soareschen:6137d09.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2330,11 +2330,12 @@ interface RTCPeerConnection : EventTarget  {
                     steps to <var>connection</var>'s operation queue:</p>
                     <ol>
                       <li>
-                        <p>If <var>connection</var>'s <code><a data-for=
-                        "RTCPeerConnection">remoteDescription</a></code> is
-                        <code>null</code> return a promise rejected with a newly
-                        <a data-link-for="exception" data-lt="create">created
-                        </a> <code>InvalidStateError</code>.</p>
+                        <p>If <var>connection</var>'s <a>signaling stated</a>
+                        is not <code>"have-remote-offer"</code> or
+                        <code>"have-local-pranswer"</code>, return a promise
+                        rejected with a newly <a data-link-for="exception"
+                        data-lt="create">created</a>
+                        <code>InvalidStateError</code>.</p>
                       </li>
                       <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2330,11 +2330,12 @@
                     steps to <var>connection</var>'s operation queue:</p>
                     <ol>
                       <li>
-                        <p>If <var>connection</var>'s <code><a data-for=
-                        "RTCPeerConnection">remoteDescription</a></code> is
-                        <code>null</code> return a promise rejected with a newly
-                        <a data-link-for="exception" data-lt="create">created
-                        </a> <code>InvalidStateError</code>.</p>
+                        <p>If <var>connection</var>'s <a>signaling stated</a>
```
jan-ivar: s/stated/state/

soareschen:
Fixed those now.

stefhak:
lgtm.

PR #1551, opened 2017-08-23
Clarify the meaning of session description sdp need not match

### LINKED ISSUE #1228: Confusion on currentRemoteDescription.sdp need not match remoteDescription.sdp
I was reading #921 and found #1004 that added:

> Note that the value of currentRemoteDescription.sdp need not match the value of remoteDescription.sdp.

When I read the sentence I interpreted it as they may be different because localDescription may point to pendingLocalDescription instead. Didn't expect that to be related to #921.

soareschen:
Fixes #1228.

***
[Preview](https://s3.amazonaws.com/pr-preview/soareschen/webrtc-pc/issue-1228-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/f798246...soareschen:55d8546.html)


soareschen:
AFAIK ICE candidates are also added to the local/remote descriptions after setting. For instance local ICE candidates are gathered and added to `localDescription` only after `setLocalDescription` is called.

I have added another sentence to clarify both cases.

Issue #1259, opened 2017-05-29
How to find what keyGenAlgorithm values the browser supports?


fluffy:
Is there a way to do this with other WebCryptoAPI? Should we have a note pointing that out here ?

aboba:
@fluffy If getAlgorithm() could be applied to the built-in certificates, you could at least know what the default algorithm was. But as it is, I think the answer is "no".

aboba:
This was discussed at the virtual interim on June 28.  Way forward is to develop a PR to return supported algorithm values.

aboba:
From: https://www.w3.org/2017/06/28-webrtc-minutes

Bernard to create a PR based on Tim's suggestion

PR #1553, opened 2017-08-23
What algorithms the browser supports

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1259

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/issue-1259-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/f798246...23dc398.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4309,8 +4309,8 @@ interface RTCPeerConnectionIceErrorEvent : Event {
             class="methods">
               <dt><dfn><code>getSupportedAlgorithms</code></dfn></dt>
               <dd>
-                <p>Returns a sequence of supported and recommended certificate algorithms.
-                At least one algorithm MUST be returned.</p>
+                <p>Returns a sequence providing a representative set of supported
+                certificate algorithms. At least one algorithm MUST be returned.</p>
               </dd>
               <dt><dfn><code>getFingerprints</code></dfn></dt>
               <dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -4309,7 +4309,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
             class="methods">
               <dt><dfn><code>getAlgorithms</code></dfn></dt>
               <dd>
-                <p>Returns a sequence of recommended certificate algorithms.</p>
+                <p>Returns a sequence of supported certificate algorithms.</p>
               </dd>
               <dt><dfn><code>getFingerprints</code></dfn></dt>
               <dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -4283,6 +4283,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
         <div>
           <pre class="idl">interface RTCCertificate {
     readonly        attribute DOMTimeStamp expires;
+    static sequence&lt;AlgorithmIdentifier&gt; getAlgorithms();
     sequence&lt;RTCDtlsFingerprint&gt; getFingerprints ();
 };</pre>
           <section>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4306,6 +4307,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             <h2>Methods</h2>
             <dl data-link-for="RTCCertificate" data-dfn-for="RTCCertificate"
             class="methods">
+              <dt><dfn><code>getAlgorithms</code></dfn></dt>
+              <dd>
+                <p>Returns a sequence of recommended certificate algorithms.</p>
```
fluffy: I think this would be a bit better if you said "certificate fingerprint algorithms" instead of "certificate algorithms".

fluffy: On the "recommended", it seems that it might be better to say "supported" or "supported and recommended". Lets say we had some crappy hash that was no longer recommended by standards, but the browser still supported it. Would we return it? I don't care too much how we go on this but clearly we can't return things which are recommended by specs but not supported by browser.

martinthomson: This isn't fingerprints, I assume that it is signature algorithms.

fluffy: Ah, I see, let me retract fingerprint and double down on my point that I think we need to be clear what it is :-)

webrtc.html
```diff
@@ -4283,6 +4283,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
         <div>
           <pre class="idl">interface RTCCertificate {
     readonly        attribute DOMTimeStamp expires;
+    static sequence&lt;AlgorithmIdentifier&gt; getAlgorithms();
```
pthatcherg: I'd prefer getSupportedAlgorithms()

webrtc.html
```diff
@@ -4306,6 +4307,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             <h2>Methods</h2>
             <dl data-link-for="RTCCertificate" data-dfn-for="RTCCertificate"
             class="methods">
+              <dt><dfn><code>getAlgorithms</code></dfn></dt>
+              <dd>
+                <p>Returns a sequence of supported certificate algorithms.</p>
+              </dd>
```
pthatcherg: If a implementation returns an empty sequence, is that compliant with this?

webrtc.html

webrtc.html
aboba: Added a requirement that at least one supported algorithm be returned.

ekr:
I think this should probably say something about this being a representative set.

For the specific case of RSA, I think we would pick one or two plausible key sizes.

webrtc.html
```diff
@@ -4306,6 +4307,11 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
             <h2>Methods</h2>
             <dl data-link-for="RTCCertificate" data-dfn-for="RTCCertificate"
             class="methods">
+              <dt><dfn><code>getSupportedAlgorithms</code></dfn></dt>
+              <dd>
+                <p>Returns a sequence of supported and recommended certificate algorithms.
```
taylor-b: I'm not sure "recommended" is the right term. It would be reasonable for an implementation to return an algorithm that uses a SHA-1 hash, if it's supported, but it wouldn't be recommended. So I'd suggest saying something like "Returns a representative set of supported certificate algorithms."

It would also help to have a note section that clarifies further. Here's a suggestion:

> For example, the `"RSASSA-PKCS1-v1_5"` algorithm dictionary, `RsaHashedKeyGenParams`, contains fields for the modulus length, public exponent, and hash algorithm. Implementations are likely to support a wide range of modulus lengths and exponents, but a finite number of hash algorithms. So in this case, it would be reasonable for the implementation to return one `AlgorithmIdentifier` for each supported hash algorithm that can be used with RSA, using default/recommended values for `modulusLength` and `publicExponent` (like 1024 and 65537, respectively).

PR #1558, opened 2017-08-24
Throw if a DataChannel, negotiated by the script, lacks id

### LINKED ISSUE #1529: What happens if RTCDataChannelInit.negotiated is true, but id is absent?
This is the inverse of #1413: an application negotiating the data channel out-of-band, but _not_ providing an ID. This is guaranteed to not work, since one side would select an even ID and one side would select an odd ID, and they wouldn't match up. So unless I'm missing something, I think we should throw an error when this is attempted.

adam-be:
Fix #1529

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/dc-negotiate-no-id.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/2d424aa...642597a.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7672,14 +7672,14 @@ interface RTCTrackEvent : Event {
                   </div>
                 </li>
                 <li>
-                  <p>Let <var>channel</var> have an
-                  <dfn>[[\DataChannelPriority]]</dfn> internal slot initialized
-                  to <var>option</var>'s <code>priority</code> member.</p>
-                </li>
-                <li>
                   <p>If <a>[[\Negotiated]]</a> is <code>true</code> and
                   <a>[[\DataChannelId]]</a> is <code>null</code>, <a>throw</a>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7677,6 +7677,11 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                   to <var>option</var>'s <code>priority</code> member.</p>
                 </li>
                 <li>
+                  <p>If <a>[[\Negotiated]]</a> is <code>true</code> and
+                  <a>[[\DataChannelId]]</a> is <code>null</code>, <a>throw</a>
+                  a <code>TypeError</code>.</p>
+                </li>
+                <li>
```
taylor-b: nit: Doesn't really make a difference, but for organizational purposes it would make sense to move this up one step, so it's right after `[[DataChannelId]]` being set.

PR #1559, opened 2017-08-24
Reference rtcweb-data-channel for RTCPriorityType enum.

### LINKED ISSUE #1439: RTCPriorityType should reference rtcweb-data-channel
As described by rtcweb-qos, a single enum value is used both for the SCTP level priority and for the DSCP markings used for the whole SCTP transport. So the enum descriptions should reference both the relevant IETF standards, and also clarify that "very-low" maps to "below normal", etc.

taylor-b:
Fixes #1439.

This enum is used to determine both the default DSCP marking and the
SCTP priority, if used for a data channel. So it needs to reference both
the relevant standards, and clarify that "very-low" corresponds to
"below normal", and so on.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1439_rtcweb_data_channel_ref.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/7d0e7ab...taylor-b:64f7ffc.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4118,19 +4118,23 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               </tr>
               <tr>
                 <td><dfn><code>very-low</code></dfn></td>
-                <td>See [[!RTCWEB-TRANSPORT]], Section 4.</td>
+                <td>See [[!RTCWEB-TRANSPORT]], Section 4. Corresponds to "below
+                normal" as defined in [[!RTCWEB-DATA]].</td>
               </tr>
               <tr>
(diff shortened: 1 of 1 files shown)


stefhak:
Would we need to discuss, or is this everyone's understanding (i.e. that very_low corresponds to below_normal and so on?

taylor-b:
We already decided on the mapping in the other direction (https://github.com/w3c/webrtc-pc/pull/1523), so I think this is even less controversial. Both enums have 4 values, I believe we just somehow didn't end up with matching names in the two IETF documents.

PR #1560, opened 2017-08-24
setParameters: Do argument checks in sync section and specify parellel steps

### LINKED ISSUE #1520: setParameters does input validation in parallel, and doesn't actually apply parameters?
The only steps listed in the `setParameters` algorithm do input validation. It's not clear why these steps are done in parallel, especially where they interact with `getParameters` (which is synchronous), which introduces race conditions.

If the reason we made `setParameters` promise-based was because it may involve I/O operations that can fail, then a more sensible algorithm to me would be:

1. Do input validation synchronously.
2. In parallel, attempt to start using parameters, resolving or rejecting promise appropriately.

adam-be:
First stab, do not merge yet. We need to figure out what goes into the parallel steps.

fix #1520

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/setparams-sync-part.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/eb517f2...9bc2856.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5413,7 +5413,8 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                 <code>RangeError</code>.</li>
                 <li>Let <var>p</var> be a new promise.</li>
                 <li>In parallel, configure the media stack to use
-                <var>parameters</var> to transmit <var>sender</var>.
+                <var>parameters</var> to transmit <var>sender</var>'s
+                <a>[[\SenderTrack]]</a>.
                   <ol>
                     <li>If the media stack is successfully configured with
                     <var>parameters</var>, queue a task to run the following
@@ -5431,8 +5432,11 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                         <li>If an error occurred due to hardware resources not
                         being available, reject <var>p</var> with a newly
                         created <code><a>RTCError</a></code> whose
-                        <code>errorDetail</code> set to
-                        "hardware-encoder-error".</li>
+                        <code>errorDetail</code> is set to
+                        "hardware-encoder-error" and abort these steps.</li>
+                        <li>For all other errors, <a>reject</a> <var>p</var>
+                        with a newly <a data-link-for="exception" data-lt=
+                        "create">created</a> <code>OperationError</code>.</li>
                       </ol>
                     </li>
                   </ol>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5423,14 +5423,14 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                         <li>Set <var>sender</var>'s internal
                         <a>[[\SendEncodings]]</a> slot to <code>
                         <var>parameters</var>.encodings</code>.</li>
-                        <li>Resolve <var>p</var> with <code>undefined</code>.
+                        <li><a>Resolve</a> <var>p</var> with <code>undefined</code>.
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -11224,6 +11248,11 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
                 attribute is set to the line number in the SDP where the syntax
                 error was detected.</td>
               </tr>
+              <tr>
+                <td><dfn><code>hardware-encoder-error</code></dfn></td>
+                <td>The hardware encoder resources required for the requested
+                operation are not available.</td>
```
taylor-b: "... or the hardware encoder doesn't support the provided set of parameters"? Those could feasibly be two separate errors.

webrtc.html
```diff
@@ -5331,22 +5330,46 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                <var>parameters</var> argument has a value less than 1.0,
+                return a promise rejected with a newly
                 <a data-link-for="exception" data-lt="create">created</a>
                 <code>RangeError</code>.</li>
-                <li>Set <var>sender</var>'s internal <a>[[\SendEncodings]]</a>
-                slot to <code><var>parameters</var>.encodings</code>.</li>
-                <li>Resolve <var>p</var> with <code>undefined</code>.</li>
+                <li>Let <var>p</var> be a new promise.</li>
+                <li>In parallel, configure the media stack to use
+                <var>parameters</var> to transmit <var>sender</var>.
```
taylor-b: nit: It's not `sender` that's being transmitted, it's `sender`'s track.

webrtc.html
```diff
@@ -5331,22 +5330,46 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                    </li>
+                    <li>If any error occurred while configuring the media stack,
+                    queue a task to run the following steps:
+                      <ol>
+                        <li>If an error occurred due to hardware resources not
+                        being available, reject <var>p</var> with a newly
+                        created <code><a>RTCError</a></code> whose
+                        <code>errorDetail</code> set to
+                        "hardware-encoder-error".</li>
+                      </ol>
```
taylor-b: Do we need a catchall "OperationError" if something non-hardware-related failed in the async steps?

webrtc.html
```diff
@@ -5331,22 +5330,46 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+                        <li>Resolve <var>p</var> with <code>undefined</code>.
+                      </ol>
+                    </li>
+                    <li>If any error occurred while configuring the media stack,
+                    queue a task to run the following steps:
+                      <ol>
+                        <li>If an error occurred due to hardware resources not
+                        being available, reject <var>p</var> with a newly
+                        created <code><a>RTCError</a></code> whose
+                        <code>errorDetail</code> set to
```
taylor-b: "is set to"

webrtc.html
adam-be: Let's add that. Hopefully it wouldn't be used much if we have a rather complete list of specific errors.

webrtc.html
adam-be: Raised a separate issue for this question #1602

adam-be:
Thanks for reviewing @taylor-b. New iteration coming up.

PR #1561, opened 2017-08-25
Clarify which session description to check for if negotiation is needed

### LINKED ISSUE #1331: Session description attribute not specified for some steps in check if negotiation is needed
In Section 4.7.3, the following steps do not specify how the SDP comparison is done:

- If connection has created any RTCDataChannels, and no m= section has been negotiated yet for data, return "true".
- If t isn't stopped and isn't yet associated with an m= section according to [JSEP] (section 3.4.1.), return "true".
- If t isn't stopped and is associated with an m= section according to [JSEP] (section 3.4.1.), then perform the following checks:

The remaining steps are clear that the comparison is done against `currentLocalDescription` or `currentRemoteDescription`.

soareschen:
Fixes #1331.

In this patch I clarify checking of the m= sections against `currentLocalDescription` only. But I am not entirely sure if it is correct to check for `currentLocalDescription` only. Would be great if anyone can verify and correct me if I am wrong.

I also done some minor improvements to the formatting to make it more readable.

***
[Preview](https://s3.amazonaws.com/pr-preview/soareschen/webrtc-pc/issue-1331-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/48532fb...soareschen:5ba6c78.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3445,53 +3445,55 @@ interface RTCSessionDescription {
         </ol>
         <p>To <dfn>check if negotiation is needed</dfn> for
         <var>connection</var>, perform the following checks:</p>
-        <ul>
+        <ol>
           <li>
             <p>If any implementation-specific negotiation is required, as
             described at the start of this section, return <code>true</code>.
             </p>
           </li>
+            <li>
+              <p>Let <var>description</var> be <var>connection</var>'s
+              <a data-for="RTCPeerConnection"
+              ><code>currentLocalDescription</code></a>.</p>
+            </li>
           <li>
             <p>If <var>connection</var> has created any
-            <code><a>RTCDataChannel</a></code>s, and no m= section has been
-            negotiated yet for data, return <code>true</code>.</p>
+            <code><a>RTCDataChannel</a></code>s, and no m= section in
+            <var>description</var> has been negotiated yet for data, return
+            <code>true</code>.</p>
           </li>
           <li>
-            <p>For each transceiver <var>t</var> in <var>connection</var>'s
+            <p>For each <var>transceiver</var> in <var>connection</var>'s
             <a>set of transceivers</a>, perform the following checks:</p>
-            <ul>
+            <ol>
               <li>
-                <p>If <var>t</var> isn't <a data-for="RTCRtpTransceiver">
-                stopped</a> and isn't yet <a>associated</a> with an m= section,
-                return <code>true</code>.</p>
+                <p>If <var>transceiver</var> isn't <a data-for="RTCRtpTransceiver">
+                stopped</a> and isn't yet <a>associated</a> with an m= section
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3445,62 +3445,68 @@
...
             <a>set of transceivers</a>, perform the following checks:</p>
-            <ul>
+            <ol>
               <li>
-                <p>If <var>t</var> isn't <a data-for="RTCRtpTransceiver">
-                stopped</a> and isn't yet <a>associated</a> with an m= section,
-                return <code>true</code>.</p>
+                <p>If <var>transceiver</var> isn't <a data-for="RTCRtpTransceiver">
+                stopped</a> and isn't yet <a>associated</a> with an m= section
+                in <var>description</var>, return <code>true</code>.</p>
```
taylor-b: By the way, it now may be redundant to say "in description" here, since we define what an "associated" transceiver is independently. But I don't see that it hurts.

soareschen: I am thinking whether there are transceivers associated with session descriptions other than `currentLocalDescription`, e.g. `pendingLocalDescription`. Should the result be true or false in such case?

webrtc.html
```diff
@@ -3445,62 +3445,68 @@
...
                     "initial-answers">[[!JSEP]]</span>), return
                     <code>true</code>.</p>
                   </li>
-                </ul>
+                </ol>
               </li>
               <li>
-                <p>If <var>t</var> is <a data-for="RTCRtpTransceiver">
+                <p>If <var>transceiver</var> is <a data-for="RTCRtpTransceiver">
                 stopped</a> and is <a>associated</a> with an m= section, but the
```
soareschen: @taylor-b I leave the session description here unspecified, because this seem to suggest it can be associated with any of the 4 session descriptions.

webrtc.html
```diff
@@ -3445,62 +3445,68 @@
...
                     slot, return <code>true</code>.</p>
                   </li>
                   <li>
-                    <p>If <var>connection</var>'s <code><a data-for=
-                    "RTCPeerConnection">currentLocalDescription</a></code> is of
-                    type <code>"answer"</code>, and the direction of the <a>associated</a> m=
-                    section in the answer does not match <var>t</var>'s
+                    <p>If <var>description</var> is of type <code>"answer"</code>,
+                    and the direction of the <a>associated</a> m=
+                    section in the answer does not match <var>transceiver</var>'s
```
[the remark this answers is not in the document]
soareschen: Fixed.

webrtc.html
taylor-b: These steps are only run in the `"stable"` signaling state, where there won't be any `pendingLocalDescription`, so that's not an issue.

PR #1566, opened 2017-08-30
Add a section summarizing different ICE candidate events.

### LINKED ISSUE #1213: addIceCandidate end of candidates woes
there has been much back-and-forth on this... @taylor-b you might know.
How do I make an end-of-candidates indications for all transports that need one?

http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addicecandidate
> If candidate.candidate is an empty string, process candidate as an end-of-candidates indication for the corresponding media description and ICE candidate generation.

This assumes a end of candidates indication for a particular sdpMid/sdpMLineIndex however.

However, http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectioniceevent still says that the candidate (and therefore the sdpMId and sdpMLineIndex) is null on end-of-candidates and there is only one.

taylor-b:
Fixes #1213.

We've ended up with three different types of events that all use the
same "icecandidate" event, and are fired in different places. Which is,
understandably, confusing to most readers of the specification.

This PR adds a note to the RTCPeerConnectionIceEvent section summarizing
the different types of events, describing when they're fired, how
they're represented in the RTCPeerConnectionIceEvent, and what the
application should do with them.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1213_candidate_events.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/48532fb...taylor-b:3ee9539.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3966,6 +3966,7 @@ interface RTCIceCandidate {
               for backwards compatibility, and this event does not need to be
               signaled to the remote peer.</p>
             </li>
+          </ul>
         </div>
         <div>
           <pre class="idl">
--- a/webrtc.html
+++ b/webrtc.html
@@ -3936,7 +3936,7 @@ interface RTCIceCandidate {
           different types of indications:
           <ul>
             <li>
-              <p>An candidate has been gathered. The <code><a
+              <p>A candidate has been gathered. The <code><a
               data-link-for="RTCPeerConnectionIceEvent">candidate</a></code>
               member of the event will be populated normally. It should be
               signaled to the remote peer and passed into
@@ -3958,13 +3958,16 @@ interface RTCIceCandidate {
             <li>
               <p>All <code><a>RTCIceTransport</a></code>s have finished
               gathering candidates, and the <code><a>RTCPeerConnection</a></code>'s
-              <code><a>RTCIceConnectionState</a></code> has transitioned to
-              <code>"<a data-link-for="RTCIceConnectionState">completed</a>"</code>.
+              <code><a>RTCIceGatheringState</a></code> has transitioned to
+              <code>"<a data-link-for="RTCIceGatheringState">complete</a>"</code>.
               This is indicated by the
               <code><a data-link-for="RTCPeerConnectionIceEvent">candidate</a></code>
-              member of the event being set to <code>null</code>. This only exists
-              for backwards compatibility, and this event does not need to be
-              signaled to the remote peer.</p>
+              member of the event being set to <code>null</code>. This only
+              exists for backwards compatibility, and this event does not need
+              to be signaled to the remote peer. It's equivalent to an
+              <code>"<a>icegatheringstatechange</a>"</code> event with the
+              <code>"<a data-link-for="RTCIceGatheringState">complete</a>"</code>
+              state.</p>
             </li>
           </ul>
         </div>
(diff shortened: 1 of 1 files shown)


taylor-b:
@soareschen and @fippo, can you take a look and let me know if this clears things up enough?

webrtc.html
```diff
@@ -3930,6 +3930,44 @@
         type <code>relay</code>, the <code>url</code> property of the event
         MUST be set to the URL of the ICE server from which the candidate was
         obtained.</p>
+
+        <div class="note">
+          The <code><a>icecandidate</a></code> event is used for three
+          different types of indications:
+          <ul>
+            <li>
+              <p>An candidate has been gathered. The <code><a
```
fippo: a?

webrtc.html
```diff
@@ -3930,6 +3930,44 @@
...
+              end-of-candidates indication to the remote peer.</p>
+            </li>
+            <li>
+              <p>All <code><a>RTCIceTransport</a></code>s have finished
+              gathering candidates, and the <code><a>RTCPeerConnection</a></code>'s
+              <code><a>RTCIceConnectionState</a></code> has transitioned to
+              <code>"<a data-link-for="RTCIceConnectionState">completed</a>"</code>.
+              This is indicated by the
+              <code><a data-link-for="RTCPeerConnectionIceEvent">candidate</a></code>
+              member of the event being set to <code>null</code>. This only exists
```
fippo: wdyt about saying this is functionally equvialent to an icegatheringstatechange to completed?

webrtc.html
taylor-b: I must have gone "an"-crazy from the "an RTC" thing.

webrtc.html
```diff
@@ -3930,6 +3930,47 @@
...
+              <code><a>RTCIceGatheringState</a></code> has transitioned to
+              <code>"<a data-link-for="RTCIceGatheringState">complete</a>"</code>.
+              This is indicated by the
+              <code><a data-link-for="RTCPeerConnectionIceEvent">candidate</a></code>
+              member of the event being set to <code>null</code>. This only
+              exists for backwards compatibility, and this event does not need
+              to be signaled to the remote peer. It's equivalent to an
+              <code>"<a>icegatheringstatechange</a>"</code> event with the
+              <code>"<a data-link-for="RTCIceGatheringState">complete</a>"</code>
+              state.</p>
```
soareschen: I think one more clarification needed: there could be more than one `icecandidate` event with `null` candidate, correct? i.e. `iceGatheringState` can change from `complete` back to `new` or `gathering` and then go back to `complete` again.

Since the third case is now associated with `RTCIceGatheringState`, we can put the clarification there instead of here.

fippo: no, unless the developer does an ice restart.

nils-ohlmeier: @soareschen there can be multiple candidates with the empty string, see the paragraph above. But this is the indication that gathering has finished for all ICE transports (in the current offer-answer cycle).

soareschen: Hmm ok then. I am thinking of whether application code can assume that no more `icecandidate` event will be fired after a `null` candidate event is fired. Having ICE restart or new ICE transport added could break this assumption.

PR #1567, opened 2017-08-30
Adding `remoteTrackId` attribute to represent SDP-signaled track ID.

### LINKED ISSUE #1128: does MSID still work? (was: example 13: getReceivers semantics)

taylor-b:
Fixes #1128.

This is a pretty small addition that should resolve the remaining
concerns about track ID signaling.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1128_remoteTrackId.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/48532fb...taylor-b:58714b7.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6459,6 +6459,7 @@ sender.setParameters(params)
     readonly        attribute boolean                     stopped;
     readonly        attribute RTCRtpTransceiverDirection  direction;
     readonly        attribute RTCRtpTransceiverDirection? currentDirection;
+    readonly        attribute DOMString?                  remoteTrackId;
     void setDirection (RTCRtpTransceiverDirection direction);
     void stop ();
     void setCodecPreferences (sequence&lt;RTCRtpCodecCapability&gt; codecs);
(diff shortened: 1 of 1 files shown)


stefhak:
How is this id different from the one you get by checking the id of the received track?

taylor-b:
@stefhak Tried to explain this in the note:

> If a transceiver is created by a method other than `setRemoteDescription`, `receiver.track.id` will be set to a randomly generated value instead of one that comes from SDP.

And again, this is because the receiver and its track are created before the remote track ID is known (for the "early media" use case), and the ID can't change after creation. Should the note mention that too?

ibc:
I don't fully understand why `receiver` cannot be nullable in `RTCRtpTransceiver`. I assume there is a good reason for that, so this PR provides a valid "workaround" for the "unmatched remote track ID" problem.

PR #1570, opened 2017-09-04
Resizing video (fix for Issue 1283)

### LINKED ISSUE #1283: Section 5.2: centering, scaling, cropping
From EKR: https://lists.w3.org/Archives/Public/public-webrtc/2017May/0166.html

The advice here (center, scale, then crop) differs from a MUST in JSEP S 3.6.

   If the original resolution exceeds the size limits in the attribute,
   the sender SHOULD apply downscaling to the output of the
   MediaStreamTrack in order to satisfy the limits.  Downscaling MUST
   NOT change the track aspect ratio.

stefhak:
Fixes #1283

***
[Preview](https://s3.amazonaws.com/pr-preview/stefhak/webrtc-pc/issue-1283-patch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/35a3a09...stefhak:deb3c6a.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4464,6 +4464,15 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     envelope negotiated by SDP. The rules outlined in
     <span data-jsep="interpreting-imageattr">[[!JSEP]]</span> MUST be followed
     when resizing the video.</p>
+    <div class="note">
+      <p>The procedures outlined in JSEP preserve the aspect ratio if the video
+      is resized. If the video track is rendered in a <code>video</code>
+      element on the receiving side, and that <code>video</code> element
+      has a different aspect ratio than the track, the video will be shown
+      letterboxed or pillarboxed as specified in [[!HTML51]] regardless of
+      whether it has been resized or not.</p>
+    </div>
+
     <p>When video is rescaled, for example for certain combinations
     of width or height and
     <code><a data-link-for="RTCRtpEncodingParameters">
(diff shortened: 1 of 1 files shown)


stefhak:
This PR aligns with JSEP. Even if we decide to not do that we need to update the current text as it now says ```desired size``` without ever defining what that is (it seems implicitly to say that is the maximum size even if that requires cropping, is not always be the case I guess).

aboba:
@stefhak Currently "sec.imageattr" references JSEP Section 3.6.  Should we be referencing Section 3.6.2?

stefhak:
Regarding: https://github.com/w3c/webrtc-pc/pull/1570#issuecomment-327572065: I agree, the reference should be more specific. However, only top level (3.6) is linkable now. @dontcallmedom could entries for subsections of 3.6 be added?

dontcallmedom:
@stefhak we need for JSEP to add an `anchor` to the sections we want to link to  - could raise issues there as needed? Also maybe worth letting them know that the [XML version on github.io](https://rtcweb-wg.github.io/jsep/draft-ietf-rtcweb-jsep.xml) is stuck to a 2 years old version.

stefhak:
@dontcallmedom https://github.com/rtcweb-wg/jsep/issues/835 filed!

stefhak:
I also filed https://github.com/rtcweb-wg/jsep/issues/838 on JSEP since that seems like a situation that can occur.

stefhak:
The VI October 12 2017 decided this PR should _not_ be merged. An alternative proposal is being worked on.

PR #1576, opened 2017-09-07
Editorial: Link to 'Firing a something event' definitions

### LINKED ISSUE #1141: "Firing an RTCTrackEvent event named e" is never referenced
https://w3c.github.io/webrtc-pc/#fire-track-event

This definition is never referenced, and searching for "track event" the only relevant hit is in the introductory and seemingly non-normative text for https://w3c.github.io/webrtc-pc/#rtp-media-api: "... are surfaced to the application via the ontrack event."

It was when looking at https://github.com/w3c/webrtc-pc/pull/1129 that I wanted to know if the streams member would always have at least one member, but from the spec I cannot tell.

adam-be:
We have a paragraph for each type of event that describes how it should be fired. We should link to those sections when we fire events. There's currently only one such link.

Fix #1141

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/link-fire-event.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/a004acc...730cd40.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1212,7 +1212,7 @@
             <var>connection</var>.</p>
           </li>
           <li>
-            <p>If <var>newState</var> is <code>"completed"</code>, fire an event
+            <p>If <var>newState</var> is <code>"completed"</code>, <a>fire an ice candidate event</a>
             named <code><a>icecandidate</a></code> with <code>null</code> at
             <var>connection</var>.</p>
           </li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1212,7 +1212,7 @@
             <var>connection</var>.</p>
           </li>
           <li>
-            <p>If <var>newState</var> is <code>"completed"</code>, fire an event
+            <p>If <var>newState</var> is <code>"completed"</code>, <a>fire an ice candidate event</a>
```
taylor-b: nit: It sounds a little redundant to say "fire an ice candidate event named `icecandidate`".

This may be unavoidable. But is there a reason we need to specify the event name here, and not in the "fire an ice candidate event" section? The name will always be "icecandidate" (correct?), so it seems like we could remove the "event name" argument entirely, and reference the name "icecandidate" in the "fire an ice candidate event" section instead of here.

adam-be: In most cases it's redundant. But sometimes there are two types/names of events that use the same interface. For example, "addtrack" and "removetrack".

taylor-b: Ok, makes sense. If we have to do that for the "track" event, might as well be consistent.

PR #1577, opened 2017-09-07
scaleResolutionDownBy: Specify how the User Agent should behave when scaling video

### LINKED ISSUE #1564: Deciding on resolutions with scaleResolutionDownBy
The spec mentions constraints on the value of `scaleResolutionDownBy` but not how it applies when actually scaling video frames.

I have a couple of cases in mind where it is unclear what resolution will be sent over the wire:

- Common input `640x480`, odd `scaleResolutionDownBy` `x9.0`.
  - You can't scale to `71.111x53.333`
- Odd input (say canvas or application sharing), e.g., `705x529`, `scaleResolutionDownBy` `x2.0`.
  - You can't scale to `78.333x58.778`
- The encoder doesn't handle scaling to certain resolutions
  - webrtc.org requires that all layers have identical aspect ratio, see [1], so a scaling factor where the result has to be rounded (like the two above) need a bit more massaging to be accepted. Not sure about other implementations or hardware encoders.

[1] http://searchfox.org/mozilla-central/rev/b318c7dca7392bd16c0b11929f55b1be133f0b31/media/webrtc/trunk/webrtc/modules/video_coding/codecs/vp8/simulcast_encoder_adapter.cc#62

stefhak:
Would fix #1564 if we agree this is the way to go.

***
[Preview](https://s3.amazonaws.com/pr-preview/stefhak/webrtc-pc/patch-issue-1564.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/a004acc...stefhak:9caf541.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4465,8 +4465,8 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     of width or height and
     <code><a data-link-for="RTCRtpEncodingParameters">
     scaleResolutionDownBy</a></code> values, situations when the result must
-    be rounded may occur. In such situations the user agent is allowed to
-    round as suitable.</p>
+    be rounded to obtain an integer result may occur. In such situations the
+    user agent MUST round up to the nearest integer value.</p>
     <p>The actual encoding and transmission of <code>MediaStreamTrack</code>s
     is managed through objects called <code><a>RTCRtpSender</a></code>s.
     Similarly, the reception and decoding of <code>MediaStreamTrack</code>s is
--- a/webrtc.html
+++ b/webrtc.html
@@ -4461,6 +4461,12 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     is designed to minimize occurrence of images with with letter
     box or or pillow boxing. The media MUST NOT be upscaled to
     create fake data that did not occur in the input source.</p>
+    <p>When video is rescaled, for example for certain combinations
+    of width or height and
+    <code><a data-link-for="RTCRtpEncodingParameters">
+    scaleResolutionDownBy</a></code> values, situations when the result must
+    be rounded may occur. In such situations the user agent is allowed to
+    round as suitable.</p>
     <p>The actual encoding and transmission of <code>MediaStreamTrack</code>s
     is managed through objects called <code><a>RTCRtpSender</a></code>s.
     Similarly, the reception and decoding of <code>MediaStreamTrack</code>s is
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4461,6 +4461,12 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
     is designed to minimize occurrence of images with with letter
     box or or pillow boxing. The media MUST NOT be upscaled to
     create fake data that did not occur in the input source.</p>
+    <p>When video is rescaled, for example for certain combinations
+    of width or height and
+    <code><a data-link-for="RTCRtpEncodingParameters">
+    scaleResolutionDownBy</a></code> values, situations when the result must
+    be rounded may occur. In such situations the user agent is allowed to
```
fluffy: I think we should be less wishy washy here and say how it must round - basically we can choose from floor(), ceil(), or round(). It should not be different on different browsers.

stefhak: Perhaps that's what we should do. I made this a bit open since I was unsure if that would cause problems (perhaps 71 pixels as in the example in #1564 is hard to do, while 70 or 68 is easy?). Surely sending a video with width 70 (or even 68) is preferred over not sending when the division gives 71.3333333, so I gave some room for implementations. But if we do not foresee that being a problem I totally agree to your comment.

fluffy: All the specs for any video codec anyone uses any more work fine with 71 and though there is rumor of some implementations that might have bugs at for sizes that are not even, it seems  those are largely fixed so I really don't see this as a issue any more. I think we would be fine to say it gets rounded to 71. I think my vote would be for "round up to nearest integer"

stefhak:
Updated PR according to proposal in https://github.com/w3c/webrtc-pc/pull/1577/files/c172105ae775a05b02985849348c23e92cc9944c#r137785280.

webrtc.html
```diff
@@ -4461,6 +4461,12 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
     is designed to minimize occurrence of images with with letter
     box or or pillow boxing. The media MUST NOT be upscaled to
     create fake data that did not occur in the input source.</p>
+    <p>When video is rescaled, for example for certain combinations
+    of width or height and
+    <code><a data-link-for="RTCRtpEncodingParameters">
+    scaleResolutionDownBy</a></code> values, situations when the result must
+    be rounded to obtain an integer result may occur. In such situations the
+    user agent MUST round up to the nearest integer value.</p>
```
taylor-b: Decision at the virtual interim was to change from "ceil" to "floor". The reasoning is that it's easier for scaling algorithms to handle weird resolutions by cropping off lines of pixels at the edge as needed. For example, it's easy to scale a 101x101 image down to 50x50 by removing one row/column of pixels, then doing 2:1 scaling (where every output pixel is determined by 4 input pixels).

stefhak:
I think I got all comments addressed now. Perhaps one of the editors (if agreeing) can press "merge".

Issue #1237, opened 2017-05-24
[[associated MediaStreams]] internal slot for RTCRtpSender is not used


soareschen:
The only time the [[associated MediaStreams]] and [[send encodings]] internal slots are referenced in the spec is when setting the value. There is no mention of when these fields are read and how they are being used.

aboba:
The [[associated MediaStreams]] internal slot does not seemed to be used. 

However, the[[send encodings]] internal slot is used in both setParameters and getParameters:

In `setParameters`:
"Let `N` be the number of `RTCRtpEncodingParameters` stored in sender's internal [[send encodings]] slot.... If `parameters.encodings.length` is different from `N` ..."

In `getParameters`: 
"`encodings` is set to the value of the [[send encodings]] internal slot."

soareschen:
Yes, just saw the usage of `[[send encodings]]` was added to the 20170605 draft after this issue was raised.

aboba:
@jan-ivar Any idea where this internal slot should be used?

jan-ivar:
@aboba The [[[associated MediaStreams](http://w3c.github.io/webrtc-pc/#dfn-associated-mediastreams)]] is used to create the MSIDs in the SDP.

See the [final steps to create an offer](http://w3c.github.io/webrtc-pc/#dfn-steps-to-create-an-offer):
 * *"3. Given the information that was obtained from previous inspection and the current state of connection and its RTCRtpTransceiver s, generate an SDP offer, sdpString, as described in [JSEP] (section 5.2.)."*

[Section 5.2.1](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-20#section-5.2.1) says: *"For each MediaStream that was associated with the transceiver when it was created via addTrack or addTransceiver, an "a=msid" line, as specified in [I-D.ietf-mmusic-msid]"*

That seems to refer to the same thing. We should probably make that more explicit if we can.

jan-ivar:
Nevermind that there's no way to associate a MediaStream with a transceiver using addTransceiver.

stefhak:
@jan-ivar can you do a PR?

adam-be:
As currently specified `[[AssociatedMediaStreams]]` is part of the information bundle that's used to create the SDP (as @jan-ivar points out above). I don't think we should start listing all the pieces of that bundle in step 3 of the algorithm. I've made a PR that adds some information about `[[AssociatedMediaStreams]]` and that it's used in the SDP generation process.

[...]

PR #1591, opened 2017-09-11
Add some text about how the [[AssociatedMediaStreams]] slot is used

adam-be:
Proposed fix for #1237

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/use-associated-ms-slot.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/4f0ef95...9100c94.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5186,7 +5186,9 @@ interface RTCPeerConnectionIceErrorEvent : Event {
           internal slot, representing a list of
           <code><a>MediaStream</a></code> objects that the
           <code><a>MediaStreamTrack</a></code> object of this sender is
-          associated with.</p>
+          associated with. The <a>[[\AssociatedMediaStreams]]</a> slot is used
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5186,7 +5186,9 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
           internal slot, representing a list of
           <code><a>MediaStream</a></code> objects that the
           <code><a>MediaStreamTrack</a></code> object of this sender is
-          associated with.</p>
+          associated with. The <a>[[\AssociatedMediaStreams]]</a> slot is used
+          when <var>sender</var> is represented in SDP as described in <span
+          data-jsep="initial-offers">[[!JSEP]]</span>.</p>
```
taylor-b: I don't think JSEP needs an internal slot though; it can rely on prose. So I'd recommend just removing `[[AssociatedMediaStreams]]` completely.

PR #1592, opened 2017-09-11
Select sent codec via "codecPayloadType" field rather than reordering.

### LINKED ISSUE #1207: Setting a remote description may cause discontinuity due to codec switching.
If an application is using `sender.setParameters` to select the sent codec, this gets wiped out when `setRemoteDescription` is called:

> The effect of reordering or removing codecs lasts until the codecs are renegotiated. After the codecs are renegotiated, they are set to the value negotiated, and setParameters needs to be called again to re-apply codec preferences.

It's a little inconvenient from an API perspective to need to call `setParameters` every time `setRemoteDescription`'s promise resolves. But the bigger issue is that one frame might end up encoded with an unintended codec before the application has a chance to switch. And it wouldn't be surprising if this causes a blip on the receive side.

Is there any reason we can't use [`codecPayloadType`](http://draft.ortc.org/#dom-rtcrtpencodingparameters-codecpayloadtype) like ORTC has, and say that once set, it doesn't change (unless the PT is removed from the remote description)? Neither Chrome nor Firefox have implemented the `setParameters` codec reordering yet, so I don't think there are backwards compatibility concerns.

taylor-b:
Fixes #1207.

Instead of selecting a codec by reordering the list from
`getParameters`, an application will now pick the payload type of
its preferred codec and set it in `RTCRtpEncodingParameters`. This
matches ORTC, and fixes the linked issue, because as long as the
selected codec is still present in a new offer/answer exchange, it will
stay selected.

If the selected codec is *not* present in a new offer/answer exchange,
the implementation will go back to its normal policy, which is the same
behavior as before this PR.

Note that this PR puts the `codecPayloadType` field on
`RTCRtpEncodingParameters`, which means the codec may be different
per-encoding, which wasn't possible before. If this is an issue, it
could be moved to the `RTCRtpParameters` level instead.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1207_add_codecpayloadtype.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/4f0ef95...taylor-b:7971859.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5809,7 +5809,7 @@ sender.setParameters(params)
             <tbody>
               <tr id="def-codecpayloadtype*">
                 <td><dfn>codecPayloadType</dfn></td>
-                <td><code>unsigned short</code></td>
+                <td><code>octet</code></td>
                 <td>Sender</td>
                 <td>Read/Write</td>
               </tr>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5346,8 +5346,8 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                   (if set) corresponds to a value of
                   <code><var>parameters</var>.codecs[<var>j</var>].payloadType</code> where
                   <var>j</var> goes from 0 to the number of codecs. If there is no
-                  correspondence, or if
-                  <code><var>parameters</var>.codecs[<var>j</var>].name</code> is equal to
+                  correspondence, or if the MIME subtype portion of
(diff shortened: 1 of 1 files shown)


taylor-b:
Note that the FLEXFEC reference is missing; I'll fix that.

taylor-b:
Decision at virtual interim was to go ahead with this PR. I went ahead and addressed Bernard's comments.

PR #1608, opened 2017-09-19
Validate protocol string in IdP operations

### LINKED ISSUE #1500: setIdentityProvider should throw error if protocol contains invalid characters

soareschen:
Fixes #1500.

This is based the protocol field defined in [ietf-rtcweb-security-arch](https://tools.ietf.org/html/draft-ietf-rtcweb-security-arch-12#section-5.6.5):

> protocol:  The specific IdP protocol which the IdP is using.  This is a completely opaque IdP-specific string, but allows an IdP to implement two protocols in parallel.  This value may be the empty string.  If no value for protocol is provided, a value of "default" is used.

> Note that the separator characters '/' (%2F) and '\\' (%5C) MUST NOT be permitted in the protocol field, lest an attacker be able to direct requests outside of the controlled "/.well-known/" prefix.  Query and fragment values MAY be used by including '?' or '#' characters.

The spec is not clear of how "opaque" the protocol can be in respect to the well-formness of the URI format. They way I interpret this is that the provider string may contain the path, query, and fragment components of a URI, and each component will be normalized accordingly when forming the well-known URI for the IdP proxy, e.g. invalid characters will be percent-encoded.

Since anything can be percent encoded, that means protocol may contain arbitrary characters other than "/" and "\\" which are explicitly forbidden. Here the percent encoded strings "%2F" and "%5C" are not banned, because my understanding is that percent encoding will be done by the browser instead of the application.

This PR differs with the approach in #1538. #1538 does not cover percent encoding and the query/fragment component. We might want to discuss more on whether it is the browser or application responsibility to encode the protocol field as well-formed URI.

***
[Preview](https://s3.amazonaws.com/pr-preview/soareschen/webrtc-pc/protocol-validation.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/67270e5...soareschen:8ea135e.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9943,8 +9943,9 @@ interface RTCIdentityProviderRegistrar {
               "idlMemberType"><a>DOMString</a></span>, defaulting to
               <code>"default"</code></dt>
               <dd>
-                <p>The protocol parameter used for the IdP. This attribute MUST
-                contain only characters legal for inclusion in URIs [[!RFC3986]].</p>
+                <p>The protocol parameter used for the IdP. The string
+                MUST NOT include the character <code>'/'</code> or
+                <code>'\'</code>.</p>
               </dd>
             </dl>
           </section>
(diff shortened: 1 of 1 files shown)


stefhak:
@martinthomson can you a take a look?

aboba:
@soareschen Can you rebase?

soareschen:
Rebased now.

stefhak:
I think we can merge this one now. Thanks @soareschen

PR #419, opened 2015-12-03
fix setParameters invalid webidl

jan-ivar:
0:21.52 WebIDL.WebIDLError: error: Dictionary argument or union argument containing a dictionary not followed by a required argument must be optional, /Users/Jan/moz/mozilla-central/dom/webidl/RTCRtpSender.webidl line 73:39
 0:21.52  void setParameters (RTCRtpParameters parameters);

PR #1610, opened 2017-09-20
Improve definition of required fields in RTCRtpParameters

### LINKED ISSUE #1493: Required fields in RTCRtpParameters returned from getParameters
This is similar to  #1476.

Currently all fields in RTCRtpParameters are defined to be optional. This means when writing applications and tests we cannot make assumption of any of the field returned from `getParameters()` is defined.

I have a feeling that `encodings`, `headerExtensions`, and `codecs` should at least be empty list, and that all if not most of the readonly fields in `RTCRtpCodecParameters`, `RTCRtpHeaderExtensionParameters`, and `RTCRtcpParameters` should be required?

soareschen:
Fixes #1493.

This PR attempts to clarify which fields are required in the `RTCRtpParameters` dictionary returned from `getParameters()`. I have also factor out the `transactionId` and `degradationPreference` fields to a new dictionary `RTCRtpSenderParameters`, since these two fields are required only for `RTCRtpSender`.

It would be great if someone with more experience on RTP parameters can verify whether some of the fields marked required make sense. Note that we'd also have to consider the case when no SDP has been set yet, e.g. `pc.addTransceiver().sender.getParameters()` before any offer is generated.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4524,7 +4524,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
       given as an argument is already being sent to avoid sending
       the same <code>MediaStreamTrack</code> twice, the other ways
       do not, allowing the same <code>MediaStreamTrack</code> (possibly
-      using different <code><a>RTCRtpParameters</a></code> with different
+      using different <code><a>RTCRtpSenderParameters</a></code> with different
       <code><a>RTCRtpSender</a></code>s) to be sent several times
       simultaneously. Doing this implies that at the receiving end of
       the peer-to-peer connection there are several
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4524,7 +4524,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       given as an argument is already being sent to avoid sending
       the same <code>MediaStreamTrack</code> twice, the other ways
       do not, allowing the same <code>MediaStreamTrack</code> (possibly
-      using different <code><a>RTCRtpParameters</a></code> with different
+      using different <code><a>RTCRtpSenderParameters</a></code> with different
```
taylor-b: I don't know that introducing the RTCRtpSenderParameters dictionary is really worth it. It's one more dictionary to keep track of and one more link to follow, vs. two fields that are described as "sender-only."

webrtc.html
```diff
@@ -5972,8 +5981,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       </div>
       <div>
         <pre class="idl">dictionary RTCRtcpParameters {
-             DOMString cname;
-             boolean   reducedSize;
+             required DOMString cname;
```
taylor-b: We don't need these to be marked required, or anything else in RtpParameters. The input validation is handled by comparing against `[[LastReturnedParameters]]`.

webrtc.html
```diff
@@ -5271,8 +5271,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
     readonly        attribute RTCDtlsTransport?  transport;
     readonly        attribute RTCDtlsTransport? rtcpTransport;  // Feature at risk
     static RTCRtpCapabilities getCapabilities (DOMString kind);
-    Promise&lt;void&gt;             setParameters (optional RTCRtpParameters parameters);
-    RTCRtpParameters          getParameters ();
+    Promise&lt;void&gt;             setParameters (RTCRtpSenderParameters parameters);
```
taylor-b: This needs to be optional. It's a weird WebIDL thing. https://github.com/w3c/webrtc-pc/pull/419

aboba:
As noted in the above WebIDL link, `required` is generally used for inputs to the browser, rather than outputs.  If there is ambiguity, we should fix this in the specification text.

PR #1614, opened 2017-09-28
createDataChannel: Queue a task to update negotiation needed state

### LINKED ISSUE #1597: createDataChannel touches DOM "in the background"
[createDataChannel](https://rawgit.com/w3c/webrtc-pc/master/webrtc.html#dom-rtcpeerconnection-createdatachannel) says: *"Return channel and continue the following steps in the background."*

First, it should be "[in parallel](https://html.spec.whatwg.org/#in-parallel)".

Second, the in parallel steps must [queue a task](https://html.spec.whatwg.org/#queue-a-task) to run the [update the negotiation-needed flag for connection](https://rawgit.com/w3c/webrtc-pc/master/webrtc.html#dfn-update-the-negotiation-needed-flag) algorithm which touches DOM stuff (we can't touch DOM stuff in parallel with JS).

adam-be:
Fixes #1597

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/create-datachannel-fixes.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/19827c9...e999a18.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7969,6 +7969,12 @@ interface RTCTrackEvent : Event {
                   </div>
                 </li>
                 <li>
+                  <p>If <var>channel</var> is the first
+                  <code><a>RTCDataChannel</a></code> created on
+                  <var>connection</var>, <a>update the
+                  negotiation-needed flag</a> for <var>connection</var>.</p>
+                </li>
+                <li>
                   <p>Return <var>channel</var> and continue the following steps
                   in parallel.</p>
                 </li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7932,8 +7932,8 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                 <li>
                   <p>If <var>channel</var> was the first
                   <code><a>RTCDataChannel</a></code> created on
-                  <var>connection</var>, <a>update the negotiation-needed
-                  flag</a> for <var>connection</var>.</p>
+                  <var>connection</var>, queue a task to <a>update the
```
taylor-b: This algorithm will already queue a task to fire the event, so why is this necessary? So that interactions with state and internal slots don't happen "in parallel"? I think we could just move this up before the "in parallel" part then.

taylor-b: By which I mean, make the order:

1. Return channel
2. Update negotiation-needed flag
3. Run steps to set up the underlying SCTP association in parallel

adam-be: If we return in step 1. above, I guess we need to wait for stable state before we can continue and do step 2. on the main loop and then do step 3 in parallel. To avoid that I think we could do 2, 1, 3.

PR #1618, opened 2017-10-05
format & fixup Example code

Jxck:
(same work with https://github.com/w3c/mediacapture-main/pull/486)

cleanup code example in HTML help us understand features more easily.
so I cleanup some code like below.

- format indent
- use basic new feature of ES6^
  - const
  - arrow function
  - async/await
- simplify but save meaning and semantics

if there are some format rule or feature regulation in spec.
please mention me and I'll fix them.

***
[Preview](https://s3.amazonaws.com/pr-preview/Jxck/webrtc-pc/cleanup-example.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/19827c9...Jxck:dd33e4f.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6889,7 +6889,8 @@ interface RTCPeerConnectionIceErrorEvent : Event {
         <p>To respond to a remote peer's "sendonly" offer:</p>
         <pre class="example highlight">
           try {
-              // Apply the sendonly offer first, to ensure the receiver is ready for ICE candidates.
+              // Apply the sendonly offer first,
+              // to ensure the receiver is ready for ICE candidates.
               await pc.setRemoteDescription(sendonlyOffer);
               // Stop sending audio
               await audio.sender.replaceTrack(null);
--- a/webrtc.html
+++ b/webrtc.html
@@ -10976,13 +10976,10 @@ var gameDataChan = peerConn.createDataChannel("data", { reliable: false, priorit
       <p>Send the DTMF signal "123" and abort after sending "2".</p>
       <pre class="example highlight">
         if (sender.dtmf.canInsertDTMF) {
-            sender.dtmf.ontonechange = event =&gt; {
-                if (event.tone == "2") {
-                    // empty the buffer to not play any tone after "2"
-                    sender.dtmf.insertDTMF("");
-                }
-            };
             sender.dtmf.insertDTMF("123");
+            await new Promise(r =&gt; sender.dtmf.ontonechange = e =&gt; e.tone == "2" &amp;&amp; r());
+            // empty the buffer to not play any tone after "2"
+            sender.dtmf.insertDTMF("");
         } else {
             console.log("DTMF function not available");
         }
--- a/webrtc.html
+++ b/webrtc.html
@@ -9343,10 +9343,10 @@ interface RTCDTMFToneChangeEvent : Event {
 }
 
 try {
-    const selector = pc.getSenders()[0].track;
-    const baselineReport = await pc.getStats(selector);
+    const sender = pc.getSenders()[0];
+    const baselineReport = await sender.getStats();
     await new Promise((resolve) =&gt; setTimeout(resolve, aBit)); // ... wait a bit
-    const currentReport = await pc.getStats(selector);
+    const currentReport = await sender.getStats();
     processStats(baselineReport, currentReport);
 } catch (err) {
     console.error(err);
--- a/webrtc.html
+++ b/webrtc.html
@@ -6889,11 +6889,11 @@ audio.setDirection("sendonly");</pre>
 }</pre>
         <p>To stop sending music and send audio captured from a microphone,
         as well to render received audio:</p>
-        <pre class="example highlight">//assume we have an audio transceiver and a microphone track named micTrack
+        <pre class="example highlight">// Assume we have an audio transceiver and a microphone track named micTrack
 audio.sender.replaceTrack(micTrack);
 // Unmute received audio
 audio.receiver.track.enabled = true;
-// // Set the direction to sendrecv (requires negotiation)
+// Set the direction to sendrecv (requires negotiation)
 audio.setDirection("sendrecv");</pre>
         <p>To respond to being taken off hold by a remote peer:</p>
         <pre class="example highlight">try {
--- a/webrtc.html
+++ b/webrtc.html
@@ -10629,8 +10629,7 @@ interface RTCIdentityAssertion {
               // let the "negotiationneeded" event trigger offer generation
               pc.onnegotiationneeded = async () =&gt; {
                   try {
-                      const offer = await pc.createOffer();
-                      await pc.setLocalDescription(offer);
+                      await pc.setLocalDescription(await pc.createOffer());
                       // send the offer to the other peer
                       signalingChannel.send(JSON.stringify({ desc: pc.localDescription }));
                   } catch (err) {
@@ -10701,10 +10700,8 @@ interface RTCIdentityAssertion {
                       // if we get an offer, we need to reply with an answer
                       if (desc.type == "offer") {
                           await pc.setRemoteDescription(desc);
-                          const answer = await pc.createAnswer();
-                          await pc.setLocalDescription(answer);
-                          const str = JSON.stringify({ desc: pc.localDescription });
-                          signalingChannel.send(str);
+                          await pc.setLocalDescription(await pc.createAnswer());
+                          signalingChannel.send(JSON.stringify({ desc: pc.localDescription }));
                       } else {
                           await pc.setRemoteDescription(desc);
                       }
@@ -10749,8 +10746,7 @@ interface RTCIdentityAssertion {
               // let the "negotiationneeded" event trigger offer generation
               pc.onnegotiationneeded = async () =&gt; {
                   try {
-                      const offer = await pc.createOffer();
-                      await pc.setLocalDescription(offer);
+                      await pc.setLocalDescription(await pc.createOffer());
                       // send the offer to the other peer
                       signalingChannel.send(JSON.stringify({ desc: pc.localDescription }));
                   } catch (err) {
@@ -10789,10 +10785,8 @@ interface RTCIdentityAssertion {
                       // if we get an offer, we need to reply with an answer
                       if (desc.type == "offer") {
                           await pc.setRemoteDescription(desc);
-                          const answer = await pc.createAnswer();
-                          await pc.setLocalDescription(answer);
-                          const str = JSON.stringify({ desc: pc.localDescription });
-                          signalingChannel.send(str);
+                          await pc.setLocalDescription(await pc.createAnswer());
+                          signalingChannel.send(JSON.stringify({ desc: pc.localDescription }));
                       } else {
                           await pc.setRemoteDescription(desc);
                       }
@@ -10823,8 +10817,7 @@ interface RTCIdentityAssertion {
               // let the "negotiationneeded" event trigger offer generation
               pc.onnegotiationneeded = async () =&gt; {
                   try {
-                      const offer = pc.createOffer();
-                      await pc.setLocalDescription(offer);
+                      await pc.setLocalDescription(pc.createOffer());
                       // send the offer to the other peer
                       signalingChannel.send(JSON.stringify({ desc: pc.localDescription }));
                   } catch (err) {
@@ -10892,8 +10885,7 @@ interface RTCIdentityAssertion {
               // let the "negotiationneeded" event trigger offer generation
               pc.onnegotiationneeded = async () =&gt; {
                   try {
-                      const offer = await pc.createoffer();
-                      await pc.setlocaldescription(offer);
+                      await pc.setlocaldescription(await pc.createoffer());
                       // send the offer to the other peer
                       signalingchannel.send(json.stringify({ desc: pc.localdescription }));
                   } catch (err) {
@@ -10925,8 +10917,7 @@ interface RTCIdentityAssertion {
                       // if we get an offer, we need to reply with an answer
                       if (desc.type == "offer") {
                           await pc.setremotedescription(desc);
-                          const answer = await pc.createanswer();
-                          pc.setlocaldescription(answer);
+                          pc.setlocaldescription(await pc.createanswer());
                           const str = json.stringify({ desc: pc.localdescription });
                           signalingchannel.send(str);
                       } else {
--- a/webrtc.html
+++ b/webrtc.html
@@ -10839,17 +10839,9 @@ interface RTCIdentityAssertion {
                   pc.addTransceiver(stream.getVideoTracks()[0], {
                       direction: "sendonly",
                       sendEncodings: [
-                          {
-                              rid: "f",
-                          },
-                          {
-                              rid: "h",
-                              scaleResolutionDownBy: 2.0
-                          },
-                          {
-                              rid: "q",
-                              scaleResolutionDownBy: 4.0
-                          }
+                          { rid: "f" },
+                          { rid: "h", scaleResolutionDownBy: 2.0 },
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -12927,5 +12871,3 @@ <h3 id="sec-constraints">Constraints</h3>
     the <a href="https://www.w3.org/community/ortc/">W3C ORTC CG</a>, and have
     been adapted for use in this specification.</p>
   </section>
-</body>
```
dontcallmedom: why removing these closing tags?

dontcallmedom:

webrtc.html
Jxck: oh, unexpected miss operation.

webrtc.html
```diff
@@ -5519,13 +5518,10 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
               <p><code>getParameters</code> may be used with
               <code>setParameters</code> to change the parameters in the
               following way:</p>
-              <pre class="example highlight">
-              var params = sender.getParameters();
+              <pre class="example highlight">const params = sender.getParameters();
```
taylor-b: Is there a reason for starting/ending the example code on the same line as the `<pre>`/`</pre>` tags? I think it's more readable the other way.

webrtc.html
```diff
@@ -6872,44 +6868,46 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
         as well to render received audio:</p>
-        <pre class="example highlight">
-            //assume we have an audio transceiver and a microphone track named micTrack
-            audio.sender.replaceTrack(micTrack);
-            // Unmute received audio
-            audio.receiver.track.enabled = true;
-            // // Set the direction to sendrecv (requires negotiation)
-            audio.setDirection("sendrecv");
-          </pre>
+        <pre class="example highlight">//assume we have an audio transceiver and a microphone track named micTrack
```
taylor-b: "// Assume..." instead of "//assume"

webrtc.html
```diff
@@ -6872,44 +6868,46 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+audio.sender.replaceTrack(micTrack);
+// Unmute received audio
+audio.receiver.track.enabled = true;
+// // Set the direction to sendrecv (requires negotiation)
```
taylor-b: Extra "//"

webrtc.html
```diff
@@ -9323,50 +9321,36 @@ <h2 id="sec.stats-model">Statistics Model</h2>
...
-            var fractionLost = (packetsSent - packetsReceived) / packetsSent;
+            const fractionLost = (packetsSent - packetsReceived) / packetsSent;
         }
-    }
+    });
 }
-</pre>
+
+try {
+    const selector = pc.getSenders()[0].track;
```
taylor-b: Can change this to just call getStats on the sender directly.

webrtc.html
```diff
@@ -10614,127 +10587,124 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
-            if (started &amp;& videoSendTrack) {
-              video.sender.replaceTrack(videoSendTrack);
+    pc.ontrack = async (evt) =&gt; {
+        try {
+            if (evt.track.kind === "audio") {
+                if (answerer) {
+                    audio = evt.transceiver;
+                    audio.setDirection("sendrecv");
+                    if (started &amp;&amp; audioSendTrack) {
+                        await audio.sender.replaceTrack(audioSendTrack);
```
taylor-b: Does it make sense to await `replaceTrack` here, which will delay setting `remoteView.srcObject`?

webrtc.html
Jxck: `<pre>` shows input as-is, so if adding extra line break between tag and code, it shows extra space in rendered HTML.
like below (L: before, R: after)

screen shot 2017-10-13 at 16 04 58

webrtc.html
Jxck: thats true if replaceTrack succeed, but what will happen if failed ?
I think handle error when failed without replace srcObject seems better.

taylor-b: Alright. I guess it's just an example so no reason to get too optimized.

webrtc.html
taylor-b: Hmm, I tested with Chrome and Firefox and don't see a difference.

The [HTML5 spec](https://www.w3.org/TR/html5/grouping-content.html#the-pre-element) says:

> a leading newline character immediately following the pre element start tag is stripped

> A single newline may be placed immediately after the start tag of pre and textarea elements. If the element's contents are intended to start with a newline, two consecutive newlines thus need to be included by the author.

So I wonder how you were able to get different behavior?

Jxck: I found that this caused by `<code>` after `<pre>` only inserted in Example 1 and 2.
other example doesn't has `<code>` so no effect for line break.
(this `<code>` seems for build highlighting and remove after render)

I'll insert break line except Ex 1, 2.

webrtc.html
```diff
@@ -10898,92 +10886,82 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
-    } else
-        pc.addIceCandidate(message.candidate).catch(logError);
-};
+              if (isinitiator) {
+                  // create data channel and setup chat
+                  channel = pc.createdatachannel("chat");
+                  setupchat();
+              } else {
+                  // setup chat on incoming data channel
+                  pc.ondatachannel = (evt) =&gt; {
```
jan-ivar: Not in this PR, but this reminds me we really should show an example for `negotiated` data channels as well, maybe even instead of this. cc @taylor-b

webrtc.html
```diff
@@ -11017,74 +10995,76 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
-          // empty the buffer to not play any tone after "2"
-          sender.dtmf.insertDTMF("");
-  };
-  sender.dtmf.insertDTMF("123");
-} else
-    log("DTMF function not available");
-</pre>
-      
+        if (sender.dtmf.canInsertDTMF) {
+            sender.dtmf.ontonechange = (e) =&gt; {
```
jan-ivar: Lose `()` around `e` and s/`e`/`event`/

webrtc.html
```diff
@@ -11017,74 +10995,76 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
+                if (e.tone == "2") {
+                    // empty the buffer to not play any tone after "2"
+                    sender.dtmf.insertDTMF("");
+                }
+            };
+            sender.dtmf.insertDTMF("123");
```
jan-ivar: I think we should generally encourage methods before onxxx handlers, since we queue a task for all events.

Also, if we wanted to get fancy we could do something like:
```js
if (sender.dtmf.canInsertDTMF) {
    sender.dtmf.insertDTMF("123");
    await new Promise(r => sender.dtmf.ontonechange = e => e.tone == "2" && r());
    // empty the buffer to not play any tone after "2"
    sender.dtmf.insertDTMF("");
} else {
    console.log("DTMF function not available");
}
```

webrtc.html
```diff
@@ -11017,74 +10995,76 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
+        if (sender.dtmf.canInsertDTMF) {
+            const duration = 500;
+            sender.dtmf.ontonechange = (e) =&gt; {
+                if (!e.tone) return;
+                // light up the key when playout starts
+                lightKey(e.tone);
+                // turn off the light after tone duration
+                setTimeout(lightKey, duration, "");
+            };
+            sender.dtmf.insertDTMF(sender.dtmf.toneBuffer + "1234");
```
jan-ivar: I suppose we could use `await` here with some finagling:
```js
const wait = ms => new Promise(resolve => setTimeout(resolve, ms));

if (sender.dtmf.canInsertDTMF) {
    const duration = 500;
    sender.dtmf.insertDTMF(sender.dtmf.toneBuffer + "1234", duration);
    sender.dtmf.ontonechange = async event => {
        if (!event.tone) return;
        lightKey(event.tone); // light up the key when playout starts
        await wait(duration);
        lightKey("");         // turn off the light after tone duration
    };
} else {
    console.log("DTMF function not available");
}
```

webrtc.html
Jxck: > I think we should generally encourage methods before onxxx handlers, since we queue a task for all events.

is it better calling method which fires X event should invoke after onX handler has set ?

webrtc.html
```diff
@@ -10528,83 +10527,77 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
+              pc = new RTCPeerConnection(configuration);
 
-    // once remote track arrives, show it in the remote video element
-    pc.ontrack = function (evt) {
-        // don't set srcObject again if it is already set.
-        if (!remoteView.srcObject)
-          remoteView.srcObject = evt.streams[0];
-    };
+              // send any ice candidates to the other peer
+              pc.onicecandidate = (evt) =&gt; {
```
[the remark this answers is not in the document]
Jxck: > Also, Is the evt abbreviation here to save horizontal space? I'd prefer s/signalingChannel/signaling/ to that I think.

not my style.
I agree.

webrtc.html
```diff
@@ -10615,126 +10608,125 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
-          }
-        }
-        // don't set srcObject again if it is already set.
-        if (!remoteView.srcObject)
-          remoteView.srcObject = evt.streams[0];
-    };
+              // once remote track arrives, show it in the remote video element
+              pc.ontrack = async (evt) =&gt; {
+                  try {
+                      if (evt.track.kind === "audio") {
```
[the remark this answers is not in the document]
Jxck: s/`==`/`===`/ rest of all instead.

webrtc.html
jan-ivar: Either works, is my point. There's a common misunderstanding that you *have to* declare the onxxx first, which is untrue, modulo [bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1394072). This lets us use `await` as I show.

webrtc.html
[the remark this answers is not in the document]
jan-ivar: Aah! Let's not have this discussion here, and leave it alone then. Crockford is wrong btw.

alvestrand:
What's the formal status of async/await? Is it in a version of JS that is published, or is it just in the "latest draft"?

dontcallmedom:
it's in the [latest **standard** of EcmaScript}(http://www.ecma-international.org/ecma-262/8.0/index.html#sec-async-function-definitions) (known as ES8 or ES2017)

alvestrand:
Thanks for the pointer! Then I have no objection to using it.

Jxck:
@jan-ivar I've done fixing. check please.

webrtc.html
```diff
@@ -10898,92 +10859,71 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
+                      channel = event.channel;
+                      setupChat();
+                  };
+              }
+          }
 
-function sendChatMessage(msg) {
-    channel.send(msg);
-}
+          signaling.onmessage = async {desc, candidate} =&gt; {
```
jan-ivar: `()`

webrtc.html
```diff
@@ -10528,83 +10533,67 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
+          }
 
-</pre>
+          signaling.onmessage = async {desc, candidate} =&gt; {
+              try {
+                  if (desc) {
+                      // if we get an offer, we need to reply with an answer
+                      if (desc.type == "offer") {
+                          await pc.setRemoteDescription(desc);
+                          const stream = await navigator.mediaDevices.getUserMedia(constraints);
```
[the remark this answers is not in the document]
Jxck: > could deal with wordwrap in another issue
+1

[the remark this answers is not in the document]
Jxck: but I'll fixing indent by two.

adam-be:
Very nice PR (again). Running my linting script (mentioned in https://github.com/w3c/mediacapture-main/pull/486) produces a bunch of warnings. Mainly because of the spaces between braces and content. But I'm almost in favor of keeping the spaces.

adam-be:
After we merge this PR, let's line up the examples with the left margin like we have in mediacapture-main.

webrtc.html
```diff
@@ -10898,92 +10860,71 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
-        .then(function () {
-            // send the offer to the other peer
-            signalingChannel.send(JSON.stringify({ desc: pc.localDescription }));
-        })
-        .catch(logError);
-    };
+          const signaling = new SignalingChannel(); // handles JSON.stringify/parse
+          const configuration = { iceServers: [{ urls: "stuns:stun.example.org" }] };
+          const pc;
+          const channel;
```
adam-be: These needs to be "let".

webrtc.html
```diff
@@ -10528,83 +10534,67 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
+          const pc = new RTCPeerConnection(configuration);
 
-    // once remote track arrives, show it in the remote video element
-    pc.ontrack = function (evt) {
-        // don't set srcObject again if it is already set.
-        if (!remoteView.srcObject)
-          remoteView.srcObject = evt.streams[0];
-    };
+          // send any ice candidates to the other peer
+          pc.onicecandidate = {candidate} =&gt; signaling.send({ candidate });
```
adam-be: We might want to use parenthesis around the `{candidate}`

webrtc.html
```diff
@@ -10615,126 +10604,123 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
-        audio.sender.replaceTrack(audioSendTrack);
-      }
-      if (video &amp;& videoSendTrack) {
-        video.sender.replaceTrack(videoSendTrack);
-      }
-    } else
-        pc.addIceCandidate(message.candidate).catch(logError);
-};
+          signaling.onmessage = event =&gt; {
+              if (!pc) warmup(true);
```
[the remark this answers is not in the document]
adam-be: We need `async` here for the `await`s below.

webrtc.html
```diff
@@ -11017,74 +10958,74 @@ <h2 id="sec.idp-error-handling">IdP Error Handling</h2>
...
-      lightKey(e.tone);
-      // turn off the light after tone duration
-      setTimeout(lightKey, duration, "");
-  };
-  sender.dtmf.insertDTMF(sender.dtmf.toneBuffer + "1234");
-} else
-    log("DTMF function not available");
-</pre>
-      
+        const wait = ms =&gt; new Promise(resolve =&gt; setTimeout(resolve, ms));
```
adam-be: s/wait/await/

Jxck:
@adam-be same for this PR with https://github.com/w3c/mediacapture-main/pull/486#issuecomment-339171317

Jxck:
I'm still working hard on this PR

adam-be:
Sorry, It was a bit of a misunderstanding and this PR got merged. Can you rebase your latest commits on a new PR branch?

Jxck:
ok

webrtc.html
jan-ivar: Yes, this is a syntax error otherwise, due to Javascript parsing limitation with destructuring.

adam-be:
@Jxck, when do you plan to make a PR of your pending changes? I would like to make the examples consistent between this and mediacapture-main (style, indentation and so on), but I don't want to break your upcoming PR. I have an other PR that needs to update the examples, but I'm waiting with that as well.

Jxck:
It's hard for me to solve every review done immediately.
so please go forward. I'll rebase them and make more PR later.

Issue #1612, opened 2017-09-22
No way to tell when SCTP association is closed


taylor-b:
This was brought up in https://github.com/w3c/ortc/issues/755, for a call forking use case. You can see the discussion there for more details, but the short explanation is: there are situations where the SCTP association may be closed and the application needs to generate a new offer/answer to restore it. And there's currently no way to tell when this happens. In ORTC, RTCSctpTransport has a state enum and event, so I'd propose using that. The states are very straightforward so I don't anticipate it being an issue for implementers.

taylor-b:
JSEP actually does call out this case of the answerer changing the SCTP port:

```
         If an SCTP association exists, and the remote SCTP port has
         changed, discard the existing SCTP association.  This includes
         the case when the PeerConnection state is "have-remote-
         pranswer".
```

So maybe we should reconsider marking this as CR-blocking.

The issue occurs when the SCTP port and TLS ID are changing due to a call being transitioned to a new remote endpoint (which is why "pranswer" would be used). But there's a chance the new endpoint uses the same SCTP port number as the previous endpoint. In which case the WebRTC endpoint would just see its SCTP association close, and would need to do a re-offer to establish the new association.

stefhak:
@taylor-b perhaps you can provide a little more context/summary? (I tried to read the ortc conversation, but it was hard to follow)

At the face of it, this looks like something that could be handled between the ```neogtiationneeded``` event, the ```RTCDataChannelState``` and the different events on ```RTCDataChannel``` (```error```, ```close```).

taylor-b:
I tried to summarize it in my previous comment; what part specifically needs more clarity?

Also, my reason for preferring an `RTCSctpTransport` state change as opposed to `negotiationneeded`:

> I'd prefer [an `RTCSctpTransport` state change event], because there also may be corner cases where the SCTP association is closed for some other reason [(besides this call forking scenario)], and renegotiating won't help. So better to let the application decide, given what it knows about the call topology. Plus, the state change event is already there in ORTC. 

And the `RTCDataChannel` events don't completely work, because they don't distinguish between an individual SCTP stream being closed and the whole association being closed.

aboba:
@taylor-b Addition of an `RTCSctpTransport` state attribute and change event makes sense to me.

PR #1620, opened 2017-10-05
Add RTCSctpTransportState

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1612

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/close-patchv13.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/293176b...6b32f38.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8038,7 +8038,7 @@ interface RTCTrackEvent : Event {
               <dt><dfn><code>state</code></dfn> of type <span class=
               "idlAttrType"><a>RTCSctpTransportState</a></span>, readonly</dt>
               <dd>
-                <p>The current state of the SCTP transport. The attribute returns
+                <p>The current state of the SCTP transport. On getting, it returns
                 the value of the <dfn>[[\SctpTransportState]]</dfn> internal slot.</p>
               </dd>
               <dt><dfn><code>maxMessageSize</code></dfn> of type <span class=
--- a/webrtc.html
+++ b/webrtc.html
@@ -8023,6 +8023,7 @@ interface RTCTrackEvent : Event {
     readonly        attribute RTCDtlsTransport transport;
     readonly        attribute RTCSctpTransportState state;
     readonly        attribute unsigned long    maxMessageSize;
+                    attribute EventHandler     onstatechange;
 };</pre>
           <section>
             <h2>Attributes</h2>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -8033,17 +8035,84 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
           </section>
         </div>
       </section>
+      <section id="rtcsctptransportstate*">
+        <h3><dfn>RTCSctpTransportState</dfn> Enum</h3>
+        <p><code>RTCSctpTransportState</code> indicates the state of the SCTP
+        transport.</p>
+        <div>
+          <pre class="idl">enum RTCSctpTransportState {
+    "new",
```
taylor-b: I'm not sure we need "new", since we decided the RTCSctpTransport is only created on applying an answer, in which case it already has enough information to connect? Though maybe we should revisit that decision.

webrtc.html
```diff
@@ -8033,17 +8035,84 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
                 <p>The transport over which all SCTP packets for data channels
                 will be sent and received.</p>
               </dd>
+              <dt><dfn><code>state</code></dfn> of type <span class=
+              "idlAttrType"><a>RTCSctpTransportState</a></span>, readonly</dt>
+              <dd>
+                <p>The current state of the SCTP transport. The attribute returns
+                the value of the <dfn>[[\SctpTransportState]]</dfn> internal slot.</p>
```
taylor-b: Similar to my comment on the PR that added [[DtlsTransportState]]: we need to either describe when the slot's value is changed, or not use a slot.

In this case it's probably pretty simple:

1. When it's first created, the value is `connecting`? (See below comment)
2. When the four-way startup handshake finishes, queue a task to update the state to `connected`
3. When a SHUTDOWN or ABORT chunk is received, queue a task to update the state to `closed`
4. Also can go to `closed` from applying a description that rejects the data m= section, or changes the sctp-port. Or if the application closes the PeerConnection,

webrtc.html
aboba: There might be an advantage to having the `RTCSctpTransport` created earlier, in which case the "new" state might be used. But I agree that if it is only created on applying an answer it should be initialized to "connecting".

aboba: Opening another issue relating to the transition text.

Issue #1406, opened 2017-06-18
 When ICE restarts results in connection to a new endpoint


aboba:
From fippo in Issue https://github.com/w3c/webrtc-pc/issues/1357: 

In the case of the remote fingerprint changing, when does the JS layer learn about the new DTLS transport object? The first chance is after SRD and I would also expect it to be available in [onselectedcandidatepairchange](http://w3c.github.io/webrtc-pc/#dom-rtcicetransport-onselectedcandidatepairchange).

I would expect the ice transport not to change during an ice restart. This is also consistent with the ability to query past candidates in getStats. This gets somewhat awkward in the case of the remote fingerprint changing :-/

aboba:
@fippo It makes sense that the changes would become visible after SRD.  However, exactly what happens is not well specified at present.  For example, does the RTCDtlsTransport remain the same and make an (odd) transition back to "new"? 

Or does the old RTCDtlsTransport.state transition to "closed", along with a new RTCDtlsTransport created in the "new" state?  If so, does the onstatechange EventHandler for the old RTCDtlsTransport get moved to the new RTCDtlsTransport automagically, or does the application somehow need to figure out that the RTCDtlsTransport has changed and set the EventHandler itself?

aboba:
Related Issue https://github.com/w3c/webrtc-pc/issues/1178

fippo:
> For example, does the RTCDtlsTransport remain the same and make an (odd) transition back to "new"?

I think there should be a new DTLS transport object since the internal keying material on a DTLS transport is immutable once set. Also packets on the old dtls transport still need to be decrypted (
spec pointer from [this old comment](https://github.com/w3c/ortc/issues/93#issuecomment-47406008) [here](https://tools.ietf.org/html/rfc5245#section-9.1.1.1)
Ideally at least, getting this use-case 100% seamless might be too ambitious.

@taylor-b maybe you have a chance to look at [this bug](https://bugs.chromium.org/p/webrtc/issues/detail?id=5768) again while figuring out how your lib handles this?

But creating a new object makes things worse for RTPSender and RTPReceiver:
* when does the transport change? Once its up as no media is received on the new transport before then? Note that we do not wait for media initially so I think the answer is "during SLD/SRD".
* should there be a way to subscribe to the transport change? I do not see a use-case for that.

[...]

PR #1623, opened 2017-10-10
Describe how transport objects are assigned to senders/receivers.

### LINKED ISSUE #1178: Need to describe when ICE and DTLS transport objects are created/changed
The specification mentions that:

> RTCDtlsTransport objects are constructed as a result of calls to setLocalDescription() and setRemoteDescription().

But it needs to go into more detail. Which `setLocalDescription`/`setRemoteDescription` calls cause the transports to be created (and returned by the sender/receiver `transport` attribute)? Only local descriptions? Any offer? The "applying a description" steps should answer this question.

We should also describe when the transports are changed as a result of bundling/RTCP muxing being negotiated.

taylor-b:
Fixes #1178.

As decided in the June virtual interim, DTLS/ICE transport objects will
only be created as a result of setting a local description. Also,
applying an answer may change the mapping between media descriptions and
transports (according to the rules in BUNDLE), which may result in
`sender.transport` (for example) pointing to a different object than
before.

This PR puts the actual responsibility for figuring out the media/
transport mapping on BUNDLE, using the "media transport" term, which in
this case is something that encapsulates the pair of ICE and DTLS
transports. Note that BUNDLE itself does not use that term (yet), but it
will, optimistically, if
https://github.com/cdh4u/draft-sdp-bundle/pull/19 is merged.

Note that this PR does not yet handle rollbacks, or RTCP mux
negotiation.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1178_transport_creation.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/c5eb099...taylor-b:a212628.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -78,6 +78,7 @@
     <dfn>MediaStreamConstraints</dfn> are defined in [[!GETUSERMEDIA]].</p>
     <p>The term <dfn>Blob</dfn> is defined in [[!FILEAPI]].</p>
     <p>The term <dfn>media description</dfn> is defined in [[!RFC4566]].</p>
+    <p>The term <dfn>media transport</dfn> is defined in [[!RFC7656]].</p>
     <p>The term <dfn>generation</dfn> is defined in [[!TRICKLE-ICE]] Section 2.</p>
     <p>The term <dfn data-cite="!WEBRTC-STATS#dom-rtcstatstype">RTCStatsType</dfn> is defined in [[!WEBRTC-STATS]].
     <p>When referring to exceptions, the terms <dfn
@@ -1523,19 +1524,19 @@
                   </li>
                   <li>
                     <p>If <var>description</var> is set as a local description,
-                    then run the following steps for each media description
+                    then run the following steps for each <a>media description</a>
                     in <var>description</var> that is not yet <a>associated</a> with
                     an <code><a>RTCRtpTransceiver</a></code> object:</p>
                     <ol>
                       <li>
                         <p>Let <var>transceiver</var> be the <code>
                         <a>RTCRtpTransceiver</a></code> used to create the
-                        media description.</p>
+                        <a>media description</a>.</p>
                       </li>
                       <li>
                         <p>Set <var>transceiver</var>'s <code><a data-for=
                         "RTCRtpTransceiver">mid</a></code> value to the mid of
-                        the corresponding media description.</p>
+                        the corresponding <a>media description</a>.</p>
                       </li>
                       <li>
                         <p>If <var>transceiver</var>'s <a>[[\Stopped]]</a> slot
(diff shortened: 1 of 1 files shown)


taylor-b:
As mentioned before, we really need to refactor the "set a description" steps (we're now 6 levels of ordered list deep), but I'd rather do that in a standalone PR, and maybe talk a bit about it amongst editors first. Here's an issue for that: https://github.com/w3c/webrtc-pc/issues/1624

Anyway, I can definitely extend this PR to handle rollbacks and RTCP mux being negotiated in an answer, but I thought an initial review pass may be valuable first.

stefhak:
Thanks @taylor-b ! Does this PR also address #1406 or would that be a separate PR?

taylor-b:
In a way, since it doesn't say a new RTCIceTransport _should_ be created on an ICE restart. But it would probably be worthwhile to add a paragraph to the RTCIceTransport/RTCDtlsTransport descriptions that clarifies their scope. I was going to do that in a separate PR.

webrtc.html
```diff
@@ -78,6 +78,7 @@
     <dfn>MediaStreamConstraints</dfn> are defined in [[!GETUSERMEDIA]].</p>
     <p>The term <dfn>Blob</dfn> is defined in [[!FILEAPI]].</p>
     <p>The term <dfn>media description</dfn> is defined in [[!RFC4566]].</p>
+    <p>The term <dfn>media transport</dfn> is defined in [[!RFC7656]].</p>
```
taylor-b: I'm realizing now that "media transport" may not be the right term. This RFC also says "An RTP session uses at least two media transports: one for sending and one for receiving." I'm not sure if this is just because it was written with unidirectional transports in mind (5-tuples), or if a media transport is fundamentally intended to be unidirectional.

Also, who's to say that a new DTLS association is not a new media transport? There's nothing that defines whether that's the case or not.

Really, I'd like a term that means "the transport layer of a media description, which (using BUNDLE) multiple media descriptions may share". But I don't know if that term already exists.

alvestrand: The term that ICE uses is "component", I found when doing that review.

webrtc.html
```diff
@@ -1547,7 +1548,58 @@
...
                         to an <code><a>RTCRtpTransceiverDirection</a></code>
                         value representing the direction of the corresponding
-                        media description.</p>
+                        <a>media description</a>.</p>
+                      </li>
+                      <li>
+                        <p>
+                          If the <a>media description</a> is indicated as using
+                          an existing <a>media transport</a> according to
+                          [[!BUNDLE]], let <var>transport</var> and
```
alvestrand: Travis didn't like BUNDLE as a reference.

taylor-b: I added it since initially uploading (the latest Travis build passed).

webrtc.html
```diff
@@ -1547,7 +1548,58 @@
...
+                          <var>rtcpTransport</var> be the
+                          <code><a>RTCDtlsTransport</a></code> objects
+                          representing the RTP and RTCP components of that
```
gloinul: The use of "components" without any qualifier make me react, but I can't for my life think of a better term to use here, unless one want to expand this sentence quite a lot. I do wonder if it would be clearer to add a parenthesis after "components" saying "(as defined by ICE).

aboba: I might say "transporting RTP and RTCP, respectively".

aboba: It actually might be cleaner if creating a transceiver did create the underlying DtlsTransport and IceTransport objects.

PR #1631, opened 2017-10-11
Explain the scope of DTLS/ICE transport objects in more detail.

### LINKED ISSUE #1406: When ICE restarts results in connection to a new endpoint

taylor-b:
Fixes #1406.

In the September virtual interim, we decided that new DTLS associations
and ICE restarts should *not* result in new objects, since this
introduces more complexity to the API for arguably little benefit.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1406_transport_objects_scope.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/c5eb099...taylor-b:6c33f6c.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6953,7 +6953,19 @@ sender.setParameters(params)
       about the underlying transport and the security added.
       <code><a>RTCDtlsTransport</a></code> objects are constructed as a result
       of calls to <code>setLocalDescription()</code> and
-      <code>setRemoteDescription()</code>.</p>
+      <code>setRemoteDescription()</code>. Each
+      <code><a>RTCDtlsTransport</a></code> object represents the DTLS transport
+      layer for the RTP or RTCP <code><a data-link-for="RTCIceTransport">component</a></code>
+      of a specific <code><a>RTCRtpTransceiver</a></code>, or a group of
+      <code><a>RTCRtpTransceiver</a></code>s if such a group has been
+      negotiated via [[!BUNDLE]].</p>
+
+      <div class="note">A new DTLS association for an existing
+      <code><a>RTCRtpTransceiver</a></code> will be represented by an existing
+      <code><a>RTCDtlsTransport</a></code> object, whose <code><a
+      data-link-for="RTCDtlsTransport">state</a></code> will be updated accordingly,
+      as opposed to being represented by a new object.</div>
+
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6953,7 +6953,19 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       about the underlying transport and the security added.
       <code><a>RTCDtlsTransport</a></code> objects are constructed as a result
       of calls to <code>setLocalDescription()</code> and
-      <code>setRemoteDescription()</code>.</p>
+      <code>setRemoteDescription()</code>. Each
+      <code><a>RTCDtlsTransport</a></code> object represents the DTLS transport
+      layer for the RTP or RTCP <code><a data-link-for="RTCIceTransport">component</a></code>
+      of a specific <code><a>RTCRtpTransceiver</a></code>, or a group of
+      <code><a>RTCRtpTransceiver</a></code>s if such a group has been
+      negotiated via [[!BUNDLE]].</p>
```
taylor-b: My intent here was for the prose above to definitively describe the scope of the objects, linking it to SDP (by linking it to transceivers). Where the note below explains the possibly-not-obvious implication of this definition.

stefhak:
To me it looks like this captures what was agreed at the June VI, and then confirmed at the September VI.

webrtc.html
aboba: I think more detail is needed (e.g. whether the objects are created on application of an Offer or an Answer).

taylor-b: I'd hoped that https://github.com/w3c/webrtc-pc/pull/1623 would do this, by specifying exactly when the objects should be created (and by omission, when they shouldn't be).

But if we need a note to mention that addTransceiver doesn't create the objects, should that go in the addTransceiver description itself, or here?

PR #1632, opened 2017-10-11
Adding relativeBitrate parameter to RTCRtpEncodingParameters.

### LINKED ISSUE #1625: RTCPriorityType combines relative bitrate with QoS priority, which applications may not want.
This may really be an issue with https://datatracker.ietf.org/doc/draft-ietf-rtcweb-transports/, but also may be a result of me interpreting it in a way other than intended, so I thought I'd raise the issue here first.

Anyway, rtcweb-transports says:

>    The WebRTC prioritization model is that the application tells the
>    WebRTC endpoint about the priority of media and data that is
>    controlled from the API.
>    ...
>    The priority settings affect two pieces of behavior: Packet send
>    sequence decisions and packet markings.  Each is described in its own
>    section below.

The sections below are "Local prioritization" and "Usage of Quality of Service - DSCP and Multiplexing". The "local prioritization" section gives this guidance:

>    When an WebRTC endpoint has packets to send on multiple streams that
>    are congestion-controlled under the same congestion control regime,
>    the WebRTC endpoint SHOULD cause data to be emitted in such a way
>    that **each stream at each level of priority is being given
>    approximately twice the transmission capacity (measured in payload
>    bytes) of the level below.**
>
>    Thus, when congestion occurs, a "high" priority flow will have the
>    ability to send 8 times as much data as a "very-low" priority flow if
>    both have data to send.  This prioritization is independent of the
>    media type.  The details of which packet to send first are
>    implementation defined.
> 
>    For example: If there is a high priority audio flow sending 100 byte
>    packets, and a low priority video flow sending 1000 byte packets, and
>    outgoing capacity exists for sending >5000 payload bytes, it would be
>    appropriate to send 4000 bytes (40 packets) of audio and 1000 bytes
>    (one packet) of video as the result of a single pass of sending
>    decisions.

This has two significant issues I can see:

* It only allows ratios of 1:2:4:8. This is not granular enough to be very useful.
* A single enum is used for both relative bitrate, and relative QoS priority. But in my experience it's common for an application to want a flow that uses _less_ bitrate to have a _higher_ QoS priority. This may even be the _more_ common situation.

So, why do we have one enum for both things? I'd propose defining RTCPriorityType as something that only impacts the priority in network queues (QoS-level priority and SCTP priority, as previously discussed), and use something else to control the relative bitrate allocation.

For example, a floating point value attached to each `RTCEncodingParameters` (say, `relativeCapacity`), where an encoding with twice the `relativeCapacity` of another encoding will be given twice the transmission capacity. If implementations can handle ratios of 1:2:4:8, I don't see any reason why they shouldn't be able to handle arbitrary ratios.

taylor-b:
Fixes #1625.

This provides more fine-grained control over the relative amount of bits
different encodings should be allocated. This allows the webrtc
implementation to respond quickly to changes in available capacity; the
alternative would be for an application to inspect the stats
periodically, and continually adjust `maxBitrate`, which would result in
wasted bandwidth and slower responsiveness to changing bandwidth
estimates (since polling stats and calling getParameters/setParameters
is not instantaneous).

This assumes that rtcweb-transports will update its definition of
"priority," such that it only controls the priority of streams in
network queues (via DSCP and SCTP priority values), and not the
relative amount of bitrate they're granted.

***
[Preview](https://s3.amazonaws.com/pr-preview/taylor-b/webrtc-pc/issue_1625_relativeBitrate.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/c5eb099...taylor-b:634fcc5.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5794,6 +5794,7 @@ sender.setParameters(params)
              boolean             active = true;
              RTCPriorityType     priority = "low";
              unsigned long       ptime;
+             double              relativeBitrate = 1.0;
              unsigned long       maxBitrate;
              double              maxFramerate;
              DOMString           rid;
@@ -5863,6 +5864,21 @@ sender.setParameters(params)
               data-jsep= "applying-a-remote-desc">[[!JSEP]]</span>. Note that
               the user agent MUST still respect the limit imposed by any
               "maxptime" attribute, as defined in [[!RFC4566]], Section 6.</p>
+            </dd>
+            <dt><dfn><code>relativeBitrate</code></dfn> of type <span class=
+            "idlMemberType"><a>double</a></span></dt>
+            <dd>
+              <p>Indicates the relative amount of bitrate that this encoding
+              SHOULD be allocated when congestion occurs, relative to other
+              encodings being sent under the same congestion control regime.
+              For example, if two encodings use values of 1.0 and 1.5,
+              respectively, and the congestion controller determines that 5Mbps
+              are available to allocate, the encodings SHOULD be allocated
+              2Mbps and 3Mbps, respectively. The encoding may also be further
+              constrained by other limits (such as
+              <code><a>maxBitrate</a></code> or per-transport or per-session
+              bandwidth limits), resulting in it using less than its available
+              share.</p>
             </dd>
             <dt><dfn><code>maxBitrate</code></dfn> of type <span class=
             "idlMemberType"><a>unsigned long</a></span></dt>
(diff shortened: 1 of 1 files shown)


stefhak:
I think it would be premature to merge this one now. Perhaps a discussion at TPAC?
I'm also afraid of adding one thing at a time (such as we did with ```maxBitrate```, and here ```relativePriority``` is proposed) could lead to us having a patchwork that does not accomplish what is needed in the end.

alvestrand:
Discussion at TPAC suggested heading in a different direction.
Once another PR is available, this will be closed.

stefhak:
Looks like this PR is (to be) replaced by https://github.com/alvestrand/webrtc-dscp-exp.

PR #1638, opened 2017-10-20
Let setLocalDescription use [[LastAnswer/Offer]] internal slots

### LINKED ISSUE #1164: Missing detail on obtaining lastOffer/lastAnswer in setLocalDescription
In `setLocalDescription`, the sentence "Let lastOffer be the result returned by the last call to `createOffer`" does not specify how lastOffer can be obtained. In the definition of `createOffer`, there is no instruction that the created offer have to be saved in some internal slot for later comparison by `setLocalDescription`. The same issue also applies to lastAnswer returned by `createAnswer`.

I'd also like some clarification for testing the behavior of `setLocalDescription`:

- If `createOffer`/`createAnswer` is never called before, what is the value of lastOffer/lastAnswer? In those case should the steps check the sdp value against `undefined`, or reject the promise with the reason `createOffer`/`createAnswer` have never been called?

- By saying that the sdp matches `lastOffer`/`lastAnswer`, is that string equality or structural equality of the parsed SDP?

- If `createOffer`/`createAnswer` is called multiple times with different parameters before `setLocalDescription`, is it right to say that only the last offer returned is valid and the earlier results are rejected?

- Is it right to say that even modification or addition of informational field, such as `s=` and `i=`, are to be rejected by `setLocalDescription`?

soareschen:
Fixes #1164.

***
[Preview](https://s3.amazonaws.com/pr-preview/soareschen/webrtc-pc/sld-last-offer-answer.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webrtc-pc/da266a1...soareschen:951dd1b.html)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2615,19 +2615,16 @@ interface RTCPeerConnection : EventTarget  {
                 <ol>
                   <li>Let <var>description</var> be the first argument to
                   <code>setLocalDescription</code>.</li>
-                  <li>Let <var>lastOffer</var> be the result returned by the
-                  last call to <code>createOffer</code>.</li>
-                  <li>Let <var>lastAnswer</var> be the result returned by the
-                  last call to <code>createAnswer</code>.</li>
                   <li>If <code><var>description</var>.sdp</code> is the
                   empty string and <code><var>description</var>.type</code>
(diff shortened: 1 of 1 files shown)


alvestrand:
LGTM

PR #1647, opened 2017-10-26
Replace setDirection() with writable direction attribute

### LINKED ISSUE #1643: Why is setDirection a method?
Sure, [transceiver.setDirection(direction)](http://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver) can throw, but so can attributes!

`transceiver.direction = direction` seems superior. I checked with our DOM folks, and they agree.

The algorithm seems like a shoo-in, setting the (today readonly) `direction` attribute, so no change there.

Also, since setting `direction` merely sets the negotiationneeded flag, there's zero chance of us ever wanting this method to return a promise or anything like that.

We're going over final review of our implementation slated for Firefox 59, and thought this made sense.

@taylor-b @adam-be WDYT?

adam-be:
Work in progress. I don't want to update the examples until the big example rewrite is landed.

Fixes #1643

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6775,9 +6775,8 @@ async function updateParameters() {
     [SameObject]
     readonly        attribute RTCRtpReceiver              receiver;
     readonly        attribute boolean                     stopped;
-    readonly        attribute RTCRtpTransceiverDirection  direction;
+                    attribute RTCRtpTransceiverDirection  direction;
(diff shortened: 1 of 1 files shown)


adam-be:
This patch looks much better with the patience diff algorithm.

stefhak:
Would any apps break as a result of this change?

alvestrand:
@stefhak I don't think anyone has implemented the API yet, so no. Chrome certainly hasn't.

PR #1649, opened 2017-10-31
Add a new value to RTCIceTransportPolicy.

### LINKED ISSUE #1644: Adding more values to RTCIceTransportPolicy Enum
[1] defines 4 specific modes for candidate collection. However, RTCIceTransportPolicy Enum only has 2 values. Currently, user agents choose which mode is going to be applied in certain circumstances. Should we add more values to RTCIceTransportPolicy Enum so applications have the chance to change UA's default behavior? UA may request permission from the end user before changes are made.

Here is an example of different candidate collection behaviors between Chrome and Firefox:
Run data channel sample [2] in a machine equipped with two network adapters. Chrome collects candidates on all network adapters. Firefox only collects candidates on default route and associated local addresses.

[1] https://tools.ietf.org/html/draft-ietf-rtcweb-ip-handling-04
[2] https://webrtc.github.io/samples/src/content/datachannel/basic/

jianjunz:
This PR adds a new value "default-route-and-relay" which corresponding to Mode 3 of RTCWEB-IP-HANDLING.

Ref: #1644.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -117,7 +117,7 @@
         <div>
           <pre class="idl">dictionary RTCConfiguration {
              sequence&lt;RTCIceServer&gt;   iceServers;
-             RTCIceTransportPolicy    iceTransportPolicy = "all";
+             RTCIceTransportPolicy    iceTransportPolicy = "default-route-and-relay";
              RTCBundlePolicy          bundlePolicy = "balanced";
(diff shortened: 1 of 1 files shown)


aboba:
Notes from TPAC (https://www.w3.org/2017/11/06-webrtc-minutes.html) do not indicate consensus to accept this change.

alvestrand:
Closing per TPAC minutes.

PR #1656, opened 2017-11-06
RTCSctpTransport: Specify special cases for maxMessageSize

### LINKED ISSUE #1446: RTCSctpTransport.maxMessageSize 0 case
This spec's `RTCSctpTransport.maxMessageSize` description says:

> The maximum size of data that can be passed to RTCDataChannel's send() method.

In [the related SDP spec](https://tools.ietf.org/html/draft-ietf-mmusic-sctp-sdp-26#page-8), it's specified that

> If the SDP 'max-message-size' attribute contains a maximum message
size value of zero, it indicates the SCTP endpoint will handle
messages of any size, subject to memory capacity etc.

So, there is a `maxMessageSize = 0` case we need to handle in some way.

While messages of arbitrary size may not be possible in the browser as this specification does not provide a way to deliver partial messages (e.g. for this spec, all messages need to be buffered until they are complete), other implementations may support this (one already does).

This has been discussed in https://github.com/w3c/ortc/issues/626 as well.

adam-be:
(including updating via O/A exchange)

Fixes #1446

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8250,8 +8250,8 @@ interface RTCTrackEvent : Event {
           <li>
             <p>Let <var>remoteMaxMessageSize</var> be the value of the
             "max-message-size" SDP attribute read from the remote description,
-            as described in [[!SCTP-SDP]], or 65535 if the attribute is
-            missing.</p>
+            as described in [[!SCTP-SDP]] (section 6), or 65535 if the
+            attribute is missing.</p>
           </li>
           <li>
             <p>Let <var>canSendSize</var> be the number of bytes that this
--- a/webrtc.html
+++ b/webrtc.html
@@ -8259,7 +8259,12 @@ interface RTCTrackEvent : Event {
             the implementation can handle messages of any size.</p>
           </li>
           <li>
-            <p>If either <var>remoteMaxMessageSize</var> or
+            <p>If both <var>remoteMaxMessageSize</var> and
+            <var>canSendSize</var> are 0, set <a>[[\MaxMessageSize]]</a> to the
+            positive Infinity value.</p>
+          </li>
+          <li>
+            <p>Else, if either <var>remoteMaxMessageSize</var> or
             <var>canSendSize</var> is 0, set <a>[[\MaxMessageSize]]</a> to the
             larger of the two.</p>
           </li>
@@ -8272,7 +8277,7 @@ interface RTCTrackEvent : Event {
           <pre class="idl">[Exposed=Window] interface RTCSctpTransport {
     readonly        attribute RTCDtlsTransport transport;
     readonly        attribute RTCSctpTransportState state;
-    readonly        attribute unsigned long    maxMessageSize;
+    readonly        attribute unrestricted double maxMessageSize;
                     attribute EventHandler     onstatechange;
 };</pre>
           <section>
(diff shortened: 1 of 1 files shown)


adam-be:
@lgrahl, can you take a look please?

webrtc.html
```diff
@@ -8225,9 +8231,43 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+        <a>RTCSctpTransport</a></code> run the following
+        steps:</p>
+        <ol>
+          <li>
+            <p>Let <var>transport</var> be the <code><a>RTCSctpTransport</a>
+            </code> object to be updated.</p>
+          </li>
+          <li>
+            <p>Let <var>remoteMaxMessageSize</var> be the value of the
+            "max-message-size" SDP attribute read from the remote description,
```
lgrahl: Do we need to add *from the SCTP transport's SDP media section* or is this clear to the reader?

webrtc.html
```diff
@@ -8225,9 +8231,43 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
...
+            as described in [[!SCTP-SDP]], or 65535 if the attribute is
+            missing.</p>
+          </li>
+          <li>
+            <p>Let <var>canSendSize</var> be the number of bytes that this
+            client can send (i.e. the size of the local send buffer) or 0 if
+            the implementation can handle messages of any size.</p>
```
lgrahl: Note: In the W3C API, sending messages of any size is impossible at the moment (the API lacks the possibility to send partial messages) but I think it's good to have it in there already to be future-proof (and as a note for implementations outside of the browser).

webrtc.html
adam-be: Most of our references with specific section numbers are automatically generated to not get out of sync with the external document. But there are "hard-coded" one too, so I guess it woudn't be too bad to add a reference to a top-level section here.

alvestrand:
@adam-be will look into writing a test for this before merging.

adam-be:
Test: https://github.com/w3c/web-platform-tests/pull/8495

webrtc.html
```diff
@@ -8224,10 +8230,44 @@ <h3><dfn>RTCSctpTransport</dfn> Interface</h3>
...
+          </li>
+          <li>
+            <p>If either <var>remoteMaxMessageSize</var> or
+            <var>canSendSize</var> is 0, set <a>[[\MaxMessageSize]]</a> to the
+            larger of the two.</p>
```
lgrahl: Although this is algorithmically fine for the *both `0` case* as the result of `max(0, 0) == 0` is correct, it might be a bit confusing to the reader.

alvestrand:
API style level question: Is there an argument for why a dedicated value (0) for maxMessageSize is a more user-friendly API than maxMessageSize = Number.MAX_SAFE_INTEGER, Number.MAX_VALUE or Number.POSITIVE_INFINITY?
Using these large numbers would have the advantage of requiring no checking against zero in the common case of "is this message too large to send?".

adam-be:
@alvestrand, we borrow the magic number zero from the signaling. But I agree that `Number.MAX_SAFE_INTEGER` would make a much better API.

lgrahl:
Agreed, API-wise, `Number.POSITIVE_INFINITY` is much nicer (I wouldn't use `.MAX_SAFE_INTEGER` or `.MAX_VALUE` though). The conversion from `0` to `Number.POSITIVE_INFINITY` should be at the very end of the description.

lgrahl:
Another note: This would probably require changing the type of the `maxMessageSize` field from `unsigned long` to something else?

alvestrand:
UNSIGNED LONG is fine for MAX_SAFE_INTEGER but not for the infinities.

lgrahl:
Okay. But I'd really like to avoid those discrete integers as they semantically don't fit into the scope of *arbitrary sized messages*. While I agree that messages of that size are an academic problem, I'd prefer a semantically correct API.

adam-be:
I'll update the PR to use `Number.POSITIVE_INFINITY` when no size-limit in enforced.

Issue #1617, opened 2017-10-04
Adopt "test as you commit" policy


foolip:
An increasing number of groups are beginning to make web-platform-tests part of their day-to-day work, and more are considering it, for example in https://github.com/w3c/dap-charter/issues/48.

In a [Improving interop with web-platform-tests](https://bit.ly/hackfest-wpt) talk (slides 16-17 and 23) I gave here at the https://webengineshackfest.org/ I outlined the strategy, "tipping point" idea and the remaining work.

WebRTC is at the same time challenging and in an unusually good situation testing-wise. The challenge is that http://wpt.fyi/webrtc is very red because of a lack of automation. Chromium's ecosystem infra team is investing in [fixing this in Q4](https://docs.google.com/document/d/1GZquS8Jra47E5hIMH2ObOb_7oQFgk8f9MptUAjF0LFI/edit?usp=sharing) (search for "WebRTC").

What makes things better for WebRTC than most places is that testing activity is already very high (see [visualization](https://foolip.github.io/day-to-day/)) and a lot of work was done recently by @soareschen and @rwaldron to improve the test suite. See https://rwaldron.github.io/webrtc-pc/ for an an annotated version of the spec.

It could be something similar to the [SVG policy](https://github.com/w3c/svgwg/blob/master/CONTRIBUTING.md#tests), meaning:
 - Those who review spec changes ask for a corresponding PR in web-platform-tests.
 - If testing is currently impossible or impractical, issues with the type:untestable or type:missing-coverage labels are filed.
 - There are no hard rules, just a common understand of what's expected, analogous to code review practicies.

WDYT?

stefhak:
My personal view is that this sounds like a good WoW, but before we can adopt it the current test suite must decently covering (though it is in pretty good shape in this case as you point out).

Our plan is to discuss the test suite at TPAC (proposed agenda: https://lists.w3.org/Archives/Public/public-webrtc/2017Oct/0012.html), perhaps we can talk about this subject in conjunction with that? 

I also think we should not adopt a new WoW before we've transited to CR (something I hope we will manage real soon).

foolip:
> but before we can adopt it the current test suite must decently covering (though it is in pretty good shape in this case as you point out)

[...]

PR #1657, opened 2017-11-06
Document "test as you commit" policy in CONTRIBUTING.md

foolip:
At TPAC it was resolved to try this policy for webrtc-pc.

Fixes https://github.com/w3c/webrtc-pc/issues/1617.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,3 +45,6 @@ Example: `<div class='note'>FYI</div>`
 **Internal Slots**. Use upper camel case (initial upper case letter) with a leading backslash and enclosed by two brackets. The leading backslash prevents the text in double brackets from being interpreted as a reference. If a slot name is very general, like "Id", it might be useful to add more information to it to avoid name collisions and make it easier to identify its owning entity. For example, "Id" could be extended to "ComponentId".\
 Example (defining): `<dfn>[[\FooBar]]</dfn>`\
 Example (referencing): `<a>[[\FooBar]]</a>`
+
+### Tests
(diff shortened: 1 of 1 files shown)


foolip:
The wording here is from https://github.com/WICG/feature-policy/pull/91, and https://github.com/foolip/testing-in-standards/blob/master/policy.md links to many others specs also using this wording.

foolip:
@adam-be, will you take care of merging as well? I want to claim on Wednesday that WebRTC is doing this :)

adam-be:
I'll try to get some more opinions, but let's aim for merging before Wednesday.

adam-be:
I'm thinking about how all this will be done in practice. Probably I'll take an existing test, copy it and update it to do what I want. So, it would be really great to have some pointers to some of our tests that are really good examples of how tests should be written.

foolip:
I think the very first test in order is quite good:
https://wpt.fyi/webrtc/RTCCertificate.html

There was a typo (https://github.com/w3c/web-platform-tests/pull/8087) but otherwise looks good to me, testing both things that should work and things that shouldn't.

I also like https://wpt.fyi/webrtc/RTCDataChannelEvent-constructor.html,  as an example of being pretty thorough with "boring" stuff like the difference between undefined and null in dictionaries.

In some of the existing tests you'll find large comment blocks. The group will have to judge if those are useful or not. I think @rwaldron and @soareschen used this extra bookkeeping to judge the spec coverage and produce https://rwaldron.github.io/webrtc-pc/, but it's definitely not a requirement generally in web-platform-tests. (Also, we basically have no idea about the test coverage in most areas.)

burnburn:
@foolip I think this text is a fine start (meaning we can apply it and then modify if needed).  I have the same concerns @adam-be does regarding information we could present to those we ask to write tests for their PRs.  I think the contents of your comment right above this one would be a good start for a webrtc-pc specific informative testing doc that anyone can update it as they have more info to add that would help our group write tests.  @adam-be tells me he wants to create that (small) doc right before merging this PR, and I agree.

foolip:
The other resources I would point to are:
http://web-platform-tests.org/writing-tests/index.html
http://web-platform-tests.org/reviewing-tests/index.html

PR #1664, opened 2017-11-09
Add testing guideline for naming test files and adding comments

soareschen:
I can add more about other aspects of testing later on. These are the two draft test guidelines that I come out with while still at TPAC.

We might also want to consider whether to put the main guidelines in webrtc-pc or web-platform-tests. There may be other contributors who work on web-platform-tests but not webrtc-pc. So it might be better off to put the guidelines in WPT and link to it from webrtc-pc.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -11,3 +11,40 @@ https://wpt.fyi/webrtc/RTCCertificate.html
 
 Thorough example that tests the difference between undefined and null in dictionaries.  
 https://wpt.fyi/webrtc/RTCDataChannelEvent-constructor.html
+
+## Naming Guidelines
+
+Update to WebIDL definitions in the spec must also update the IDL test file at [/interfaces/webrtc-pc.idl](https://github.com/w3c/web-platform-tests/blob/master/interfaces/webrtc-pc.idl).
+A [helper script](http://web-platform-tests.org/writing-tests/idlharness.html)
+can be pasted in the console to generate IDL code from the spec HTML. Note that
+partial dictionary definition is not supported in idlharness.js, so some manual
+editing is required to merge them as one dictionary definition.
+
+In general, a test case written for a class _RTCClass_ on a method _rtcMethod_
+should be added to a file named _RTCClass_-_rtcMethod\[-custom-postfix\]_\[.https].html.
+The `.https` extension is _required_ for tests calling `getUserMedia()`.
+
+When a test file contain too many test cases, they may be split into multiple
+files with different postfixes, e.g. RTCPeerConnection-setLocalDescription-offer.html
+
+Sometimes the test files are named based on the argument dictionary name and the
+tested field, e.g. RTCConfiguration-iceServers.html.
+
+A helper JavaScript file may be added to share helper functions across multiple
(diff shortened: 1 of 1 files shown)


alvestrand:
@fippo can you review this?
If you know of guidelines that cover parts of this that we should link to, please tell!

fippo:
philip(p) confusion. Paging @foolip

stefhak:
@foolip could you take a look?

foolip:
It is really up to the people doing most of the work here to decide what amount of metadata and structure is useful. If I were to recommend anything, it would be to err on the side of less metadata to update, to optimize for ease of adding new tests.

Given the permission given to not bother with metadata in new tests or when updating tests, unless someone is continuously going to update the metadata, then it will get out of sync. As soon as that begins to happen, I think it might be better to delete the metadata.

soareschen:
@foolip yes agree we can remove a number of metadata from the individual test files, in particular the todo list. I think some coverage accounting is still essential until we reach 100% coverage. Some considerations are how essential are such metadata to others, and where should we put the metadata. Do we put them in web-platform-tests (w3c/web-platform-tests#8051)? or as unofficial forks (https://rwaldron.github.io/webrtc-pc/)? or in webrtc-pc?

foolip:
@soareschen, all the attempts to determine coverage from test-to-spec linkage so far don't seem very successful. Linking to just sections isn't granular enough. If you search for "data-tested-assertations" you can see such an attempt dating back to https://github.com/w3c/web-platform-tests/commit/6bb53056b2b9127bec0ddc5995d094421f0dc0a3. It's brittle approach, and I can only assume it's bitrotted.

The approach in https://github.com/tabatkins/bikeshed/issues/1116 is more promising to me, and should allow for views similar to https://rwaldron.github.io/webrtc-pc/. But that will be a Bikeshed feature, and WebRTC uses ReSpec. AFAICT, the least amount of work at this point would be to keep https://rwaldron.github.io/webrtc-pc/ up to date manually, by noting which wpt commit it's in sync with, and going through the changes since then for each update.

soareschen:
Updating the coverage can be pretty tedious. If not done incrementally, it could take a whole week to track changes to new editor's draft. I guess that means the best approach is to pull out the metadata and put them in unofficial forks. If the fork gets bitrotten, it is simply forgotten and affect neither web-platform-tests nor webrtc-pc.

TESTING.md
```diff
@@ -11,3 +11,40 @@ https://wpt.fyi/webrtc/RTCCertificate.html
...
+become widely available for WPT) Helper files should have as little dependency
+to other helper files, as script tags are not good for dependency management.
+
+## Comments
+
+The current tests have detailed comments referencing sections of the spec that
+is being tested. New tests may be written with minimal or no comments up to the
+author's preference. If a test case is testing specific steps in a spec, it is
+recommended to leave at least one line of comment referencing the spec to help
+readers quickly understand the objective of the test.
```
adam-be: How does this relate to the "help link" that should point to the tested section in the spec?

adam-be: Mention about "help link": https://github.com/w3c/web-platform-tests/pull/8450#discussion_r153404434

Issue #1533, opened 2017-08-12
Clarify whether RTCRtpContributingSource members are live.


jan-ivar:
@taylor-b Even with [recent clarifications](https://github.com/w3c/webrtc-pc/pull/1149), I still find two ways of reading [getContributingSources()](http://w3c.github.io/webrtc-pc/#dom-rtcrtpreceiver-getcontributingsources). Even though it returns a [sequence](https://heycam.github.io/webidl/#idl-sequence) of platform object values, and the sequence itself is passed by value, it's not clear whether the objects in the sequence are the same each time or new ones.

Basically, assuming no external changes for a couple of seconds, which is right:
 1. Each call returns a sequence of unique static snapshots of the most recent values at the time, or
 2. Each call returns a sequence of the same platform objects, whose readonly members are "live"?

Where "live" means "may change on the next event loop". What's the intent; can we make it clearer?

taylor-b:
> Each call returns a sequence of the same platform objects, whose readonly members are "live"?

That's how I interpreted it. Otherwise why would they be interfaces and not just dictionaries? When we've talked about this API in virtual interims (for example, when discussing https://github.com/w3c/webrtc-pc/issues/1091), I mentioned this assumption and there wasn't any disagreement.

What would you suggest adding to make this more clear? Maybe add something to the description of `getContributingSources` like:

> Only one `RTCRtpContributingSource` object will be created for each CSRC identifier received by this `RTCRtpReceiver`. Subsequent calls to `getContributingSources` will return the same objects if the set of contributing sources has not changed.

And/or we could add this somewhere:

> Note that the `RTCRtpContributingSource` and `RTCRtpSynchronizationSource` objects will be updated regardless of whether `getContributingSources`/`getSynchronizationSources` is called again. Calls to `getContributingSources` and `getSynchronizationSources` are only needed to retrieve an updated list of objects, in case an additional CSRC or SSRC identifier has been seen since the last call.

Just some ideas.

na-g:
@jan-ivar & @taylor-b To get one of these potentially live objects, one has to poll the getContributingSources, or getSynchronizationSources interfaces until the CSRC or SSRC has contributed to playout (which may never happen). Without a way to ask for one to ask for one of these live objects by source identifier, is there a added value in its liveness when one is already polling for discovery purposes?

na-g:
[...]

PR #1668, opened 2017-11-21
RTCRtpContributingSource & RTCRtpSynchronizationSource -> dictionaries

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1533.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6610,42 +6610,43 @@ async function updateParameters() {
         </section>
       </div>
       <p>The <dfn>RTCRtpContributingSource</dfn> and
-      <dfn>RTCRtpSynchronizationSource</dfn> objects contain information about
-      a given contributing source (CSRC) or synchronization source (SSRC),
-      including the most recent time a
+      <dfn>RTCRtpSynchronizationSource</dfn> dictionaries contain information
+      about a given contributing source (CSRC) or synchronization source (SSRC)
+      respectively, including the most recent time a
       packet that the source contributed to was played out. The browser MUST
       keep information from RTP packets received in the previous 10 seconds.
       When the first audio frame contained in an RTP packet is delivered to the
       <code><a>RTCRtpReceiver</a></code>'s <code><a>MediaStreamTrack</a></code>
       for playout, the user agent MUST queue a task to update the relevant
-      <code><a>RTCRtpContributingSource</a></code> and
-      <code><a>RTCRtpSynchronizationSource</a></code> objects based on the
-      contents of the packet. The
-      <code><a>RTCRtpSynchronizationSource</a></code> object corresponding to
-      the SSRC identifier is updated each time, and if the RTP packet contains
-      CSRC identifiers, then the <code><a>RTCRtpContributingSource</a></code>
-      objects corresponding to those CSRC identifiers are also updated.</p>
+      information for the <code><a>RTCRtpContributingSource</a></code> and
+      <code><a>RTCRtpSynchronizationSource</a></code> dictionaries based on the
+      contents of the packet. The information relevant to the
+      <code><a>RTCRtpSynchronizationSource</a></code> dictionary corresponding
+      to the SSRC identifier, is updated each time, and if the RTP packet
+      contains CSRC identifiers, then the information relevant to the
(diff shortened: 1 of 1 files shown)


alvestrand:
Needs test update to account for the interface -> dictionary change.
@aboba will check.

foolip:
Chrome has already shipped RTCRtpContributingSource, so there should be some tests that starts failing with this change. @henbos, do you think it's safe to change the interface into a dictionary?

henbos:
This would affect apps relying on them being live, but I'm not sure they do since it's debatable how useful that fact is, it would make more sense to just poll it and look at the result. Considering getContributingSources() is relatively new and not a common use case my gut feeling this is safe to change.

foolip:
Good. When it comes to tests, in addition to verifying the new behavior (that an object and not an instance of RTCRtpContributingSource is returned), an addition to https://github.com/w3c/web-platform-tests/blob/master/webrtc/historical.html is needed, to make sure the RTCRtpContributingSource interface object doesn't exist.

jan-ivar:
FWIW the WPT tests for getContributingSources() look incorrect to us. AFAIK this API would only return a non-zero array if connected to an MCU server emitting CSRCs. Frankly we're having a hard time testing it. @henbos How is Chrome testing this? We'd love some help.

henbos:
When this was implemented, Chrome had mocks which emulated receving CSRCs. The tests relying on mocks in Chrome was horrible, and for almost all other cases did not add any value, our tests simply did not test the real thing and maintaining a fake implementation was burdensome. When we finally removed mocking so that WPT and other tests would run with the real thing -- greatly improving our test coverage -- this particular feature lost its test coverage on the JavaScript layer.

I don't see how this is testable on a JavaScript layer without emitting CSRCs. We currently only have tests for the lower layers and some plumbing.

If we want WPT coverage for this maybe we need WebDriver support for pretty low level stuff like packets? Sounds like another case of high-effort low-impact work.

foolip:
Filing a type:untestable for this on web-platform-tests and deleting the wrong tests might be the best use of everyone's time, then?

henbos:
SGTM

alvestrand:
The KITE project has on its wishlist the integration of an MCU into tests - preferably picking an opensource  MCU that can be just integrated into the project's builds.
Such an MCU could test CSRCs (if it did server-side mixing).
I agree that it's out of scope for WPT, but @agouaillard-cosmo may want to comment on behalf of KITE.

foolip:
I'd like to be clear that something like that wouldn't be out of scope due to WPT as a whole having a clear delineation of scope, and that if someone did the work inside of WPT it would be most welcome. However, first there has to be someone to do the work, and that someone has to think that WPT is the best way to achieve whatever goals they have.

webrtc.html
```diff
@@ -6610,105 +6610,84 @@ <h2>Methods</h2>
...
-      CSRC identifiers, then the <code><a>RTCRtpContributingSource</a></code>
-      objects corresponding to those CSRC identifiers are also updated.</p>
+      information for the <code><a>RTCRtpContributingSource</a></code> and
+      <code><a>RTCRtpSynchronizationSource</a></code> dictionaries based on the
+      contents of the packet. The information relevant to the
+      <code><a>RTCRtpSynchronizationSource</a></code> dictionary corresponding
+      to the SSRC identifier, is updated each time, and if the RTP packet
+      contains CSRC identifiers, then the information relevant to the
+      <code><a>RTCRtpContributingSource</a></code> dictionaries corresponding to
+      those CSRC identifiers is also updated.</p>
```
henbos: This still talks about updating dictionaries upon receiving RTP packets. But since we don't want a "live" state anymore, what is updated is some internal set of dictionaries, correct? The ones returned by getContributingSources() being copies not affected by this update.

We still want the "queue a task to update the dictionaries" type of language to indicate that if an RTP packet is received midst-JS task execution cycle we don't see this until the next cycle?
getContributingSources() is always value-equal to getContributingSources() in a single cycle?

Just checking to make sure Chromium implementation still requires a cache.

henbos: This has already been answered - the answer is yes.

PR #1671, opened 2017-11-24
Editorial: Add IANA-HASH-FUNCTION reference

### LINKED ISSUE #1663: Point to the Hash Function Textual Names registry
https://w3c.github.io/webrtc-pc/#dom-rtcdtlsfingerprint-algorithm says "One of the the hash function algorithms defined in the 'Hash function Textual Names' registry, initially specified in [RFC4572] Section 8."

That registry now exists:
https://www.iana.org/assignments/hash-function-text-names/hash-function-text-names.xml

It would be helpful to refer to this directly, since that's where additional values will be added.

I noticed this when looking at a failing test:
https://github.com/w3c/web-platform-tests/blob/8c5d6017f9d992c5f97737c41e43e18f1fb682af/webrtc/RTCCertificate.html#L175

@soareschen FYI

adam-be:
Fixes #1663

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7267,8 +7267,7 @@ async function onOffHold() {
               "idlMemberType"><a>DOMString</a></span></dt>
               <dd>
                 <p>One of the the hash function algorithms defined in the 'Hash
-                function Textual Names' registry, initially specified in
-                [[!RFC4572]] Section 8.</p>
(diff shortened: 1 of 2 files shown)


alvestrand:
lgtm

foolip:
Thanks!

PR #1672, opened 2017-11-27
Set direction of transceiver created by offerToReceive* option

### LINKED ISSUE #1640: offerToReceiveAudio/Video processing creates sendrecv transceivers
The spec says to behave as if addTransceiver was called, which by default uses "sendrecv". We should probably be creating a "recvonly" transceiver instead here.

adam-be:
Fixes #1640

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3284,39 +3284,59 @@ interface RTCPeerConnection : EventTarget  {
             <dl data-link-for="RTCOfferOptions" data-dfn-for="RTCOfferOptions" class="dictionary-members">
               <dt><dfn>offerToReceiveAudio</dfn> of type <span class="idlMemberType"><a>boolean</a></span></dt>
               <dd>
-                <p>
-                  Whenever this is given a non-false value, and the
-                  <code><a>RTCPeerConnection</a></code> has no non-stopped
-                  "sendrecv" or "recvonly" audio transceivers, createOffer()
-                  MUST as its first step invoke the equivalent of
-                  <code>addTransceiver("audio")</code> on the
-                  <code><a>RTCPeerConnection</a></code> object, except that this
-                  MUST NOT <a>Update the negotiation-needed flag</a>,  and,
-                  provided this does not fail, proceed with createOffer()'s
-                  regular steps.
-                </p>
-                <p>
-                  In all other situations, it will be disregarded.
-                </p>
+                <p>Offer to receive audio even though no audio will be sent.
+                When this dictionary member is present as an option to <a href=
+                "#dom-rtcpeerconnection-createoffer">createOffer</a>, run the
+                <a>offer to receive</a> steps with a <var>mediaType</var> of
+                "audio" instead of the regular <a href=
+                "#dom-rtcpeerconnection-createoffer">createOffer</a> steps.</p>
               </dd>
               <dt><dfn>offerToReceiveVideo</dfn> of type <span class="idlMemberType"><a>boolean</a></span></dt>
               <dd>
-                <p>
-                  Whenever this is given a non-false value, and the
-                  <code><a>RTCPeerConnection</a></code> has no non-stopped
-                  "sendrecv" or "recvonly" video transceivers, createOffer()
-                  MUST as its first step invoke the equivalent of
-                  <code>addTransceiver("video")</code> on the
-                  <code><a>RTCPeerConnection</a></code> object, except that this
-                  MUST NOT <a>Update the negotiation-needed flag</a>,  and,
-                  provided this does not fail, proceed with createOffer()'s
-                  regular steps.
-                </p>
-                <p>
-                  In all other situations, it will be disregarded.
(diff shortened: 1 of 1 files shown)


adam-be:
Test: https://github.com/w3c/web-platform-tests/pull/8450

webrtc.html
```diff
@@ -3284,39 +3284,59 @@ <h2>Attributes</h2>
...
+                <a>RTCPeerConnection</a></code> object.</p>
+              </li>
+              <li>
+                <p>If the value of the dictionary member is false, jump to the
+                step labeled <em>create offer</em>.</p>
+              </li>
+              <li>
+                <p>If <var>connection</var> has any stopped "sendrecv" or
+                "recvonly" transceivers of type <var>mediaType</var>, jump to
+                the step labeled <em>create offer</em>.</p>
```
jan-ivar: @adam-be This flips the logic from *"Whenever this ... has no* ***UN-stopped*** *... transceivers"*, to *"Whenever this ... has no* ***STOPPED*** *... transceivers"*. Seems wrong (out of scope for fixing https://github.com/w3c/webrtc-pc/issues/1640 at least).

jan-ivar: Found by running the [wpt test](https://github.com/w3c/web-platform-tests/pull/8450#issuecomment-348316922). 👍

adam-be: Good catch! Let's revert that language back to what it was.

webrtc.html
```diff
@@ -3284,39 +3284,59 @@ <h2>Attributes</h2>
...
+              <li>
+                <p>If <var>transceiver</var> is unset because the previous
+                operation threw an error, abort these steps.</p>
+              </li>
+              <li>
+                <p>Set <var>transceiver</var>'s <a>[[\Direction]]</a> slot to
+                "recvonly".</p>
+              </li>
+              <li>
+                <p><em>create offer:</em> Run the steps specified by <a href=
```
fippo: I am a bit late and probably nitpicking but... when calling `createOffer({offerToReceiveAudio: true, offerToReceiveVideo: true})` won't this only process offerToReceiveAudio before running the normal createOffer steps?

webrtc.html
```diff
@@ -3284,39 +3284,59 @@ <h2>Attributes</h2>
...
+                <p>Let <var>connection</var> be the current <code>
+                <a>RTCPeerConnection</a></code> object.</p>
+              </li>
+              <li>
+                <p>If the value of the dictionary member is false, jump to the
+                step labeled <em>create offer</em>.</p>
+              </li>
+              <li>
+                <p>If <var>connection</var> has any stopped "sendrecv" or
+                "recvonly" transceivers of type <var>mediaType</var>, jump to
```
fippo: how is the type of a transceiver defined? Both the transceivers sender and receiver have a kind but the transceiver itself does not. Might need an internal slot for the kind?

fippo: actually the kind is only guaranteed to be on transceiver.receiver.track

fippo: I think the correct thing to say here is the transceiver kind defined in http://w3c.github.io/webrtc-pc/#dfn-transceiver-kind

fippo: done in https://github.com/w3c/webrtc-pc/pull/1681

webrtc.html
```diff
@@ -3284,39 +3284,59 @@ <h2>Attributes</h2>
...
-                  <code>addTransceiver("audio")</code> on the
-                  <code><a>RTCPeerConnection</a></code> object, except that this
-                  MUST NOT <a>Update the negotiation-needed flag</a>,  and,
-                  provided this does not fail, proceed with createOffer()'s
-                  regular steps.
-                </p>
-                <p>
-                  In all other situations, it will be disregarded.
-                </p>
+                <p>Offer to receive audio even though no audio will be sent.
```
fippo: actually it does not. When you add a track and set offerToReceiveAudio: true audio will be sent. While this is helpful as a  rationale I'd suggest removing it.

adam-be: I just wanted a short descriptive line about what it does before we start defining the behavior. But I agree that it must be correct. Do you have a suggestion for a sentence?

webrtc.html
adam-be: True.

webrtc.html
fippo: i've removed it in #1683 after not being happy with all the "controls the receive behaviour" variants I came up with :-/

fippo: what about a short description at the beginning of the section? Also a good place to say "use transceivers instead" ;-)

adam-be: That probably works. I was thinking about some common section to make it easier to deal with the problem you mention here https://github.com/w3c/webrtc-pc/pull/1672/files#r155310994

webrtc.html
adam-be: Proposed fix #1686 for this.

PR #1676, opened 2017-12-01
Editorial: Add jib as editor

adam-be:


webrtc.js
```diff
@@ -65,6 +65,9 @@ var respecConfig = {
     },
     { name: "Taylor Brandstetter", company: "Google",
       w3cid: "82908"
+    },
+    { name: "Jan-Ivar Bruaroey", company: "Mozilla",
```
alvestrand: Can we use ø here, or does he prefer to be Bruaroey?

adam-be: I took the same from the W3C records.

adam-be: @jan-ivar, any preference on this topic? :)

jan-ivar: I use Jan-Ivar Bruaroey.

PR #1678, opened 2017-12-04
remove idl comment about rtcpTransport being at risk

fippo:
the linebreak is confusing and without checking the source its not
clear that rtcpTransport is at risk and not getCapabilities

atrisk


alvestrand:
I think warning about feature-at-risk in the big red box further down the page is sufficient.
LGTM.

aboba:
LGTM.  I put in the comment originally as a reminder to complete the feature-at-risk text, and forgot to remove it ;(

PR #1681, opened 2017-12-06
offerToReceive: s/transceiver type/transceiver kind/ (and do it for kinds, not mediaTypes)

fippo:
clarifies https://github.com/w3c/webrtc-pc/pull/1672#discussion_r155347548
Also changes mediaType to kind for consistency.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3287,7 +3287,7 @@ interface RTCPeerConnection : EventTarget  {
                 <p>Offer to receive audio even though no audio will be sent.
                 When this dictionary member is present as an option to <a href=
                 "#dom-rtcpeerconnection-createoffer">createOffer</a>, run the
-                <a>offer to receive</a> steps with a <var>mediaType</var> of
+                <a>offer to receive</a> steps with a <var>kind</var> of
                 "audio" instead of the regular <a href=
                 "#dom-rtcpeerconnection-createoffer">createOffer</a> steps.</p>
               </dd>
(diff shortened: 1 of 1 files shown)


fippo:
@adam-be can you please take a look?

webrtc.html
```diff
@@ -3314,7 +3314,7 @@ <h2>Attributes</h2>
               </li>
               <li>
                 <p>If <var>connection</var> has any stopped "sendrecv" or
-                "recvonly" transceivers of type <var>mediaType</var>, jump to
+                "recvonly" transceivers of kind <var>mediaType</var>, jump to
```
jan-ivar: Maybe s/`mediaType`/`mediaKind` too? We might want to be even more specific. How about:

 "recvonly" transceivers whose `receiver.kind` is *mediaKind*, jump to

?

fippo: whose *transceiver kind* (with link to the definition)?

I agree that mediaKind would be more consistent language, will update.

fippo: just kind instead of mediaKind is even more consistent :-)

jan-ivar: Thanks, I forgot we had the [transceiver kind](http://w3c.github.io/webrtc-pc/#dfn-transceiver-kind) definition.

fippo:
squashed after some great feedback from jib :-)

webrtc.html
adam-be: So did I, obviously :)

fippo:
needs rebase after #1686

alvestrand:
Same problem was solved in #1686, so closing this one.

Issue #1586, opened 2017-09-10
behaviour of offerToReceive* set to false when there is a local track


fippo:
related to #1383 but that is complicated enough already.
Suppose I do this:
```
var pc = new RTCPeerConnection();
navigator.mediaDevices.getUserMedia({audio: true})
.then(stream => {
  pc.addTrack(stream.getAudioTracks()[0], stream);
  return pc.createOffer({offerToReceiveAudio: false})
})
.then(offer => {
  console.log(offer.sdp);
});
```
I want to send audio but not receive it. This means the sdp should say "sendonly".
But we currently specify this as "leave it alone".

This needs something like "if there is a transceiver for this kind set its direction to sendonly".

fippo:
I knew this was somewhat familiar, see #1461 and #1466. But both were closed leaving the poor guy trying to shim this 😭

stefhak:
I feel your pain, but I wonder, was ```{offerToReceiveAudio: false}``` ever defined to make the m-line say ```sendonly```? My recollection says the use of ```offerToReceiveX``` was only to make the offer contain an m-line of kind ```X``` even if there was no outgoing media of type ```X```.

jan-ivar:
> was {offerToReceiveAudio: false} ever defined to make the m-line say sendonly

@stefhak The spec [used to say](https://dev.w3.org/2011/webrtc/editor/archives/20130603/webrtc.html#constraints): *"This constraint allows an application to indicate its preferences for receiving video when creating an offer."*

and 

*"This is an enum type constraint that can take the values "true" and "false". The default is a non mandatory "true" for an RTCPeerConnection object that has a video stream at the point in time when the constraints are being evaluated and is non mandatory "false" otherwise."*

I interpreted it at the time to mean that, which is why Firefox works that way. Chrome does too, as [I pointed out](https://github.com/w3c/webrtc-pc/issues/1461#issue-241153792).

jan-ivar:
What convinced me to not pursue this is @taylor-b's [comment](https://github.com/w3c/webrtc-pc/pull/1466#issuecomment-315227981). The scope of createOffer isn't to discover whom you're dealing with, rather just to discover which machines will be involved. Specifically: apps are generally on both sides of a connection, and should know which side intends to send what ahead of OA.

I could be wrong about this, of course. So others feel free to raise this issue if there's an important use-case left in the cold by this.

taylor-b:
[...]

PR #1683, opened 2017-12-06
fix offerToReceive(Audio|Video): false

fippo:
specifies that offerToReceive* stops all non-stopped transceivers from sending.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3292,6 +3292,58 @@ interface RTCPeerConnection : EventTarget  {
         </section>
       <section>
         <h4>Legacy configuration extensions</h4>
+        <p>This section describes a set of legacy extensions that may be used to
+        influence how an offer is created, in addition to the media added to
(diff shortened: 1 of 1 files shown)


fippo:
A stab at https://github.com/w3c/webrtc-pc/issues/1586

Ironically Firefox Nightly just "broke" this: https://jsfiddle.net/dyytbfxk/
is sendrecv in nightly, sendonly in stable.

Firefox stable sets (to my surprise) the direction on *all* m-lines.
This shows that this feature is 💩 💩 💩  :-)

jan-ivar:
@fippo By "broke" you mean we now implement the spec as mandated?

fippo:
given that the spec decided to change existing behaviour without even bothering to investigate usage (I have not seen httparchive usage data) yes.

It is very much unclear how many people depend on this and the lack of feedback from consumers of the peerconnection API (both service providers and opensource library vendors) is quite concerning.

fippo:
needs rebase after #1686

PR #1686, opened 2017-12-07
offerToReceive: Rewrite to handle two options to createOffer

adam-be:
I think this is cleaner. Question is if we want to have a description of the individual options.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3273,6 +3273,54 @@ interface RTCPeerConnection : EventTarget  {
         </section>
       <section>
         <h4>Legacy configuration extensions</h4>
+        <p>When <a href="#dom-rtcpeerconnection-createoffer">createOffer</a>
+        is called with any of the legacy options specified in this section,
+        run the followings steps instead of the regular <a href=
+        "#dom-rtcpeerconnection-createoffer">createOffer</a> steps:</p>
+        <ol>
+          <li>
+            <p>Let <var>options</var> be the methods second argument.</p>
+          </li>
+          <li>
+            <p>Let <var>connection</var> be the current <code>
+            <a>RTCPeerConnection</a></code> object.</p>
+          </li>
+          <li>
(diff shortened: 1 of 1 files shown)


adam-be:
Better looking diff (--patience): https://dpaste.de/heHB

adam-be:
@fippo, sorry that this breaks some of your PRs, but it should address the problem you brought up with two options at the same time.

webrtc.html
```diff
@@ -3273,70 +3273,72 @@ <h2>Callback <a class="idlType">RTCSessionDescriptionCallback</a>
...
-            </dl>
-            <p>To <dfn>offer to receive</dfn> with <var>mediaType</var>, run
-            the following steps:</p>
+        <p>When <a href="#dom-rtcpeerconnection-createoffer">createOffer</a>
+        is called with any of the legacy options specified in this section,
+        run the followings steps instead of the regular <a href=
+        "#dom-rtcpeerconnection-createoffer">createOffer</a> steps:</p>
+        <ol>
+          <li>
+            <p>Let <var>options</var> be the methods second argument.</p>
```
fippo: first argument? (I am glad we're dealing with the legacy callbacks elsewhere)

fippo:
no worries, I expected that i'll have to update my PRs.

This looks pretty good, I was afraid it would be more ugly!

webrtc.html
adam-be: Will fix!

adam-be:
I've added an intro text and a short description of the attributes. Can you take a look @fippo 

Readable diff: https://dpaste.de/0jb8

adam-be:
Rendered preview: https://rawgit.com/w3c/webrtc-pc/otr-both-options/webrtc.html#legacy-configuration-extensions

adam-be:
Extended test to use two options: https://github.com/w3c/web-platform-tests/pull/8450

PR #1692, opened 2017-12-08
Rephrase RTCDataChannel.bufferedAmount description

### LINKED ISSUE #1680: RTCDataChannel.bufferedAmount description confusing
I don't understand the following description of `RTCDataChannel.bufferedAmount`:

> [...] If the channel is closed, this attribute's value will only increase with each call to the send() method (the attribute does not reset to zero once the channel closes).

Let me go through that:

> If the channel is closed [...]

So, the channel is closed already.

> [...] this attribute's value will only increase with each call to the send() method [...]

That doesn't make any sense to me as the channel is already closed. Sending in a non-open state will result in an exception being thrown.

> (the attribute does not reset to zero once the channel closes)

Fine.

---

If I understand correctly, this is what it should say:

> [...] This attribute's value will only increase with each call to the send() method as long as the channel is in the `open` state. However, the attribute does not reset to zero once the channel closes.

lgrahl:
Resolves #1680

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8736,10 +8736,11 @@ interface RTCTrackEvent : Event {
               whether the user agent is able to transmit text asynchronously
               with script execution.) This does not include framing overhead
               incurred by the protocol, or buffering done by the operating
-              system or network hardware. If the channel is closed, this
-              attribute's value will only increase with each call to the
-              <code><a data-for="RTCDataChannel">send()</a></code> method (the
-              attribute does not reset to zero once the channel closes).</p>
+              system or network hardware. This attribute's value will only
+              increase with each call to the <code><a data-for=
+              "RTCDataChannel">send()</a></code> method as long as the <a>
+              [[\ReadyState]]</a> slot is <code>open</code>. However, the
+              attribute does not reset to zero once the channel closes.</p>
             </dd>
             <dt><code>bufferedAmountLowThreshold</code> of type <span class=
             "idlAttrType"><a>unsigned long</a></span></dt>


webrtc.html
```diff
@@ -8736,10 +8736,11 @@ <h2>Attributes</h2>
...
               with script execution.) This does not include framing overhead
               incurred by the protocol, or buffering done by the operating
-              system or network hardware. If the channel is closed, this
-              attribute's value will only increase with each call to the
-              <code><a data-for="RTCDataChannel">send()</a></code> method (the
-              attribute does not reset to zero once the channel closes).</p>
+              system or network hardware. This attribute's value will only
+              increase with each call to the <code><a data-for=
+              "RTCDataChannel">send()</a></code> method as long as the <a>
+              [[\ReadyState]]</a> slot is <code>open</code>. However, the
```
taylor-b: Isn't the part about the value increasing already stated above?

webrtc.html
```diff
@@ -8736,10 +8736,11 @@ <h2>Attributes</h2>
...
+              attribute does not reset to zero once the channel closes.</p>
```
taylor-b: Does this only affect non graceful closures, or closures that race with data being sent?

webrtc.html
lgrahl: I can't see it.

webrtc.html
lgrahl: I'd say both and this is my interpretation:
* If the data channel closes abruptly, the value remains the same and does not reset because the state goes directly from whatever to `closed`. I believe that implementation could flush the underlying buffer. They just need to make sure the value does not reset.
* If the data channel closes gracefully, the value will be `0` once the channel reaches the `closed` state. Before that, all pending data has been written during the `closing` state. Even though the value decreases, I wouldn't call this a *reset*.

I agree that the description isn't crystal clear in that regard. Suggestions welcome.

lgrahl: By the way, graceful shutdown of data channels isn't something that's really specified right now in the W3C spec (even though it is specified in [draft-ietf-rtcweb-data-channel-13](https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13#section-6.7)). See for example [this bug in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1413198). I guess I just forgot to open an issue (once again).

lgrahl: Suggestion:

> During the `closing` state, the attribute's value may decrease until the underlying buffer is completely sent. However, the attribute's value is frozen once the channel reaches the `closed` state.

webrtc.html
taylor-b: I mean by "The bufferedAmount attribute must return the number of bytes of application data (UTF-8 text and binary data) that have been queued using send() but that, as of the last time the event loop started executing a task, had not yet been transmitted to the network."

webrtc.html
taylor-b: Your suggestion, and your interpretation of graceful/non-graceful closing behavior, sounds good to me. Though I thought we were ok with regards to graceful closure? "An RTCDataChannel object's underlying data transport may be torn down in a non-abrupt manner by running the closing procedure."

webrtc.html
lgrahl: I think it's okay to be redundant here. It reasserts that the value can only increase by a call to `send` and nothing else.

But I have a question. Is the following part really necessary?

> [...] as of the last time the event loop started executing a task [...]

Isn't that always the case for pretty much everything in the world of JS (in browsers)? Or am I missing something important?

webrtc.html
lgrahl: Ok, I'll update the description in the PR.

I'll open a new issue for the closing procedure stuff and will mention you.

webrtc.html
taylor-b: I think it's just reasserting that. Maybe it wasn't obvious to the authors of the spec. Also, I'm fine being redundant; better than omitting information.

stefhak:
This PR changes the behavior IIUC, before:
- each call to `send()` increases `bufferedAmount` when the channel is closed
after:
- calling `send()` does not increase `bufferedAmount` when the channel is closed

Will this give any legacy issues?

lgrahl:
I don't think that's the case. The `send` method states:

> 1. Let channel be the RTCDataChannel object on which data is to be sent.
> 2. If channel's [[ReadyState]] slot is not open, throw an InvalidStateError.
> ...

Step 2 raises an `InvalidStateError`, so `bufferedAmount` cannot increase.

PR #1693, opened 2017-12-08
Specify legacy behavior when offerToReceive* option is set to false

fippo:
specifies that offerToReceive* stops all non-stopped transceivers from sending.

rebase of #1683 (which accidentially got closed while squashing :-/)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3313,8 +3313,22 @@ interface RTCPeerConnection : EventTarget  {
             kind, <var>kind</var>, run the following steps:</p>
             <ol>
               <li>
-                <p>If the value of the dictionary member is false, continue
-                with the next option, if any.</p>
+                <p>If the value of the dictionary member is false,
+                <ol>
+                  <li>
+                    <p>For each non-stopped "sendrecv" transceiver of
+                    <a>transceiver kind</a> <var>kind</var>, set
+                    <var>transceiver</var>'s <a>[[\Direction]]</a> slot to
+                    "sendonly".</p>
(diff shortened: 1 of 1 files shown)


alvestrand:
Is this intended to apply only to the initial offer, or to subsequent offers too?

fippo:
both. At least that is what the Chrome does, see [here](https://jsfiddle.net/mc5j66zk/3/) (firefox probably does that as well but doesn't like createOffer->SLD->createOffer->SLD)

alvestrand:
@fippo is this what you want this functionality to do?
@taylor-b says this is not what Chrome is currently doing - Chrome stops the transceivers instead (sets port zero) on transceivers that have nothing to send. It's not unreasonable to change this.

aboba:
@fippo Can you provide a test?

alvestrand:
Test: https://github.com/w3c/web-platform-tests/pull/8450

fippo:
@taylor-b https://jsfiddle.net/mc5j66zk/4/ disagrees with you?
I think setting the port to 0 was something Firefox did and someone filed a bug... @nils-ohlmeier does that ring a bell?

Will see about tests

taylor-b:
Chrome only sets the port to 0 if offerToReceive is 0 and the current direction is `recvonly`. In other words, it sets a port of 0 instead of using `inactive`; it will never generate an `inactive` offer as far as I know.

fippo:
@taylor-b it does inactive (see fiddle) but sets the port to 0 (which I missed)

PR #1696, opened 2017-12-14
replaceTrack: Clarify how the UA determines if negotiation is needed

### LINKED ISSUE #1389: No normative definition for "Determine if negotiation is needed to transmit withTrack in place of the sender's existing track" in replaceTrack
My understanding is that the step should be different from the section for _"check if negotiation is needed for connection"_, but the words are too similar to easily confuse one with another. Can there be better wordings to distinguish the two?

There is a note section below that describe cases that _may_ require negotiation, but there is no normative definition or reference to other specs on how exactly renegotiation is determined. Is it up to the implementer to decide what requires renegotiation?

adam-be:
Fixes #1389

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5815,7 +5815,9 @@ async function updateParameters() {
                     <li>
                       <p>Determine if negotiation is needed to transmit
                       <var>withTrack</var> in place of the sender's existing
-                      track. Negotiation is not needed if <var>withTrack</var>
+                      track. The details of this decision is left to
+                      implementations with exception that
+                      negotiation is not needed if <var>withTrack</var>
                       is null or if the sender's existing
                       track is ended (which appears as though the track was
                       muted). Ignore which <code>MediaStream</code> the track


webrtc.html
```diff
@@ -5815,7 +5815,9 @@ <h2>Methods</h2>
                     <li>
                       <p>Determine if negotiation is needed to transmit
                       <var>withTrack</var> in place of the sender's existing
-                      track. Negotiation is not needed if <var>withTrack</var>
+                      track. The details of this decision is left to
+                      implementations with exception that
```
fluffy: I don't think this is  right. They are defined by SDP offer / answer and JSEP. Basically if what is being sent would be outside the currently negotiated SDP offer/answer envelope, then negotiation is needed *before* it can be sent.

adam-be:
@fluffy, I like your proposal better since it would make this more consistent between implementations.

webrtc.html
aboba: Agree with Cullen. Given the divergence in replaceTrack implementations, more specific guidance is probably preferable to endorsing "implementation specific" behavior.  Can we discuss this at the January Virtual Interim?

alvestrand:
@adam-be  will review the rules in JSEP and see if we can get a good reference to "satisfy these criteria, or else negotiation is needed".

alvestrand:
Closing, new PR with JSEP reference is expected.

Issue #705, opened 2016-06-15
Missing sender identifier attribute (msid)


jcaden:
`replaceTrack` documentation says: 

> To avoid track identifiers changing on the remote receiving end when a track is replaced, the sender must retain the original track identifier and stream associations and use these in subsequent negotiations.

At this point, I can't see a way to get the identifier that a sender is using. If you don't execute `replaceTrack`, you can use track's identifier, but if you change the track there is no way to get it.

It will be great to add a public attribute in the API providing access to the identifier that is being used by the sender.

stefhak:
The way to get that info over to the receiver would be an SDP O/A procedure, and that is what we wanted to avoid with replaceTrack. If the id is important, send it over with some app proprietary signaling, or use removeTrack+addTrack to get the new track, and its Id, to the remote end-point.

jcaden:
But, as long as we are defining a API to avoid making modifications in SDP, shouldn't it be important to let programmers access this information using this API?

Maybe my question was not clear enough. What I meant is that there is no way for the sender peer to know what msid is being used for sending once the `replaceTrack` method is called. Of course, the receiver peer is not affected at all by this changes, as the msid won't change for them.

Without the `msid` property in the sender there is no other way to get this information but parsing the SDP, what in my opinion should not be necessary, that's why this API makes sense, to not require SDP parsing or modifications.

stefhak:
Ah, then I get it. I still think that what matters is the id of the MediaStreamTrack, because that is what would be referred to. I mean that the app of the sending peer can tell the app of the receiving peer that "id xx represents front facing camera and id yy represents room facing camera" as an example. This means that if app of sender peer does not know xx and yy this would not work.
I have not looked into this in detail, but are there situations this could happen?

stefhak:
Closing as we've not heard back from @jcaden. Feel free to re-open if you think that is needed.

jcaden:
Sorry I didn't replied because I need to find better examples and use cases. I'll reopen the issue when I have a clear example.

stefhak:
Sounds good. BTW, note that replaceTrack() is specified to fail if a re-negotiation would be needed. The idea is that nothing in the SDP negotiated session would change, just the source of the media sent. This means that msid would remain unchanged (not sure if it makes a difference in your example, just wanted to highlight).

jcaden:
[...]

PR #1698, opened 2017-12-15
RTPRtcSender: provided properties to get related MediaStreamTrackId and media type

mparis:
This properties are quite useful (even needed) to several use cases where application metadata is attached to an specific msid (as commented in several discussions), but "addTransceiverByKind" is used, so no track is attached to the RtpSender yet.

***
<a href="https://pr-preview.s3.amazonaws.com/mparis/webrtc-pc/pull/1698.html" title="Last updated on Feb 6, 2018, 9:47 AM GMT (1bf977a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1698/f0d453c...mparis:1bf977a.html" title="Last updated on Feb 6, 2018, 9:47 AM GMT (1bf977a)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6881,6 +6881,7 @@ async function updateParameters() {
       <div>
         <pre class="idl">[Exposed=Window] interface RTCRtpTransceiver {
     readonly        attribute DOMString?                  mid;
+    readonly        attribute DOMString                   kind;
     [SameObject]
     readonly        attribute RTCRtpSender                sender;
     [SameObject]
@@ -6905,6 +6906,11 @@ async function updateParameters() {
               negotiation is complete, the <code>mid</code> value may be null.
               After rollbacks, the value may change from a non-null value
               to null.</p>
+            </dd>
+            <dt><dfn><code>kind</code></dfn> of type <span class=
+            "idlAttrType"><a>DOMString</a></span>, readonly</dt>
+            <dd>
+              <p>Media type of the related SDP mline.</p>
             </dd>
             <dt><dfn><code>sender</code></dfn> of type <span class=
             "idlAttrType"><a>RTCRtpSender</a></span>, readonly</dt>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5509,6 +5509,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     readonly        attribute MediaStreamTrack? track;
     readonly        attribute RTCDtlsTransport?  transport;
     readonly        attribute RTCDtlsTransport? rtcpTransport;
+    readonly        attribute DOMString mediaStreamTrackId;
     static RTCRtpCapabilities getCapabilities (DOMString kind);
     Promise&lt;void&gt;             setParameters (optional RTCRtpParameters parameters);
     RTCRtpParameters          getParameters ();
@@ -5564,6 +5565,13 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               <code>transport</code>.</p>
               <p>On getting, the attribute MUST return the value of the
               <a>[[\SenderRtcpTransport]]</a> slot.</p>
+            </dd>
+            <dt><dfn><code>mediaStreamTrackId</code></dfn> of type <span class=
+            "idlAttrType"><a>DOMString</a></span>, readonly, nullable</dt>
+            <dd>
+              <p>msid-appdata token from draft-ietf-mmusic-msid attribute as carried
+              in related SDP mline or null if there is not draft-ietf-mmusic-msid attribute.
+              Useful for senders created without a track (or detached from a track).</p>
             </dd>
           </dl>
         </section>


mparis:
Related to #705

webrtc.html
```diff
@@ -5536,6 +5538,16 @@ <h2>Attributes</h2>
               <p>On getting, the attribute MUST return the value of the
               <a>[[\SenderRtcpTransport]]</a> slot.</p>
             </dd>
+            <dt><dfn><code>msid</code></dfn> of type <span class=
+            "idlAttrType"><a>DOMString</a></span>, readonly</dt>
+            <dd>
+              <p>msid value of the related SDP mline.</p>
```
alvestrand: What do you mean by "msid value" here? Both fields, or just one of them?
(note - the track part used to be available as track.id, but after replaceTrack(), the relationship broke, so making it available makes sense.)

webrtc.html
```diff
@@ -5536,6 +5538,16 @@ <h2>Attributes</h2>
...
+            </dd>
+            <dt><dfn><code>kind</code></dfn> of type <span class=
+            "idlAttrType"><a>DOMString</a></span>, readonly</dt>
+            <dd>
+              <p>Media type of the related SDP mline..</p>
```
alvestrand: This is useful for senders created without a track (or detached from a track).

webrtc.html
```diff
@@ -5482,6 +5482,8 @@ <h3><dfn>RTCRtpSender</dfn> Interface</h3>
     readonly        attribute MediaStreamTrack? track;
     readonly        attribute RTCDtlsTransport?  transport;
     readonly        attribute RTCDtlsTransport? rtcpTransport;
+    readonly        attribute DOMString msid;
```
alvestrand: msid needs to be nullable for the case where the SDP has no msid.

webrtc.html
mparis: I mean the MediaStreamTrack id.
Updating it ;)

fluffy:
I can see this being sort of useful but given this data is not carried in the RTCP, do we have all the info to easily implement this?

mparis:
Hello @fluffy,
first of all thanks for your interesting question.

Regarding _mstId_ attr, it could be easily implemented following [1]:
`
         If a MediaStreamTrack is attached to the transceiver's RtpSender,
         the "a=msid" lines MUST use that track's ID.  If no
         MediaStreamTrack is attached, a valid ID MUST be generated, in
         the same way that the implementation generates IDs for local
         tracks.
`

We have already implemented it in the next way:
When an RtpSender is created [1]:
- If track is null, just generate a mstID
- Otherwise, the track.id will be forever attached even if replaceTrack() is called.

In relation to _kind_ attr:
When RtpSender is created from addTransceiver:
- if a track is provided, _kind_ will be the track.kind.
- Otherwise, kind is provided when addTransceiver is called, so it will be set to _kind_

When RtpSender is created from addTrack,  _kind_ will just be the track.kind.

Refs
[1] https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-5.2.1
[2] https://w3c.github.io/webrtc-pc/#dfn-create-an-rtcrtpsender

webrtc.html
```diff
@@ -5536,6 +5538,19 @@ <h2>Attributes</h2>
               <p>On getting, the attribute MUST return the value of the
               <a>[[\SenderRtcpTransport]]</a> slot.</p>
             </dd>
+            <dt><dfn><code>mstId</code></dfn> of type <span class=
+            "idlAttrType"><a>DOMString</a></span>, readonly, nullable</dt>
+            <dd>
+              <p>MediaStreamTrack id that appears in the msid attribute of the
+              related SDP mline or null if there is not msid attribute.
+              Useful for senders created without a track (or detached from a track).</p>
```
stefhak: Exactly how would this be useful?

webrtc.html
```diff
@@ -5536,6 +5538,19 @@ <h2>Attributes</h2>
...
+            </dd>
+            <dt><dfn><code>kind</code></dfn> of type <span class=
+            "idlAttrType"><a>DOMString</a></span>, readonly</dt>
+            <dd>
```
stefhak: the Transceiver seems like a more natural place for kind IMO.

aboba: Agree, since transceiver.sender.kind would have to be the same as transceiver.kind anyway.

mparis: I agree, so I will change this property to the RtpTransceiver

webrtc.html
mparis: There should be some use cases where this is quite useful.
I am exposing why we needed to implement that:
- Before having any track, you perform addTransceiverByKind(), so an RtpTransceiver with an RtpSender is created
- An SDP offer is created and signaled to the peer attaching some application metadata (needed for APP functionality) related to the MediaStreamTrack id being in the SDP a=msid attr.
- Then the a track is attached to the RtpSender.
- In addition, the track (or others tracks) can be detached and attached again and again.

jan-ivar:
See also https://github.com/w3c/webrtc-pc/issues/1702.

alvestrand:
My original question about what's in the MSID attribute is still not clarified.
Also still blocked on IPR.

mparis:

mparis:
Hello @alvestrand,
regarding IPR is now unblocked as I marked it as non-substantive to accelerate the review process.

In relation to "MSID attr" is now replaced with "mstId attr" and description changed following [1]:
`
If a MediaStreamTrack is attached to the transceiver's RtpSender, the "a=msid" lines MUST use that track's ID. If no MediaStreamTrack is attached, a valid ID MUST be generated, in the same way that the implementation generates IDs for local tracks.
`

Refs
[1] https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-5.2.1

mparis:
Maybe this PR should be aligned with #1718

dontcallmedom:
Marked as  substantive for IPR from ash-nazg.

dontcallmedom:
I have re-marked the pull request as substantive, since that is definitely a substantive change to the specification. @mparis please see get in touch with me (dom@w3.org) to see if and how we can accept your pull request from an IPR perspective.

fluffy:
I just want to add that I totally agree with Dom and this is a substantive from IPR point of view.

webrtc.html
```diff
@@ -5541,6 +5542,13 @@ <h2>Attributes</h2>
               <p>On getting, the attribute MUST return the value of the
               <a>[[\SenderRtcpTransport]]</a> slot.</p>
             </dd>
+            <dt><dfn><code>mstId</code></dfn> of type <span class=
+            "idlAttrType"><a>DOMString</a></span>, readonly, nullable</dt>
+            <dd>
+              <p>MediaStreamTrack id that appears in the msid attribute of the
```
alvestrand: This means the second part of the MSID line.
Why don't you say that this is the msid-appdata token from draft-ietf-mmusic-msid, as carried in the SDP that is shown in the currentLocalDescription for this media section?
That makes it unambiguous.

(I also dislike the name "mstId", because it's an acronym - if you want to call it mediaStreamTrackId, I'll be happier - even though it's sometimes a bit misleading.)

mparis: It makes sense, so I am updating it.
Thanks for the feedback ;).

mparis: Done, but I did not mention anything about currentLocalDescription because IMO this attribute should have a value independently of the negotiation state [1].

Refs
[1] https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-currentlocaldescription

jan-ivar:
I don't think we should add this. You can always get the kind from `transceiver.receiver.track.kind`, and using msid as a correlator is broken. See https://github.com/rtcweb-wg/jsep/issues/842#issuecomment-355634662. Use `transceiver.mid` instead. See near the end of https://blog.mozilla.org/webrtc/the-evolution-of-webrtc/

taylor-b:
Reminder that we did this before and decided to roll it back: https://github.com/w3c/webrtc-pc/pull/1567

IIRC, the reasoning was something along the lines of: "If we support reading the msid-appdata field, we should also support writing it; otherwise it's not a full solution."

dontcallmedom:
given the technical concerns, and the lack of clear path forward wrt IPR, closing this pull request

PR #1701, opened 2017-12-17
Check that transceiver is not stopped before using.

### LINKED ISSUE #1494: addTrack can reuse a sender whose transceiver is stopped
The checks that a sender has to pass in order to be reused by addTrack do not include a check that its transceiver is not stopped. addTrack setting the track on a stopped transceiver is probably not what we want.

stefhak:
Fixes #1494.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4913,7 +4913,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                     </li>
                     <li>
                       <p>The transceiver is not <code>stopped</code>. More
-                      precisely, the <a>[[\stopped]]</a> slot of the
+                      precisely, the <a>[[\Stopped]]</a> slot of the
                       <code><a>RTCRtpTransceiver</a></code> associated with the
                       sender is <code>false</code>.</p>
                     </li>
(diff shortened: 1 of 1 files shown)


fluffy:
One side comment - feel free to ignore  .... the bullet point after this point seem to to make this check sort of redundant. Not sure we need this but seems fine to add anyways.

stefhak:
Thanks @fluffy! I must admit I did not check very carefully, I just wanted to get an issue closed. But on closer inspection I think that if you create a transceiver, and then stop it, its `currentDirection` would never have been anything other than `null` but it is useless so we need the check introduced here.

webrtc.html
```diff
@@ -4911,6 +4911,12 @@ <h2>Methods</h2>
                       <code><a>RTCRtpTransceiver</a></code>, associated with
                       the sender, matches <var>track</var>'s kind.</p>
                     </li>
+                    <li>
+                      <p>The transceiver is not <code>stopped</code>. More
+                      precisely, the <a>[[\stopped]]</a> slot of the
```
adam-be: s/[[\stopped]]/[[\Stopped]]/

PR #1704, opened 2017-12-19
offerToReceive: Tweak attribute descriptions to be a bit more generic

### LINKED ISSUE #1703: offerToReceive: Attribute description is only mentioning the the "true case"
The attribute descriptions could be made a bit more generic since we support "false" now.

adam-be:
...since we support the "false case" now.

fixes #1703


adam-be:
@fippo, can you take a look?

Issue #1700, opened 2017-12-16
Maximum message size slightly incorrect


lgrahl:
> Let remoteMaxMessageSize be the value of the "max-message-size" SDP attribute read from the remote description, as described in [SCTP-SDP] (section 6), or 65535 if the attribute is missing.

It should be `65536` instead of `65535`.

(Sorry for messing this up.)

alvestrand:
@adam-be will update the test and close the bug.

adam-be:
PR for updated test https://github.com/w3c/web-platform-tests/pull/8996

foolip:
I've reviewed https://github.com/w3c/web-platform-tests/pull/8996 on the assumption that the spec change is correct, please merge together with the spec PR.

adam-be:
@foolip, the PR is already merged. It's just the issue that's kept open. Not the ideal workflow, but we'll try to do it right next time.

foolip:
Oh, it's https://github.com/w3c/webrtc-pc/pull/1713, I didn't see that. Having either the spec PR point to the test PR or vice versa is useful for reviewers and people coming along after the fact like me :)

PR #1713, opened 2018-01-03
Maximum message size slightly incorrect

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1700


adam-be:
We need to update the test as well. I can do that.

Issue #1492, opened 2017-07-17
Validation of reordered readonly parameters in setParameters


soareschen:
Step 7 of setParameters says:

> If parameters.encodings.length is different from N, or if any parameter in the parameters argument, marked as a Read-only parameter, has a value that is different from the corresponding parameter value returned from sender.getParameters(), abort these steps and return a promise rejected with a newly created InvalidModificationError. Note that this also applies to transactionId.

And the description for `codec` field for `RTCRtpParameters` says:

> When using the setParameters method, the codecs sequence from the corresponding call to getParameters can be reordered and entries can be removed, but entries cannot be added, and the RTCRtpCodecParameters dictionary members cannot be modified.

This seems to require step 7 to handle reordering/dropping of readonly parameters as well, but it is a bit ambiguous in step 7 whether reordering/dropping is allowed. Even if it allows, does this mean that `encodings` elements can be reordered, and `headerExtensions` elements can be reordered/dropped as well?

aboba:
My take would be that reordering/dropping of `encodings` should not be allowed.  Reordering `headerExtensions` does not make a difference, so I'm not clear this (or dropping) should be allowed either. 

@taylor-b What do you think?

taylor-b:
The order of the encodings is equivalent to their precedence, which is negotiated in SDP via "a=simulcast". So it doesn't make sense to try to reorder them, because it would require renegotiation, which none of the other things you can change in `setParameters` require.

We _could_ say "the order in 'a=simulcast' isn't absolute, it's a 'proposed preference', so the sender can choose to override it." Similar to what we did for codecs. But this would still be getting into "new feature" territory, so unless there's a good use case I'd suggest just not allowing this.

soareschen:
One more thing to note for reordering validation is that current algorithm do not take into account repeated elements in the array. The clause _"If parameters.encodings.length is different from N"_ seems to be specifically added to prevent repeated elements in `encodings`, but the same is not applied to `codec`.

aboba:
@stefhak Can we close this issue since there is a merged PR?

PR #1714, opened 2018-01-03
Validation of reordered readonly parameters in setParameters

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1492

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5595,17 +5595,21 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                 has never been called, return a promise <a>rejected</a> with a newly
                 <a data-link-for="exception" data-lt="create">created</a>
                 <code>InvalidStateError</code>.</li>
-                <li>If <code><var>parameters</var>.encodings.length</code>
-                is different from <var>N</var>, or if
-                <code><var>parameters</var>.encodings</code> has been re-ordered,
-                or if any parameter in <var>parameters</var> is
-                marked as a <dfn>Read-only parameter</dfn> or has a value that is
-                different from the corresponding parameter value in <var>sender</var>'s
-                <a>[[\LastReturnedParameters]]</a> internal slot,
-                return a promise <a>rejected</a> with a newly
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5597,8 +5597,9 @@ <h2>Methods</h2>
                 <code>InvalidStateError</code>.</li>
                 <li>If <code><var>parameters</var>.encodings.length</code>
                 is different from <var>N</var>, or if
-                any parameter in <var>parameters</var>, is
-                marked as a <dfn>Read-only parameter</dfn>, has a value that is
+                <code><var>parameters</var>.encodings</code> has been re-ordered,
+                or if any parameter in <var>parameters</var> is
+                marked as a <dfn>Read-only parameter</dfn> or has a value that is
```
taylor-b: I think "or has a value" is meant to be "and has a value"? In other words, "is read-only *and* was changed"?

stefhak: Yes, agree.

alvestrand: Might be better to rephrase - the sequence "A or B and C" isn't necessarily parsed as "A or (B and C)".

webrtc.html
```diff
@@ -5595,17 +5595,21 @@ <h2>Methods</h2>
...
+                <code>InvalidModificationError</code>:
+                <ol>  
+                   <li><code><var>parameters</var>.encodings.length</code>
+                   is different from <var>N</var>.</li>
+                   <li>Any parameter in <var>parameters</var> is
+                   marked as a <dfn>Read-only parameter</dfn> and has a value that is
+                   different from the corresponding parameter value in <var>sender</var>'s
+                   <a>[[\LastReturnedParameters]]</a> internal slot. Note that this also
+                   applies to <var>transactionId</var>.</li>
+                   <li><code><var>parameters</var>.encodings</code> has been re-ordered.</li>
```
taylor-b: nit: May want to swap the second and third conditions, so that the ones dealing with `parameters.encodings` are next to each other.

Issue #1712, opened 2018-01-03
Should we require a reference from RTPSender to RTCPeerConnection?


alvestrand:
One of the design objectives of the "non-PC" model of WebRTC was that we should be able to work without a PeerConnection if we wanted to. So I'm leery of things that require references to the PC.

It seems that we've added such a requirement to RTPSender by this statement in the description of replaceTrack:

"4. If connection's [[IsClosed]] slot is true, return a promise rejected with a newly created InvalidStateError and abort these steps."

For cleanliness, I'd prefer to keep all the pointers going the other way, which can be achieved by either depending on the "stopped" slot of the transceiver (already checked in step 5) or defining that the procedure "Stop sending media with sender" (referenced from the "close" method of PC) leaves a state marker in a sender's internal slot that can be tested instead.

The last approach has the advantage that senders don't have to know about transceivers either, which is a Good Thing (if we're not using SDP, transceivers are a liability not a feature).

fippo:
Are you looking for [RtpSenderIsStopped](http://draft.ortc.org/#dom-rtcrtpsender-stop)?

aboba:
@alvestrand Step 4 could be removed since step 5 checks whether the `transceiver` has been stopped.  Removing transceiver dependencies as well would require adding `RtpSenderIsStopped` and `RtpReceiverIsStopped` slots as in ORTC.

alvestrand:
@fippo yes, that's a solution
@aboba the difference between step 8 (replace track only) and step 10 (replace track and start sending from it) seems to be rather small; we could treat "not associated with SDP" as a special case of "negotiation not needed".
For completely indepenedent operation, I suspect that we need RTPSender to have a complete representation of the allowable operation parameter envelope, and reject ReplaceTrack if the new track is not within the currently set operation parameter envelope. The initial envelope would then be "sender's maximal abilities", and the post-negotiation envelope would be "SDP-negotiated envelope". Bigger task to specify; ORTC does this as part of the send() function.
Getting rid of one level of reference is good for now.

PR #1716, opened 2018-01-04
Remove reference from RTPSender to RTCPeerConnection

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1712

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5775,17 +5775,6 @@ async function updateParameters() {
                   <var>sender</var>.</p>
                 </li>
                 <li>
-                  <p>Let <var>connection</var> be the
-                  <code><a>RTCPeerConnection</a></code> object that created
-                  <var>sender</var>.</p>
-                </li>
-                <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5842,8 +5831,8 @@ <h2>Methods</h2>
                       <p>Queue a task that runs the following steps:</p>
                       <ol>
                         <li>
-                          <p>If <var>connection</var>'s <a>[[\IsClosed]]</a>
-                          slot is <code>true</code>, abort these steps.</p>
+                          <p>If <var>transceiver</var>'s <a>[[\Stopped]]</a> slot is
+                          <code>true</code>, abort these steps.</p>
```
alvestrand: Why not reject p?
(I know it said "abort" before. Leaving dangling promises when we don't have to seems ugly.)

alvestrand: According to Jan-Ivar, we're being consistent that closed objects drop promises rather than rejecting them.

Issue #1669, opened 2017-11-21
order of transceivers, senders/receivers?


fippo:
split from [this comment](https://github.com/w3c/webrtc-pc/issues/1586#issuecomment-3460708949)

http://w3c.github.io/webrtc-pc/#dfn-collecttransceivers says:
```
The conversion from the transceiver set to the sequence is user agent defined and the 
order does not have to be stable between calls.
```
(there is similar text in mediacapture-main [wrt to the track set](https://w3c.github.io/mediacapture-main/#track-set)

However [JSEP](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-20#section-5.2) implies that the offer is created 
```
in the order the RtpTransceivers were added to the PeerConnection.
```

Given that I would assume that the result of getTransceivers, getSender or getReceivers is in the order they were added to the peerconnection which is stable and defined by JSEP, not by the user agent.

Is there any particular reason not to assume order?

alvestrand:
The reason for the non-ordering of tracks in streams was a long discussion about how it should work - if I remember rightly, the issue is that if the ordering matters, you have to keep it consistent and stable in the face of reordering.
(For tracks in streams, this matters because tracks on a video element (at least at that time) had ordering, so the mapping would have an observable effect.)
This means that you have to define not only whehter stream.add(a); stream.add(b); stream.add(c) produces (a, b, c) or not, but you have to define whether streaam.add(a); stream.add(b); stream.add(c); stream.remove(b) produces (a, <blank>, c) or (a, c) - and if you re-add b, do you get (a, b, c) or (a, c, b)?
Other ideas included sorting on "id" (consistent, but unpredictable).
In the end we decided to abandon the idea of ordering and leave it implementation-defined.

I don't remember a specific decision to not impose order on other sets, but similar arguments may have applied - or the language may have gotten inherited.

The JSEP text seems to require that a specific piece of data is kept with transceivers - addition order. Transceivers in the SDP (the place where ordering matters) can't be removed, or reordered once they're added to the SDP, so the reordering arguments in MedaCapture don't seem to apply as much.
But given the *way* transceivers are added to the SDP, their order won't necessarily be addition order (they're the same for initial offers, but in answers and subsequent offers, they are assigned to "appropriate" m-sections, not assigned in 1-2-3 order), so the create-time ordering and the media-section ordering won't be consistent with each other.

[...]

PR #1719, opened 2018-01-04
Define pc.getTransceivers() et al to be in insertion order.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1669.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4785,19 +4785,14 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               <ol>
                 <li>Let <var>transceivers</var> be the result of executing the
                 <code><a>CollectTransceivers</a></code> algorithm.</li>
-                <li>Let <var>senderset</var> be a new empty set.</li>
+                <li>Let <var>senders</var> be a new empty sequence.</li>
                 <li>For each <var>transceiver</var> in <var>transceivers</var>,
                   <ol>
                     <li>Let <var>sender</var> be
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4847,13 +4837,10 @@ <h2>Methods</h2>
...
-                represents a snapshot of all the
-                <code><a>RTCRtpTransceiver</a></code> objects in this <code><a>
-                  RTCPeerConnection</a></code> object's <a>set of
-                  transceivers</a>. The conversion from the transceiver set to
-                  the sequence is user agent defined and the order does not
-                  have to be stable between calls.
+                <li>Let <var>transceivers</var> be a new sequence consisting of
+                all <code><a>RTCRtpTransceiver</a></code> objects in this
+                <code><a>RTCPeerConnection</a></code> object's
+                <a>set of transceivers</a>, in insertion order.
```
stefhak: Is `insertion order` defined somewhere?

alvestrand:
Ready to merge. (Github is acting up)

webrtc.html
jan-ivar: It's the order our algorithms add transceivers to the *set of transceivers*. This seems well-defined in our spec for local "add*" methods, and in JSEP for ones added by remote media descriptions.

stefhak: works for me!

stefhak:
Just noting that the order of transceivers may be different between the two endpoints, and may differ from m-line order. I guess that's ok.

Issue #1691, opened 2017-12-08
When to fire events triggered by setRemoteDescription.


henbos:
In the spec, we have explored two approaches to firing events as a result of processing the result of an SRD call:
1. Schedule to fire an event.
2. Fire the event immediately midst-processing.

Both are problematic. 1) is problematic because when we fire later, the states of the event might no longer represent the state that caused the event to fire, and 2) is problematic because the event handler could influence the SRD algorithm.

This issue proposes handling all events triggered by SRD - including pc.ontrack, stream.onaddtrack/onremovetrack, track.onmute/onunmute - the way we handle ontrack. That is: the SRD algorithm creates and adds each event to a list, then when the algorithm is done but before the promise is resolved, we do "for each event, fire event".

For more details, here's essentially what I wrote on https://github.com/w3c/webrtc-pc/pull/1667#discussion_r155505342:

Generally speaking, events are typically scheduled to fire later, so by the time they do the effects of multiple events have already occurred. I think this is bad because the event handler gets fired with objects whose state are not even guaranteed to represent the state that caused them to fire.

Firing the events before the last step is not only important because we want the events to fire before the promise resolves, but because if we don't then calling SRD twice or doing other things before the events fire will yield track information in the event handlers that don't correlate with the event that caused them to fire.

I also argue that we cannot fire events midst-processing, or else the event handler can both inspect and influence the state of processing algorithm. Something as simple as "for each transceiver, fire an event on the transceiver" would be problematic because the event handler could modify the set of transceiver (or any other state of the PC that the processing algorithm might use) while we are iterating them.

What if you close the pc or stop or add a track in one of the event handlers? The SRD algorithm would have to handle substeps of the algorithm failing. Even if we know we are safe for now, this is not resilient to future changes and implementations are error prone if they don't think about this stuff, like copying or making sure any container is iterable even if it's modified during iteration or else we might crash.

I argue we must:
1. Fire all events after any processing steps.
2. Have fired all events before the last step (before promise is resolved).

So, if a stream with two tracks are both removed, stream.onremovetrack fires twice and in both events stream.getTracks().length == 0, even on the first track's removal event.
The guarantee is: the state of any object in an event is up-to-date with the result of the SRD call. Meaning the stream is empty.

This is consistent with ontrack. If you add two tracks to a stream, inspecting the stream in the first ontrack yields it contains two tracks. If this wasn't the case you would also have a confusing problem:
```
[...]

PR #1720, opened 2018-01-05
Defer SRD add/remove tracks until right before firing track events.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1691.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5415,6 +5415,11 @@ interface RTCPeerConnectionIceErrorEvent : Event {
             <var>msids</var>.</p>
           </li>
           <li>
+            <p>Let <var>track</var> be <var>receiver</var>'s
+            <a>[[\ReceiverTrack]]</a>.
+            </p>
+          </li>
+          <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5345,32 +5363,38 @@ <h4>Processing Remote MediaStreamTracks</h4>
...
+            <p>Let <var>msids</var> and <var>addList</var> be empty lists.
             </p>
           </li>
           <li>
-            <p><a>Set the associated remote streams</a> for the media
-            description, given <var>receiver</var> and an empty
-            list.</p>
+            <p><a>Set the associated remote streams</a>, given
+            <var>receiver</var>, <var>msids</var>, <var>addList</var>, and
+            <var>removeList</var>.</p>
```
henbos: Since adding only adds to addList and remove only adds to removeList, I think it would be clearer not to have a separate section for "Set the associated remote streams" now that the firing of events has been moved to above.

henbos: Nevermind it's also creating the media stream and other stuff, keep it.

Issue #1694, opened 2017-12-11
RTCCertificate Interface should (or should not) be backed up.


steely-glint:
in "4.10.2 RTCCertificate Interface"
we say 
"The RTCCertificate object can be stored and retrieved from persistent storage by an application. " 
It isn't clear if this should be backed when the rest of a browser is backed up or indeed sync'd (i.e. bookmarks etc).

In contrast to the rest of localStorage, there is no mechanism for a service to access the privateKeys - so no way for them to be backed up and restored (e.g. in case of a phone being immersed in water).

What is the intended behaviour?

aboba:
@steely-glint Section 4.10.2 states: 
"Note that an `RTCCertificate` might not directly hold private keying material, this might be stored in a secure module."

The secure module will generally not support retrieval of the private keying material so that we can't assume that private keying material can be stored and retrieved from persistent storage, only the public keys.

steely-glint:
I guess I'm taking issue with the 'persistent storage' - I'd argue that something that _can't_ be backed up shouldn't be described as persistent.

So I feel we should either change the 'persistent' to something else or we should specify a mechanism to back up the keys - ideally both public and private.

aboba:
@steely-glint Here is my take on the intended behavior. The specification says that the RTCCertificate represents a handle to both the X.509 certificate and the private key. That would seem to imply that when the RTCCertificate is stored and retrieved, that always includes both elements, regardless of whether the private key is stored in a secure module. The format for storage/retrieval isn't specified, but any format that can represent both the certificate and private key (e.g. PKCS12, PEM, etc.) would suffice. The storage is "persistent" across reboots, but while there is an internal slot for the private key, it cannot be retrieved in JS. 

Does this make sense?

steely-glint:
Yes, I think that is the intended behavior. 
I do question however if it is the **expected** behavior - especially for web devs. 
**Everything** that represents state I can think of in a browser can either be backed-up or at least sync'd with other instances of the same browser on different devices that are owned by the same user.

From the user's perspective, if I have my phone stolen and I remote reset it, I expect to be able to restore all my apps and their state onto a new phone (assuming the replacement supports the same os). 

These certs are the single exception I can think of to that experience. If that is the intention, we should clearly state it, possibly flagging their semi-ephemeral nature with a warning to developers.

Issue #1089, opened 2017-03-20
Update for structured cloning changes in HTML


domenic:
In https://github.com/whatwg/html/commit/97d644c97335956610a31e8ad98d1a388c063e84 we changed HTML's formulation of structured cloning significantly for the better. Web RTC should update to have separate serialization and deserialization steps. This can also eliminate having to deal with the memory variable.

I can work on a pull request for this shortly. Please let me know if there are any other Web RTC specs that have structured clone integration.

adam-be:
I don't believe that there are other WebRTC docs affected by this. A PR would be much appreciated.

stefhak:
Re-opening as #1171 reverted #1108 and I would like to make sure we do not forget this.

alvestrand:
This is actually directly relevant to the discussion about what the internal representation of the private key part of a stored certificate is (for some meaning of the word "is").

stefhak:
It seems the W3C HTML spec (5.2) now _is_ updated, so PR https://github.com/w3c/webrtc-pc/pull/1108/files could be re-applied.

PR #1723, opened 2018-01-08
Whether KeyingMaterial internal slot can be stored and retrieved

aboba:
Clarification for Issues https://github.com/w3c/webrtc-pc/issues/1694 and https://github.com/w3c/webrtc-pc/issues/1739

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4623,14 +4623,17 @@ interface RTCPeerConnectionIceErrorEvent : Event {
           </section>
         </div>
         <p>For the purposes of this API, the <a>[[\Certificate]]</a> slot
-        contains unstructured binary data.</p>
-        <p>The <code>RTCCertificate</code> object can be stored and retrieved
-        from persistent storage by an application. Since the <code>RTCCertificate</code>
-        represents a handle to both the certificate and the private key,
-        implementations MUST support storing and retrieving both the certificate
-        and private key, even if the <code>RTCCertificate</code> stores the private
-        key in a secure module. When a <a>user agent</a> is required to obtain a
-        structured clone [[!HTML51]] of an <code>RTCCertificate</code> object,
+        contains unstructured binary data. No mechanism is provided for
+        applications to access the <a>[[\KeyingMaterial]]</a> internal slot.
+        Implementations MUST support applications storing and retrieving
+        <code>RTCCertificate</code> objects from persistent storage.
+        In implementations where an <code>RTCCertificate</code> might not
+        directly hold private keying material (it might be stored in a
+        secure module), a reference to the private key can be held in
+        the <a>[[\KeyingMaterial]]</a> internal slot, allowing the
+        private key to be stored and retrieved.</p>
+        <p>When a <a>user agent</a> is required to obtain a structured
+        clone [[!HTML51]] of an <code>RTCCertificate</code> object,
         it performs the following steps:</p>
         <ol>
           <li>Let <var>input</var> and <var>memory</var> be the corresponding
--- a/webrtc.html
+++ b/webrtc.html
@@ -4626,7 +4626,8 @@ interface RTCPeerConnectionIceErrorEvent : Event {
         contains unstructured binary data.</p>
         <p>Note that an <code>RTCCertificate</code> might not directly hold
         private keying material, this might be stored in a secure module.</p>
-        <p>The <code>RTCCertificate</code> object can be stored and retrieved
+        <p>The <code>RTCCertificate</code> object, with the exception of the
+        the <a>[[\KeyingMaterial]]</a> internal slot, can be stored and retrieved
         from persistent storage by an application. When a <a>user agent</a> is
         required to obtain a structured clone [[!HTML51]] of an
         <code>RTCCertificate</code> object, it performs the following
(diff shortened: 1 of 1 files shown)


aboba:
@ekr Can you review?

webrtc.html
```diff
@@ -4626,7 +4626,8 @@ <h2>Methods</h2>
         contains unstructured binary data.</p>
         <p>Note that an <code>RTCCertificate</code> might not directly hold
         private keying material, this might be stored in a secure module.</p>
-        <p>The <code>RTCCertificate</code> object can be stored and retrieved
+        <p>The <code>RTCCertificate</code> object, with the exception of the
+        the <a>[[\KeyingMaterial]]</a> internal slot, can be stored and retrieved
```
martinthomson: This isn't necessary.  The object can be stored and retrieved, including the KeyingMaterial internal slot.

Firefox implements this.  An RTCCertificate can be stored to indexedDB and later retrieved and used.  This is exactly the same process that we use for keys in WebCrypto.  See https://w3c.github.io/webcrypto/Overview.html#concepts-key-storage

alvestrand: So the keying material can be stored on storage that is persistent to the browser, but there's no way of getting it out so that it can be accessed by the Javascript client, is that right?
#1694 was complaining about using the word "persistent" for that kind of storage.

aboba: If private keying material is held in a secure module, does this affect the ability to store and retrieve the KeyingMaterial internal slot?  Just trying to understand what the requirements are for storing and retrieving certificates. Typically the term "persistent" implies the ability to persist across reboots.

martinthomson: What happens here is that the keying material is stored however the browser can.  If it is in a secure module, the browser needs to store a handle to the stored keying material so that it can be used again.  If the secure module goes away, then it won't be available until it comes back.  It's still persistent though in the sense that if the machine reboots, it's still there unless the user explicitly wipes the storage (or the module).

Javascript can never directly use these keys, it can only ask the browser to use them in certain predefined ways.  Just as the browser can't use the keys directly if there is a secure module in use, it can only ask the secure module to use them in predefined ways.  For these keys, they can only be used to make new WebRTC connections.

webrtc.html
```diff
@@ -4624,10 +4624,12 @@ <h2>Methods</h2>
...
         contains unstructured binary data.</p>
-        <p>Note that an <code>RTCCertificate</code> might not directly hold
-        private keying material, this might be stored in a secure module.</p>
         <p>The <code>RTCCertificate</code> object can be stored and retrieved
-        from persistent storage by an application. When a <a>user agent</a> is
+        from persistent storage by an application. Note that an
+        <code>RTCCertificate</code> might not directly hold
+        private keying material, this might be stored in a secure module,
+        in which case the <a>[[\KeyingMaterial]]</a> internal slot might not
+        be storable and retrievable. When a <a>user agent</a> is
```
alvestrand: If we store it and retrieve it on the same host, would we assume that the KSM can still be accessed?

fluffy: all the web crpto seem pretty vague about this. I think the answer is "maybe". AFAIC this is all way under specified. So for example, if the user and told the browser to clear the locally storage, it would almost certainly be gone - I think it is often tied to saving in a one of the web databases / local storage options. So if that is gone, it is probably gone even if it is still in the KSM. How memory on the KSM is ever reclaimed is not clear to me. But .... I don't think we can fix this in WebRTC spec.

aboba:
@fluffy @martinthomson Can you review at the updated PR?

martinthomson:
The current text implies that the private keys would need to be extracted from the KSM when doing a structured clone.  That's not possible in the general case.  I think that instead we need some sort of failure mode for the case when one of these is *used*.  That would also cover the case where you made a cert and the KSM with the private key was removed/flushed before use.

webrtc.html
```diff
@@ -4624,13 +4624,14 @@ <h2>Methods</h2>
...
         <p>The <code>RTCCertificate</code> object can be stored and retrieved
-        from persistent storage by an application. When a <a>user agent</a> is
-        required to obtain a structured clone [[!HTML51]] of an
-        <code>RTCCertificate</code> object, it performs the following
-        steps:</p>
+        from persistent storage by an application. Since the <code>RTCCertificate</code>
+        represents a handle to both the certificate and the private key,
+        implementations MUST support storing and retrieving both the certificate
+        and private key, even if the <code>RTCCertificate</code> stores the private
+        key in a secure module. When a <a>user agent</a> is required to obtain a
```
alvestrand: Suggested text:
Note that there is no way to access the private key, so it doesn't matter to the user whether the private key is stored directly or if it is stored as a reference to some other key storage such as a KSM.

aboba:
@martinthomson @alvestrand  PR updated to clarify that the [[KeyingMaterial]] slot is not accessible by the application, and could represent a reference to the key material stored in a KSM, permitting storage and retrieval as well as structured cloning.

webrtc.html
```diff
@@ -4623,14 +4623,18 @@ <h2>Methods</h2>
...
-        steps:</p>
+        contains unstructured binary data. No mechanism is provided for
+        applications to access the <a>[[\KeyingMaterial]]</a> internal slot.
+        Implementations MUST support applications storing and retrieving
+        <code>RTCCertificate</code> objects from persistent storage.
+        In implementations where an <code>RTCCertificate</code> might not
+        directly hold private keying material (it might be stored in a
+        secure module), a reference to the private key can be held in
+        the <a>[[\KeyingMaterial]]</a> internal slot, allowing the
+        private key to be stored and retrieved.</p>
```
fluffy: Perhaps change "retrieved" to "used"

alvestrand:
See also #1089

PR #1725, opened 2018-01-08
replaceTrack "negotiation needed" clarification

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1389

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5826,8 +5826,8 @@ async function updateParameters() {
                     <li>
                       <p>Determine if negotiation is needed to transmit
                       <var>withTrack</var> in place of the sender's existing
-                      track. Negotiation is not needed if no codec changes
-                      are required, or <var>withTrack</var>
+                      track. Negotiation is not needed if no codec envelope
+                      changes are required, or <var>withTrack</var>
                       is null. Ignore which <code>MediaStream</code> the track
                       resides in and the <code>id</code> attribute of the track
                       in this determination. If negotiation is needed, then
--- a/webrtc.html
+++ b/webrtc.html
@@ -5826,9 +5826,10 @@ async function updateParameters() {
                     <li>
                       <p>Determine if negotiation is needed to transmit
                       <var>withTrack</var> in place of the sender's existing
-                      track. Negotiation is not needed if no codec envelope
-                      changes are required, or <var>withTrack</var>
-                      is null. Ignore which <code>MediaStream</code> the track
+                      track. Negotiation is not needed if no changes are
+                      required to the currently negotiated SDP offer/answer
+                      envelope, or <var>withTrack</var> is null.
+                      Ignore which <code>MediaStream</code> the track
                       resides in and the <code>id</code> attribute of the track
                       in this determination. If negotiation is needed, then
                       <a>reject</a> <var>p</var> with a newly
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5826,7 +5826,8 @@ <h2>Methods</h2>
                     <li>
                       <p>Determine if negotiation is needed to transmit
                       <var>withTrack</var> in place of the sender's existing
-                      track. Negotiation is not needed if <var>withTrack</var>
+                      track. Negotiation is not needed if no codec changes
```
alvestrand: Do we want the word "envelope" in there?
ie changing from 640x480 to 320x240 doesn't require changes if the negotiated limits are 0x0 to 1024x768.

webrtc.html
```diff
@@ -5826,8 +5826,10 @@ <h2>Methods</h2>
                     <li>
                       <p>Determine if negotiation is needed to transmit
                       <var>withTrack</var> in place of the sender's existing
-                      track. Negotiation is not needed if <var>withTrack</var>
-                      is null. Ignore which <code>MediaStream</code> the track
+                      track. Negotiation is not needed if no changes are
+                      required to the currently negotiated SDP offer/answer
+                      envelope, or <var>withTrack</var> is null.
```
taylor-b: Something that just came up, in https://github.com/w3c/webrtc-pc/issues/1728: what if the current negotiated direction is "inactive"? Changes *are* required to send the track, technically. But I don't think we want to force the application to wait until the direction changes before calling `replaceTrack`. Do you agree? If so, do we make a specific exception for the direction, or is there some other way this could be worded...

stefhak:
LGTM

burnburn:
This addition seems to be the general definition of what we always assumed would trigger NN, namely that the SDP has changed.  Is this the right place for that general definition?

webrtc.html
```diff
@@ -5827,9 +5827,11 @@ <h2>Methods</h2>
                       <p>Determine if negotiation is needed to transmit
                       <var>withTrack</var> in place of the sender's existing
                       track. Negotiation is not needed if <var>withTrack</var>
-                      is null. Ignore which <code>MediaStream</code> the track
-                      resides in and the <code>id</code> attribute of the track
-                      in this determination. If negotiation is needed, then
+                      is null or if no changes are required to the currently
+                      negotiated SDP offer/answer envelope. Ignore which
```
alvestrand: Somewhere this paragraph needs to define "the currently negotiated SDP offer/answer envelope" before we have negotiated anything - should probably be = the local capabilities (what's in createOffer).

webrtc.html
```diff
@@ -5827,9 +5827,11 @@ <h2>Methods</h2>
                       <p>Determine if negotiation is needed to transmit
                       <var>withTrack</var> in place of the sender's existing
                       track. Negotiation is not needed if <var>withTrack</var>
-                      is null. Ignore which <code>MediaStream</code> the track
-                      resides in and the <code>id</code> attribute of the track
-                      in this determination. If negotiation is needed, then
+                      is null or if no changes are required to the currently
```
taylor-b: nit: Would be nice to avoid the double negative, "negotiation is not needed if no changes are required."

aboba:
This has been obsoleted by PR https://github.com/w3c/webrtc-pc/pull/1772

PR #1743, opened 2018-01-22
Properly defined direction when creating transceiver.

stefhak:
I think this is the way it should be (though we seem to mix using quotation signs or not for direction).

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1670,8 +1670,9 @@
                               <li>
                                 <p><a>Create an RTCRtpTransceiver</a> with
                                 <var>sender</var>, <var>receiver</var> and
-                                <var>direction</var> set to <code>"recvonly"</code>,
-                                and let <var>transceiver</var> be the result.</p>
+                                an <code><a>RTCRtpTransceiverDirection</a></code>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1670,8 +1670,9 @@ <h4>Set the RTCSessionDescription</h4>
                               <li>
                                 <p><a>Create an RTCRtpTransceiver</a> with
                                 <var>sender</var>, <var>receiver</var> and
-                                <var>direction</var> set to <code>"recvonly"</code>,
-                                and let <var>transceiver</var> be the result.</p>
+                                an <code><a>RTCRtpTransceiverDirection</a></code>
+                                value of <code>"recvonly"</code>, and let
```
alvestrand: Hm. This section is part of "apply a description", not of "create transceiver".
Shouldn't the "direction" be the negotiated direction from the SDP?
(it's not really easy to see where this line sits...)

taylor-b: The reason it's "recvonly" is that if you receive a "sendrecv" offer but don't have anything to send (no `addTrack`-matched transceiver), you want to reply with a "recvonly" answer by default. So this is correct/intentional.

Issue #1737, opened 2018-01-17
Add direction change as reason for "mute" event


stefhak:
Section 5.3, description of the `track` attribute has
```When one of the SSRCs for RTP source media streams received by receiver is removed (either due to reception of a BYE or via timeout), the mute event is fired at track.```

in its text. However, https://w3c.github.io/webrtc-pc/#set-description can also lead to the track becoming muted ("`For each track in muteTracks, set the muted state to the value true.`") for certain changes of `direction`. Should we add this reason for firing `mute` to the `track` attribute description in 5.3?

alvestrand:
I think the problematic thing here was that if the track is muted due to setRemoteDescription, and there were packets straggling in late, it could be unmuted while the transceiver direction is invalid for unmuting.

I think the proper solution is to guard the unmuting in 5.3 by a check on the transceiver state: "If packets are received again, the track is muted, and the associated transceiver's direction is still sendrecv or recvonly, unmute the track".

jan-ivar:
I believe that's what https://github.com/w3c/webrtc-pc/issues/1707 is about.

Issue #1707, opened 2017-12-21
onmute then onunmute can fire before negotiation completes


docfaraday:
Let's say we're renegotiating, and the offerer has muted a track using "recvonly" or "inactive" in the offer.

answerer.setRemoteDescription(offer) causes onmute to fire, but let's say an RTP packet for that track arrives, causing the track to be unmuted. Then negotiation completes (which definitively tells us that track is muted), but we don't fire onmute again.

Is this the desired behavior?

taylor-b:
I'd assume the RTP packet would be ignored. It was either sent before the "inactive" offer, in which case it's a case of packet reordering, or afterwards, which would be strange.

When you talk about negotiation completing, is this the point at which the answerer decides "I'm going to either accept this offer, or roll it back?" This seems to relate to #1706 then.

docfaraday:
By "negotiation completing", I mean answerer.setLocalDescription(answer).

I don't think we can assume this packet will be ignored,  because the _offerer_ might need to roll back too, and it doesn't make sense to drop packets until it is sure this will not happen (ie; at offerer.setRemoteDescription(answer)).

docfaraday:
Maybe we could specify that unmuting happens only in the stable state? Although that would preclude rendering pre-answer media from new tracks. Hmm.

docfaraday:
Or, we could just do away with the idea of using RTP to drive unmute, and fire it at the same time as ontrack.

alvestrand:
@docfaraday have you observed this behavior in a browser?
I assume the spec line you're thinking of is this one:
"When one of the SSRCs for RTP source media streams received by receiver is removed (either due to reception of a BYE or via timeout), the mute event is fired at track. If and when packets are received again, the unmute event is fired at track." - from the attribute description of RTPReceiver.

docfaraday:
@alvestrand I have not observed this, but it does seem there is room for this to occur under the spec. It would seem sensible to avoid unmuting on reception of RTP if the current negotiation state says that the track ought to be muted.

stefhak:
I'd agree to https://github.com/w3c/webrtc-pc/issues/1707#issuecomment-355335685, essentially meaning that we say that "direction" overrides reception of (perhaps spurious) RTP packets.

alvestrand:
It seems that #1737 and this both go away if you check against the current direction before unmuting the track.

PR #1748, opened 2018-01-30
Gating 'unmute' events on track with transceiver direction.

stefhak:
Fix for #1737 and #1707

***
<a href="https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1748.html" title="Last updated on Mar 8, 2018, 1:13 PM GMT (bc78c39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1748/312b635...stefhak:bc78c39.html" title="Last updated on Mar 8, 2018, 1:13 PM GMT (bc78c39)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -10869,7 +10869,7 @@ async function consumeIdentityAssertion() {
       one and only one <code>MediaStreamTrack</code> to the recipient. A peer
       is defined as a user agent that supports this specification. In addition,
       the sending side application can indicate what <code>MediaStream</code>
-      object(s) the <code>MediaStreamTrack</code> is member of. The
+      object(s) the <code>MediaStreamTrack</code> is a member of. The
       corresponding <code>MediaStream</code> object(s) on the receiver side
       will be created (if not already present) and populated accordingly.</p>
       <p>As also described earlier in this document, the objects
@@ -10943,7 +10943,7 @@ async function consumeIdentityAssertion() {
       source whose corresponding <code><a>MediaStreamTrack</a></code> is muted,
       and the <a>[[\Receptive]]</a> slot of the 
       <a><code>RTCRtpTransceiver</code></a> object the
-      <a><code>RTCRtpReceiver</code></a> is member of is <code>true</code>,
+      <a><code>RTCRtpReceiver</code></a> is a member of is <code>true</code>,
       it MUST queue a task to <a>set the muted state</a> of the corresponding
       <code><a>MediaStreamTrack</a></code> to <code>false</code>.
       </p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -6614,11 +6614,10 @@ async function updateParameters() {
               by <var>receiver</var> is removed (either due to reception of
               a BYE or via timeout), <a>set the muted state</a> of the
               <code>track</code> to the value <code>true</code>.
-
-              <a data-lt="set the RTCSessionDescription">
-              Setting an <code>RTCSessionDescription</code></a> can also 
-              lead to <a data-lt="set the muted state"> setting the muted
-              state</a> of the <code>track</code> to the
+              Note that <a data-lt="set the RTCSessionDescription">
+              setting an <code>RTCSessionDescription</code></a>
+              can also lead to <a data-lt="set the muted state"> setting
+              the muted state</a> of the <code>track</code> to the
               value <code>true</code>.
               If packets are received again, and the <code>track</code> is
               muted, and either the [[\PendingDirection]] or
--- a/webrtc.html
+++ b/webrtc.html
@@ -1635,7 +1635,7 @@
                       <li>
                         <p>If <var>description</var> is of type
                         <code>"offer"</code> 
-                        set <var>transceiver</var>'s <a>[[\PendingDirection]]</a> slot
+                        set <var>transceiver</var>'s <a>[[\OfferedDirection]]</a> slot
                         to an <code><a>RTCRtpTransceiverDirection</a></code>
                         value representing the direction of the <a>media description</a>.</p>
                       </li>
@@ -1645,7 +1645,7 @@
                         set <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot
                         to an <code><a>RTCRtpTransceiverDirection</a></code>
                         value representing the direction of the <a>media description</a>
-                        and set <var>transceiver</var>'s <a>[[\PendingDirection]]</a>
+                        and set <var>transceiver</var>'s <a>[[\OfferedDirection]]</a>
                         slot to the value <code>null</code>.</p>
                       </li>
                     </ol>
@@ -1739,7 +1739,7 @@
                                 <p>Set <var>transceiver</var>'s
                                 <a>[[\CurrentDirection]]</a> slot to
                                 <var>direction</var> and set <var>transceiver</var>'s
-                                <a>[[\PendingDirection]]</a> slot to <code>null</code>.</p>
+                                <a>[[\OfferedDirection]]</a> slot to <code>null</code>.</p>
                               </li>
                               <li>
                                 <p>
@@ -6619,7 +6619,7 @@ async function updateParameters() {
               the muted state</a> of the <code>track</code> to the
               value <code>true</code>.
               If packets are received again, and the <code>track</code> is
-              muted, and either the [[\PendingDirection]] or
+              muted, and either the [[\OfferedDirection]] or
               [[\CurrentDirection]] slot of the 
               <a><code>RTCRtpTransceiver</code></a> the
               <a><code>RTCRtpReceiver</code></a> is member of is either 
@@ -6926,7 +6926,7 @@ async function updateParameters() {
           <var>direction</var>.</p>
         </li>
         <li>
-          <p>Let <var>transceiver</var> have a <dfn>[[\PendingDirection]]</dfn> internal slot,
+          <p>Let <var>transceiver</var> have a <dfn>[[\OfferedDirection]]</dfn> internal slot,
           initialized to null.</p>
         </li>
         <li>
@@ -7143,7 +7143,7 @@ async function updateParameters() {
                   slot to <code>true</code>.</p>
                 </li>
                 <li>
-                  <p>Set <var>transceiver</var>'s <a>[[\PendingDirection]]</a>
+                  <p>Set <var>transceiver</var>'s <a>[[\OfferedDirection]]</a>
                   slot to <code>null</code>.</p>
                 </li>
                 <li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -10939,13 +10939,12 @@ async function consumeIdentityAssertion() {
       source, as is the case for each <code>MediaStreamTrack</code>
       associated with
       an <code><a>RTCRtpReceiver</a></code>) is always strong.</p>
-      <p>When an <code><a>RTCRtpReceiver</a></code> receives data on an RTP
-      source for the first time, and the [[\Receptive]]
-      slot of the 
+      <p>Whenever an <code><a>RTCRtpReceiver</a></code> receives data on an RTP
+      source whose corresponding <code><a>MediaStreamTrack</a></code> is muted,
+      and the <a>[[\Receptive]]</a> slot of the 
       <a><code>RTCRtpTransceiver</code></a> object the
-      <a><code>RTCRtpReceiver</code></a> is member of is either 
-      <code>true</code>, it MUST queue a task to
-      <a>set the muted state</a> of the corresponding
+      <a><code>RTCRtpReceiver</code></a> is member of is <code>true</code>,
+      it MUST queue a task to <a>set the muted state</a> of the corresponding
       <code><a>MediaStreamTrack</a></code> to <code>false</code>.
       </p>
       <p>When one of the SSRCs for RTP source media streams received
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6530,8 +6555,17 @@ <h2>Attributes</h2>
...
+              <code>track</code>. <a data-lt="set the RTCSessionDescription">
+              Setting an <code>RTCSessionDescription</code></a> can also
+              lead to the firing of the <code>mute</code> event at
+              <code>track</code>.
+              If and when packets are received again, and either the
+              [[\PendingDirection]] or [[\CurrentDirection]] slot of the
+              <a><code>RTCRtpTransceiver</code></a> the
+              <a><code>RTCRtpReceiver</code></a> is member of is either 
+              <code>"recvonly"</code> or <code>"sendrecv"</code>, the
+              the <code>unmute</code> event is fired at <code>track</code>.
```
fluffy: If it was muted by a BYE, an out of order packet that comes after the BYE should not unmute it.

stefhak: That part is unchanged, filed #1748 to track resolution.

webrtc.html
```diff
@@ -1549,81 +1549,104 @@ <h4>Set the RTCSessionDescription</h4>
...
                     then run the following steps for each <a>media description</a>
-                    in <var>description</var> that is not yet <a>associated</a> with
-                    an <code><a>RTCRtpTransceiver</a></code> object:</p>
+                    in <var>description</var>:</p>
                     <ol>
                       <li>
-                        <p>Let <var>transceiver</var> be the <code>
-                        <a>RTCRtpTransceiver</a></code> used to create the
-                        <a>media description</a>.</p>
+                        If <a>media description</a> in <var>description</var> is not yet
```
taylor-b: nit: "If *the* media description..."

webrtc.html
```diff
@@ -1549,81 +1549,104 @@ <h4>Set the RTCSessionDescription</h4>
...
-                        <p>
-                          Set <var>transceiver</var>.<a>[[\Receiver]]</a>.<a>[[\ReceiverRtcpTransport]]</a>
-                          to <var>rtcpTransport</var>.
-                        </p>
+                        <p>If <var>description</var> is of type
+                        <code>"answer"</code> or <code>"pranswer"</code>, then
+                        set <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot
+                        to an <code><a>RTCRtpTransceiverDirection</a></code>
+                        value representing the direction of the corresponding
+                        <a>media description</a>.</p>
```
taylor-b: What about [[PendingDirection]]?

webrtc.html
```diff
@@ -6529,9 +6553,22 @@ <h2>Attributes</h2>
...
+              a BYE or via timeout), <a>set the muted state</a> of the
+              <code>track</code> to the value <code>true</code>.
+              Note that <a data-lt="set the RTCSessionDescription">
+              setting an <code>RTCSessionDescription</code></a>
+              can also lead to <a data-lt="set the muted state"> setting
+              the muted state</a> of the <code>track</code> to the
+              value <code>true</code>.
+              If packets are received again, and the <code>track</code> is
+              muted, and either the [[\PendingDirection]] or
+              [[\CurrentDirection]] slot of the
```
taylor-b: Is this 100% correct? Suppose `[[CurrentDirection]]` is `inactive`, and I receive a `sendrecv` remote offer. The remote endpoint isn't supposed to send anything until it receives my answer saying "yes, I agree to receiving." So if I receive packets in this state, shouldn't they be ignored?

This is different than the case where I *send* a `sendrecv` offer. Because in that case, it would be legal for media to arrive before the answer.

So, do we need to differentiate the cases where `[[PendingDirection]]` is set from a local offer vs. set from a remote offer? Or is my interpretation of the SDP direction rules wrong?

webrtc.html
stefhak: I think it ([[PendingDirection]]) should be set to null. I did not do that because it is not used here until it's set again, but forgot its role in 'unmute'.

webrtc.html
stefhak: As done now [[PendingDirection]] is _only_ used on the offering side, and its sole purpose is to (as you say) allow for media arriving before the answer to unmute.

We could perhaps expand to use [[PendingDirection]] on the answering side, but that's not part of this PR (and not needed for what it is trying to fix).

webrtc.html
stefhak: Good catch!

webrtc.html
taylor-b: I assume you mean "only on the local side" rather than "only on the offering side"? I guess this works then. Though it's not obvious from the name "PendingDirection," so maybe this could be clarified where it's defined.

jan-ivar:
Patch has a lot of white space changes. I recommend [this view](https://github.com/w3c/webrtc-pc/pull/1748/files?w=1) of Files changed.

webrtc.html
```diff
@@ -6570,9 +6594,22 @@ <h2>Attributes</h2>
...
+              a BYE or via timeout), <a>set the muted state</a> of the
+              <code>track</code> to the value <code>true</code>.
+              Note that <a data-lt="set the RTCSessionDescription">
+              setting an <code>RTCSessionDescription</code></a>
+              can also lead to <a data-lt="set the muted state"> setting
+              the muted state</a> of the <code>track</code> to the
+              value <code>true</code>.
+              If packets are received again, and the <code>track</code> is
+              muted, and either the [[\OfferedDirection]] or
+              [[\CurrentDirection]] slot of the
```
jan-ivar: If `[[OfferedDirection]]` is non-null, then I think it must trump [[CurrentDirection]] or it won't work right.

webrtc.html
```diff
@@ -1549,81 +1549,104 @@ <h4>Set the RTCSessionDescription</h4>
...
                       <li>
                         <p>If <var>transceiver</var>'s <a>[[\Stopped]]</a> slot
                         is <code>true</code>, abort these sub steps.</p>
                       </li>
                       <li>
                         <p>If <var>description</var> is of type
-                        <code>"answer"</code> or <code>"pranswer"</code>, then
-                        set <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot
+                        <code>"offer"</code> 
+                        set <var>transceiver</var>'s <a>[[\OfferedDirection]]</a> slot
```
jan-ivar: I think we also need to set `[[OfferedDirection]]` in SRD(offer). Otherwise, if packets arrive inbetween SRD(offer) and SLD(answer) then we might accidentally unmute just-muted tracks, because `[[CurrentDirection]]` isn't updated yet (it gets updated very soon in SLD(answer) but this leaves a timing-hole).

I also worry about rollback.

webrtc.html
```diff
@@ -1549,81 +1549,104 @@ <h4>Set the RTCSessionDescription</h4>
...
+                        with an <code><a>RTCRtpTransceiver</a></code> object then run
+                        the following steps:
+                        <ol>
+                          <li>
+                            <p>Let <var>transceiver</var> be the <code>
+                            <a>RTCRtpTransceiver</a></code> used to create the
+                            <a>media description</a>.</p>
+                          </li>
+                          <li>
+                            <p>Set <var>transceiver</var>'s <code><a data-for=
```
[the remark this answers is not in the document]
stefhak: Thanks, will update.

webrtc.html
```diff
@@ -6570,9 +6594,22 @@ <h2>Attributes</h2>
...
-              a BYE or via timeout), the <code>mute</code> event is fired at
-              <code>track</code>.  If and when packets are received again,
-              the <code>unmute</code> event is fired at <code>track</code>.</p>
+              a BYE or via timeout), <a>set the muted state</a> of the
+              <code>track</code> to the value <code>true</code>.
+              Note that <a data-lt="set the RTCSessionDescription">
+              setting an <code>RTCSessionDescription</code></a>
+              can also lead to <a data-lt="set the muted state"> setting
+              the muted state</a> of the <code>track</code> to the
+              value <code>true</code>.
```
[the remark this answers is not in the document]
stefhak: I wanted to safe guard so that a future update doesn't make setLocal update the muted state. But fair enough, I will update.

webrtc.html
stefhak: I think the proposed text actually works, if at least one of [[OfferedDirection]] and [[CurrentDirection]] includes "receive" incoming packets can lead to "unmute".

webrtc.html
```diff
@@ -6570,9 +6594,22 @@ <h2>Attributes</h2>
...
+              If packets are received again, and the <code>track</code> is
+              muted, and either the [[\OfferedDirection]] or
+              [[\CurrentDirection]] slot of the 
+              <a><code>RTCRtpTransceiver</code></a> the
+              <a><code>RTCRtpReceiver</code></a> is member of is either
```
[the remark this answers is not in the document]
stefhak: will update

webrtc.html
```diff
@@ -6570,9 +6594,22 @@ <h2>Attributes</h2>
               <code><a>RTCRtpReceiver</a></code> object <var>receiver</var>.
               When one of the SSRCs for RTP source media streams received
               by <var>receiver</var> is removed (either due to reception of
-              a BYE or via timeout), the <code>mute</code> event is fired at
-              <code>track</code>.  If and when packets are received again,
-              the <code>unmute</code> event is fired at <code>track</code>.</p>
+              a BYE or via timeout), <a>set the muted state</a> of the
+              <code>track</code> to the value <code>true</code>.
```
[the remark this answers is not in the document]
stefhak: Yes, that's something I was thinking of. Though, 10.3 needs some updates since it is just mentioning PeerConnection (and has not been updated to reflect the introduction of receiver and transceiver).

Also, 'RTP source' is undefined.

webrtc.html
stefhak: I worry about rollback as well.

But I don't think we should set [[OfferedDirection]] in SRD at all. [[OfferedDirection]] just indicates that the offerer has an outstanding offer (and should be prepared to receive if the [[OfferedDirection]] says so).
The answerer doesn't need to use (or set [[OfferedDirection]]), instead it decides if it's going to accept the offered session at SLD(answer). And SLD(answer) updates [[CurrentDirection]], and before [[CurrentDirection]] says so the answerer should not accept media IIUC.

docfaraday: This does not solve the problem in issue #1707 at all. This issue arises when:

answerer->SRD(offer); <--- offer that mutes a track
Stray RTP packet for track arrives at answerer, unmuting the track
answerer->SLD(answer); <--- doesn't alter the muted state of the track

I really do think the simplest/easiest/most likely to work fix is to disable  "unmute on RTP" when the signaling state isn't "stable".

stefhak: @docfaraday yes, that's an oversight by me. I just changed the PR. Would this work as intended?

stefhak:
@jan-ivar @fluffy @docfaraday @taylor-b it would be good if you could take a look at this updated approach. I'm not married to it in any way, it's more that I'm eager to decide to either move forward with this or kill it in favor of some other approach.

webrtc.html
```diff
@@ -1689,6 +1718,12 @@ <h4>Set the RTCSessionDescription</h4>
                           </li>
                           <li>
                             <p>If <var>direction</var> is <code>"sendrecv"</code> or
+                            <code>"recvonly"</code>,
+                            set <var>transceiver</var>'s <a>[[\Receptive]]</a> slot
+                            to <code>true</code>, otherwise set it to <code>false</code>.</p>
```
taylor-b: This is in the "remote description" steps, but don't we only want the slot set to `true` from *local* descriptions, as per [this comment thread](https://github.com/w3c/webrtc-pc/pull/1748#discussion_r165148079)?

Though I think it would still be correct to set the slot to `false` here.

stefhak: I agree, that's a mistake by me. I'll update.

alvestrand:
@jan-ivar will review; looks ready for merge to others.

webrtc.html
```diff
@@ -1549,81 +1553,106 @@ <h4>Set the RTCSessionDescription</h4>
...
-                      </li>
-                      <li>
-                        <p>
-                          Set <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SenderRtcpTransport]]</a>
-                          to <var>rtcpTransport</var>.
-                        </p>
+                        <p>If <var>direction</var> is <code>"sendrecv"</code> or
+                        <code>"recvonly"</code>,
+                        set <var>transceiver</var>'s <a>[[\Receptive]]</a> slot
+                        to <code>true</code>, otherwise set it to <code>false</code>.</p>
```
jan-ivar: Since this is in SLD, do we need to handle rollback?

I'm a bit concerned subsequent SLD calls with new offers can now mess with track state from a successful negotiation. E.g. The following is no longer benign:
```js
await pc.setRemoteDescription(incomingOfferSansTrack);
await pc.setLocalDescription(await pc.createAnswer());
signalingChannel.send({description: pc.localDescription});
// receiver.track.muted is true here, and can't be unmuted by any stray incoming rtp packets
pc.direction = "sendrecv";
await pc.setLocalDescription(await pc.createOffer());
await pc.setLocalDescription({type: "rollback"});
// stray incoming rtp packets can now cause unmute to fire on receiver.track 
```
Unsure what the way out here is.

stefhak: I guess the direction of the m-line in the remoteDescription is either 'inactive' or 'receiveonly' (meaning that the other endpoint does not intend to send)?

Then you change the direction (should be transceiver.direction, not pc.) indicating that you're willing to receive, and apply that offer locally before rolling back.

Would not the rollback essentially apply a description that (direction) says you're not willing to receive, and thus [[Receptive]] would go back to false? I'm not sure this PR makes this problem any worse.

stefhak: Ah. I opened #1783 cause I had the feeling you should in fact do the steps in 4.4.1.6, only with the previous description. That may be wrong, but it seems strange that SLD(offer) can change CurrentDirection, but a following SLD(rollback) would not change it back. Perhaps we should expand the text you mention to include [[Receptive]] but then also perhaps add [[CurrentDirection]].

Issue #1749, opened 2018-01-30
Need <section> around dictionaries under RTCRtpSender


alvestrand:
As it stands, the spec reads oddly after the RTCRtpSender section; the definition of the dictionaries comes one after the other, with no bracketing <section> binding each dictionary IDL to its list of attributes.

This concerns RTCRtpParameters, RTCRtpEncodingParameters, RTCDegradationPreference, RTCRtcpParameters, RTCRtpHeaderExtensionParameters, RTCRtpCodecParameters, RTCRtpCapabilities, RTCRtpCodecCapability, RTCRtpHEaderExtensionCapability. These actually live outside the RTCRtpSender section altogether, kind of in limbo.

They would be better off as subsections under RTCRtpSender.

alvestrand:
kudos to @orphis for noting the issue.

PR #1750, opened 2018-01-30
Need <section> around dictionaries under RTCRtpSender and RTCIceTransport

aboba:
Fix for Issues https://github.com/w3c/webrtc-pc/issues/1749 and https://github.com/w3c/webrtc-pc/issues/1751

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6458,8 +6458,7 @@ async function updateParameters() {
              DOMString uri;
 };</pre>
         <section>
-          <h2>Dictionary <code><a>RTCRtpHeaderExtensionCapability</a></code></dfn>
-          Members</h2>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5946,7 +5948,7 @@ <h2>Methods</h2>
              RTCDegradationPreference                  degradationPreference;
 };</pre>
         <section>
-          <h2>Dictionary <dfn>RTCRtpParameters</dfn> Members</h2>
+          <h2>Dictionary <code><a>RTCRtpParameters</a></code> Members</h2>
```
alvestrand: I'm not sure it matters, either in TOC-gen or in styling, but is it really proper to use a h2 inside a h3, and not a h4?

aboba: h2 is used consistently throughout the spec for these kind of sections.  And yes, it does seem odd.

PR #1753, opened 2018-01-31
Add "statsended" event and surrounding machinery

alvestrand:
Fixes https://github.com/w3c/webrtc-stats/issues/235

This adds an event that is emitted every time a stats object lifetime ends.
A companion PR to the stats spec will remove the "objectDeleted" flag.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/1753.html" title="Last updated on Feb 23, 2018, 8:48 AM GMT (0df866d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1753/35248b7...0df866d.html" title="Last updated on Feb 23, 2018, 8:48 AM GMT (0df866d)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9557,7 +9557,16 @@ interface RTCDTMFToneChangeEvent : Event {
       queries can be linked by the <code><a>RTCStats</a></code> <a data-link-for=
       "RTCStats">id</a> dictionary member. Thus, a Web application can make
       measurements over a given time period by requesting measurements at the
-      beginning and end of that period.</p>
+        beginning and end of that period.</p>
+      <p>
+        Stats objects may have a limited lifetime. Until the end of
+        their lifetime, they are always present in the result from
+        getStats(). When their lifetime ends, a record of the
+        statistics for that object is emitted through a
+        "statslifetimeended" event, containing an RTCStats
+        dictionary. The object descriptions in [[WEBRTC-STATS]]
+        describe the lifetime of each stats object type.
+      </p>
     </section>
     <section>
       <h3>RTCPeerConnection Interface Extensions</h3>
@@ -9566,7 +9575,15 @@ interface RTCDTMFToneChangeEvent : Event {
       <div>
         <pre class="idl">partial interface RTCPeerConnection {
     Promise&lt;RTCStatsReport&gt; getStats (optional MediaStreamTrack? selector = null);
-};</pre>
+    attribute EventHandler onstatslifetimeended;
+          };</pre>
+        <section>
+          <h2>Attributes</h2>
+          <dl data-link-for="RTCPeerConnection"
+              data-dfn-for="RTCPeerConnection" class="attributes">
+            <dt><dfn><code>onstatslifetimeended</code></dfn> of type
+              <span class="idlAttrType"><a>EventHandler</a></span></dt>
+            <dd>The event type of this event handler is <code><a>statslifetimeended</a></code>.</dd>
         <section>
           <h2>Methods</h2>
           <dl data-link-for="RTCPeerConnection" data-dfn-for=
@@ -9732,6 +9749,58 @@ interface RTCDTMFToneChangeEvent : Event {
       <p>The set of valid values for <code><a>RTCStatsType</a></code>, and the dictionaries derived
       from RTCStats that they indicate, are documented in
       [[!WEBRTC-STATS]].</p>
+    </section>
+    <section>
+      <h3><dfn>RTCStatsLifetimeEvent</dfn></h3>
+      <p>The <code><a>statslifetimeended</a></code> event uses
+        the <code><a>RTCStatsLifetimeEvent</a></code>.</p>
+      <div>
+        <pre class="idl">
+          [ Constructor (DOMString type, RTCDataChannelEventInit
+          eventInitDict), Exposed=Window]
+          interface RTCStatsLifetimeEvent : Event {
+            readonly attribute any stat; // Must be RTCStats subtype
+          };</pre>
+        <section>
+          <h2>Constructors</h2>
+          <dl data-link-for="RTCStatsLifetimeEvent"
+              data-dfn-for="RTCStatsLifetimeEvent" class="constructors">
+            <dt><code>RTCStatsLifetimeEvent</code></dt>
+            <dd>
+            </dd>
+          </dl>
+        </section>
+        <section>
+          <h2>Attributes</h2>
+          <dl data-link-for="RTCStatsLifetimeEvent"
+              data-dfn-for="RTCStatsLifetimeEvent" class="attributes">
+            <dt><code>stat</code> of
+              type <span class="idlAttrType">any</span>
+            </dt>
+            <dd>
+              <p>The <dfn id=dom-statslifetimeevent-stat><code>stat</code></dfn>
+              attribute represents the object of the appropriate
+              subclass of <code><a>RTCStats</a></code>
+              object giving the value of the statistics for the object
+                whose lifetime has ended, at the time that it ended.</p>
+            </dd>
+          </dl>
+        <section>
+          <h2>Dictionary <dfn>RTCStatsLifetimeEventInit</dfn>
+            members</h2> 
+          <dl  data-link-for="RTCStatsLifetimeEventInit"
+            data-dfn-for="RTCStatsLifetimeEventInit"
+               class="dictionary-members">
+            <dt><dfn><code>stat</code></dfn> of
+            type <span class="idlMemberType"><a>RTCStats</a></span>,
+              required</dt>
+            <dd>
+              <p>The <code><a>RTCStats</a></code> object giving the
+                stats for the object whose lifetime has ended.</p>
+            </dd>
+          </dl>
+          </section>
+       </div>
     </section>
     <section>
       <h3>The stats selection algorithm</h3>
@@ -12165,6 +12234,11 @@ interface RTCErrorEvent : Event {
           the <var>isolated</var> attribute on a <code>MediaStreamTrack</code>
           changes.</td>
         </tr>
+        <tr>
+          <td><dfn id="event-statslifetimeended"><code>statslifetimeended</code></dfn></td>
+          <td><code><a>RTCStatsLifetimeEvent</a></code></td>
+          <td>A new <code><a>RTCStatsLifetimeEvent</a></code> is dispatched to
+          the script in response to the end of a stats object's lifetime.</td>
       </tbody>
     </table>
     <p>The following events fire on <code><a>RTCDTMFSender</a></code>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -9671,6 +9691,59 @@ <h2>Dictionary <a class="idlType">RTCStats</a> Members</h2>
...
+    <section>
+      <h3><dfn>RTCStatsLifetimeEvent</dfn></h3>
+      <p>The <code><a>statslifetimeended</a></code> event uses
+        the <code><a>RTCStatsLifetimeEvent</a></code>.</p>
+      <div>
+        <pre class="idl">
+          [ Constructor (DOMString type, RTCStatsLifetimeEventInit
+          eventInitDict), Exposed=Window]
+          interface RTCStatsLifetimeEvent : Event {
+            readonly attribute any stat; // Must be RTCStats subtype
```
henbos: Should we change this to sequence&lt;any&gt;?

For example if there is both a sender stat and RTP stream stat that are simultaneously removed by stopping a transceiver (maybe a bad example, if transceivers are never removed then I guess neither are these objects). Or any other example where two objects referencing each other are both removed due to the same event.

I'm thinking it is a problem if a stats object references an object that neither exists amongst "deleted stats" (because that event has not fired yet) or "current stats" (because it has already been removed and event queued). I will refer to a stats object that has been deleted but its event has not been processed yet as a "limbo stats object", stuck between being added to the application's "deleted stats" list and the PC's "current stats".

On the other hand it might not be a problem. Because if you want to do lookup in both "deleted stats" and "current stats" you have to call getStats() to get the "current stats", and by the time you have the "current stats" the limbo stats object's event will already have fired and the object added to "deleted stats".

```
deletedStats = new Set();
pc.onstatslifetimeended = e => {
  deletedStats.add(e.stat);
  // Trying to process e.stat by looking up references to other stats objects in deletedStats
  // would be a problem for "limbo stats".

  pc.getStats().then(r => {
    currentStats = r;
    // Process e.stat, it may reference things in both deletedStats and currentStats.
    // By the time we reach this line, deletedStats will contain any "limbo stats".
  });
}
```

Maybe we don't need to change it to a sequence? On the other hand you might want to process it without forcing a getStats() call. You can of course always do setTimeout() and you're safe, but it's not obvious to developers and it might be considered a hack?

This should be discussed before merge. If we decide to not do a sequence we need to write a comment about this in the description.

henbos: I don't like having to defensively getStats() because "what if it references an object still in there", that means if 10 objects get deleted we may do 10 getStats(). setTimeout() is also hacky, because there is no way to tell "was this the last event?" you would do things multiple times here as well.

You could try to deduce if multiple events belong together by looking at their timestamp, but multiple objects deleted at the same time might not have the same timestamp. Some objects have timestamp for the last SR or RR received, not for when the object was deleted.

I think we need a `sequence<any>` to inform the application of all things removed associated with the same removal event.

alvestrand:
WebIDL insists that attributes can't return sequences. What's the better option?

henbos:
Can you do this?
```
[SameObject]
readonly attribute Array<any> stats;
```

webrtc.html
jan-ivar: `any` is too broad. We should at least use `object` since it is guaranteed to be an object.

I agree with @henbos we should return all simultaneously deleted stats at once rather than fire multiple events. But why add another way to enumerate stats?

Why don't we simply return an [RTCStatsReport](http://w3c.github.io/webrtc-pc/#dom-rtcstatsreport)? That'd be consistent and it'd work as an attribute.

webrtc.html
```diff
@@ -9671,6 +9691,59 @@ <h2>Dictionary <a class="idlType">RTCStats</a> Members</h2>
...
+          };</pre>
+        <section>
+          <h2>Constructors</h2>
+          <dl data-link-for="RTCStatsLifetimeEvent"
```
jan-ivar: Shouldn't it be `RTCStatsLifetimeEndedEvent` ?

webrtc.html
```diff
@@ -9495,7 +9495,16 @@ <h3>Introduction</h3>
...
       "RTCStats">id</a> dictionary member. Thus, a Web application can make
       measurements over a given time period by requesting measurements at the
-      beginning and end of that period.</p>
+        beginning and end of that period.</p>
+      <p>
+        Stats objects may have a limited lifetime. Until the end of
+        their lifetime, they are always present in the result from
+        getStats(). When their lifetime ends, a record of the
+        statistics for that object is emitted through a
+        "statslifetimeended" event, containing an RTCStats
```
jan-ivar: I'm fond of algorithms. E.g. *"When their lifetime ends, the UA MUST queue a task to run the following steps:"*

webrtc.html
```diff
@@ -9495,7 +9495,16 @@ <h3>Introduction</h3>
...
+        dictionary. The object descriptions in [[WEBRTC-STATS]]
+        describe the lifetime of each stats object type.
```
jan-ivar: Should we add some guidance here? I know we discussed quite conservative definitions of lifetime termination that left out all terminal cases like `transceiver.stop()` and `pc.close()`. The purpose of these events seem limited to distinguishing old data from new data in subsequent getStats() calls, e.g. replaced tracks, and restarted ice candidates.

alvestrand: Yes, there should be guidance.

webrtc.html
alvestrand: Makes sense.

webrtc.html
alvestrand: RTCStatsReport for the win. Then the need for "any" disappears into what we've already done.
Have to make clear in the text that it's *only* the newly-dead data that appears here.

webrtc.html
alvestrand: I think that belongs in the "stats object design guidelines" in the stats document.

webrtc.html
alvestrand: Added the algorithm under the "onstatslifetimeended" handler. Looked to me like a more natural place than in an introduction section.
Have to make sure I'm phrasing it right, though.

webrtc.html
alvestrand: Yeah :-( long names!

vr000m:
@lennart-csio @karthikbr82 PTAL

webrtc.html
```diff
@@ -9504,7 +9513,39 @@ <h3>RTCPeerConnection Interface Extensions</h3>
...
+                <ul>
+                  <li>Create a new RTCStatsReport object called <var>r</var></li>
+                  <li>For each monitored object, create a new RTCStats
+                    object with the statatistics for that object, and
+                    add it to <var>r</var>
+                  </li>
+                  <li>
+                    Create a new <a>RTCStatsLifetimeEndedEvent</a>
+                    called <var>e</var>, with its name set to
+                    "statsended" and its "report" being set to <var>r</var>
```
taylor-b: Is the event named "statsended" or "statslifetimeended"?

alvestrand: What do you think it should be named? @jan-ivar suggested "statsended", but it seems that I didn't change it everywhere.

taylor-b: Both sound fine to me, I was just noting the inconsistency

alvestrand:
I'm naming the event "statsended" and naming the event IDL type "RTCStatsEvent". This leaves the naming consistent if we later decide to emit events containing stats that don't happen when stats end.

alvestrand:
Changes pushed. Please take another look.

vr000m:
I suppose the lifetime of the objects and the corresponding stats will be added to webrtc-stats and not here.

alvestrand:
@jan-ivar you're the one with comments in "changes requested" state.

webrtc.html
```diff
@@ -9566,7 +9578,39 @@ <h3>RTCPeerConnection Interface Extensions</h3>
...
+            <dt><dfn><code>onstatsended</code></dfn> of type
+              <span class="idlAttrType"><a>EventHandler</a></span></dt>
+            <dd>
+              <p>The event type of this event handler
+                is <code><a>statsended</a></code>.
+              </p>
+              <p>When the lifetime of one or more <a>monitored object</a>s ends, the UA MUST
+                queue a task to run the following steps:
+                <ul>
+                  <li>Create a new RTCStatsReport object called <var>r</var></li>
```
henbos: nit: Usually in the spec variable names are full words, with the exception of "e" for "event". Prefer "report" to "r".

webrtc.html
```diff
@@ -9566,7 +9578,39 @@ <h3>RTCPeerConnection Interface Extensions</h3>
...
+          <dl data-link-for="RTCPeerConnection"
+              data-dfn-for="RTCPeerConnection" class="attributes">
+            <dt><dfn><code>onstatsended</code></dfn> of type
+              <span class="idlAttrType"><a>EventHandler</a></span></dt>
+            <dd>
+              <p>The event type of this event handler
+                is <code><a>statsended</a></code>.
+              </p>
+              <p>When the lifetime of one or more <a>monitored object</a>s ends, the UA MUST
+                queue a task to run the following steps:
```
henbos: Can you clarify that "if the same event causes multiple monitored objects' lifetime to end, they MUST be added to the same report." Alternatively, "any monitored objects' ending in the same event loop execution cycle must be fired in the same event" but that phrasing would require different algorithmic steps.

I think there is room for interpretation here, but a UA that implemented queuing one such task per ended object would hit the issues previously discussed.

webrtc.html
```diff
@@ -9665,8 +9709,8 @@ <h3><dfn>RTCStatsReport</dfn> Object</h3>
     </section>
     <section>
       <h3><dfn>RTCStats</dfn> Dictionary</h3>
-      <p>An <code>RTCStats</code> dictionary represents the stats
-      gathered by inspecting a specific object relevant to a <a>selector</a>.
+      <p>An <code>RTCStats</code> dictionary represents the <a>stats object</a>
+      constructed by inspecting a specific <a>monitored object</a> relevant to a <a>selector</a>.
```
henbos: "a specific monitored object relevant to a selector"
nit: There are multiple monitored objects relevant to a selector (multiple RTCStats), change to plural.

webrtc.html
vr000m: +1 for the clarification. Additionally,  this should be covered in the webrtc-stats spec in the lifetime of the objects.

alvestrand:
Harald will address nits and merge.

webrtc.html
alvestrand: actually, removing the "relevant to a selector". The stats object is constructed based on a single monitored object; the selector is irrelevant to that part of the operation.

webrtc.html
alvestrand: Done.

webrtc.html
alvestrand: Changed to defining a "delete stats" operation, which can be linked to from other places in the spec.

Issue #1756, opened 2018-02-01
need to clear [[AssociatedMediaStreams]] in removeTrack?


fippo:
[addTrack](http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtrack) adds to the [[AssociatedMediaStreams]] but [removeTrack](http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-removetrack) does not set them to an empty list.

Should it? This came up in [in firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1427162) but even ignoring getLocalStreams() legacy one doesn't want to keep a reference to the streams anymore.

cc @henbos too

jan-ivar:
@fippo It shouldn't, because the stream association is still tied to the transceiver. E.g. If I then do
```js
await sender.replaceTrack(trackB);
sender.direction = "sendrecv";
```
Then remotely (after renegotiation) the receiver.track would again pop up in the same-id remote stream.

Confused? See https://github.com/w3c/webrtc-pc/issues/1758 which I just filed.

What I think is true however, is that we probably don't need to keep a reference to the streams themselves, just their ids. Since keeping a stream open has side-effects (local camera staying on) that seems like an improvement (remember what's being sent may come from an entirely different stream).

alvestrand:
Why should the stream association be tied to the (local) transceiver?

The fact that if you're using replaceTrack, you're having odd effects on the stream associations is the biggest reason why using replaceTrack() together with streams is going to be a headache. The WG accepted that in the name of making replaceTrack() possible without a negotiation cycle.

If you let removeTrack() remove the track and clear the streams list, you let addTrack/removeTrack-based applications continue to behave consistently, and make it easy to shim addStream/removeStream on top of them - in fact perfectly in the case of "one track is added only as part of one stream".

removeTrack() triggers a negotiation cycle.

jan-ivar:
> Why should the stream association be tied to the (local) transceiver?

Because you added it with `tc = addTransceiver(track, {streams})` and you haven't `tc.stop()`ed it?

fippo:
if you call removeTrack you can call replaceTrack on that sender? Surprising but I can see how your code would work.

But how does this work then:
```
When the other peer stops sending a track in this manner, the track is removed from any remote MediaStreams that were initially revealed in the track event
```
I assume that the stream ids come back when you set the sender direction to active?

jan-ivar:
[...]

PR #1759, opened 2018-02-01
Change [[AssociatedMediaStreams]] to [[AssociatedMediaStreamIds]]

jan-ivar:
Avoid keeping local streams alive unnecessarily. Fix for https://github.com/w3c/webrtc-pc/issues/1756.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/1759.html" title="Last updated on Feb 1, 2018, 9:36 PM GMT (e01c833)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1759/d682ab2...jan-ivar:e01c833.html" title="Last updated on Feb 1, 2018, 9:36 PM GMT (e01c833)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4940,8 +4940,16 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                       <var>track</var>.</p>
                     </li>
                     <li>
-                      <p>Set <var>sender</var>'s <a>[[\AssociatedMediaStreams]]</a>
-                      to <var>streams</var>.</p>
+                      <p>Set <var>sender</var>'s
+                      <a>[[\AssociatedMediaStreamIds]]</a> to an empty set.
+                      </p>
+                    </li>
+                    <li>
+                      <p>For each <var>stream</var> in <var>streams</var>, add
+                      <var>stream.id</var> to
+                      <a>[[\AssociatedMediaStreamIds]]</a> if it's not already
+                      there.
+                      </p>
                     </li>
                     <li>
                       <p>Let <var>transceiver</var> be the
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4940,8 +4940,16 @@ <h2>Methods</h2>
...
-                      to <var>streams</var>.</p>
+                      <p>Set <var>sender</var>'s
+                      <a>[[\AssociatedMediaStreamIds]]</a> to an empty set.
+                      </p>
+                    </li>
+                    <li>
+                      <p>For each <var>stream</var> in <var>streams</var>, add
+                      <var>stream.id</var> to
+                      <a>[[\AssociatedMediaStreamIds]]</a> if it's not already
+                      there.
```
taylor-b: Since `[[AssociatedMediaStreamIds]]` was just set to an empty set, won't the IDs *always* be "not already there"? Unless there's a duplicate stream I guess.

alvestrand:
I am not happy with this, because the mapping from ID to MediaStream is not unique.
(The collision cases are corner cases)

webrtc.html
jan-ivar: Yes it's to handle duplicate ids. E.g.
```js
pc.addTrack(track, stream, stream);
```
The old prose didn't remove duplicates, which it probably should?

henbos:
> Let sender have an [[AssociatedMediaStreamIds]] internal slot, representing a list of Ids of MediaStream objects that this sender is to be associated with.

Conceptually we have "`MediaStream` objects that this sender is associated with" and `MediaStream` IDs are _not unique_, so I think this relationship should be explicit rather than implicit, even if SDP only signals the IDs and implementation-wise IDs work.

I think this for three reasons, including symmetry between senders and receivers and usefulness of exposing streams.

1. If we were to do the same thing for receivers, i.e. have an `[[AssociatedRemoteMediaStreamIds]]` slot, because IDs are not unique and tracks are added to/removed from stream objects, implementing this would either be buggy or implicitly have an `[[AssociatedRemoteMediaStreams]]` slot, in which case `[[AssociatedRemoteMediaStreamIds]]` would add no value.

I would prefer to maintain the symmetry between "senders and streams" and "receivers and streams"; I think changing one of the cases to Ids just adds confusion. If a UA really want to get rid of references (which I question too) that can be an implementation detail, not advised by the spec.

2. Since legacy `getLocalStreams()` has to be implemented as "all streams of all active senders", and there is great usage of `getLocalStreams()` and `getRemoteStreams()`, this shows that there is an application need for the stream references. You could force the application to keep track of these references instead, but this seems unnecessary, I would rather expose a `readonly  sequence<MediaStream> attribute streams` to both senders and receivers. This would make it possible to shim `getLocalStreams()` and `getRemoteStreams()`, which would take us one step closer to deprecating these APIs in the UA implementation.

Even if we don't do 2. I don't think this change adds value in and of itself, just confusion. I think it was motivated by [this](https://github.com/w3c/webrtc-pc/issues/1756#issuecomment-362353997):

> What I think is true however, is that we probably don't need to keep a reference to the streams themselves, just their ids. Since keeping a stream open has side-effects (local camera staying on) that seems like an improvement (remember what's being sent may come from an entirely different stream).

Which brings me to my third point.

3. If a stream is associated with a sender it is in use by the PC. The stream should not be garbage collected, the camera should not be closed. With this change, what happens if the application does not have a reference to the stream after attaching it to the PC? Camera stops and stream is GC'd? This could happen midst conference.

I disapprove of this change. I don't like implicit relationships.

henbos:
My 3rd point is mute I realize, the only camera in use is kept open by the track reference, and the stream references could be totally unrelated, only their IDs matter, so not referencing them makes sense.

It mainly boils down to the relationships and application needs for these relationships to be exposed or not.

I'm on the fence here between having the stream relationships explicit and abandoning streams altogether, but addTrack() and friends takes stream _objects_, not stream IDs... WebRTC 2.0 - nuke streams?

PR #1805, opened 2018-03-16
Split out RTCSendRtpParameters and mark parameters as required.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1493.


webrtc.html
```diff
@@ -6070,18 +6064,37 @@ <h2>Dictionary <code><a>RTCRtpParameters</a></code> Members</h2>
               "video/rtx", and an <code>sdpFmtpLine</code> attribute (providing
               the "apt" and "rtx-time" parameters). <a>Read-only parameter</a>.</p>
             </dd>
+          </dl>
+        </section>
+      </div>
+      </section>
+     <section id="rtcrtpparameters">
```
alvestrand: wrong section name?

aboba:
@jan-ivar Why are header extension parameters required?  What if no header extensions are in use?

aboba:
@jan-ivar The PRs are layered on each other, so I guess they have to be examined in totality.  But it is a bit confusing to see RTCRtpEncodingParameters in RTCRtpParameters when it seems like it belongs in RTCRtpSendParameters.  This gets fixed in another PR, right?

jan-ivar:
@aboba Right, though note the name remains `receiver.getParameters().encodings` for symmetry. 

The split of `RTCRtpEncodingParameters` into that and `RTCRtpDecodingParameters` is https://github.com/w3c/webrtc-pc/pull/1807, so we can consider it separately.

stefhak:
LGTM.

jan-ivar:
Rebased.

jan-ivar:
@taylor-b See https://github.com/w3c/webrtc-pc/issues/1493#issuecomment-373808387. Normatively, `required` does make a difference here because it affects how `setParameters` responds to modifications of what you got from `getParameters`.

taylor-b:
Meaning that you'd get a `TypeError` instead of `InvalidModificationError` if you remove something?

taylor-b:
I assumed "no value" was different than "some value" (`null` != "foo") and so removing a read-only parameter would result in `InvalidModificationError`. Though the way it's worded ("*has a value* that is different...") may not make that clear enough. Is that the hole you're referring to? If so, doesn't it still exist, since there are still things (like `RTCRtpCodecParameters` fields) that aren't `required`?

jan-ivar:
I think the problem is *"Any parameter in parameters"*, which implies [presence](https://github.com/heycam/webidl/issues/381).

If I do `setParameters({})` how many parameters are in *parameters*? It's vague at best. `null` is a value.

Issue #1806, opened 2018-03-16
Inconsistency over active member in receiver's "encoding" parameters.


jan-ivar:
While working on https://github.com/w3c/webrtc-pc/issues/1493, I found this inconsistency over whether
```js
pc.getReceivers()[0].getParameters().encodings[0].active // is true/false or undefined
```
Receiver's [getParameters()](https://w3c.github.io/webrtc-pc/#dom-rtcrtpreceiver-getparameters) says:

*"Every member of the RTCRtpEncodingParameters dictionaries other than the RID fields is left undefined."*

whereas [active](https://w3c.github.io/webrtc-pc/#dom-rtcrtpencodingparameters-active) says:

*"For an RTCRtpReceiver, a value of true indicates that this encoding is being decoded. A value of false indicates this encoding is no longer being decoded."*

The latter seems on purpose.

aboba:
@jan-ivar I think the latter text may have been left over from a time before we clarified how direction worked. Since there is no RTCRtpReceiver.setParameters, it is not possible to set encodings[].active and the specification doesn't say much if anything about when encodings[].active would toggle. So I'm not clear that active is of much use for an RTCRtpReceiver in WebRTC 1.0 (though in ORTC it does have the meaning indicated by the latter sentence).

PR #1807, opened 2018-03-16
Split out RTCRtpDecodingParameters and fix active.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1806 (sits on-top of https://github.com/w3c/webrtc-pc/pull/1805).

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/1807.html" title="Last updated on Apr 12, 2018, 2:24 AM GMT (5fd8052)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1807/1b42a97...jan-ivar:5fd8052.html" title="Last updated on Apr 12, 2018, 2:24 AM GMT (5fd8052)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5579,8 +5579,8 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     readonly        attribute RTCDtlsTransport?  transport;
     readonly        attribute RTCDtlsTransport? rtcpTransport;
     static RTCRtpCapabilities getCapabilities (DOMString kind);
-    Promise&lt;void&gt;             setParameters (optional RTCSendRtpParameters parameters);
-    RTCSendRtpParameters          getParameters ();
+    Promise&lt;void&gt;             setParameters (optional RTCRtpSendParameters parameters);
+    RTCRtpSendParameters          getParameters ();
     Promise&lt;void&gt;             replaceTrack (MediaStreamTrack? withTrack);
     Promise&lt;RTCStatsReport&gt;   getStats();
 };</pre>
@@ -5772,7 +5772,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               <p><code>setParameters</code> does not cause SDP renegotiation
               and can only be used to change what the media stack is sending or
               receiving within the envelope negotiated by Offer/Answer. The
-              attributes in the <code><a>RTCSendRtpParameters</a></code> dictionary
+              attributes in the <code><a>RTCRtpSendParameters</a></code> dictionary
               are designed to not enable this, so attributes like
               <code>cname</code> that cannot be changed are read-only. Other
               things, like bitrate, are controlled using limits such as
@@ -5790,11 +5790,11 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               to a remote <code><a>RTCRtpReceiver</a></code>.</p>
 
               <p>When <code>getParameters</code> is called, the
-              <code><a>RTCSendRtpParameters</a></code> dictionary is
+              <code><a>RTCRtpSendParameters</a></code> dictionary is
               constructed as follows:</p>
               <ul>
                 <li>
-                  <code><a data-link-for="RTCSendRtpParameters">transactionId</a></code>
+                  <code><a data-link-for="RTCRtpSendParameters">transactionId</a></code>
                   is set to a new unique identifier, used to match this
                   <code>getParameters</code> call to a
                   <code>setParameters</code> call that may occur later.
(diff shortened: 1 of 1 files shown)


jan-ivar:
Only the last commit is this PR (the other three belong to https://github.com/w3c/webrtc-pc/pull/1805).

aboba:
@jan-ivar I don't think that header extensions or encodings should necessarily be required for RtpReceivers.

webrtc.html
```diff
@@ -6070,18 +6056,104 @@ <h2>Dictionary <code><a>RTCRtpParameters</a></code> Members</h2>
...
+            </dd>
+          </dl>
+        </section>
+      </div>
+      </section>
+      <section id="rtcrtpdecodingparameters">
+      <h3><dfn>RTCRtpDecodingParameters</dfn> Dictionary</h3>
+      <div>
+        <pre class="idl">dictionary RTCRtpDecodingParameters {
+             boolean             active = true;
```
taylor-b: We don't need an `active` field for the receiver since it has no `setParameters`. Note the definition from ORTC:

> For an RTCRtpReceiver, indicates that this encoding is being decoded. Setting it to false causes this encoding to no longer be decoded. Setting it to true causes this encoding to be decoded.

So, this is meant to be something you can set to false to turn off decoding (saving resources I guess). Without the ability to set it to false it has no use.

jan-ivar:
> I don't think that header extensions or encodings should necessarily be required for RtpReceivers.

@aboba With no semantic difference between absence and empty list, requiring empty lists simplifies reading. Same argument as with `sender.getParameters()` (presumably symmetry was a design factor here).

With different types for sender vs receiver, there's no longer any need to clarify what applies where, and the table is redundant.

webrtc.html
jan-ivar: OK so there's no other way for it to be inactive? I'll remove it.

jan-ivar: @taylor-b When a sender sets `active = false`, what happens on the receiver's end?

taylor-b: By "the receiver", do you mean an SFU receiving the simulcasted streams, or an `RTCRtpReceiver` on the other end receiving a continuous stream from the SFU?

If only one encoding is deactivated, I'd expect the SFU to switch to forwarding another encoding, and the downstream endpoints may notice a decrease in quality. If all encodings are deactivated, I guess eventually an RTP timeout would occur and the track `muted` state would be set, as described [here](https://www.w3.org/TR/webrtc/#mediastreamtrack).

jan-ivar: OK removing it.

aboba:
@jan-Ivar Agree that the sender/receiver entry in the table is redundant, but don’t we still need the read/write entry?

jan-ivar:
@aboba The only read-only member in that table was [rid](https://rawgit.com/jan-ivar/webrtc-pc/5d4ecd0837e08b647d2d66e87b442346a36818e1/webrtc.html#dom-rtcrtpdecodingparameters-rid), which I've marked as a [Read-only parameter](http://w3c.github.io/webrtc-pc/#dfn-read-only-parameter), which seems to be the convention we're using.

aboba:
@jan-ivar Does inheritance make sense here?  I realize that there is not much in RTCRtpDecodingParameters, but in future there is likely to be more attributes there, so it might make sense to have both encoding/decoding parameters inherit from a common dictionary.

jan-ivar:
@aboba We can arguably fix that in the future, since there's very little observable from messing with dictionary inheritance (though member order may change: inheritance first, alphabetical second). That said, a common base sounds cleaner today, so I'll add that.

aboba:
@jan-ivar Should dtx have a default value ("enabled")?

jan-ivar:
@aboba We can't give `dtx` a default value in WebIDL without it then showing up in `getParameters()` for video encodings, which seems wrong. We could if we split out a separate dictionary for video, but I think that would be too much.

jan-ivar:
Rebased.

aboba:
@taylor-b I am not clear that we should remove `RTCRtpReceiveParameters.encodings` altogether. For example, if a conference server is sending multiple video streams to a browser, all using the same payload type, how does the receiver de-multiplex them?  In 'Unified Plan' will the SSRCs of the streams always be signaled in the remote description?

taylor-b:
@aboba I'm sorry, I don't follow. The browser will demux based on information in the remote description (whether it be MIDs or SSRCs). `RTCRtpReceiveParameters` just provides information to the web application. But if `RTCRtpDecodingParameters` only has `rid`, which will never be present, it's not actually providing any information.

aboba:
Perhaps the question is whether any browsers are planning to support simulcast reception. The current APIs don't provide a way to set up reception of simulcast, but if I recall discussion correctly, we didn't prohibit reception of simulcast. 

With respect to SSRCs, my impression was that they wouldn't be present in Unified Plan SDP.

PR #1810, opened 2018-03-20
Update structured cloning for recent changes to HTML

### LINKED ISSUE #1089: Update for structured cloning changes in HTML

stefhak:
Fixes #1089 (eventually, perhaps).

Do not merge - not ready (respec errors among other things).

@domenic PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1810.html" title="Last updated on Apr 5, 2018, 11:34 AM GMT (e1ac9e5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1810/1cc5bfc...stefhak:e1ac9e5.html" title="Last updated on Apr 5, 2018, 11:34 AM GMT (e1ac9e5)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4584,7 +4584,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
         certificate used to authenticate WebRTC communications. In addition to
         the visible properties, internal slots contain a handle to the
         generated private keying materal (<dfn>[[\KeyingMaterial]]</dfn>), a certificate
-        (<dfn>[[\Certificate]]</dfn>]]) that <code>RTCPeerConnection</code>
+        (<dfn>[[\Certificate]]</dfn>) that <code>RTCPeerConnection</code>
         uses to authenticate with a peer, and the origin (<dfn>[[\Origin]]</dfn>)
         that created the object.</p>
         <div>
--- a/webrtc.html
+++ b/webrtc.html
@@ -86,6 +86,13 @@
     <p><code><dfn data-cite="!HIGHRES-TIME#dom-performance-timeOrigin">performance.timeOrigin</dfn></code>
     and <code><dfn data-cite="!HIGHRES-TIME#dom-performance-now">performance.now()</dfn></code>
     are defined in [[!HIGHRES-TIME]].</p>
+    <p>The terms <dfn><a href=
+    "https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects">
+    serializable objects</a></dfn>, <dfn><a href=
+    "https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps">
+    serialization steps</a></dfn>, and <dfn><a href=
+    "https://html.spec.whatwg.org/multipage/infrastructure.html#deserialization-steps">
+    deserialization steps</a></dfn> are defined in [[!HTML]].</p>
     <p>The terms <dfn>MediaStream</dfn>, <dfn>MediaStreamTrack</dfn>, and
     <dfn>MediaStreamConstraints</dfn> are defined in [[!GETUSERMEDIA]].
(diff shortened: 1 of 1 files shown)


domenic:
This seems unlikely to be correct, as it references a specification that has known bugs in the area of serialization and cloning.

stefhak:
Aha, I assumed the spec was decent given it is a Recommendation.

domenic:
No. The root problem is that the editors of that specification seem to copy our work (in the HTML Standard) blindly and incompletely, which leads to lots of bugs.

fluffy:
Could you say more about what the bugs are so we can decide if we can use this in WebRTC.

domenic:
I'm sorry, I'm not in the business of helping specs add buggy dependencies that browsers cannot feasibly ship. I would suggest not using this in Web RTC, and instead using my original pull request, which depends on the actual maintained definition of structured cloning that is implemented in browsers.

stefhak:
Updated, now identical to @domenic's original PR.

foolip:
As for why it matters, the HTML Living Standard is the one we consider "authoritative" in Blink, i.e. we try to match it or fix it when it's wrong. It's also what people working in web-platform-tests will need to look at, since it's the most upstream and well maintained.

stefhak:
@domenic can you take a look now?

webrtc.html
```diff
@@ -4688,29 +4695,35 @@ <h2>Methods</h2>
...
-          <li>Let the <a>[[\KeyingMaterial]]</a> internal slot of <var>output</var>
-          refer to the same private keying material represented by the
-          <a>[[\KeyingMaterial]]</a> internal slot of <var>input</var>.
-          </li>
+          <li>Set <var>serialized</var>.[[\expires]] to the value of
+          <var>value</var>'s <code>expires</code> attribute.</li>
+
+          <li>Set <var>serialized</var>.[[\certificate]] to
+          a copy of the unstructured binary data in
+          <var>value</var>.[[<a>certificate</a>]].</li>
```
jan-ivar: It's [[Certificate]] with upper-case 'C' right? Also, we probably want `<var>value</var>.[[\Certificate]]`. Otherwise I see squiggly red underlines in the spec.

webrtc.html
```diff
@@ -4688,29 +4695,35 @@ <h2>Methods</h2>
...
+
+          <li>Set <var>serialized</var>.[[\handle]] be a serialization of the
```
jan-ivar: s/handle/KeyingMaterial/ ?

stefhak:
I need to look into your comments @jan-ivar

webrtc.html
```diff
@@ -4641,29 +4648,35 @@ <h2>Methods</h2>
...
+          of the private keying material represented by
+          <var>value</var>'s <a>[[\KeyingMaterial]]</a> slot.</li>
+        </ol>
+
+        <p>Their <a>deserialization steps</a>, given <var>serialized</var> and
+        <var>value</var>, are:</p>
+
+        <ol>
+          <li>Initialize <var>value</var>'s <code>expires</code> attribute to
+          contain <var>serialized</var>.[[\Expires]].</li>
```
adam-be: nit: Should we link these slots as well?

stefhak: I don't think that is necessary - the definition is like 10 lines before. WDYT?

stefhak:
I think this one is ready. However we should probably hold off merging until we've issued a new CR (if we do that anytime soon).

Issue #1813, opened 2018-03-22
Simulcast makes no sense for audio.


jan-ivar:
[setParameters()](http://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-setparameters) (and JSPEP) appear kind-agnostic when it comes to allowing multiple encodings.

We should probably limit multiple encodings to video.

I see two options:
 1. TypeError on `sendEncodings.length > 0` in `addTransceiver` for audio, or
 2. TypeError on `sendEncodings.length > 1` in `addTransceiver` for audio.

The first option seems the simplest. The second may add some convenience but would let `rid` be set.

Also, we might want to ignore `active`, and remove `active` and `priority` from `getParameters()`.

jan-ivar:
We need to nail down what this API must support today so that we can test it. E.g. if someone calls:
```js
pc.addTransceiver("audio", {sendEncodings: [{rid: 'f'}, {rid: 'h', codecPayloadType: 1}]});
```
browser vendors need to know what to do. I hope we agree all browsers should do the same thing.

juberti:
I don't think simulcast is nonsensical for audio. Consider a high-bitrate and a low-bitrate encoding of the same audio source, wouldn't this make sense for the same reasons as video simulcast?

aboba:
Consensus from the April VI was to develop a PR supporting a "best efforts" approach:

a. Requesting more simulcast streams than are supported will not result in an error. 
b. sender.getParameters will return the number of simulcast streams that can be supported.
c. If simulcast is not supported, 0 streams is returned (rather than 1, so that RID is not sent).

alvestrand:
If we return 0 streams, we have nowhere to set the bandwidth, which seems like a wrong thing.
I think it is OK to say that if 1 stream is returned, simulcast is off and no RID is sent.

aboba:
@alvestrand. Yes, if no simulcast is supported, would be better to say 1 stream is returned, simulcast is off and no RID is sent.

PR #1814, opened 2018-03-22
Limit sendEncodings to video.

jan-ivar:
Potential fix for  https://github.com/w3c/webrtc-pc/issues/1813.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/1814.html" title="Last updated on Mar 22, 2018, 3:53 AM GMT (56b14bb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1814/1cc5bfc...jan-ivar:56b14bb.html" title="Last updated on Mar 22, 2018, 3:53 AM GMT (56b14bb)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5159,7 +5159,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               non-null value, as defined in <span data-jsep=
               "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>.</p>
               <p>The <code>sendEncodings</code> argument can be used to
-              specify the number of offered simulcast encodings, and
+              specify the number of offered simulcast encodings for video, and
               optionally their RIDs and encoding parameters.</p>
               <p>When this method is invoked, the user agent MUST run the
               following steps:</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5159,7 +5159,7 @@ <h2>Methods</h2>
               non-null value, as defined in <span data-jsep=
               "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>.</p>
               <p>The <code>sendEncodings</code> argument can be used to
-              specify the number of offered simulcast encodings, and
+              specify the number of offered simulcast encodings for video, and
```
fluffy: So outside of WebRTC, we o simulcast for audio too. Why would be limit it to video ?

jan-ivar: @fluffy Because of costs of implementing, testing and supporting any feature? Not to mention that features without strong use-cases driving them tend to end up broken anyway. I think the question must be flipped around: Why wouldn't simulcast be limited to video?

fluffy:
Sure, it's a trade off. Seems like a good thing for the WG to have consensus on not just get changed by the editors. Perhaps that happened and I just don't recall. There is also a big differences between requiring browsers to do this and forbidding them to do this. I would be against saying this was forbidden as it turns out to be really easy to do and often useful.

jan-ivar:
I was asked by @aboba to put up this PR. I also solicited opinions internally at Mozilla and the opinion was simulcast for audio made no sense. We should probably move discussion to https://github.com/w3c/webrtc-pc/issues/1813.

webrtc.html
fluffy: I'm fine with it being optional for browser to implement, but I don't think it should be forbidden. Many people in Africa are working across satellite links and they often can't use anything except very low bit rate audio while people on the thick pipes want high bandwidth. If you want to do this in a way where the media is encrypted, some clients need to send a high and low bandwidth version of the audio. Then the cloud can forward the appropriate bandwidth one to people that are listening.

jan-ivar:
I'm closing PR until discussion concludes in https://github.com/w3c/webrtc-pc/issues/1813.

PR #1815, opened 2018-03-22
Describe how ICE restarts affect the RTCIceTransportState.

### LINKED ISSUE #1630: Example ICE state transition appears incorrect
One of the transitions is "(any state, ICE restart occurs): new". However, this isn't what I'd expect, since during an ICE restart the existing candidate pair continues being used. So I'd actually expect that the transitions would be:

* connected -> no change
* completed -> connected
* disconnected -> checking?
* failed -> new -> checking

Does that sound right? I guess we may need to revisit the ICE state table such that its definitions take ICE restarts into account, if the behavior is not clear enough as-is.

taylor-b:
Fixes #1630.

This attempts to clarify that an ICE restart will cause these state
transitions:

failed -> checking
disconnected -> checking
completed -> connected

***
<a href="https://pr-preview.s3.amazonaws.com/taylor-b/webrtc-pc/pull/1815.html" title="Last updated on Mar 22, 2018, 5:40 AM GMT (adb7a66)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1815/1cc5bfc...taylor-b:adb7a66.html" title="Last updated on Mar 22, 2018, 5:40 AM GMT (adb7a66)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7996,6 +7996,12 @@ async function onOffHold() {
           </tbody>
         </table>
       </div>
+      <p>An ICE restart causes candidate gathering and connectity checks to
+      begin anew, causing a transition to <code>connected</code> if begun in the
+      <code>completed</code> state. If begun in the transient
+      <code>disconnected</code> state, it causes a transition to
+      <code>checking</code>, effectively forgetting that connectivity was
+      previously lost.</p>
       <p>The <code>failed</code> and <code>completed</code> states require an
       indication that there are no additional remote candidates. This can be
       indicated by calling <code><a href=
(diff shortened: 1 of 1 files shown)


taylor-b:
Something to point out is that is that this bends the definition of `disconnected` slightly, such that an ICE restart will cause a transition from `disconnected` to `checking`, effectively forgetting "there was a previous connection that was lost". It seems sensible to have a state transition to indicate that new connectivity checks have begun, and this is what Chrome has historically done.

If reviewers prefer, I can remove that caveat, meaning the state would stay `disconnected` until transitioning to `connected` or `failed`.

aboba:
IceTransport.state is only “new” at the time it is constructed?

taylor-b:
@aboba Hmm, maybe it would make sense if it was `new` in between `setLocalDescription(ice_restart_sdp)` and the actual start of connectivity checks? It's not truly "new" though; there's state carried over. And I don't know that the extra state transition would provide much value. So I'd be inclined to keep things simple and say "only `new` when first constructed". What do you think?

stefhak:
LGTM.

Issue #1734, opened 2018-01-11
RTCRtpContributingSource.audioLevel has different type and range than similar fields in webrtc-stats 


na-g:
The [audioLevel](https://w3c.github.io/webrtc-pc/#dom-rtcrtpcontributingsource-audiolevel) field of RTCRtpContributingSource is a byte with a range range of 0 to 127. The audioLevel field of [RTCRTPContributingSourceStats](https://w3c.github.io/webrtc-stats/#dom-rtcrtpcontributingsourcestats-audiolevel) and [RTCAudioStreamTrackStats](https://w3c.github.io/webrtc-stats/#dom-rtcaudiostreamtrackstats-audiolevel) are doubles with a range of 0.0 to 1.0. Is there compelling reason that RTCRtpContributingSource.audioLevel should differ in this regard? Should the same transform noted in the fields in the audioLevel fields webrtc-stats be applied to RTCRtpContributingSource.audioLevel so that they are directly comparable?

jan-ivar:
@alvestrand @taylor-b Thoughts?

Firefox 59 goes to beta in 10 days, so we're considering pref'ing off `getContributingSources` and `getSynchronizationSources` until this is resolved.

taylor-b:
I can't recall a good reason for making it a `byte`. I'd be fine with changing it; it's not implemented in Chrome yet.

dontcallmedom:
the reason for making it a `byte` is documented in the spec:
> Both RFCs define the level as an integral value from 0 to 127 representing the audio level in negative decibels relative to the loudest signal that the system could possibly encode. Thus, 0 represents the loudest signal the system could possibly encode, and 127 represents silence.

(I have no opinion on whether this is a good or bad reason - @fluffy @alvestrand and @aboba [seem to have been involved in the design discussion back then](https://lists.w3.org/Archives/Public/public-webrtc/2015Jul/0109.html)

foolip:
Hi all,

@rakuco and I are looking at [normative spec change without tests](https://docs.google.com/document/d/1kFYTqUMEbo9p87i6wAzORoicMWgHsuYwL2CgtvdhDaE/edit?usp=sharing) every week this quarter, and this week I took a look at https://github.com/w3c/webrtc-pc/pull/1736, since I'm a bit familiar with the tests that were written.

[...]

PR #1816, opened 2018-03-22
Require RTCRtpHeaderExtensionParameters members.

jan-ivar:
Another fix for https://github.com/w3c/webrtc-pc/issues/1493.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/1816.html" title="Last updated on Mar 28, 2018, 12:54 AM GMT (6dbebf2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1816/1cc5bfc...jan-ivar:6dbebf2.html" title="Last updated on Mar 28, 2018, 12:54 AM GMT (6dbebf2)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6373,9 +6373,9 @@ async function updateParameters() {
       <h3><dfn>RTCRtpHeaderExtensionParameters</dfn> Dictionary</h3>
       <div>
         <pre class="idl">dictionary RTCRtpHeaderExtensionParameters {
-             DOMString      uri;
-             unsigned short id;
-             boolean        encrypted;
+             required DOMString      uri;
+             required unsigned short id;
+             required boolean        encrypted;
 };</pre>
         <section>
           <h2>Dictionary <code><a>RTCRtpHeaderExtensionParameters</a></code>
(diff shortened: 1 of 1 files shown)


aboba:
@jan-ivar Should encrypted have a default value?  In ORTC it defaults to “false”.

foolip:
Hi all,

@rakuco and I are looking into normative spec changes that don't have tests this quarter, in this [tracking doc](https://docs.google.com/document/d/1kFYTqUMEbo9p87i6wAzORoicMWgHsuYwL2CgtvdhDaE/edit?usp=sharing). This is the last week of the this effort, and I ended up here this week.

After looking around for a bit it looks like there is an existing test which could fairly easily be modified to cover this change:
https://github.com/w3c/web-platform-tests/blob/172fe9d52634f70dd2810c4a859acd1aa6dc365a/webrtc/RTCRtpParameters-headerExtensions.html#L58

I think the new test would just have to omit the `encrypted` member and then use `getParameters()` to verify it behaves the same as when an explicit false was given. I see "The headerExtensions sequence is populated based on the header extensions that have been negotiated for sending" in the spec so maybe a loopback call needs to be set up to test this, I'm not sure.

foolip:
Oh, I overlooked that in addition to the `encrypted` default value, the import change was making two members required. That's very easy to test, just try to use a dictionary with each of them missing in turn and assert that TypeError is thrown.

foolip:
I've filed https://github.com/w3c/web-platform-tests/issues/10253.

@alvestrand @aboba @stefhak, it's a very small and non-random sample, but I also came across https://github.com/w3c/webrtc-pc/issues/1734#issuecomment-359023684 back in January. It seems like maybe https://github.com/w3c/webrtc-pc/blob/master/CONTRIBUTING.md#tests isn't quite working as intended, or at least not as I *hoped*. (It's deliberately worded in a pretty noncommital way, so hope seems right.)

Elsewhere, @plehegar and I have speculated about whether it might be useful to have a status check on PRs similar to https://github.com/w3c/ash-nazg, but which would *just* check if tests were mentioned/linked anywhere in a PR, and allow anyone with a GitHub account make a failing status pass by checking a "no tests required" checkbox. Do anyone think that might be useful, or more of a nuisance?

PR #1822, opened 2018-03-31
Data channel closing procedure

### LINKED ISSUE #1699: Data channel closing procedure
The [data channel closing procedure](https://rawgit.com/w3c/webrtc-pc/master/webrtc.html#data-transport-closing-procedure) lacks information how implementations should gracefully close data channels. It only mentions:

> An RTCDataChannel object's underlying data transport may be torn down in a non-abrupt manner by running the closing procedure. When that happens the user agent must, unless the procedure was initiated by the close method, queue a task that sets the object's [[ReadyState]] slot to closing. This will eventually render the data transport closed.

As far as I can see, that is the whole closing procedure description the `close` method references, too.

We need to add a few detailed steps and should probably reference [draft-ietf-rtcweb-data-cannel, section 6.7](https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13#section-6.7) which describes how a data channel can be closed gracefully.

Related bug in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1413198

---

I'm also confused by the following statement of the closing procedure:

> [...] This will eventually render the data transport closed.

I don't believe that's actually true. Why would that be the case? Only the data channel will be closed, not the transport.

---

Mention: @taylor-b

lgrahl:
This adds steps for the data channel *closing* procedure. Explicitly states that pending messages must be sent before the channel can be closed.

Resolves #1699 

I will write a WPT test for this but that will be bundled in one big data channel PR towards WPT (yes, sorry for that but I'm running out of time).

***
<a href="https://pr-preview.s3.amazonaws.com/lgrahl/webrtc-pc/pull/1822.html" title="Last updated on Mar 31, 2018, 10:06 PM GMT (99c487a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1822/a86cc50...lgrahl:99c487a.html" title="Last updated on Mar 31, 2018, 10:06 PM GMT (99c487a)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8738,22 +8738,26 @@ interface RTCTrackEvent : Event {
           <code>closing</code>.</p>
         </li>
         <li>
-          <p>Finish sending all currently pending messages of the
-          <var>channel</var>.</p>
-        </li>
-        <li>
-          <p>Follow the closing procedure defined for the <var>channel</var>'s
-          underlying <a data-lt="data transport">transport</a>:</p>
+          <p>Run the following steps in parallel:</p>
           <ol>
             <li>
-              <p>In the case of an SCTP-based <a data-lt="data transport">
-              transport</a>, follow [[!RTCWEB-DATA]], section 6.7.</p>
+              <p>Finish sending all currently pending messages of the <var>channel</var>.</p>
+            </li>
+            <li>
+              <p>Follow the closing procedure defined for the <var>channel</var>'s
+              underlying <a data-lt="data transport">transport</a>:</p>
+              <ol>
+                <li>
+                  <p>In the case of an SCTP-based <a data-lt="data transport">
+                  transport</a>, follow [[!RTCWEB-DATA]], section 6.7.</p>
+                </li>
+              </ol>
+            </li>
+            <li>
+              <p>Render the <var>channel</var>'s <a>data transport</a> <a>closed</a> by following
+              the associated procedure.
             </li>
           </ol>
-        </li>
-        <li>
-          <p>Render the <var>channel</var>'s <a>data transport</a>
-          <a>closed</a> and follow the associated procedure.
         </li>
       </ol>
       <p>When an <code><a>RTCDataChannel</a></code> object's <a>underlying data
--- a/webrtc.html
+++ b/webrtc.html
@@ -8723,10 +8723,39 @@ interface RTCTrackEvent : Event {
       <p>An <code><a>RTCDataChannel</a></code> object's <a>underlying data
       transport</a> may be torn down in a non-abrupt manner by running the
       <dfn id="data-transport-closing-procedure">closing procedure</dfn>. When
-      that happens the user agent MUST, unless the procedure was initiated by
-      the <code><a data-link-for="RTCDataChannel">close</a></code> method, queue a
-      task that sets the object's <a>[[\ReadyState]]</a> slot to <code>closing</code>.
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM. Would be good if @adam-be had a look.

webrtc.html
```diff
@@ -8723,10 +8723,39 @@ <h3><dfn>RTCDataChannel</dfn></h3>
...
+          <p>Let <var>channel</var> be the <code><a>RTCDataChannel</a></code>
+          object whose <a data-lt="data transport">transport</a> was
+          closed.</p>
+        </li>
+        <li>
+          <p>Unless the procedure was initiated by the <var>channel</var>'s
+          <code><a data-link-for="RTCDataChannel">close</a></code> method, set
+          <var>channel</var>'s <a>[[\ReadyState]]</a> slot to
+          <code>closing</code>.</p>
+        </li>
```
adam-be: I think the state update is the only operation that needs to queued on the main thread. A simple fix for that would be to bring back the text above that conditionally queued the task to fire the event, and then just "run these steps" for the rest.

adam-be: The task updating the state should fire the event at channel as well.

webrtc.html
```diff
@@ -8723,10 +8723,39 @@ <h3><dfn>RTCDataChannel</dfn></h3>
...
+        <li>
+          <p>Finish sending all currently pending messages of the
+          <var>channel</var>.</p>
```
taylor-b: This can't be done directly from a queued task on the main thread; I think we need the whole "in parallel..." deal here.

lgrahl: Would it be enough to mention that in the task? E.g.

> Finish sending all currently pending messages of the channel in parallel and wait for that procedure to complete.

webrtc.html
lgrahl: > The task updating the state should fire the event at channel as well.

IIRC that would clash with the rule to not fire events due to user action (even though I don't like that rule).

lgrahl:
I added tests for this as part of https://github.com/w3c/web-platform-tests/pull/10468.

webrtc.html
adam-be: You're right that there's no event.

webrtc.html
taylor-b: That may imply that the main thread is blocked while waiting, though.

lgrahl: Mhh...

> Finish sending all currently pending messages of the channel in parallel and, in the task, wait for that procedure to complete.

Or do you have another suggestion?

taylor-b: "In the task, wait for that procedure to complete" may still imply the main thread is blocked. We follow a model of "queued task = main thread", "in parallel = other threads". So I think it would be more consistent to structure it as:

- In parallel:
  - Finish sending all pending messages.
  - Follow the closing procedure.
  - Queue a task to render the channel closed.

Or in other words, there's one task that changes the state to "closing", then stuff happens in parallel, then there's another task that changes the state to "closed". Rather than one task that waits.

lgrahl:
@taylor-b Updated based on your comments. The [closed procedure](https://rawgit.com/w3c/webrtc-pc/master/webrtc.html#data-transport-closed) already states that a task must be queued, so I omitted that from the description.

alvestrand:
@adam-be @taylor-b please indicate approval (or not).

Issue #1830, opened 2018-04-11
The "Create an RTCRtpSender" algorithm needs a `kind` argument


jan-ivar:
The [Create an RTCRtpSender](http://w3c.github.io/webrtc-pc/#dfn-create-an-rtcrtpsender) algorithm uses undeclared "kind" variable: *"If sender is of kind "audio" "*

Sender has no kind yet, and algorithm needs to tolerate *track* being null.

fippo:
ORTC has a [constructor with "trackOrKind"](http://draft.ortc.org/#rtcrtpsender-interface-definition*) as the first argument for exactly that reason.

aboba:
@fippo In WebRTC 1.0, RTCRtpSender does not have a constructor.

PR #1831, opened 2018-04-11
Pass kind into algorithms that rely on it.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1830.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4924,6 +4924,9 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                   method's first argument.</p>
                 </li>
                 <li>
+                  <p>Let <var>kind</var> be <var>track.kind</var>.</p>
+                </li>
+                <li>
                   <p>Let <var>streams</var> be a list of
                   <code><a>MediaStream</a></code> objects constructed from the
                   method's remaining arguments, or an empty list if the method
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5495,7 +5497,8 @@ <h3><dfn>RTCRtpSender</dfn> Interface</h3>
       changed appropriately.</p>
 
       <p>To <dfn>create an RTCRtpSender</dfn> with a
-      <code><a>MediaStreamTrack</a></code>, <var>track</var>, a list of
+      <code><a>MediaStreamTrack</a></code>, <var>track</var>, a string,
+      <var>kind</var>, a list of
```
alvestrand: the language "a string, kind" reads oddly to me. it's kind of the verbal equivalent of 

function createRtcRtpSender(DOMString kind, .....)

but emphasizing the stringiness seems weird to me. Not going to object too much, though.

jan-ivar: I prefer to not list types at all (we're inconsistent throughout the spec). But if we're going to list types then, kind is not a type, its type is string, and we are listing types here.

jan-ivar: "...and optionally an id string, id" was already there. Should I remove the types altogether?

webrtc.html
```diff
@@ -6559,7 +6562,7 @@ <h2>Dictionary <code><a>RTCRtpHeaderExtensionCapability</a></code> Members</h2>
       <h3><dfn>RTCRtpReceiver</dfn> Interface</h3>
       <p>The <code>RTCRtpReceiver</code> interface allows an application to
       inspect the receipt of a <code>MediaStreamTrack</code>.</p>
-      <p>To <dfn>create an RTCRtpReceiver</dfn> with kind, <var>kind</var>, and
```
jan-ivar: My comment was in reference to this line, which uses `kind` as both type and name.

webrtc.html
alvestrand: I'm happy with leaving the types out for these non-formal interfaces.
(the reason why audio/video isn't an enum is an artifact of WebIDL's treatment of enums and extensibility, not because we didn't want it to be, I think.)

Issue #1240, opened 2017-05-24
editorial: media api introduction


fippo:
http://w3c.github.io/webrtc-pc/#rtp-media-api

> The RTP media API lets a web application send and receive MediaStreamTracks over a peer-to-peer 
connection.

This is odd. Tracks and streams are an abstraction that is used by this API to describe how media is sent over a p2p connection.

> Tracks, when added to a RTCPeerConnection, result in signaling; when this signaling is forwarded to a remote peer, it causes corresponding tracks to be created on the remote side.

This is even more weird. It is correct but boiled down a bit too much.
If an application adds a track to a peerconnection it signals its intent to transmit media over that connection. This triggers the negotiation process described in 4.7 because we need to negotiate how this media is going to be transmitted over the network in a way that the other peer understands. As part of the negotiation, signaling messages are exchanges between the two applications which make the stream pop up at the remote side. No data is sent until the p2p connection is up though

Shall I try to come up with something along those lines?

taylor-b:
I'd hold off on doing anything until we settle how tracks are signaled. See issue #1161.

aboba:
@taylor-b Now that Issue #1161 has been closed, can we move forward?

aboba:
@fippo Can you submit a PR?

taylor-b:
> Now that Issue #1161 has been closed, can we move forward?

There are still discussions ongoing about track ID signaling (https://github.com/rtcweb-wg/jsep/issues/842), but I guess the introduction could ignore that aspect.

fippo:
Actually with the transceiver model 
> when this signaling is forwarded to a remote peer, it causes corresponding tracks to be created on the remote side.

is not even true anymore as addTransceiver('audio') will generate an SDP that creates a track on the remote side. At least that is my understanding of transceivers and jsep now (and I still don't think this is a good idea)

taylor-b:
Is it just the default direction of "sendrecv" you have a problem with? Or more generally, do you not like that it's possible to have a remote track pop up on the receive side before one is set on the send side?

Anyway, I agree the "corresponding tracks" sort of language is misleading, since tracks don't map 1:1 between the send/receive side. It would be good if this were explained somewhere, though it's not exactly straightforward so I don't know if the intro is the best place.

fippo:
[...]

PR #1832, opened 2018-04-11
rtp media api: remove note

fippo:
removes the note which no longer applies in the transceiver model
as track ids have become meaningless.

Originally I wanted to remove only `Doing this implies that at the receiving end of the peer-to-peer connection there are several MediaStreamTracks with an identical id.` which is no longer true in the transceiver model. But the whole note seems to be an attempt to explain "there can be tracks with identical id if the sender adds the same track multiple times"...

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-pc/pull/1832.html" title="Last updated on Apr 11, 2018, 12:15 PM GMT (7dc9dc7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1832/1b42a97...fippo:7dc9dc7.html" title="Last updated on Apr 11, 2018, 12:15 PM GMT (7dc9dc7)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4779,27 +4779,6 @@ interface RTCPeerConnectionIceErrorEvent : Event {
       <code>MediaStreamTrack</code> and <code><a>RTCRtpReceiver</a></code> are
       surfaced to the application via the <code><a>track</a></code> event.
     </p>
-    <div class="note">
-      <p>There are several ways to initiate the sending of a
-      <code>MediaStreamTrack</code> over a peer-to-peer connection.
-      One way is to use the <code>addTrack</code> method on the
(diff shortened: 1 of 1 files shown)


jan-ivar:
I agree with the criticism of the note, however I think it also explained that it's possible to send the same bits twice. There's a lot of exposition here not available anywhere else. Any way to go with your original suggestion?

fippo:
I think this needs to be more explicit than just as a note. But that is a good point to add to #1240

alvestrand:
seems this PR should wait for outcome of ID assignment discussion.

fippo:
this is already not true in Firefox 59+

PR #1844, opened 2018-04-20
Fix setCodecPreferences note regarding offer/answer negotiation.

### LINKED ISSUE #1838: setCodecPreferences contradicts JSEP wrt answer codec list being a subset?
https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setcodecpreferences says
```
Note that calls to createAnswer will use only the common subset of these codecs and 
the codecs that appear in the offer.
```
seems to contradict the last sentence of this paragraph from http://rtcweb-wg.github.io/jsep/#rfc.section.5.3.1
```
Otherwise, the media formats on the m= line MUST be generated in the same order as 
those offered in the current remote description, excluding any currently unsupported 
formats. Any currently available media formats that are not present in the current 
remote description MUST be added after all existing formats.
```
However, this paragraph won't apply when setCodecPreferences has been used as the one before it says
```
If codec preferences have been set for the associated transceiver, media formats MUST
 be generated in the corresponding order, regardless of what was offered, and MUST 
exclude any codecs not present in the codec preferences.
```

I think removing that informal "Note" is the way to go.

taylor-b:
Fixes #1838.

The note said that, even if codec preferences are set, the list of
codecs in the answer will be a subset of those in the offer. Which JSEP
required before, but now does not:

  [T]he media formats in the answer MUST include at least one format
  that is present in the offer, but MAY include formats that are locally
  supported but not present in the offer.

So now, the subset rule is only a SHOULD in RFC3264, and not a MUST in
JSEP. So although most implementations are expected to honor the
"SHOULD", implementations can technically offer codecs in the answer
that were not in the offer. Which is now clarified in this note, with a
brief example.

Also moving the note out of the normative portion of `setCodecPreferences`,
and putting in an actual "note" block.

***
<a href="https://pr-preview.s3.amazonaws.com/taylor-b/webrtc-pc/pull/1844.html" title="Last updated on Apr 25, 2018, 10:57 PM GMT (7308fbe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1844/670a792...taylor-b:7308fbe.html" title="Last updated on Apr 25, 2018, 10:57 PM GMT (7308fbe)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7188,15 +7188,14 @@ async function updateParameters() {
               fulfill these requirements, the user agent MUST <a>throw</a> an
               InvalidAccessError.</p>
               <p class="note">
-              Calls to <code>createAnswer</code> will likely use only the
-              common subset of the codec preferences and the codecs that appear
-              in the offer, due to a recommendation in [[!RFC3264]]. For
-              example, if codec preferences are "C, B, A", but only codecs "A,
-              B" were offered, the answer will probably only contain codecs "B,
-              A". Although <span data-jsep="initial-answers">[[!JSEP]]</span>
-              does allow adding codecs that were not in the offer, so this may
-              differ by implementation.</p>
-
+              Due to a recommendation in [[!RFC3264]], calls to
+              <code>createAnswer</code> SHOULD use only the common subset of
+              the codec preferences and the codecs that appear in the offer.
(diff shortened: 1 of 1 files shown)


taylor-b:
@aboba Done.

PR #1845, opened 2018-04-20
Treat rejected m= sections (port of 0) as having "inactive" direction.

### LINKED ISSUE #1812: Does direction matter for rejected m= sections?
The direction of an m= section in a remote description affects what happens to remote tracks; for example:

> If direction is "sendonly" or "inactive", and transceiver's [[CurrentDirection]] slot is neither "sendonly" nor "inactive", process the removal of a remote track for the media description, given transceiver.

But what if the m= section is rejected? In that case, it seems odd that the direction would make a difference, since it affects nothing underneath. It doesn't matter if a rejected m= section has "a=sendrecv"; none of its streams will be sent or received regardless.

So, would it be appropriate to treat rejected m= sections as implicitly having an `inactive` direction?

taylor-b:
Fixes #1812.

This will:

- Cause `currentDirection` to change to `"inactive"`
- Cause the remote track to be removed from its stream(s), if
  applicable.

This seems pretty sensible, as the direction has no relevance for a
rejected m= section. No packets will be sent or received for a rejected
m= section (aka, stopped transceiver), so it's effecitvely `"inactive"`;
it would be odd if a different direction in a rejected m= section
produced different behavior.

This also resolves a problem where remote tracks were possibly never
being removed from their streams, which would happen if you set a remote
description with a rejected m= section with "sendrecv" or "sendonly"
direction.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1640,7 +1640,8 @@
                         <p>Let <var>direction</var> be an <code>
                         <a>RTCRtpTransceiverDirection</a></code> value
                         representing the direction from the <a>media
-                        description</a>.</p>
+                        description</a>, or <code>"inactive"</code> if the
+                        <a>media description</a> section was rejected.</p>
                       </li>
                       <li>
                         <p>If <var>direction</var> is <code>"sendrecv"</code> or
@@ -1675,7 +1676,8 @@
                             representing the direction from the <a>media
                             description</a>, but with the send and receive
                             directions reversed to represent this peer's point
-                            of view.</p>
+                            of view, or <code>"inactive"</code> if the
+                            <a>media description</a> was rejected.</p>
                           </li>
                           <li>
                             <p>As described by <span data-jsep=


jan-ivar:
See [this blog](https://blog.mozilla.org/webrtc/rtcrtptransceiver-explored/) for a refresher on what events we fire today (examples work in Firefox). In short, `ended` and `muted` are orthogonal. This works out somewhat ok in practice for playback at least, since video elements know to ignore `ended` tracks, so it doesn't matter much that they stay in their stream(s).

With this change we'd be firing not just `ended`, but also `muted` on remote tracks, as well as removing them from any remote streams and firing`removetrack` on said streams. Is that the intent?

This may make sense, but we'd need to update Firefox to comply, and we just shipped it.

Also, if we we want to do this, I think it would make sense to preserve the symmetry of doing this on the offerer side as well. Currently, this is done in prose under [stop()](http://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-stop):

 *  *" In addition, since the receiver's [[ReceiverTrack]] has ended, the steps described in track ended MUST be followed."*

We probably want to extend this to do `muted` and `removetrack` then too, right?

taylor-b:
> With this change we'd be firing not just ended, but also muted on remote tracks, as well as removing them from any remote streams and firingremovetrack on said streams. Is that the intent?

Yes. The intent is that there should be no observable difference between `setRemoteDescription(rejected_sendrecv_answer)` and `setRemoteDescription(rejected_inactive_answer)`.

> Also, if we we want to do this, I think it would make sense to preserve the symmetry of doing this on the offerer side as well. Currently, this is done in prose under [stop()](http://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-stop):
> 
>  *  *" In addition, since the receiver's [[ReceiverTrack]] has ended, the steps described in track ended MUST be followed."*
> 
> We probably want to extend this to do `muted` and `removetrack` then too, right?

So, you're suggesting to do the `muted` and `removetrack` in the `stop()` steps, rather than `setRemoteDescription` steps? That would also resolve #1178. So that seems ok by me. I'll make another PR for that.

jan-ivar:
> Yes. The intent is that there should be no observable difference between setRemoteDescription(rejected_sendrecv_answer) and setRemoteDescription(rejected_inactive_answer).

Is there high value in that? Today, a pc has discrete control over what happens on the remote end.

E.g. if the app wants to remotely remove tracks on stop, it can do that easily:
```js
transceiver.direction = "inactive";
transceiver.stop();
```

Also, note that peer connections only remove tracks from streams it itself manages, so if apps do:
```js
video.srcObject = new MediaStream([transceiver. receiver.track]);
```
then we don't remove tracks from video.srcObject, so there's no panacea against JS having to deal with ended tracks.

jan-ivar:
Since there was a question about what I was suggesting, my preference would be to stick with existing behavior. We shipped this in Firefox 59.

PR #1849, opened 2018-04-21
Data channel receive procedure

### LINKED ISSUE #1825: Data channel receive procedure missing?
It seems there is no data channel receive procedure that can be followed, e.g. nothing tells us when a *message* event is invoked and what validation steps are required (check for *open* [or *closing*], ...).

The WPT tests reference [this WS spec section](https://html.spec.whatwg.org/multipage/web-sockets.html#feedback-from-the-protocol) (*When a WebSocket message has been received [...]*) and I would propose to copy that.

lgrahl:
This adds steps for receiving a data channel message. Also clarifies that messages can only be received as long as `.readyState` is `open`.

Resolves #1825 

Since implications on the closing procedure aren't trivial, here's an explanation:

1. Let peers A and B have a common data channel
2. B sends data in a tight loop.
3. A enqueues data to send and then closes the channel (which will enqueue to send an outgoing stream reset after all messages have been delivered). A's channel is now in the `closing` state.
    - **Important:** A cannot receive any more messages from B at this point! All incoming data from B will be discarded.
4. B's channel is still in the `open` state until it has received all messages that have been queued by A.
5. After having received all messages by A, B's channel will move into the `closing` state (due to having received an incoming stream reset and acknowledging it). B is allowed to discard any pending outgoing messages (if possible) and will send an outgoing stream reset.
6. A's channel will move into the `closed` state (due to having received acknowledgement for its outgoing stream reset and received an incoming stream reset).
7. B's channel will move into the `closed` state (due to having received acknowledgement for its outgoing stream reset).

Tests for *enqueue messages and close* exist as part of https://github.com/w3c/web-platform-tests/pull/10468.

***
<a href="https://pr-preview.s3.amazonaws.com/lgrahl/webrtc-pc/pull/1849.html" title="Last updated on Apr 21, 2018, 11:25 PM GMT (af41c10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1849/670a792...lgrahl:af41c10.html" title="Last updated on Apr 21, 2018, 11:25 PM GMT (af41c10)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8787,6 +8787,54 @@ interface RTCTrackEvent : Event {
           <var>channel</var>.</p>
         </li>
       </ol>
+      <p>When an <dfn data-lt="receive an rtcdatachannel message">
+      <code><a>RTCDataChannel</a></code> message has been received</dfn> via
+      the <a>underlying data transport</a> with type <var>type</var> and data
+      <var>rawData</var>, the user agent MUST queue a task to run the following
+      steps:</p>
+      <ol>
+        <li>
+          <p>Let <var>channel</var> be the <code><a>RTCDataChannel</a></code>
+          object for which the user agent has received a message.</p>
+        </li>
+        <li>
+          <p>If <var>channel</var>'s <a>[[\ReadyState]]</a> slot is not
+          <code>open</code>, abort these steps and discard <var>rawData</var>.
+          </p>
+        </li>
+        <li>
+          <p>Execute the sub step by switching on <var>type</var> and the
+          <var>channel</var>'s <code>binaryType</code>:</p>
+          <ul>
+            <li>
+              <p>If <var>type</var> indicates that <var>rawData</var> is a
+              <code>string</code>:</p>
+              <p>Let <var>data</var> be a <span class=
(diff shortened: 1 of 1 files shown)


lgrahl:
@domenic I roughly copied this from the [WebSocket receive procedure](https://html.spec.whatwg.org/multipage/web-sockets.html#feedback-from-the-protocol) which states:

> [...] a new Blob object, **created in the relevant Realm of the WebSocket object** [...]

I was wondering what that *relevant Realm* is about and whether we need it? I tried figuring it out but that's hard stuff to digest, so I ignored it for now. (Maybe you can ping me in IRC.)

domenic:
In general when creating objects, you need to state what realm (= global object) they are created in. Consider scenarios like the example in https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects where there are four global objects involved. In each of them, `Blob` (or `ArrayBuffer`) are different values. Which one should be used for constructing this object?

Relevant is probably the right choice indeed. So you'll want to add that to both the Blob and ArrayBuffer-creation steps.

PR #1851, opened 2018-04-22
In-band negotiated channel should be open when dispatching

### LINKED ISSUE #1761: Initial data channel state is inconsistent
In section [6.2](https://w3c.github.io/webrtc-pc/#rtcdatachannel) of the spec we have the following sentence:  
"An RTCDataChannel, created with createDataChannel or dispatched via an RTCDataChannelEvent, MUST initially be in the connecting state."

But then in the [RTCDataChannelState Enumeration state](https://w3c.github.io/webrtc-pc/#dom-rtcdatachannelstate) for 'open' it says:
"This is the initial state of an RTCDataChannel object dispatched as a part of an RTCDataChannelEvent."

I'm guessing that these two sentences were written before the "underlying data transport" concept got introduced.

To me it would make the most sense if these get changed to describe that 'connecting' is the initial state if the underlying data transport is not yet connected, and 'open' should be state if the underlying data transport is connected.

lgrahl:
These changes move the data channel that is about to be announced into the `open` state prior to firing the `datachannel` event. (Also explicitly announces the data channel as open now.)

Resolves #1761

I can write a WPT test for this once merged.

***
<a href="https://pr-preview.s3.amazonaws.com/lgrahl/webrtc-pc/pull/1851.html" title="Last updated on Apr 22, 2018, 12:49 AM GMT (6648f92)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1851/670a792...lgrahl:6648f92.html" title="Last updated on Apr 22, 2018, 12:49 AM GMT (6648f92)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8654,6 +8654,10 @@ interface RTCTrackEvent : Event {
           object to be announced.</p>
         </li>
         <li>
+          <p>If <var>channel</var>'s <a>[[\ReadyState]]</a> is <code>closing</code> or
+          <code>closed</code>, abort these steps.</p>
+        </li>
+        <li>
           <p>Set <var>channel</var>'s <a>[[\ReadyState]]</a> slot to
           <code>open</code>.</p>
         </li>
@@ -8735,9 +8739,8 @@ interface RTCTrackEvent : Event {
           <code><a>RTCPeerConnection</a></code> object.</p>
         </li>
         <li>
-          <p>If the <var>channel</var>'s <a>[[\ReadyState]]</a> is still
-          <code>open</code>, <a data-lt="announce the rtcdatachannel as open">
-          announce the data channel as open</a>.</p>
+          <p><a data-lt="announce the rtcdatachannel as open">Announce the data channel as
+          open</a>.</p>
         </li>
       </ol>
       <p>An <code><a>RTCDataChannel</a></code> object's <a>underlying data
--- a/webrtc.html
+++ b/webrtc.html
@@ -8722,9 +8722,20 @@ interface RTCTrackEvent : Event {
           </table>
         </li>
         <li>
+          <p>Set <var>channel</var>'s <a>[[\ReadyState]]</a> to
+          <code>open</code> (but do not fire the <code><a>open</a></code>
+          event, yet).</p>
+          <div class="note">This allows to start sending messages inside of the
+          <code><a>datachannel</a></code> event handler prior to the
+          <code><a>open</a></code> event being fired.</div>
+        </li>
+        <li>
           <p><a>Fire a datachannel event</a> named
           <code><a>datachannel</a></code> with <var>channel</var> at the
           <code><a>RTCPeerConnection</a></code> object.</p>
+        </li>
+        <li>
+          <p>If the <var>channel</var>'s <a>[[\ReadyState]]</a> is still <code>open</code>, <a data-lt="announce the rtcdatachannel as open">announce the data channel as open</a>.</p>
         </li>
       </ol>
       <p>An <code><a>RTCDataChannel</a></code> object's <a>underlying data
--- a/webrtc.html
+++ b/webrtc.html
@@ -8735,7 +8735,9 @@ interface RTCTrackEvent : Event {
           <code><a>RTCPeerConnection</a></code> object.</p>
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


lgrahl:
Let me rephrase: I can write a WPT test for this once **approved** (didn't want to mess with https://github.com/w3c/web-platform-tests/pull/10468 before approval).

webrtc.html
```diff
@@ -8721,11 +8721,22 @@ <h3><dfn>RTCDataChannel</dfn></h3>
...
+          <code><a>datachannel</a></code> event handler prior to the
+          <code><a>open</a></code> event being fired.</div>
+        </li>
         <li>
           <p><a>Fire a datachannel event</a> named
           <code><a>datachannel</a></code> with <var>channel</var> at the
           <code><a>RTCPeerConnection</a></code> object.</p>
         </li>
+        <li>
+          <p>If the <var>channel</var>'s <a>[[\ReadyState]]</a> is still <code>open</code>, <a data-lt="announce the rtcdatachannel as open">announce the data channel as open</a>.</p>
```
jan-ivar: Normally, I'd worry about running browser code after calling content JS - we normally fire events at the end - but apparently [announce the data channel as open](http://w3c.github.io/webrtc-pc/#announce-datachannel-open) queues a separate task, so that seems fine.

But why this test? Is it to avoid firing the `open` event if JS does
```js
pc.ondatachannel = ({channel}) => channel.close();
pc.onopen = () => { /* never fires */ };
```
?

If so, it won't catch:
```js
pc.ondatachannel = async ({channel}) => { await 0; channel.close(); };
pc.onopen = ({channel}) => console.log(channel.readyState); // open, and remains so
```
If I read the spec right, this would appear to leave `channel.readyState` as `"open"`, even though the channel is closed. That seems wrong. We didn't specify it before, but if we're going to, then I think we want it to be right.
 
Should we maybe check existing state in the queued task? We check pc.[[IsClosed]] but not [[ReadyState]].

Or would it be better to not set state at all, since we've opened it already?

jan-ivar: and wordwrap.

lgrahl:
Added tests for this in https://github.com/lgrahl/web-platform-tests/commit/1de5a5425419df83edc5b2216f5e9316729aa81a.

@taylor-b Supplement from these tests: Chrome still fires the `open` event when closing the channel inside the `datachannel` event handler. Firefox does not.

webrtc.html
jan-ivar: @lgrahl Sorry for dropping the ball here. Thanks for confirming intent.

> I don't see a problem with that since `open` will fire in between `await 0`

Actually, it won't. `await 0` (or `Promise.resolve().then(() => channel.close())` if you prefer, this is not specific to await syntax), will execute on a [micro-task](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/), [immediately](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#Timing) after the current task, before the next queued task.

I think we need to modify [announce the data channel as open](http://w3c.github.io/webrtc-pc/#announce-datachannel-open) to check that readyState is not `"closed"` or `"closing"`, before setting it to `"open"`. This seems like a good idea anyway in a queued task.

lgrahl: Alright, so this is what my naive mind has assumed how it works (referring to the second code example in https://github.com/w3c/webrtc-pc/pull/1851#discussion_r185976747):
1. The `datachannel` event fires which runs some synchronous code immediately - the event handler. This event handler enqueues a new task to run the code after `await 0`.
2. Since the synchronous code could have closed the channel in the meantime, the check is required.
3. The scheduled `channel.close()` runs after all other tasks enqueued before this one ran.

Please correct me if it works differently. And sorry for the brief detour. :)

However, your suggestion to move that code into the `open` announcement sounds good to me. Only the description *announce the data channel as open* is a bit awkward if it under certain circumstances does not do so. Suggestions welcome.

jan-ivar: > announcing the channel as open enqueues a new task and that runs after

Exactly.

> announce the data channel as open is a bit awkward if it under certain circumstances does not do so

That's already the case if [[IsClosed]] is true, so I wouldn't worry about it.

jan-ivar:
@lgrahl are there more changes coming to address my comments wrt my feedback on checking readyState in the queued task?

lgrahl:
@jan-ivar Yes, I plan to address your comments as discussed. Just got distracted by other stuff. :)

Issue #1497, opened 2017-07-20
Not possible to tell how old `RTCRtpContributingSource.timestamp` is


taylor-b:
This issue was raised by an application developer, in a [chromium bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=625208#c13):

> What I'm trying to do is make [the application] more reactive - when a participant stops talking, at the moment, it's kept in the list of contributing sources for another 10 seconds (goal is to reduce that). I'd like to compare the timestamp to a current value, to see how many msec ago the source was contributing (so instead of 10 seconds, it can be easily and reliably configured to something lower), which is hard/hacky to do without knowing the reference point of the timestamp (getting current value of the timestamp). I'd like to treat each contributing source individually, not compare the timestamps between each other.

Basically, they'd like to be able to easily tell "how long ago did this speaker say something"? The only guarantee provided for `timestamp` is that it uses "a local clock". But there's no way for the application to get the current value of this clock.

I suppose the application could, say, poll the `timestamp` attribute until it sees an update, and then figure out how to correlate it with `performance.now()`. But that seems less than ideal and prone to imprecision.

PR #1854, opened 2018-04-25
Define the reference clock used for RTCRtpContributingSource.timestamp

### LINKED ISSUE #1690: RTCRtpContributingSource.timestamp needs a clearer definition
It says:

> The timestamp is defined in [HIGHRES-TIME] and corresponds to a local clock.

What does "local clock" mean here?  The definition in HIGHRES-TIME allows DomHighResTimestamp to mean one of three things:

1) A time measured from the time origin of some global (in which case you need to define which one).
2) A time measured on the global monotonic clock.
3) A time measured from some other DOMHighResTimeStamp.

I assume "local clock" is meant to preclude option 2, but it's not clear whether option 1 or option 3 is meant, and in either case more information needs to be provided as to which timebase is used.

taylor-b:
Fixes #1690 and #1497.

Following the decision at the January VI, this timestamp uses
performance.timeOrigin + performance.now() (matching
RTCStats.timestamp).

This will allow an application to tell "how long ago did media from
this source play?", whereas previously it was only defined as "a
local clock", which meant it only supported comparing timestamps
between sources.

Also replacing "playout of an RTP packet containing the source"
with "playout of media originating from this source", since you
don't really "play out" an RTP packet directly.

***
<a href="https://pr-preview.s3.amazonaws.com/taylor-b/webrtc-pc/pull/1854.html" title="Last updated on Apr 25, 2018, 11:56 PM GMT (43c2895)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1854/53e0ef7...taylor-b:43c2895.html" title="Last updated on Apr 25, 2018, 11:56 PM GMT (43c2895)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6820,10 +6820,10 @@ async function updateParameters() {
             "idlMemberType"><a>DOMHighResTimeStamp</a></span>, required</dt>
             <dd>
               <p>The timestamp of type DOMHighResTimeStamp [[!HIGHRES-TIME]],
-              indicating the most recent time of playout of media
-              originating from this source. The timestamp is defined as
-              <a>performance.timeOrigin</a> + <a>performance.now()</a> at
-              the time of playout.</p>
+              indicating the most recent time of playout of media that arrived
+              in an RTP packet originating from this source. The timestamp is
+              defined as <a>performance.timeOrigin</a> +
+              <a>performance.now()</a> at the time of playout.</p>
             </dd>
             <dt><dfn data-idl><code>source</code></dfn> of type <span class=
             "idlMemberType"><a>unsigned long</a></span>, required</dt>
--- a/webrtc.html
+++ b/webrtc.html
@@ -83,6 +83,9 @@
     <p>The terms <dfn>event</dfn>, <dfn data-cite="!HTML51/webappapis.html#events-event-handlers">event
     handlers</dfn> and <dfn data-cite="!HTML51/webappapis.html#event-handler-event-type">event
     handler event types</dfn> are defined in [[!HTML51]].</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -83,6 +83,9 @@ <h2>Terminology</h2>
     <p>The terms <dfn>event</dfn>, <dfn data-cite="!HTML51/webappapis.html#events-event-handlers">event
     handlers</dfn> and <dfn data-cite="!HTML51/webappapis.html#event-handler-event-type">event
     handler event types</dfn> are defined in [[!HTML51]].</p>
+    <code><dfn data-cite="!HIGHRES-TIME#dom-performance-timeOrigin">performance.timeOrigin</dfn></code>
```
alvestrand: Format nit: This one needs to be wrapped in <p> ... </p>, or formatting may be odd.

webrtc.html
```diff
@@ -6817,9 +6820,10 @@ <h2>Dictionary RTCRtpContributingSource Members</h2>
             "idlMemberType"><a>DOMHighResTimeStamp</a></span>, required</dt>
             <dd>
               <p>The timestamp of type DOMHighResTimeStamp [[!HIGHRES-TIME]],
-              indicating the most recent time of playout of an RTP packet
-              containing the source. The timestamp is defined in
-              [[!HIGHRES-TIME]] and corresponds to a local clock.</p>
+              indicating the most recent time of playout of media
+              originating from this source. The timestamp is defined as
```
alvestrand: This rephrasing lost the reference to the RTP packet. Was this intentional?
I'm not sure it contributes any value (is there media not contained in an RTP packet?), but we might consider "the most recent time of playout of media that arrived in an RTP packet originating from this source".
This may be "interesting" to compute in the case of media that's mixed before playout, but let's burn that bridge in another bug.

Issue #1855, opened 2018-04-26
replaceTrack algorithm lost some important text


jan-ivar:
When [#1772](https://github.com/w3c/webrtc-pc/pull/1772/files?w=1) removed [this step](https://w3c.github.io/webrtc-pc/archives/20171002/webrtc.html) from the the `replaceTrack` algorithm:

*"8. If transceiver is not yet associated with a media description, then set sender's track attribute to withTrack, and return a promise resolved with undefined."*

...the intent was merely to make this synchronous step asynchronous, i.e. to have the sender's track attribute set to withTrack asynchronously.

Unfortunately, without changes to the subsequent language around *"Determine if negotiation is needed to transmit withTrack in place of the sender's existing track"*, the net effect is typically that it rejects instead, which was not our intent.

I'll try to come up with prose to fix this.

PR #1863, opened 2018-05-01
Improve replaceTrack prose; restores lost pre-negotiation behavior.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1855.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5869,25 +5869,20 @@ async function updateParameters() {
                       <p>Run the following steps in parallel:</p>
                       <ol>
                         <li>
-                          <p>Determine if negotiation would be needed in order to
-                          replace <var>sender</var>'s existing track with
-                          <var>withTrack</var>. Ignore whether the sender already
-                          needs negotiation, in this determination.
-                          Negotiation is not needed if <var>withTrack</var>
-                          is null. Ignore which <code>MediaStream</code> the track
-                          resides in and the <code>id</code> attribute of the track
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5869,25 +5869,20 @@ <h2>Methods</h2>
...
-                          resides in and the <code>id</code> attribute of the track
-                          in this determination. Also, ignore
-                          <var>transceiver</var>'s <a>[[\Direction]]</a> slot in
-                          this determination. If negotiation is needed, then
-                          <a>reject</a> <var>p</var> with a newly
+                          If <var>withTrack</var> is not <code>null</code> and
+                          <var>transceiver</var> is already associated with a
+                          media description, determine if <var>withTrack</var>
+                          can be sent immediately by the sender without violating
+                          the sender's already-negotiated envelope, and if it
```
stefhak: Is "already-negotiated envelope" defined well enough?

taylor-b: This also loses the "ignore direction" part; if the direction is "recvonly", the track can't immediately be sent, so should the promise be rejected?

jan-ivar: @taylor-b The new text also ignores direction. We want replaceTrack to succeed whenever it is valid to replace the track, so in the "recvonly" case it should be fine to replace the track regardless of the envelope, is that right?

jan-ivar: @jan-ivar will bring back "ignore direction"

Issue #1778, opened 2018-02-21
Recycling an m= section may not remove the old transceiver's remote track from its stream


taylor-b:
There's a particular corner case that, according to the current "apply a description" algorithm, will leave a dangling track as part of a remote stream when an m= section is recycled.

In most cases this won't happen, because the track will be removed during the offer/answer that rejects the m= section. For example:

1. Local offer adding `sendrecv` transceiver.
2. Remote `sendrecv` answer.
3. Local offer stopping transceiver.
4. Remote `inactive` answer (track removed from stream here).
5. Local offer recycling m= section...

But consider this:

1. Local offer adding `sendrecv` transceiver.
2. Remote `sendrecv` answer.
3. Remote re-offer, still `sendrecv`.
4. Local answer stopping the transceiver (track not removed, because that only happens when applying remote descriptions).
5. Local offer recycling m= section.
6. Remote answer.

When the transceiver is rejected by a local answer, the "process the removal of a remote track" steps will never be followed.

So, assuming we want to fix this, when should the remote track actually be removed?

* When stopping the transceiver (in between steps 3 and 4)?
* When applying a local description with the transceiver stopped (step 4)?
* When applying a local description that recycles the m= section (step 5)?
* When applying a remote description that "accepts" the recycling (step 6)?

If done at step 6, this would match the existing pattern of "only remote descriptions change remote tracks/streams." It may be hard to specify that clearly though.

alvestrand:
at transceiver.stop seems logical to me.

jan-ivar:
> Local offer stopping transceiver.

@taylor-b I think there's some confusion here. Do you mean Local offer setting direction to `recvonly`?

`transceiver.stop()` fires the ended event on remote tracks on both ends thanks to:

*"In addition, since the receiver's [[ReceiverTrack]] has ended, the steps described in track ended MUST be followed."*

But it doesn't remove the track from its stream. That's something we only added for `direction`.

jan-ivar:
I think you mean this case (assuming live tracks in both direction):
```js
await pc1.setLocalDescription(await pc1.createOffer());
await pc2.setRemoteDescription(pc1.localDescription);
pc2.getTransceivers()[0].direction = "inactive";
await pc2.setLocalDescription(await pc2.createAnswer());
await pc1.setRemoteDescription(pc2.localDescription);
```
When I run [this](https://jsfiddle.net/jib1/s4o9ykdL/) in Firefox 59 I get:
```
[...]

Issue #1744, opened 2018-01-22
Multiple SRDs may leave streams and tracks in unexpected state.


jan-ivar:
In https://github.com/w3c/webrtc-pc/issues/1729#issuecomment-356787787 I mention stream associations in theory can change between multiple SRD calls.

There's an issue even if we don't mangle SDP:

* If the first SRD described a new track A in stream S, but the second (corrective) SRD didn't, you wouldn't expect to find track A being part of stream S, especially if you're already playing S.

You might not expect `pc.getTransceivers()` to grow at all, but that's JSEP (use rollback for that).
You might not expect a `track` event, but we agreed that's hard to avoid in https://github.com/w3c/webrtc-pc/issues/1729.

But at least, you'd expect track A to be muted, and not part of stream S.

The muted state is likely not an issue since we never set it to false, so that part of the algorithm can probably stay as it is (dependent on delta-comparison against [[CurrentDirection]]) .

But as I mentioned in https://github.com/w3c/webrtc-pc/issues/1729#issuecomment-356787787 I think we should lift the call to [set the associated remote streams](http://w3c.github.io/webrtc-pc/#set-associated-remote-streams) out one level, basically always call it with the msids (or empty list if send-bit is off).

stefhak:
I'm all for doing the proposed change, given it does not turn out that other things don't break as a consequence of it. It seems like a very marginal corner case to me (apply more than one remote offer).

jan-ivar:
@stefhak This is how Firefox 59 works now FWIW. There is one observable difference however: If the JavaScript empties the remote stream, then a renegotiation in 59 will add all the remote tracks back in, whereas before it wouldn't (by the old/existing spec you'd have to twiddle direction, i.e. it would only add/remove based on deltas compared to [[CurrentDirection]]).

jan-ivar:
Fixed by https://github.com/w3c/webrtc-pc/pull/1947.

Issue #1706, opened 2017-12-21
Should rollback fire addtrack/removetrack events? 


jan-ivar:
SRD(offer) can cause the following events to fire: `track`, `mute`, `addtrack`, and `removetrack`, as well as cause tracks to be added/removed from the remotely managed stream(s).

Should a rollback undo any/all of that? How do we undo the `track` event? E.g should we undo the `mute` event by firing the `unmute` event?

docfaraday:
I guess it depends on whether the intent is to allow sessions to continue functioning normally after a sRD(rollback), without requiring a subsequent negotiation to set the record straight. I _think_ this is the intent, right?

stefhak:
@docfaraday I have the same understanding of the intent (https://github.com/w3c/webrtc-pc/issues/1706#issuecomment-353430767) as you.

aboba:
Consensus from the April VI is for Jan-Ivar to produce a PR along the lines described in the slides.

stefhak:
@jan-ivar do you think we can close this Issue now that #1867 has been merged?

jan-ivar:
No, https://github.com/w3c/webrtc-pc/pull/1867 fixed missing events from remote rejection (https://github.com/w3c/webrtc-pc/issues/1778), not rollback.

PR #1867, opened 2018-05-02
Fire events in SLD(answer) in direction rejection edge-case.

jan-ivar:
In the special case where an answerer modifies `direction` to "reject" an incoming track only (as opposed to stopping the transceiver outright), redo the steps for track-removal and firing of `removetrack` and `muted` events - normally done in SRD(offer) - in SLD(answer) as well.

Fix for https://github.com/w3c/webrtc-pc/issues/1778.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1504,107 +1504,142 @@
                   </li>
                   <li>
                     <p>If <var>description</var> is set as a local description,
-                    then run the following steps for each <a>media description</a>
-                    in <var>description</var>:</p>
+                    then run the following steps:</p>
                     <ol>
                       <li>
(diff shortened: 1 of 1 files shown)


jan-ivar:
See [diff without whitespace](https://github.com/w3c/webrtc-pc/pull/1867/files?w=1).

jan-ivar:
I've introduced a [[FiredDirection]] slot to track exactly what direction we've fired events for. This should help with https://github.com/w3c/webrtc-pc/issues/1744, and https://github.com/w3c/webrtc-pc/issues/1706 as well when I get to it.

@taylor-b PTAL.

PR #1869, opened 2018-05-02
editorial: remove getParameters/setParameters ()

fippo:
for consistency.


aboba:
@dontcallmedom Since we have two getParameters methods (one for sender, one for receiver), would removing the parenthesis result in a warning or error?

Issue #1852, opened 2018-04-23
RTCRtpDecodingParameters


aboba:
Currently, RTCRtpDecodingParameters is defined as follows: 

dictionary RTCRtpCodingParameters {
    DOMString rid;
};

dictionary RTCRtpDecodingParameters : RTCRtpCodingParameters {
};

Some questions: 

a. Since WebRTC 1.0 API doesn't support simulcast reception, why is the rid attribute in RTCRtpDecodingParameters?
b. Does codecPayloadType belong in RTCRtpDecodingParameters?

aboba:
@taylor-b Question:  Do we need the codecPayloadType attribute on either the sender or receiver? If so, could it be read-only?

taylor-b:
It was only meant for the sender; the description mentions "This field is not used for RTCRtpReceivers."

aboba:
At the virtual interim, there were no browser vendors indicating the intention to support simulcast reception.  Resolution was to attempt to confirm this on the mailing list, and if so, to remove the rid attribute.

Issue #1840, opened 2018-04-17
What should receiver.getParameters().encodings[] return?


aboba:
Section 5.3 defines RTCRtpParameters.getParameters() as follows: 

"The getParameters() method returns the RTCRtpReceiver object's current parameters for how track is decoded. When getParameters is called, the RTCRtpReceiveParameters dictionary is constructed as follows: encodings is populated based on RIDs present in the current remote description, and whether media is actively being decoded or not. The headerExtensions sequence is populated based on the header extensions that the receiver is currently prepared to receive.  The codecs sequence is populated based on the codecs that the receiver is currently prepared to receive."

Some questions; 

a. Since WebRTC 1.0 API does not support configuration of simulcast reception, and no browsers appear to be preparing to support this, do we really expect encodings to be populated based on RIDs?   If not, should we mark simulcast reception as a feature at risk?
b. Since the active attribute is not used by receivers, should we delete the text about "whether media is actively being decoded or not"?
c. Since headerExtensions and codecs is populated based on what the receiver is prepared to receive, does this imply that they are set when SLD is called (on the Offerer) or SRD (on the Answerer)?

alvestrand:
I think it should return what's currently described in section 5.3:

The codecs sequence is populated based on the codecs that the receiver is currently prepared to receive.

NOTE
Both the local and remote description may affect this list of codecs. For example, if three codecs are offered, the receiver will be prepared to receive each of them and will return them all from getParameters. But if the remote endpoint only answers with two, the absent codec will no longer be returned by getParameters as the receiver no longer needs to be prepared to receive it.

alvestrand:
May 2018 VI decision: Delete the vestigal text.

PR #1876, opened 2018-05-23
Mark simulcast reception as a feature at risk

aboba:
Fix for Issues https://github.com/w3c/webrtc-pc/issues/1840 and https://github.com/w3c/webrtc-pc/issues/1852

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -30,7 +30,7 @@
 
     <p>To go into Proposed Recommendation status, the group expects to demonstrate implementation of each feature in at least two deployed browsers, and at least one implementation of each optional feature. Mandatory feature with only one implementation may be marked as optional in a revised Candidate Recommendation where applicable.</p>
 
-    <p>The following features are marked as risk:</p>
+    <p>The following features are marked as at risk:</p>
     <ul>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -30,9 +30,10 @@
 
     <p>To go into Proposed Recommendation status, the group expects to demonstrate implementation of each feature in at least two deployed browsers, and at least one implementation of each optional feature. Mandatory feature with only one implementation may be marked as optional in a revised Candidate Recommendation where applicable.</p>
 
-    <p>The following features are marked as risk:</p>
+    <p>The following features are marked as at risk:</p>
     <ul>
       <li>The value <a href="#dom-rtcrtcpmuxpolicy-negotiate"><code>negotiate</code></a> of RTCRtcpMuxPolicy</li>
+      <li>The <code>rid</code> attribute of <code><a>RTCRtpCodingParameters</a></code></li>
```
taylor-b: It's just its presence in RTCRtpDecodingParameters that's at risk; it would still go in RTCRtpEncodingParameters

alvestrand: The two pieces of this edit are inconsistent; one says that simulcast reception is the feature at risk, the other says that the rid field (which only makes sense as an indication of simulcast reception) is the feature at risk.
I think the IDL feature at risk should be the "encodings" parameter of the RTCRTPDecodingParameters dictionary, not the "rid" parameter - on the encoding side, it's not at risk.

Issue #1885, opened 2018-06-05
Firing signalingstatechange early and synchronously invites trouble


jan-ivar:
When we set local/remote descriptions, we _"fire a simple event named signalingstatechange at connection"_—which reduces to synchronously calling JS callbacks—and we do this *before* most of the actual work.

Suppose JS calls `pc.close()` in that event/callback.
Which of the remaining steps do we do?

We've dealt with similar problems in the past (https://github.com/w3c/webrtc-pc/pull/1720) by deferring firing of events to the end of the algorithm. I think we need to do that here as well. That way it would still fire synchronously before SLD/SRD succeeds, like now.

This would change the timing somewhat, but hopefully for the better.

PR #1886, opened 2018-06-07
Fire signalingstatechange later at end of SLD/SRD algorithm like other events

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1885.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1447,12 +1447,6 @@
                     </ul>
                   </li>
                   <li>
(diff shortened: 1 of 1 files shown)


fippo:
don't we have tests for this already -- the ones that were flaky in Chrome so @henbos had to fix stuff?
(yay test, *this* is how it is supposed to work)

alvestrand:
I think the relevant tests are in https://chromium-review.googlesource.com/c/chromium/src/+/1085302

Issue #1883, opened 2018-06-04
Is close() supposed to fire state change events? Connection state change set asynchronously?


henbos:
Supposedly, calling close() implies...
- signaling state changes to ["closed"](https://w3c.github.io/webrtc-pc/#dom-rtcsignalingstate-closed), [this is explicit](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-close) but _no signalingstatechange event is fired_.
- ICE connection state changes to ["closed"](https://w3c.github.io/webrtc-pc/#dom-rtciceconnectionstate-closed), this is not explicit but destroying the ICE agent is.
- connection state changes to ["closed"](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectionstate-closed), this is not explicit but destroying the ICE agent is.

Did we simply forget to set the ICE connection state and connection state to "closed" in the spec, or is this supposed to be covered by destroying the ICE agent?

Without setting the ICE connection state and the connection state explicitly it is unclear if this is supposed to happen synchronously or asynchronously and whether or not events should fire. For example, [update ICE connection state](https://w3c.github.io/webrtc-pc/#update-ice-connection-state) is invoked _asynchronously_ [when the ICE agent indicates the state has changed](https://w3c.github.io/webrtc-pc/#rtcicetransport) and it also fires an event. But it is not clear that these steps should be executed when the ICE agent is destroyed.

Possible spec bugs:
- ICE connection state and connection state are set asynchronously, but signaling state is set synchronously.
- ICE connection state and connection state changing fires events, but signaling state changing does not.

I would think that invoking a synchronous API should set all three of these states synchronously and not fire any events.

Otherwise if it is intended that some states are updated synchronously and some are updated asynchronously it makes sense that the asynchronous changes does still fire events, but it got me a bit confused.

jan-ivar:
Looks like we decided this before: https://github.com/w3c/webrtc-pc/issues/1799#issuecomment-372490350

jan-ivar:
> signaling state is set synchronously.

Signaling state is set in SLD/SRD and tracks the state of those methods essentially. We fire lots of events synchronously in those methods before the promise resolves, so I think we need to continue doing so to maintain invariants. I filed https://github.com/w3c/webrtc-pc/issues/1885 to adjust the ordering slightly.

The other two are ICE agent dependent, so having them be asynchronous makes sense to me.

> it is not clear that these steps should be executed when the ICE agent is destroyed.

[...]

PR #1887, opened 2018-06-07
Have pc.close() set connection states to "closed".

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1883.


alvestrand:
(too late comment)
are there tests that need updating for this?

Issue #1891, opened 2018-06-14
addTransceiver should throw on closed peer connection.


jan-ivar:
The [addTransceiver](http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtransceiver) method never checks [[isClosed]] the way e.g. [addTrack](http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtrack) does. Seems like a simple omission.

PR #1892, opened 2018-06-14
Have addTransceiver throw on closed peer connection.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1891.


fippo:
I would place it slightly further down since the check is done after checking all the input variables in addTrack

PR #1902, opened 2018-06-19
Adding note about potential consequences of stopping a transceiver.

### LINKED ISSUE #1858: What happens when an answerer stops a transceiver that others are "bundled" on?
The BUNDLE spec says:

> When an answerer wants to reject a bundled "m=" section in an answer, it MUST first check the following criterium:
> 
> -In the corresponding offer, an offerer BUNDLE address:port (previously selected [Section 8.3.1] or newly suggested [Section 8.5.1]) has been assigned to the "m=" section by the offerer.
> 
> If the criterium above is fulfilled the answerer can not reject the "m=" section in the answer (unless the answerer rejects each bundled "m=" section within the BUNDLE group).  The answerer can either reject the whole offer, reject each bundled "m=" section within the BUNDLE group, or keep the "m=" section within the BUNDLE group in the answer and later create an offer where the "m=" section is disabled within the BUNDLE group [Section 8.5.4].

Or, in [this PR](https://github.com/cdh4u/draft-sdp-bundle/pull/59/files) which changes the terminology:

> When an answerer wants to reject a bundled "m=" section in an answer, it MUST first check the following criterium:
> 
> - In the corresponding offer, the "m=" section is the offerer tagged "m=" section.
> 
> If the criterium above is fulfilled the answerer can not reject the "m=" section in the answer. The answerer can either reject the whole offer, reject each bundled "m=" section within the BUNDLE group, or keep the "m=" section within the BUNDLE group in the answer and later create an offer where the "m=" section is disabled within the BUNDLE group.

Either way, the point is the same. If you have an established bundle group, "a=BUNDLE audio video", the answerer can't reject the "audio" section without rejecting "video" as well.

So: how should this play with `RTCRtpTransceiver.stop`, specifically when called in the "have-remote-offer" state? Should stopping the audio transceiver automatically cause the video transceiver to be stopped as well? Or should the answerer keep the audio m= section (as "inactive", maybe?), then stop it in a subsequent offer triggered by `onnegotiationneeded`?

And does this require JSEP changes?

taylor-b:
Fixes #1858.

If a transceiver is stopped in between applying a remote offer and
creating an answer, and it's associated with the "offerer tagged" m=
section (the one listed first in "a=group:BUNDLE"), all other m=
sections in the BUNDLE group will be rejected as well.

The actual normative text should go in JSEP (see
https://github.com/rtcweb-wg/jsep/pull/849), but it's surprising enough
that it seems worth calling out in a note.


taylor-b:
What other surprising situations? May be nice to call them out explicitly. I added "when RTCSignalingState is stable", though.

webrtc.html
```diff
@@ -685,7 +685,7 @@ <h4><dfn>RTCSignalingState</dfn> Enum</h4>
               </tr>
               <tr>
                 <td><dfn data-idl><code>stable</code></dfn></td>
-                <td>There is no offer&shy;answer exchange in progress. This is
+                <td>There is no offer/answer exchange in progress. This is
```
alvestrand: while picking nits, can we add a comma between "initial state" and "in which case" below?

alvestrand:
There may be more follow-on work to get the description right, but let's add the note first.

aboba:
Re-based and merged.

PR #1904, opened 2018-06-19
Do "process removal of remote track" steps when stopping a transceiver.

### LINKED ISSUE #1812: Does direction matter for rejected m= sections?

taylor-b:
Fixes #1812.

This is effectively very similar to #1845, the only difference being
that this PR will remove the track as soon as the transceiver is
stopped, rather than later when setLocalDescription is called.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1596,6 +1596,12 @@
                             <a>media description</a>.</p>
                           </li>
                           <li>
+                            <p>If the <a>media description</a> is rejected, and
+                            <var>transceiver</var> is not already stopped, <a>stop
+                            the RTCRtpTransceiver</a> <var>transceiver</var> given
+                            <var>removeList</var>.</p>
+                          </li>
+                          <li>
                             <p>If <var>transceiver</var>'s <a>[[\Stopped]]</a> slot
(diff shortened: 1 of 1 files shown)


henbos:
Relates to #1919 which has list discussion needed

alvestrand:
Opinion in the editing team is that when a transceiver is stopped, the track should end, but there's no reason to remove it from the stream.

jan-ivar:
Based on what I mentioned in https://github.com/w3c/webrtc-pc/pull/1845#issuecomment-384719900 I think I'm still inclined to not do this.

henbos:
Also we typically avoid firing events on synchronous method calls (such as close()).

PR #1909, opened 2018-06-20
Update a receiver's set of remote streams when "a=msid" lines change.

### LINKED ISSUE #1609: Partial removal of a track?
It is possible for a sender to have multiple streams, resulting in the other endpoint having a receiver with multiple streams.

Let's say we signal that track1 belongs to stream1 and stream2. Is it OK to then signal that the track belongs to another set of streams - let's say stream1, but no longer stream2? In this case, it should be removed from one stream and the receiver's set of associated streams should be updated, but the track should not be muted.

The spec is currently written to "process the removal of a remote track" which set the receiver's associated streams and mutes the track (if not already muted).

If this is to be supported, I believe we need to either:
- If this counts as the removal of a track, the description should be updated to say only mute if the receiver's set of associated streams is empty.
- If this does not count as the removal of a track, we need to add a step to update the set of associated streams of receivers that are not removed.

I'm no SDP guru, is my understanding of this correct?

taylor-b:
Fixes #1609.

This PR updates the `[[AssociatedRemoteMediaStreams]]` slot in every
`setRemoteDescription` call, even if the direction isn't changing, and
also adds:

- A `getStreams()` method to `RTCRtpReceiver`, just returning the value
  of the internal slot.
- A `streamsupdated` event, to tell when the set of stremas changes.
  Fired at the end of `setRemoteDescription` like everything else.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1704,6 +1704,17 @@
                             <var>addList</var>, and <var>trackEvents</var>.</p>
                           </li>
                           <li>
+                            <p>If <var>direction</var> is <code>"sendrecv"</code> or
+                            <code>"recvonly"</code>, and
+                            <var>transceiver</var>'s
+                            <a>[[\FiredDirection]]</a> slot
+                            is also <code>"sendrecv"</code> or <code>"recvonly"</code>,
+                            <a>set the associated remote streams</a> of
+                            <var>transceiver</var>'s <a>[[\Receiver]]</a> given the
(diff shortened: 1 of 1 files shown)


alvestrand:
There's new API in here, so we need list discussion, and perhaps a new VI topic.

alvestrand:
We might fire the track event again, rather than adding a setstreams event. (Jan-Ivar's comment).

PR #1911, opened 2018-06-20
Adding note about the relationship between tracks of senders/receivers.

### LINKED ISSUE #1169: RTP Media API section should describe how tracks for senders/receivers relate to each other.
Previously, there was a 1:1 mapping between tracks added to a PeerConnection on one side and received on the other side.

The track IDs also matched (https://github.com/w3c/webrtc-pc/issues/1128).

And when you called `removeTrack`, the track on the other side would be removed and ended (https://github.com/w3c/webrtc-pc/issues/1161).

None of these things are true any more, due to the API changes to allow for "early media", transceivers, `replaceTrack`, etc.

The fundamental difference is that tracks on senders and receivers are not symmetrical. A track attached to a sender represents "a source of media being sent", whereas a track attached to a receiver represents "a stream of media received from an `RTCRtpSender` on the remote side`.

This clearly has been a source of confusion, and it's not explained in any unified way. So I'd suggest adding a paragraph to the RTP Media API section explaining this model.

taylor-b:
Fixes #1169.

Clarifies that there's not a 1:1 relationship between tracks sent by
one PeerConnection and received by the other (due to `replaceTrack`),
and the IDs are not guaranteed to match. So it's better to think of
a 1:1 relationship between senders on one side and receivers on the
other side, rather than tracks.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1,4 +1,4 @@
-e<!DOCTYPE html>
+<!DOCTYPE html>
 <html lang="en">
 <head>
   <meta charset="utf-8">
--- a/webrtc.html
+++ b/webrtc.html
@@ -4681,6 +4681,25 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     added to an <code>RTCPeerConnection</code>, result in signaling; when this
     signaling is forwarded to a remote peer, it causes corresponding tracks to
     be created on the remote side.</p>
+    <p class="note">There is not an exact 1:1 correspondence between tracks sent
+    by one <code>RTCPeerConnection</code> and received by the other. For one,
+    in many cases the ID of a track sent by one side will not match the ID of
+    the corresponding track received on the other side; whether the IDs end up
+    matching depends on the relative ordering of calls to
+    <code><a data-link-for="RTCPeerConnection">addTrack</a></code>,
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4681,6 +4681,25 @@ <h2>RTP Media API</h2>
...
+    <code><a data-link-for="RTCRtpSender">replaceTrack</a></code> is called,
+    changing the track sent by an <code><a>RTCRtpSender</a></code>, no new
+    track will be created on the receiver side; the corresponding
+    <code><a>RTCRtpReceiver</a></code> will only have a single track,
+    potentially representing multiple sources of media stiched together. Thus
+    it's more accurate to think of a 1:1 relationship between
+    <code><a>RTCRtpSender</a></code>s on one side and
+    <code><a>RTCRtpReceiver</a></code>s on the other side, matching them using
+    the <code><a>RTCRtpTransceiver</a></code>'s <code><a
+    data-link-for="RTCRtpTransceiver">mid</a></code> if necessary.</p>
```
henbos: The transceiver on one side maps 1:1 with the transceiver on the other end with the same "mid". The sender's track is what is sent to the receiver's track. The sender's track can be replaced or nulled, which affects what media shows up in the receiver's track. The receiver's track never changes.

PR #1923, opened 2018-07-05
RTCRtpReceiver.isActive added.

### LINKED ISSUE #1920: Unable to figure out which receivers are "active".
Based on discussions at #1890, we are unable to figure out which receivers are "active" or not after setRemoteDescription().
```
1. const offer = await caller.createOffer();
2. await caller.setLocalDescription(offer);
3. await callee.setRemoteDescription(offer);
4. const answer = await callee.createAnswer();
5. await callee.setLocalDescription(answer);
6. await caller.setRemoteDescription(answer);
```
The recever's currentDirection is updated at step 5 (half-way through the answer). Likely the application would like to know at 3 when the ontrack events fire.

Today the application is forced to piece together from ontrack and other events what the state of the receivers really are. It's confusing having a pc with a set of receivers and not be able to figure out what their state are without wiring up events and listening to history.

We no longer support getRemoteStreams() but it should be possible to do the equivalent, i.e. "get active receivers".

henbos:
Fixes #1920

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6673,6 +6673,7 @@ async function updateParameters() {
     readonly        attribute MediaStreamTrack  track;
     readonly        attribute RTCDtlsTransport? transport;
     readonly        attribute RTCDtlsTransport? rtcpTransport;
+    readonly        attribute boolean           isActive;
     static RTCRtpCapabilities          getCapabilities (DOMString kind);
     RTCRtpReceiveParameters            getParameters ();
     sequence&lt;RTCRtpContributingSource&gt;    getContributingSources ();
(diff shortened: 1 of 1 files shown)


henbos:
See [comment](https://github.com/w3c/webrtc-pc/issues/1920#issuecomment-402706696) for background.

webrtc.html
```diff
@@ -6726,6 +6727,56 @@ <h2>Attributes</h2>
               <p>On getting, the attribute MUST return the value of the
               <a>[[\ReceiverRtcpTransport]]</a> slot.</p>
             </dd>
+            <dt><code>isActive</code> of type <span class=
+            "idlAttrType"><a>boolean</a></span>, readonly</dt>
+            <dd>
+              <p>The <dfn data-idl><code>isActive</code></dfn> attribute
+              indicates whether the receiver is considered active. An active
+              receiver is one that the remote endpoint has offered to send to,
+              and we have not (yet) rejected, or one that we have offered the
```
stefhak: Unclear who "we" is - rephrase?

alvestrand:
We might not need it given discussion in bug.
List discussion needed.

PR #1939, opened 2018-07-17
Reject promise at replaceTrack if TRX is stopped.

### LINKED ISSUE #1938: Promise should be rejected if transceiver is stopped
Looking at [replaceTrack](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack), it seems to me that if transceiver has been stopped when reaching the point of `Queue a task that runs the following steps` (6.4.4 in [replaceTrack](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack)) the promise should be rejected with an `InvalidStateError`. Came up when looking at #1389.

stefhak:
Is the `abort these steps` part needed? If so, should is also be part of the step when `stopped` is first checked (6.1)?

Close #1938.

***
<a href="https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1939.html" title="Last updated on Jul 18, 2018, 8:24 AM GMT (3fb27d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1939/57235b3...stefhak:3fb27d1.html" title="Last updated on Jul 18, 2018, 8:24 AM GMT (3fb27d1)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5969,7 +5969,10 @@ async function updateParameters() {
                           <ol>
                             <li>
                               <p>If <var>transceiver</var>'s <a>[[\Stopped]]</a> slot is
-                              <code>true</code>, abort these steps.</p>
+                              <code>true</code>, return a promise <a>rejected</a>
+                              with a newly <a data-link-for="exception"
+                              data-lt="create"> created</a>
+                              <code>InvalidStateError</code>, and abort these steps.</p>
                             </li>
                             <li>
                               <p>Set <var>sender</var>'s <code><a data-link-for=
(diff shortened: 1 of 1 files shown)


alvestrand:
After our previous discussion of "abort these steps", I added an explicit statement up in the terminology section that "throw an XXX error" also implies aborting the steps in question. But this is returning an exception, not throwing an error. I guess it's reasonable to be explicit about it.

CL looks good.

stefhak:
If we think having `abort these steps` makes sense here, it should probably be added to the previous step as well. I'll update the PR and see what people thinks.

alvestrand:
https://infra.spec.whatwg.org/#algorithm-control-flow - we should add "return" to the terminology list along with "throw", with a reference to [[INFRA]] - it terminates processing according to that spec.
So "abort these steps" is superfluous.

alvestrand:
There seems to have been a generic advice from WebIDL people that leaving promises unclosed is a perfectly OK thing to do, and we should just do it; if we revisit the issue, we should apply the principle across the API, not piecemeal.

PR #1942, opened 2018-07-22
reject with InvalidAccessError when mux policy does not match

### LINKED ISSUE #1235: setRemoteDescription error when rtcpMuxPolicy is "require" and SDP is missing "a=rtcp-mux"
in the constructor NotSupportedError is thrown on an attempt to use a negotiate in a browser which does not support it. If such a browser attempts to set a remote RTCSessionDescription should it reject with a specific error? The current error behaviour would be a RTCError from step 3 of that algorithm.

Also, the setConfiguration algorithm also has this:
> If configuration.rtcpMuxPolicy is set and its value differs from the connection's rtcpMux policy, throw an InvalidModificationError.

If the browser does not implement 'negotiate' and an attempt is made to use setConfiguration to change this wouldn't it be more consistent to return NotSupportedError too?

This would also allow streamlining the constructor steps by merging step 3 (mux policy) into step 5 (set the configuration). From an implementation perspective i'd prefer to throw the error from a single place.
Even if we want different errors that would be possible by tweaking the language above.

@taylor-b you might have an opinion already

fippo:
fixes #1235

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-pc/pull/1942.html" title="Last updated on Aug 2, 2018, 7:45 AM GMT (5443149)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1942/dd52fdc...fippo:5443149.html" title="Last updated on Aug 2, 2018, 7:45 AM GMT (5443149)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1328,6 +1328,15 @@
                     the syntax error was detected) and abort these steps.</p>
                   </li>
                   <li>
+                    If <var>description</var> is set as a remote description,
+                    the <var>connection</var>'s <code><a>RTCRtcpMuxPolicy</a></code>
+                    is <code><a data-link-for="RTCRtcpMuxPolicy">require</a></code>
+                    and the remote description does not use RTCP mux,
+                    then <a>reject</a> <var>p</var> with a newly <a data-link-for="exception"
+                    data-lt="create">created</a>
+                    <code>NotSupportedError</code> and abort these steps.
+                  </li>
+                  <li>
                     <p>If the content of <var>description</var> is invalid,
                     then <a>reject</a> <var>p</var> with a newly
                     <a data-link-for="exception" data-lt="create">created</a>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1334,7 +1334,7 @@
                     and the remote description does not use RTCP mux,
                     then <a>reject</a> <var>p</var> with a newly <a data-link-for="exception"
                     data-lt="create">created</a>
-                    <code>NotSupportedError</code> and abort these steps.
+                    <code>InvalidAccessError</code> and abort these steps.
                   </li>
                   <li>
                     <p>If the content of <var>description</var> is invalid,


fippo:
needs test. Basically:
* create a peerconnection with muxpolicy
* set a short sdp without mux
* expect error to be thrown

webrtc.html
```diff
@@ -1327,6 +1327,15 @@ <h4>Set the RTCSessionDescription</h4>
...
                     the syntax error was detected) and abort these steps.</p>
                   </li>
+                  <li>
+                    If <var>description</var> is set as a remote description,
+                    the <var>connection</var>'s <code><a>RTCRtcpMuxPolicy</a></code>
+                    is <code><a data-link-for="RTCRtcpMuxPolicy">require</a></code>
+                    and the remote description does not use RTCP mux,
+                    then <a>reject</a> <var>p</var> with a newly <a data-link-for="exception"
+                    data-lt="create">created</a>
+                    <code>NotSupportedError</code> and abort these steps.
```
alvestrand: What is the argument for using NotSupportedError and not either InvalidAccessError (thrown for all other invalid SDP) or OperationError (thrown for all other errors)?

fippo: this is the same error as in "4.4.1.7 Set the configuration" when trying to set 'negotiate' in the constructor and this is unsupported.
I do not think this is invalidaccesserror as the SDP is valid. OperationError would wfm too

henbos: Or InvalidStateError, which we throw when trying to perform an operation that is otherwise supported but while the PC is in an invalid state (such as being closed).

But InvalidAccessError is probably the error most consistent with other usages in the spec (rejected SDP, invalid parameters, invalid initEncodings, invalid use of addTrack/removeTrack).

NotSupportedError on the other hand is typically used when what you're doing is _not implemented by the browser_, e.g. non-muxed RTCP, url scheme name, keygenAlgorithm. This usage makes it sound like non-muxed RTCP is not implemented (like the other usage) as opposed to PC not being configured that way.

fippo: you can't configure the PC to actually do non-muxed -- mux is either negotiate or require. And I think the usage of NotSupportedError is exactly what you say hbos :-)

stefhak: Given the comments above it feels like we should go with InvalidAccessError. Would you agree @fippo ?

fippo:
```
var pc = new RTCPeerConnection({rtcpMuxPolicy: 'require'}); pc.setRemoteDescription({type: 'offer', sdp: "v=0\r\no=- 166855176514521964 2 IN IP4 127.0.0.1\r\ns=-\r\nt=0 0\r\nm=audio 9 UDP/TLS/RTP/SAVPF 111\r\nc=IN IP4 0.0.0.0\r\na=rtcp:9 IN IP4 0.0.0.0\r\na=ice-ufrag:someufrag\r\na=ice-pwd:somelongpwdwithenoughrandomness\r\na=fingerprint:sha-256 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00\r\na=setup:actpass\r\na=mid:audio1\r\na=sendonly\r\na=rtcp-rsize\r\na=rtpmap:111 opus/48000/2\r\n"}).catch(e => console.error(e.name))
```
throws InvalidAccessError. Not changing implementations is a good argument :-)

fippo:
while implementing this: we don't care about the rather weird case where only one m-line does not do rtcp-mux?

If we do this would actually mean the m-line should be rejected instead.

Issue #1921, opened 2018-06-29
RTCRtpReceiver.getStreams() and RTCRtpSender.getStreams()


henbos:
The sender's set of streams is negotiated and shows up at the other end as the receiver's [[AssociatedRemoteMediaStreams]] slot. This is exposed in the ontrack event, but without listening to that event, there is no way for the application to know which streams belong to a receiver.

As of recently we even have RTCRtpSender.setStreams(), but there is no RTCRtpSender.getStreams(). Why support controlling something but not checking its current value, forcing the application to do additional bookkeeping.

As a side note, not having RTCRtpSender/RTCRtpReceiver.getStreams() makes shimming getLocalStreams()/getRemoteStreams() difficult. We may not care about supporting legacy APIs but we should care about supporting legitimate use cases highlighted by them. It's not unreasonable for an application to want to know about their streams.

henbos:
Based on discussion on #1890

alvestrand:
I support adding receiver.getStreams(). It makes supporting use cases that made use of the legacy pc.getRemoteStreams() easy.

henbos:
I went for readonly attributes instead of getters since it just dumps the value of an internal slot without any special logic.

I also went with streamIds instead of streams on the RTCRtpSender because that's what the internal slot uses, but it's odd to have a setter take MediaStreams and the getter return IDs. We could consider changing [[AssociatedMediaStreamIds]] back into a streams slot to be consistent with addTrack(), addTransceiver() and setStreams().

henbos:
There seem to be a lack of interest in RTCRtpReceiver.streams and RTCRtpSender.streamsId, does anyone care?

henbos:
Should we close this?

murillo128:
Why is this closed? 

I just needed to get the senders streams and got no way of doing so. IMHO given that the info is already there and correctly updated, we should expose it in the API.

henbos:
I remember a disinterest in adding new APIs that have to do with streams, but we can reopen the discussion if there is interest.

murillo128:
In my case I am doing an SDP-less negotiation on the wire, meaning that the O/A is done locally  and only track/stream info is transmitted on the wire.

In order to do that I needed to retrieve the stream info of the rtpsenders in order to be able to propagate it to the other end. 

The internal slot is already there, so just exposing it via a read only attribute would work for me.

henbos:
[...]

PR #1944, opened 2018-07-25
RTCRtpReceiver.streams

henbos:
Partial fix for #1921

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/1944.html" title="Last updated on Jul 25, 2018, 4:01 PM GMT (7b70cd3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1944/a22c04e...henbos:7b70cd3.html" title="Last updated on Jul 25, 2018, 4:01 PM GMT (7b70cd3)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6703,6 +6703,7 @@ async function updateParameters() {
     readonly        attribute MediaStreamTrack  track;
     readonly        attribute RTCDtlsTransport? transport;
     readonly        attribute RTCDtlsTransport? rtcpTransport;
+    readonly        attribute sequence&lt;MediaStream&gt; streams;
     static RTCRtpCapabilities?         getCapabilities (DOMString kind);
     RTCRtpReceiveParameters            getParameters ();
     sequence&lt;RTCRtpContributingSource&gt;    getContributingSources ();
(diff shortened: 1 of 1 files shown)


jan-ivar:
@henbos isn't this overtaken by events, given https://github.com/w3c/webrtc-pc/pull/1909#issuecomment-406294846 and https://github.com/w3c/webrtc-pc/pull/1947?

henbos:
If we have a track attribute we should have a streams attribute. Why obscure this part of the state to only be seen in events firing?

Whether you want to shim getLocal/RemoteStreams() or you want to do the 1.0 equivalent of those APIs - you want to iterate the PCs streams - this allows you to do that without keeping a map of receiver -> streams that you update on events firing and on any edge case where the slot might be cleared.

jan-ivar:
> Whether you want to shim getLocal/RemoteStreams() or you want to do the 1.0 equivalent of those APIs 

@henbos The spec has moved away from streams as an API surface to RTCPeerConnection, deliberately, so we should discourage such shimming, not encourage it. E.g. we replaced the `addstream` event with the `track` event, and removed the methods you mention years ago.

There is no "1.0." of the spec yet.  People are free to shim whatever they want, but `getLocal/RemoteStreams()` are from an era where RTCPeerConnection in some implementations picked up on local changes to those streams. This no longer matches how things work.

Thus they confuse where the new API line is dawn, as to which objects RTCPeerConnection is aware of and reacts to. It's the most leaky of our [leaky abstractions](https://blog.mozilla.org/webrtc/the-evolution-of-webrtc/).

Streams are even optional in the new API, and they don't support early media. E.g. we can do:

```js
video.srcObject = new MediaStream([transceiver.receiver.track]);
await new Promise(r => transceiver.receiver.track.onunmute = r);
await video.play();

// Negotiate RTCPeerConnection here
```

So the only purpose really of the `track` event IMHO is to advertise the stream associations.

Issue #1919, opened 2018-06-28
Investigate stopping SRD and createAnswer


henbos:
As pointed out in https://github.com/w3c/webrtc-pc/pull/1902/commits/6ae910eafcbf3df624703c2a8104a0b39d6e029f stopping a transceiver when in a non-stable state can have unpredictable consequences.

Investigate what to do to avoid shooting yourself in the foot.

henbos:
I need to discuss this further with you guys, there's things I don't understand yet. Me wrapping my head around bundling:

Bundling is defined in [sdp-bundle-negotiation](https://tools.ietf.org/html/draft-ietf-mmusic-sdp-bundle-negotiation-52), JSEP uses a single bundle group/transport for bundled m= sections ([jsep#4.1.1](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-4.1.1)), but can have multiple transports for non-bundled m= sections.
- `balanced`: the first transceiver (m= section) _of each type_ (audio/video/application) defines the transport parameters, all subsequent transceivers are bundle-only (bundled with the first one). Non-bundle-aware endpoints would still be able to send & receive 1 audio, 1 video and 1 application.
- `max-compat`: all transceivers (m= sections) define their own transport parameters. So one transport per transceiver?
- `max-bundle`: the first transceiver (m= section) is the only one that defines transport parameters, all subsequent transceivers regardless of type are bundle-only. Non-bundle-aware endpoints would only be able to send & receive 1 audio, 1 video OR 1 application: the first one.

Then there is the "offerer tagged m= section", which is the m= section whose BUNDLE attributes apply to all other m= sections within that bundle group.

**Questions due to my lack of understanding**:
Q1. What are BUNDLE attributes?
Q2. Is "offerer tagged" different from which transceiver defines transport parameters? Can a bundle-only m= section be the "offerer tagged" m= section?

Rejecting the [offerer tagged m= section requires rejecting all m= section](https://tools.ietf.org/html/draft-ietf-mmusic-sdp-bundle-negotiation-52#section-7.3.3). This may imply:
1. The application is interested in knowing which transceiver (m= section) is "offerer tagged".
2. The application is interested in controlling which transceiver (m = section) is "offerer tagged".
3. The application is interested in knowing which transceivers are bundled. Today this can be controlled, and thus deduced from, the bundle-policy passed at construction. Exposing the bundle groups as a set of transceivers would make this a whole lot more clear though.

And possibly,
[...]

PR #1946, opened 2018-07-25
Add warning to RTCRtpTransceiver.stop(). Stop rejected transceivers at SLD().

henbos:
** Superseded by #2001 **

---

Related to #1919.

Because of bundling, "m=" sections may have been rejected by createOffer() or createAnswer() as a result of some transceiver having been stopped, affecting all other "m=" sections in that bundle. This change make sure that setLocalDescription() updates associated transceivers to be stopped if their "m=" section have been rejected. Otherwise the JSEP transceiver and the webrtc-pc don't have a matching state.

Because of this surprising behavior, a note is added to RTCRtpTransceiver.stop().

The issue about stop() being unpredictable is still unresolved.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/1946.html" title="Last updated on Oct 4, 2018, 11:30 AM GMT (e0be50b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1946/a22c04e...henbos:e0be50b.html" title="Last updated on Oct 4, 2018, 11:30 AM GMT (e0be50b)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1654,6 +1654,11 @@
                               </li>
                             </ol>
                           </li>
+                          <li>
+                            <p>If the <a>media description</a> is rejected, and
+                            <var>transceiver</var> is not already stopped, <a>stop
+                            the RTCRtpTransceiver</a> <var>transceiver</var>.</p>
+                          </li>
                         </ol>
                       </li>
                     </ol>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1654,6 +1654,11 @@ <h4>Set the RTCSessionDescription</h4>
                               </li>
                             </ol>
                           </li>
+                          <li>
+                            <p>If the <a>media description</a> is rejected, and
+                            <var>transceiver</var> is not already stopped, <a>stop
+                            the RTCRtpTransceiver</a> <var>transceiver</var>.</p>
+                          </li>
```
jan-ivar: This is a normative change, and it appears in the wrong place I think, since it would seem to affect SLD(offer) here. Should it at least be inside the sub-steps directly above?

alvestrand: Rejecting a media section and then leaving the transceiver unstopped is certainly the wrong thing to do in my opinion. I haven't tried to reformat this to see where it's actually indented to (it's pretty deep).

henbos:
Was the decision to reject if stopping between offer and answer? I need to revisit this one.

alvestrand:
Approach (explain weirdness) accepted at interim, hbos to check for comments.

PR #1947, opened 2018-07-25
Fire track event in response to remote sender.setStreams() changes.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/1609, https://github.com/w3c/webrtc-pc/issues/1744 and https://github.com/w3c/webrtc-pc/issues/1921.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/1947.html" title="Last updated on Jul 25, 2018, 9:07 PM GMT (1e18d26)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1947/a22c04e...jan-ivar:1e18d26.html" title="Last updated on Jul 25, 2018, 9:07 PM GMT (1e18d26)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1642,9 +1642,21 @@
                                 and <var>transceiver</var>'s
                                 <a>[[\FiredDirection]]</a> slot is either
                                 <code>"sendrecv"</code> or <code>"recvonly"</code>,
-                                <a>process the removal of a remote track</a> for
-                                the <a>media description</a>, given <var>transceiver</var>,
-                                <var>removeList</var>, and <var>muteTracks</var>.</p>
+                                then run the following steps:</p>
+                                <ol>
+                                  <li>
+                                    <p><a>Set the associated remote streams</a> given
+                                    <var>transceiver</var>.<a>[[\Receiver]]</a>,
+                                    an empty list, another empty list, and
+                                    <var>removeList</var>.</p>
+                                  </li>
+                                  <li>
+                                    <a>process the removal of a remote track</a>
+                                    for the <a>media description</a>, given
+                                    <var>transceiver</var> and
+                                    <var>muteTracks</var>.</p>
+                                  </li>
+                                </ol>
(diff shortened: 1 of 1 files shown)


jan-ivar:
To clarify this PR, it does two things:

It cleans up what I mention in https://github.com/w3c/webrtc-pc/issues/1744#issue-290620037:
> I think we should lift the call to [set the associated remote streams](http://w3c.github.io/webrtc-pc/#set-associated-remote-streams) out one level, basically always call it with the msids (or empty list if send-bit is off).

And it fires the track event if a new stream association is detected.

webrtc.html
```diff
@@ -1714,14 +1726,27 @@ <h4>Set the RTCSessionDescription</h4>
...
+                          </li>
+                          <li>
+                            <p><a>Set the associated remote streams</a> given
+                            <var>transceiver</var>.<a>[[\Receiver]]</a>,
+                            <var>msids</var>, <var>addList</var>, and
+                            <var>removeList</var>.</p>
+                          </li>
+                          <li>
+                            <p>If the previous step increased the length of
+                            <var>addList</var>, or if <var>transceiver</var>'s
```
henbos: If "ontrack" is the catch-all "streams changed" event, then we need to process the addition of a remote track even if only the removeList increased in length.

webrtc.html
```diff
@@ -1714,14 +1726,27 @@ <h4>Set the RTCSessionDescription</h4>
...
                             <a>[[\FiredDirection]]</a> slot
                             is neither <code>"sendrecv"</code> nor <code>"recvonly"</code>,
                             <a>process the addition of a remote track</a> for
```
henbos: "process the addition of a remote track" now seems like an inadequate name.
How about "process the addition of a track event"? Inlining "add a new RTCTrackEvent to trackEvents" also works.

Issue #1834, opened 2018-04-11
Handling of invalid values in RTCRtpSender.setParameters()


Orphis:
Currently, a `RangeError` is reported when calling `setParameters()` with `scaleResolutionDownBy` set to a value lower than 1.0.
Shouldn't the same be done for other invalid values?

- `maxFramerate` is a double, which can be negative, very small or just 0. Huge values shouldn't be a problem.

- `maxBitrate` is an integer than can have be lower than some encoder dependent value. For Chrome's implementation, there is a 30000 bps minimum I believe.

- `ptime` can be bigger than `maxptime` defined in the SDP. This one is probably not a problem since it is only a preferred value and implementations just have to do their best.

- `scaleResolutionDownBy` can be a huge value. What happens if it forces frames to be smaller than what the encoder allows or it ends up with a fractional size smaller than 1x1?

alvestrand:
Can someone find a reference for RangeError?
It seems like it was something we defined at one point (see changelog), but the definition is lost.

alvestrand:
Answering my own question:

https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror

stefhak:
Based on the slide in https://www.w3.org/2011/04/webrtc/wiki/images/5/54/WebRTCWG-2018-05-22.pdf it seems the proposal is to:
- make `scaleResolutionDownBy` and `maxFramerate` unsigned
- check lower boundary on `maxFramerate` and `maxBitrate`
- check upper boundary on `scaleResolutionDownBy`
- check boundaries on `ptime`

To me it seems a bit hard to check the boundaries. At least for `maxBitrate` this can depend a lot on codec used and on content type (static vs quickly moving), and codec used can change e.g. due to an offer/answer exchange in the middle of a session (and content can also change).

My proposal would be to first do the changes of type to unsigned.

aboba:
@stefhak maxFramerate is a double because there are uses for non-integer maximum frames rates (e.g. 7.5 frames/second for one stream, 15 frames/second for another).  Similar arguments can be made for scaleResolutionDownBy,

aboba:
@stefhak Can we close this now?

stefhak:
@Orphis in my view we can close this one now. A checks that `maxFramerate` is not negative has been added, while the other checks discussed in https://github.com/w3c/webrtc-pc/issues/1834#issue-313346219 seems to not make sense (based on the discussion here). Any other opinion?

PR #1948, opened 2018-07-30
Made maxFramerate and scaleResolutionDownBy unsigned long (were: doub…

stefhak:
…le).

This is for #1834. I changed to `unsigned` as proposed in https://www.w3.org/2011/04/webrtc/wiki/images/5/54/WebRTCWG-2018-05-22.pdf, however I'm not sure what more that can be reliable checked, so leaving it at that for now.

***
<a href="https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1948.html" title="Last updated on Jul 30, 2018, 7:56 AM GMT (98a8c64)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1948/f385034...stefhak:98a8c64.html" title="Last updated on Jul 30, 2018, 7:56 AM GMT (98a8c64)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6238,8 +6238,8 @@ async function updateParameters() {
              RTCPriorityType     priority = "low";
              unsigned long       ptime;
              unsigned long       maxBitrate;
-             double              maxFramerate;
-             double              scaleResolutionDownBy;
+             unsigned long       maxFramerate;
+             unsigned long       scaleResolutionDownBy;
 };</pre>
         <section>
           <h2>Dictionary <code><a>RTCRtpEncodingParameters</a></code>


aboba:
@stefhak I believe that maxFramerate and scaleResolutionDownBy have already been implemented as double; changing this to unsigned long could be problematic.

stefhak:
@aboba aha. I read  https://www.w3.org/2011/04/webrtc/wiki/images/5/54/WebRTCWG-2018-05-22.pdf along with #1834 as it was agreed to do this change.
Anyway, what does existing implementations do if a negative maxFramerate is supplied? (There is already a check for scaleResolutionDownBy being >= 1.0)

aboba:
@stefhak a negative maxFramerate will result in promise rejection.

alvestrand:
maxFramerate has to be a double, not an integer type.
scaleResolutionDownBy needs to be a double too.
Instead write restrictions in prose.

Issue #1619, opened 2017-10-05
canInsertDTMF transitions not specified


aboba:
There is no text describing when `canInsertDTMF` transitions to `true`.

fluffy:
So I think the value of canInsertDTMF is based on if the remote SDP negotiated it or not. Does that sounds correct to you ?

aboba:
@fluffy  This issue is about specifying the conditions under which canInsertDTMF is true.  For example, a combination of the following: 

1. It's an audio sender. 
2. "telephone-event" is negotiated.
3.  The sender can send: 
      a. current direction is not send/recv or recv-only?
      b. sender.track is non-null?

alvestrand:
After chewing through this in real code for a while, I think that the conditions for canInsertDTMF = true (in addition to "track" being present and being an audio track) are:
1) Negotiation done
2) Direction is sendrecv or sendonly
3) DTMF-capable codec ("telephone-event") negotiated
4) PeerConnection in "connected" state
SInce the last point is a state transition that we have an event for, it may make sense to hang the value change off the text that triggers the event.

fluffy:
So if by negation done, you mean we have an offer and a PR-answer or answer, that sounds about right to me. The one case I am thinking of is you have called 1-800-go-fedex and you might only have a pr-answer but you still want to be able to send DTMF for the the IVR.

alvestrand:
Offer + pr-answer would work fine, I think. I want to make sure I can have a conformant implementation that is only able to send DTMF after the connection has been established (which also requires completing the DTLS handshake).

fluffy:
Yep, that sounds right to me.

jan-ivar:
This is a good idea. Users shouldn't have to [poll on `canInsertDTMF`](https://github.com/webrtc/adapter/pull/789#discussion_r173511934) to figure out when to send.

fippo:
reminder: [test as you commit](https://github.com/w3c/web-platform-tests/pull/9941#pullrequestreview-102874364)

The test is pretty simple: do a signaling handshake, listen to connectionstate, check for canInsertDTMF. The helper from that PR should work even.

aboba:
@jan-ivar @fippo PR submitted. Can you take a look?

aboba:
This issue was addressed by PR https://github.com/w3c/webrtc-pc/pull/1980

PR #1957, opened 2018-08-09
canInsertDTMF transitions

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1619

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/1957.html" title="Last updated on Sep 3, 2018, 8:22 PM GMT (2e53e9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1957/06345b0...2e53e9c.html" title="Last updated on Sep 3, 2018, 8:22 PM GMT (2e53e9c)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9862,7 +9862,7 @@ interface RTCDataChannelEvent : Event {
         with which <var>sender</var> is associated.</li>
         <li>Let <var>pc</var> be the <code><a>RTCPeerConnection</a></code> associated with
         <var>transceiver</var>.</li>
-        <li>Check whether <code><var>pc</var>.connectionState</code> is <code>"connected"</code>.
+        <li>Check whether <var>pc</var>'s <a>[[\ConnectionState]]</a> slot is <code>"connected"</code>.
         If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
         and abort these steps.</li>
         <li>Check whether <code><var>sender</var>.track</code> is set.
@@ -9871,10 +9871,10 @@ interface RTCDataChannelEvent : Event {
         <li>Check whether <code><var>sender</var>.track.kind</code> has the value
         <code>"audio"</code>. If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to
         <code>false</code> and abort these steps.</li>
-        <li>Check whether <code><var>transceiver</var>.currentDirection</code> is set. If not,
+        <li>Check whether <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> is set. If not,
         set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
         and abort these steps.</li>
-        <li>Check whether <code><var>transceiver</var>.currentDirection</code> has the value
+        <li>Check whether <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot has the value
         <code>"sendrecv"</code> or <code>"sendonly"</code>. If not, set <var>dtmfSender</var>'s
         <a>[[\CanInsertDTMF]]</a> slot to <code>false</code> and abort these steps.</li>
         <li>Check whether for some value of <var>j</var> ranging from 0 to the number of codecs,
--- a/webrtc.html
+++ b/webrtc.html
@@ -9775,6 +9775,7 @@ interface RTCDataChannelEvent : Event {
                 <a>throw</a> an <code>InvalidStateError</code>.</li>
                 <li>Let <var>dtmf</var> be the <code><a>RTCDTMFSender</a></code>
                 associated with <var>sender</var>.</li>
+                <li><a>update the [[\CanInsertDTMF]]</a> internal slot of <var>dtmf</var>.</li>
                 <li>If <var>dtmf</var>'s <a>[[\CanInsertDTMF]]</a> internal slot
                 is <code>false</code>, <a>throw</a> an <code>InvalidStateError</code>.</li>
                 <li>Let <var>tones</var> be the method's first argument.</li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -9860,22 +9860,22 @@ interface RTCDataChannelEvent : Event {
         <var>dtmfSender</var>.</li>
         <li>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
         associated with <var>sender</var>.</li>
-        <li>Let <var>pc</var> be the <code><a>RTCPeerConnection</a></code> associated with
+        <li>Let <var>connection</var> be the <code><a>RTCPeerConnection</a></code> associated with
         <var>transceiver</var>.</li>
-        <li>Check whether <var>pc</var>'s <a>[[\ConnectionState]]</a> slot does not have the value
-        <code>"connected"</code> or <code>"completed"</code>. If so, set <var>dtmfSender</var>'s
+        <li>Check whether <var>connection</var>'s <a>[[\ConnectionState]]</a> slot
+        has the value <code>"connected"</code>. If not, set <var>dtmfSender</var>'s
         <a>[[\CanInsertDTMF]]</a> slot to <code>false</code> and abort these steps.</li>
         <li>Check whether <code><var>sender</var>.track</code> is unset.
         If so, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to
         <code>false</code> and abort these steps.</li>
-        <li>Check whether <code><var>sender</var>.track.kind</code> does not have the value
-        <code>"audio"</code>. If so, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a>
+        <li>Check whether <code><var>sender</var>.track.kind</code> has the value
+        <code>"audio"</code>. If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a>
         slot to <code>false</code> and abort these steps.</li>
-        <li>Check whether <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot is unset.
-        If so, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
+        <li>Check whether <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot is set.
+        If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
         and abort these steps.</li>
         <li>Check whether <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot
-        does not have the value <code>"sendrecv"</code> or <code>"sendonly"</code>. If so, set
+        has the value <code>"sendrecv"</code> or <code>"sendonly"</code>. If not, set
         <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
         and abort these steps.</li>
         <li>Check whether for some value of <var>j</var> ranging from 0 to the number of
--- a/webrtc.html
+++ b/webrtc.html
@@ -9712,7 +9712,41 @@ interface RTCDataChannelEvent : Event {
             <dt><dfn data-idl><code>canInsertDTMF</code></dfn> of type <span class=
             "idlAttrType"><a>boolean</a></span>, readonly</dt>
             <dd>
-              <p>Whether the <a>RTCDTMFSender</a> is capable of sending DTMF.</p>
+              <p>Whether the <a>RTCDTMFSender</a> <var>dtmfSender</var> is capable of
+              sending DTMF. On getting, the attribute MUST return the value of
+              <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot. To set
+              <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot, run the following steps
+              when <code><var>pc</var>.onsignalingstatechange</code> or
+              <code><var>pc</var>.connectionstatechange</code> is fired:
+              <ol>
+                <li>Let <var>pc</var> be the <code><a>RTCPeerConnection</a></code> which
+                is sending DTMF.</li>
+                <li>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
+                used to send DTMF.</li>
+                <li>Check whether <code><var>pc</var>.signalingState</code> is <code>"stable"</code>.
+                If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
+                and abort these steps.</li>
+                <li>Check whether <code><var>pc</var>.connectionState</code> is <code>"connected"</code>.
+                If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
+                and abort these steps.</li>
+                <li>Check whether <code><var>transceiver</var>.sender.track</code> is set.
+                If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
+                and abort these steps.</li>
+                <li>Check whether <code><var>transceiver</var>sender.track.kind</code> has the value
+                <code>"audio"</code>. If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to
+                <code>false</code> and abort these steps.</li>
+                <li>Check whether <code><var>transceiver</var>.currentDirection</code> is set. If not,
+                set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
+                and abort these steps.</li>
+                <li>Check whether <code><var>transceiver</var>.currentDirection</code> has the value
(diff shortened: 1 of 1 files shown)


aboba:
@fippo Can you take a look?

webrtc.html
```diff
@@ -9712,7 +9712,41 @@ <h2>Attributes</h2>
...
+                <code>"audio"</code>. If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to
+                <code>false</code> and abort these steps.</li>
+                <li>Check whether <code><var>transceiver</var>.currentDirection</code> is set. If not,
+                set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
+                and abort these steps.</li>
+                <li>Check whether <code><var>transceiver</var>.currentDirection</code> has the value
+                <code>"sendrecv"</code> or <code>"sendonly"</code>. If not, set <var>dtmfSender</var>'s
+                <a>[[\CanInsertDTMF]]</a> slot to <code>false</code> and abort these steps.</li>
+                <li>Check whether for some value of <var>j</var> ranging from 0 to the number of codecs,
+                <code><var>transceiver</var>.sender.getParameters().codecs[<var>j</var>].mimetype</code> has the value
```
fippo: does the clockRate need to match a negotiated codec too? Chrome has
```
a=rtpmap:110 telephone-event/48000
a=rtpmap:112 telephone-event/32000
a=rtpmap:113 telephone-event/16000
a=rtpmap:126 telephone-event/8000
```
(and maybe @alvestrand knows why)

aboba: Yes, it does need to match a negotiated codec clock rate, so that DTMF can be sent with the same SSRC as that other codec (clockrate can't vary on an SSRC).  PR assumes that the SDP negotiation handles that correctly.

fippo: and .codecs is only set after negotiation. Now it makes sense, thank you!

webrtc.html
```diff
@@ -9712,7 +9712,41 @@ <h2>Attributes</h2>
             <dt><dfn data-idl><code>canInsertDTMF</code></dfn> of type <span class=
             "idlAttrType"><a>boolean</a></span>, readonly</dt>
             <dd>
-              <p>Whether the <a>RTCDTMFSender</a> is capable of sending DTMF.</p>
+              <p>Whether the <a>RTCDTMFSender</a> <var>dtmfSender</var> is capable of
+              sending DTMF. On getting, the attribute MUST return the value of
+              <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot. To set
+              <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot, run the following steps
+              when <code><var>pc</var>.onsignalingstatechange</code> or
+              <code><var>pc</var>.connectionstatechange</code> is fired:
```
stefhak: Perhaps we want to phrase this a bit differently. Rather than saying it should be done "when onsignalingstatechange fires" it seems to me it should be expressed in terms of "update the connection state" (dfn) and "signaling-state" (dfn) changes.

jan-ivar: I worry this reads like a shim, and I hope no-one implements it this way. It's also not clear what thread this is to happen on. Internal slots lets us move setting to the respective algorithms that will end up needing to set it. If the number of places needed to references this are too numerous, we could perhaps break out a sub-algorithm like we did for [update the negotiation-needed flag](https://w3c.github.io/webrtc-pc/#dfn-update-the-negotiation-needed-flag), but we still need to be specific about when it is called, in the implementation itself, and not piggy-back on other JS observables IMHO.

Seeing the below, I also worry there's too much state baked into this attribute...

webrtc.html
```diff
@@ -9712,7 +9712,41 @@ <h2>Attributes</h2>
...
+              <ol>
+                <li>Let <var>pc</var> be the <code><a>RTCPeerConnection</a></code> which
+                is sending DTMF.</li>
+                <li>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
+                used to send DTMF.</li>
+                <li>Check whether <code><var>pc</var>.signalingState</code> is <code>"stable"</code>.
+                If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
+                and abort these steps.</li>
```
jan-ivar: Why can't we send dtmf during re-negotiation?

webrtc.html
```diff
@@ -9712,7 +9712,37 @@ <h2>Attributes</h2>
             <dt><dfn data-idl><code>canInsertDTMF</code></dfn> of type <span class=
             "idlAttrType"><a>boolean</a></span>, readonly</dt>
             <dd>
-              <p>Whether the <a>RTCDTMFSender</a> is capable of sending DTMF.</p>
+              <p>Whether the <a>RTCDTMFSender</a> <var>dtmfSender</var> is capable of
+              sending DTMF. On getting, the attribute MUST return the value of
+              <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot. To set
```
alvestrand: I think it is better if the setting of the slot is not part of the table of attributes. It's tempting, but it looks messy.

webrtc.html
```diff
@@ -9712,7 +9712,37 @@ <h2>Attributes</h2>
...
+              <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot, run the following steps
+              when <code><var>pc</var>.connectionstatechange</code> is fired:
+              <ol>
+                <li>Let <var>pc</var> be the <code><a>RTCPeerConnection</a></code> which
+                is sending DTMF.</li>
+                <li>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
+                used to send DTMF.</li>
```
alvestrand: This should be "the transceiver to which the RTPSender is associated", since [[CanInsertDTMF]] is a sender attribute.

webrtc.html
```diff
@@ -9712,7 +9712,37 @@ <h2>Attributes</h2>
...
+                <li>Check whether <code><var>pc</var>.connectionState</code> is <code>"connected"</code>.
```
alvestrand: You'll get some hassle about this (using JS notation to describe access to internal state variables) - do we have an internal attribute for [[connectionState]]?

webrtc.html
```diff
@@ -9712,7 +9712,37 @@ <h2>Attributes</h2>
...
+                If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
```
alvestrand: This can be expressed shorter as "If PC's [[connectionState]] is not "connected" or "completed", set [[CanInsertDTMF]] to false and abort these steps". Similar for the next ones.

webrtc.html
aboba: Moved the text outside the attribute table (to its own section).

webrtc.html

webrtc.html
```diff
@@ -9849,6 +9851,40 @@ <h2>Methods</h2>
         </section>
       </div>
     </section>
+    <section>
+      <h3>canInsertDTMF update</h3>
```
henbos: This is all good, but was invokes the canInsertDTMF update? The [[CanInsertDTMF]] attribute is initially false and never touched as far as I can tell.

This needs to be invoked either on all possible places where one of the conditions might change which seems tedious and error prone, or any time where we care that the internal slot is up to date, that is:
- When the canInsertDTMF attribute is invoked.
- When insertDTMF() is invoked.

aboba: Added a step to invoke the update when insertDTMF() is called.

henbos: What about when canInsertDTMF attribute is called? The caller might decide never to call insertDTMF() in the first place if canInsertDTMF is false.

webrtc.html
```diff
@@ -9849,6 +9851,40 @@ <h2>Methods</h2>
...
+      <p>To <dfn>update the [[\CanInsertDTMF]]</dfn> slot of an <code><a>RTCDTMFSender</a></code>
+      instance <var>dtmfSender</var>, the user agent MUST queue a task that runs the following steps:</p>
+      <ol>
+        <li>Let <var>sender</var> be the <code><a>RTCRtpSender</a></code> associated with
+        <var>dtmfSender</var>.</li>
+        <li>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
+        associated with <var>sender</var>.</li>
+        <li>Let <var>pc</var> be the <code><a>RTCPeerConnection</a></code> associated with
```
stefhak: Nit: are we not usually calling it `<var>connection</var>` instead of `<var>pc</var>`?

webrtc.html
```diff
@@ -9849,6 +9851,40 @@ <h2>Methods</h2>
...
+        <var>transceiver</var>.</li>
+        <li>Check whether <var>pc</var>'s <a>[[\ConnectionState]]</a> slot does not have the value
+        <code>"connected"</code> or <code>"completed"</code>. If so, set <var>dtmfSender</var>'s
```
stefhak: I can't find `"completed"` in the connectionState [enum](https://w3c.github.io/webrtc-pc/#dfn-connection-state)

Maybe it's only me, but I think it would clearer if put something like
`<li>Check whether <var>pc</var>'s <a>[[\ConnectionState]]</a> slot has the value
        <code>"connected". If not, set <var>dtmfSender</var>'s

webrtc.html
```diff
@@ -9849,6 +9851,40 @@ <h2>Methods</h2>
...
+        <a>[[\CanInsertDTMF]]</a> slot to <code>false</code> and abort these steps.</li>
+        <li>Check whether <code><var>sender</var>.track</code> is unset.
+        If so, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to
+        <code>false</code> and abort these steps.</li>
+        <li>Check whether <code><var>sender</var>.track.kind</code> does not have the value
+        <code>"audio"</code>. If so, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a>
```
stefhak:
```
<li>Check whether <code><var>sender</var>.track.kind</code> has the value
        <code>"audio"</code>. If not, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a>
```
?

webrtc.html
```diff
@@ -9849,6 +9851,40 @@ <h2>Methods</h2>
...
+        slot to <code>false</code> and abort these steps.</li>
+        <li>Check whether <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot is unset.
+        If so, set <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
```
stefhak: This step is not needed given that the next step checks if it's sendrecv or sendonly.

webrtc.html
```diff
@@ -9849,6 +9851,40 @@ <h2>Methods</h2>
...
+        and abort these steps.</li>
+        <li>Check whether <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot
+        does not have the value <code>"sendrecv"</code> or <code>"sendonly"</code>. If so, set
```
stefhak:
```
has have the value <code>"sendrecv"</code> or <code>"sendonly"</code>. If not, set
```
?

webrtc.html
```diff
@@ -9849,6 +9851,40 @@ <h2>Methods</h2>
...
+        <var>dtmfSender</var>'s <a>[[\CanInsertDTMF]]</a> slot to <code>false</code>
+        and abort these steps.</li>
+        <li>Check whether for some value of <var>j</var> ranging from 0 to the number of
+        codecs, <code><var>sender</var>.getParameters().codecs[<var>j</var>].mimetype</code>
+        has the value <code>"audio/telephone-event"</code>. If not, set <var>dtmfSender</var>'s
+        <a>[[\CanInsertDTMF]]</a> slot to <code>false</code> and abort these steps.</li>
```
stefhak: I note that we have not added any check that the script could not do itself, so we're basically just adding convenience. Would it make sense to add something like
```
If there is any other reason why <code><var>sender</var></code> would not be able to insert DTMF , set <var>dtmfSender</var>'s  <a>[[\CanInsertDTMF]]</a> slot to <code>false</code> and abort these steps.
```

webrtc.html

webrtc.html
aboba: The text is attempting to describe behavior exactly so that browsers can behave similarly.  So if there is a missing check, I would prefer it be described explicitly.

webrtc.html
```diff
@@ -9712,7 +9712,9 @@ <h2>Attributes</h2>
             <dt><dfn data-idl><code>canInsertDTMF</code></dfn> of type <span class=
             "idlAttrType"><a>boolean</a></span>, readonly</dt>
             <dd>
-              <p>Whether the <a>RTCDTMFSender</a> is capable of sending DTMF.</p>
+              <p>Whether the <a>RTCDTMFSender</a> <var>dtmfSender</var> is capable of
+              sending DTMF. On getting, the user agent MUST <a>update the [[\CanInsertDTMF]]</a>
+              internal slot of <var>dtmfSender</var> and return it.</p>
```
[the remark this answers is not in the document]
jan-ivar: What @stefhak said. Since `insertDTMF` doesn't *update* the [[CanInsertDTMF]] slot, it would keep failing until the user tries to get the attribute in JS, so this won't work.

As I mentioned in https://github.com/w3c/webrtc-pc/pull/1957#discussion_r208940444 I think the point of internal slots would be to move the `true`/`false` determination away from the getter, to where the changes may happen.

webrtc.html
```diff
@@ -9775,6 +9775,8 @@ <h2>Methods</h2>
                 <a>throw</a> an <code>InvalidStateError</code>.</li>
                 <li>Let <var>dtmf</var> be the <code><a>RTCDTMFSender</a></code>
                 associated with <var>sender</var>.</li>
+                <li><a>update the [[\CanInsertDTMF]]</a> internal slot of
+                <var>dtmf</var>.</li>
```
jan-ivar: [[CanInsertDTMF]] seems redundant with this approach.

webrtc.html
```diff
@@ -9773,6 +9775,8 @@ <h2>Methods</h2>
...
```
jan-ivar: [[CanInsertDTMF]] seems redundant with this approach.

webrtc.html
```diff
@@ -9849,6 +9853,40 @@ <h2>Methods</h2>
...
+        <li>Let <var>sender</var> be the <code><a>RTCRtpSender</a></code> associated with
+        <var>dtmfSender</var>.</li>
+        <li>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
+        associated with <var>sender</var>.</li>
+        <li>Let <var>connection</var> be the <code><a>RTCPeerConnection</a></code> associated with
+        <var>transceiver</var>.</li>
+        <li>Check whether <var>connection</var>'s <a>[[\ConnectionState]]</a> slot
+        has the value <code>"connected"</code>. If not, set <var>dtmfSender</var>'s
+        <a>[[\CanInsertDTMF]]</a> slot to <code>false</code> and abort these steps.</li>
+        <li>Check whether <code><var>sender</var>.track</code> is unset.
```
jan-ivar: I think `null` would be clearer than *unset* here, since a track may be explicitly set to `null` with *replaceTrack*

webrtc.html
```diff
@@ -9849,6 +9853,40 @@ <h2>Methods</h2>
...
+      <ol>
+        <li>Let <var>sender</var> be the <code><a>RTCRtpSender</a></code> associated with
+        <var>dtmfSender</var>.</li>
+        <li>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
+        associated with <var>sender</var>.</li>
+        <li>Let <var>connection</var> be the <code><a>RTCPeerConnection</a></code> associated with
+        <var>transceiver</var>.</li>
+        <li>Check whether <var>connection</var>'s <a>[[\ConnectionState]]</a> slot
+        has the value <code>"connected"</code>. If not, set <var>dtmfSender</var>'s
+        <a>[[\CanInsertDTMF]]</a> slot to <code>false</code> and abort these steps.</li>
```
jan-ivar: Maybe simpler: *"If [[ConnectionState]] is not "connected", return true"* ?

Applies throughout.

webrtc.html
alvestrand: the relevant webrtc code goes

if (canInsertDtmf()) {
  insertDTMF()
} else {
  return error
}

so this makes the spec match the implementation....

jan-ivar: If you skip that test it still throw an error... 🤷🏻‍♂️

webrtc.html
stefhak: I was after that there might be a reason that is not observable for the script why DTMF can't be sent. Similar to that the UA may fire an `negotiationneeded` event at any time even if the script has not done anything. But perhaps that is not needed - let's skip it.

webrtc.html
stefhak: Yes, I like that approach (and thus removing the internal slot, and just have the algorithm return true or false).

webrtc.html
jan-ivar: @aboba As long as the algorithm is synchronous and relies solely on main-thread state, can't we run it whenever the `canInsertDTMF` getter is called, and only then?

stefhak:
I did #1980 for comparison. Not sure it is any better though.

aboba:
Closing since Stefan’s alternative PR looks more promising.

Issue #1273, opened 2017-05-31
Events not bubbling


aboba:
From EKR: https://lists.w3.org/Archives/Public/public-webrtc/2017May/0166.html

A similar comment about pulling stuff out is with the boilerplate
about events not bubbling or being cancelable.

aboba:
There are four instances in the specification with events "which does not bubble (except where otherwise stated) and is not cancelable (except where otherwise stated)".  Does it make sense to pull this text out, or just leave it as it is?

stefhak:
Four cases is not that much, I think we could leave that as is.

jan-ivar:
Yes we seem to be inconsistent here at least. Searching for "Fire", some places, like tonechange event, is a sub-algorithm that talks about bubbling, whereas many others like "RTCError event" does not.

stefhak:
Sorry for my ignorance, but what reasons are we using to select whether an event should:
- fire as a simple event (no wording about bubbling or being cancable)
- fire as an event (no wording about bubbling or being cancable)
- fire as an event that does not bubble and is not cancable?

alvestrand:
Plan: Remove all text about bubbling and cancelling, and refer to a DOM spec that defaults to "not bubble and not cancellable".

stefhak:
@aboba I think this can be closed given the merge of #1966.

PR #1966, opened 2018-08-20
A shot at aligning eventfiring with how it's specified in other docs.

stefhak:
Fix for #1273. References the WhatWG DOM spec's definition of "fire an event" (in the same way as e.g. WhatWG WebSocket spec does). Does away with "doesn't bubble and is not cancelable" as that seems only relevant when the target is a Dom object.

Note: I have no clue on what "with an OperationError exception" means in 
```
Fire an event named error using the RTCError interface with an OperationError exception at channel.
```

***
<a href="https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1966.html" title="Last updated on Sep 1, 2018, 5:22 AM GMT (410c326)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1966/1df02bb...stefhak:410c326.html" title="Last updated on Sep 1, 2018, 5:22 AM GMT (410c326)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1524,7 +1524,7 @@
                           </li>
                           <li>
                             <p><a>Fire an event</a> named <code><a data-link-for=
-                            "RTCDataChannel">error</a></code > using the
+                            "RTCDataChannel">error</a></code> using the
                             <code><a>RTCErrorEvent</a></code> interface with an
                             <code>OperationError</code> exception at
                             <var>channel</var>.</p>
@@ -9035,7 +9035,7 @@ interface RTCTrackEvent : Event {
         </li>
         <li>
           <p><a>Fire an event</a> named <code><a data-link-for=
-          "RTCDataChannel">error</a></code >using the
+          "RTCDataChannel">error</a></code> using the
           <code><a>RTCErrorEvent</a></code> interface with the
           <code><a>errorDetail</a></code> attribute set to
           "data-channel-failure" at <var>channel</var>.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1539,7 +1539,7 @@
                     </ol>
                   </li>
                   <li>
-                    <p>Let <var>trackEvents</var>, <var>muteTracks</var>,
+                    <p>Let <var>trackEventInits</var>, <var>muteTracks</var>,
                     <var>addList</var>, and <var>removeList</var> be empty
                     lists.</p>
                   </li>
@@ -1759,7 +1759,7 @@
                             is neither <code>"sendrecv"</code> nor <code>"recvonly"</code>,
                             <a>process the addition of a remote track</a> for
                             the <a>media description</a>, given <var>transceiver</var>
-                            and <var>trackEvents</var>.</p>
+                            and <var>trackEventInits</var>.</p>
                           </li>
                           <li>
                             <p>If <var>direction</var> is <code>"sendonly"</code> or
@@ -1899,10 +1899,18 @@
                     <var>track</var> to <var>stream</var>.</p>
                   </li>
                   <li>
-                    <p>For each <code><a>RTCTrackEvent</a></code>
-                    <var>trackEvent</var> in <var>trackEvents</var>,
+                    <p>For each entry <var>entry</var> in <var>trackEventInits</var>,
                     <a>fire an event</a> named <code title=
-                    "event-track"><a>track</a></code> using <var>trackEvent</var>
+                    "event-track"><a>track</a></code> using the
+                    <code><a>RTCTrackEvent</a></code> interface with its
+                    <code>receiver</code> attribute initialized to
+                    <var>entry</var>.<code>receiver</code>, its 
+                    <code>track</code> attribute initialized to
+                    <var>entry</var>.<code>track</code>, its 
+                    <code>streams</code> attribute initialized to
+                    <var>entry</var>.<code>streams</code> and its 
+                    <code>transceiver</code> attribute initialized to
+                    <var>entry</var>.<code>transceiver</code>
                     at the <var>connection</var> object.</p>
                   </li>
                   <li>
@@ -5403,7 +5411,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
         an incoming media description <span data-jsep=
         "applying-a-remote-desc">[[!JSEP]]</span> given
         <code>RTCRtpTransceiver</code> <var>transceiver</var> and
-        <var>trackEvents</var>, the user agent MUST run the following steps:</p>
+        <var>trackEventInits</var>, the user agent MUST run the following steps:</p>
         <ol>
           <li>
             <p>Let <var>receiver</var> be <var>transceiver</var>'s
@@ -5421,10 +5429,10 @@ interface RTCPeerConnectionIceErrorEvent : Event {
             </p>
           </li>
           <li>
-            <p>Add a new <code><a>RTCTrackEvent</a></code> with
+            <p>Create a new object with
             <var>receiver</var>, <var>track</var>,
-            <var>streams</var> and <var>transceiver</var> to
-            <var>trackEvents</var>.</p>
+            <var>streams</var> and <var>transceiver</var> and add
+            it to <var>trackEventInits</var>.</p>
           </li>
         </ol>
         <p>To <dfn id="process-remote-track-removal">
@@ -8995,8 +9003,10 @@ interface RTCTrackEvent : Event {
         <li>
           <p>If the <a data-lt="data transport">transport</a> was closed
           <dfn id="data-transport-closed-error">with an error</dfn>, <a>fire
-          an event</a> using the <code><a>RTCError</a></code> interface with
-          <code>errorDetail</code> set to "sctp-failure" at <var>channel</var>.</p>
+          an event</a> named <code><a>error</a></code> using the
+          <code><a>RTCErrorEvent</a></code> interface with its
+          <code>errorDetail</code> attribute set to "sctp-failure"
+          at <var>channel</var>.</p>
         </li>
         <li>
           <p><a>Fire an event</a> named <code title=
@@ -9076,8 +9086,8 @@ interface RTCTrackEvent : Event {
         </li>
         <li>
           <p><a>Fire an event</a> named <code><a>message</a></code> using the 
-          <code title="event-datachannel-message"><a>message</a></code> interface
-          with the <code>origin</code> attribute initialized
+          <code title="event-datachannel-message"><a>MessageEvent</a></code> interface
+          with its <code>origin</code> attribute initialized
           to the origin of the document that created the <var>channel</var>'s
           associated <a>RTCPeerConnection</a>, and the <code>data</code>
           attribute initialized to <var>data</var> at
@@ -11482,9 +11492,9 @@ interface RTCErrorEvent : Event {
         <tr>
           <td><dfn id=
           "event-datachannel-message"><code>message</code></dfn></td>
-          <td><code><a data-cite=
+          <td><dfn id=the-messageevent-interfaces><code><a data-cite=
           "!webmessaging#the-messageevent-interfaces">MessageEvent</a></code>
-          [[!webmessaging]]</td>
+          </dfn>[[!webmessaging]]</td>
           <td>A message was successfully received.</td>
         </tr>
         <tr>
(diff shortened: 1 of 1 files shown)


stefhak:
Would be good to get your input @annevk

webrtc.html
```diff
@@ -1520,13 +1523,14 @@ <h4>Set the RTCSessionDescription</h4>
                             to <code>"closed"</code>.</p>
                           </li>
                           <li>
-                            <p>Fire an event named <code><a data-link-for=
-                            "RTCDataChannel">error</a></code > with an
+                            <p><a>Fire an event</a> named <code><a data-link-for=
+                            "RTCDataChannel">error</a></code > using the
+                            <code><a>RTCError</a></code> interface with an
                             <code>OperationError</code> exception at
                             <var>channel</var>.</p>
```
annevk: This was already weird I suppose, but all events are supposed to use an interface that inherits from `Event`. I suspect this won't work if you actually follow all the steps "fire an event" lays out.

webrtc.html
```diff
@@ -1897,8 +1901,8 @@ <h4>Set the RTCSessionDescription</h4>
                   <li>
                     <p>For each <code><a>RTCTrackEvent</a></code>
                     <var>trackEvent</var> in <var>trackEvents</var>,
-                    <a>fire a track event</a> named <code title=
-                    "event-track"><a>track</a></code> with <var>trackEvent</var>
+                    <a>fire an event</a> named <code title=
+                    "event-track"><a>track</a></code> using <var>trackEvent</var>
                     at the <var>connection</var> object.</p>
```
annevk: This also seems weird, since it looks like _trackEvents_ would hold instances, whereas "fire an event" wants a class. (I haven't checked it out in context though.)

webrtc.html
```diff
@@ -9005,12 +8994,12 @@ <h3><dfn>RTCDataChannel</dfn></h3>
         </li>
         <li>
           <p>If the <a data-lt="data transport">transport</a> was closed
-          <dfn id="data-transport-closed-error">with an error</dfn>, fire
-          an <code><a>RTCError</a></code> event at <var>channel</var> with
-          <code>errorDetail</code> set to "sctp-failure".</p>
+          <dfn id="data-transport-closed-error">with an error</dfn>, <a>fire
+          an event</a> using the <code><a>RTCError</a></code> interface with
+          <code>errorDetail</code> set to "sctp-failure" at <var>channel</var>.</p>
```
annevk: Same as above. (And more below.)

webrtc.html
```diff
@@ -9085,11 +9075,13 @@ <h3><dfn>RTCDataChannel</dfn></h3>
           </ul>
         </li>
         <li>
-          <p>Fire an event named <code><a>message</a></code> at
-          <var>channel</var> with the <code>origin</code> attribute initialized
+          <p><a>Fire an event</a> named <code><a>message</a></code> using the 
+          <code title="event-datachannel-message"><a>message</a></code> interface
```
annevk: message is not an interface.

webrtc.html
```diff
@@ -9085,11 +9075,13 @@ <h3><dfn>RTCDataChannel</dfn></h3>
...
+          with the <code>origin</code> attribute initialized
           to the origin of the document that created the <var>channel</var>'s
           associated <a>RTCPeerConnection</a>, and the <code>data</code>
```
annevk: This seems too vague. You need to define that RTCPeerConnection holds a document in a slot somewhere and use that concept here. And define how the slot gets filled when creating RTCPeerConnection.

webrtc.html
stefhak: Also, I'm missing the "named error" part.

stefhak:
Thanks @annevk for reviewing - I'll update.

stefhak:
@aboba that is spelled out in ```process the addition of a remote track```, however, as @annevk identifies in https://github.com/w3c/webrtc-pc/pull/1966#discussion_r211494217 I need to redefine that part (as "fire an event" can't take an event object as input).

webrtc.html
stefhak: Agreed, I opened #1969. However, goes beyond the scope of this PR.

webrtc.html
```diff
@@ -1895,10 +1899,18 @@ <h4>Set the RTCSessionDescription</h4>
                     <var>track</var> to <var>stream</var>.</p>
                   </li>
                   <li>
-                    <p>For each <code><a>RTCTrackEvent</a></code>
-                    <var>trackEvent</var> in <var>trackEvents</var>,
-                    <a>fire a track event</a> named <code title=
-                    "event-track"><a>track</a></code> with <var>trackEvent</var>
+                    <p>For each entry <var>entry</var> in <var>trackEventInits</var>,
```
henbos: nit: I don't know what "type" means for the constructor, but there is an RTCTrackEvent constructor taking the RTCTrackEventInit:
```
[Constructor(DOMString type, RTCTrackEventInit eventInitDict),
 Exposed=Window]
```
So maybe you could make this shorter by saying something like "Construct an RTCTrackEvent using the entry as the RTCTrackEventInit argument" and fire it. These steps are roughly redefining the constructor.

webrtc.html
```diff
@@ -5426,10 +5429,10 @@ <h4>Processing Remote MediaStreamTracks</h4>
             </p>
           </li>
           <li>
-            <p>Add a new <code><a>RTCTrackEvent</a></code> with
+            <p>Create a new object with
```
henbos: If the RTCTrackEvent constructor above is addressed, this should say create a new RTCTrackEventInit. You might have to say trackEventInits is an empty list of RTCTrackEventInits rather than an empty list?

webrtc.html
```diff
@@ -78,9 +78,10 @@ <h2>Terminology</h2>
     <code><dfn data-cite="!HTML51/webappapis.html#errorevent-errorevent">
     ErrorEvent</dfn></code> interface are defined in [[!HTML51]].</p>
     <p>The concepts <dfn data-cite="!HTML51/webappapis.html#queuing">queue a
-    task</dfn>, <dfn data-cite="!HTML51/infrastructure.html#fire">fire a
-    simple event</dfn> and <dfn data-cite="!HTML51/webappapis.html#networking-task-source">networking
+    task</dfn> and <dfn data-cite="!HTML51/webappapis.html#networking-task-source">networking
     task source</dfn> are defined in [[!HTML51]].</p>
+    <p>The concept <dfn data-cite="!DOM#firing-events">fire an
+    event</dfn> is defined in [[!DOM]].</p>
```
henbos: What is the https:// link for this? I'm assuming this is equivalent to the "does not bubble, etc etc" language that has been removed int his PR.

webrtc.html
stefhak: That's a good idea!
Another idea I played with was to let the `trackEvents` continue to hold the events, and then here, instead of using `Fire an event` use `[dispatch](https://dom.spec.whatwg.org/#dispatching-events) event`. Not sure what way is best here.

webrtc.html
stefhak: Yes, I like that!

webrtc.html
stefhak: The link is https://dom.spec.whatwg.org/#firing-events.

To my understanding, "does not bubble etc" has no meaning when the target object is not an DOM element, so we do not need that part.

webrtc.html
annevk: Instead of fire, you can use https://dom.spec.whatwg.org/#concept-event-create and https://dom.spec.whatwg.org/#concept-event-dispatch. I'd generally try to avoid that, but if that's the only thing that works here, that's the way to go. Please don't try invoking the constructor of things yourself as there's all kinds of complicated details there that really should be abstracted over across the whole web platform.

stefhak: Reading https://dom.spec.whatwg.org/#concept-event-create I stumble on the note "This does not work if members are required", and https://w3c.github.io/webrtc-pc/#dom-rtctrackeventinit requires receiver, track and transceiver.

The easiest way out seems to be to do what the PR in its current form does - though a bit ugly.

stefhak:
Adding the "Editorial" label as the intent is just to make the spec clearer and consistent with other specs - not to change any behavior.

webrtc.html
annevk: It is important (it affects some properties) but the defaults for those are false.

webrtc.html
stefhak: @jan-ivar that is really what I wanted to do, but is it allowed? I somehow got the impression that dictionaries are not "real" things, but can be used in the process of creating something "real" such as an object. If we can use dictionaries in this way I'll change immediately!

webrtc.html
stefhak: @annevk does `does not bubble` and `is not cancelable` have a meaning even when the target object is not in the DOM tree? Currently we have e.g.
```
Firing an ice candidate event named e with an RTCIceCandidate candidate means that an event with the name e, which does not bubble (except where otherwise stated) and is not cancelable (except where otherwise stated), and which uses the RTCPeerConnectionIceEvent interface with the candidate attribute set to the new ICE candidate, MUST be created and dispatched at the given target.
```
in the spec, but I somehow concluded that the "no bubble, not cancelable" is pointless here.

webrtc.html
```diff
@@ -10011,9 +10001,10 @@ <h2>Attributes</h2>
                       for that monitored object, and add it to <var>report</var>.
                     </li>
                     <li>
-                      <p>Fire an <code><a>RTCStatsEvent</a></code> event named
-                      named <code><a>statsended</a></code> at <var>connection</var>
-                      with <code>report</code> set to <var>report</var>.</p>
+                      <p><a>Fire an event</a> named <code><a>statsended</a></code>
+                      using the <code><a>RTCStatsEvent</a></code> interface
+                      with <code>report</code> set to <var>report</var>
```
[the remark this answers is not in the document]
stefhak: Good catch.

webrtc.html
```diff
@@ -9840,10 +9837,12 @@ <h2>Methods</h2>
...
                     + <a>[[\InterToneGap]]</a> ms from now that
                     runs the steps labelled <em>Playout task</em>.</li>
-                    <li><a>Fire a tonechange event</a> named
-                    <code><a>tonechange</a></code> with a string
-                    consisting of <var>tone</var> at the
-                    <code><a>RTCDTMFSender</a></code> object.</li>
+                    <li><a>Fire an event</a> named <code><a>tonechange</a></code>
+                    using the <code><a>RTCDTMFToneChangeEvent</a></code>
+                    interface with the <code><a data-link-for=
+                    "RTCDTMFToneChangeEvent">tone</a></code> attribute set to
```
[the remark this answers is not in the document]
stefhak: If you could point out the mistakes I'd be happy to fix them.

webrtc.html
annevk: As I said, it's observable, via `bubbles` and `cancelable` on https://dom.spec.whatwg.org/#interface-event, but the defaults for these are false (as they should be for what you're doing here too) so you don't have to mention them explicitly. I was just correcting the bit where you said they have no meaning.

stefhak: Thanks for the clarification!

webrtc.html
jan-ivar: I believe so. At least the WebIDL bindings in the Firefox codebase lets us do this. This is necessary for instance to satisfy APIs that return dictionaries.

As to supporting links, I found https://github.com/heycam/webidl/issues/142 which is asking for a more compact syntax for initializing dictionaries from prose, so I assume from that that doing such a thing in the first place is expected. :)

webrtc.html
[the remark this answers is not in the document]
jan-ivar: If I search for "attribute set to" in the [Preview](https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1966.html), I get 9 hits. 5 of them are directly preceeded by links, whereas 4 of them are not (candidate x 3, and sdpLineNumber).

webrtc.html
```diff
@@ -9035,11 +9034,14 @@ <h3><dfn>RTCDataChannel</dfn></h3>
           <code>closed</code>.</p>
         </li>
         <li>
-          <p>Fire an <code><a>RTCError</a></code> event at <var>channel</var> with
-          <code>errorDetail</code> set to "data-channel-failure".</p>
+          <p><a>Fire an event</a> named <code><a data-link-for=
+          "RTCDataChannel">error</a></code >using the
```
jan-ivar: Missing space s/errorusing/error using/

Issue #1970, opened 2018-08-23
Sort out meaning of "Fire event with OperationError exception"


stefhak:
In step 2.2.5.3 of the `set an RTCSessionDescription` algorithm it is said:
`Fire an event named error with an OperationError exception at channel.`

I'm not sure what that means. Should we add the error "operation-error" to `RTCErrorDetailType` and use that?

alvestrand:
I think it's supposed to mean "fire an event containing a DOMError with the name OperationError".
We have a "throw an OperationError exception" down in createDataChannel. But I'm not sure what firing it as an event (as opposed to throwing it as an exception or rejecting a promise with it) means.

jan-ivar:
As to intent, @alvestrand's guess is also mine. But *RTCErrorEvent*'s `error` member is of type *RTCError*, and incompatible with *DOMException*, so that won't work.

However, the surrounding steps, the closing of the data channel, looks identical to [data-channel-failure](http://w3c.github.io/webrtc-pc/#dfn-data-channel-failure), so my guess is this is what we want here. Maybe we can reuse that algorithm even?

PR #1971, opened 2018-08-27
Added error detail "datachannel-id-generation-error", using in error …

stefhak:
…event. Fix for #1970.

Would any test have to be added or changed?

Also, `"sctp-id-generation-error"` might be better than `"datachannel-id-generation-error"` - WDYT?

***
<a href="https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1971.html" title="Last updated on Sep 7, 2018, 5:47 AM GMT (a45ef37)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1971/f458914...stefhak:a45ef37.html" title="Last updated on Sep 7, 2018, 5:47 AM GMT (a45ef37)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1525,9 +1525,9 @@
                           <li>
                             <p><a>Fire an event</a> named <code><a data-link-for=
                             "RTCDataChannel">error</a></code> using the
-                            <code><a>RTCErrorEvent</a></code> interface with an
-                            <code>OperationError</code> exception at
-                            <var>channel</var>.</p>
+                            <code><a>RTCErrorEvent</a></code> interface with the
+                            <code><a>errorDetail</a></code> attribute set to
+                            "data-channel-failure" at <var>channel</var>.</p>
                           </li>
                           <li>
                             <p><a>Fire an event</a> named


alvestrand:
There needs to be a test - it's probably a pretty grotty one, generating datachannels until you get the error; I'm not sure we have the controls to make the range of available IDs small enough that it's quick to test.

I think "datachannel-id" is right, since this ID is part of the datachannel interface; if we ever do datachannels on top of QUIC (for instance), the same error could occur, but it wouldn't have anything to do with SCTP.

lgrahl:
There's a test that checks for that in https://github.com/web-platform-tests/wpt/pull/10468 which will need updating with these changes.

lgrahl:
(Hint: If you comment out some asserts, Chrome's main thread is blocked for several minutes with this test.)

lgrahl:
It should be this one: https://github.com/web-platform-tests/wpt/blob/1614d70351a708f2ae9904e037631c87e8296d59/webrtc/RTCDataChannel-id.html#L543:L635

Oh, it seems the test might not even need updating with these changes.

jan-ivar:
What distinguishes *datachannel-id-generation-error* from [data-channel-failure](http://w3c.github.io/webrtc-pc/#dfn-data-channel-failure)?

The latter text already says: *"For example, the data channel's id may be outside the range negotiated by the [RTCWEB-DATA] implementations in the SCTP handshake."*

stefhak:
@jan-ivar https://github.com/w3c/webrtc-pc/pull/1971#issuecomment-416637297 has a good point. I'd be happy to change to use `data-channel-failure`.

lgrahl:
The specific error is more... specific. But really, it's such a rare corner case that I don't mind having a general error at all. :smiley_cat:

jan-ivar:
Since they're both related to invalid ids I don't really understand the difference, but if others do...

lgrahl:
[`data-channel-failure`](http://w3c.github.io/webrtc-pc/#dom-rtcerrordetailtype-data-channel-failure) is being described as *The data channel has failed.* So, very generic. But, again, I don't mind in this case.

webrtc.html
```diff
@@ -1521,9 +1521,10 @@ <h4>Set the RTCSessionDescription</h4>
                           </li>
                           <li>
                             <p>Fire an event named <code><a data-link-for=
-                            "RTCDataChannel">error</a></code > with an
-                            <code>OperationError</code> exception at
-                            <var>channel</var>.</p>
+                            "RTCDataChannel">error</a></code > using the
+                            <code><a>RTCErrorEvent</a></code> interface with its
+                            <code>errorDetail</code> attribute set to
+                            "data-channel-failure" at <var>channel</var>.</p>
```
jan-ivar: When I search for "data-channel-failure", I find two places that look inconsistent:

*"Fire an event named error using the RTCErrorEvent interface with its errorDetail attribute set to "data-channel-failure" at channel."*

vs. [here](https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1971.html#dfn-data-channel-failure):

*"Fire an RTCError event at channel with errorDetail set to "data-channel-failure"."*

We should probably make it consistent.

stefhak: I _think_ that is already part of #1966

webrtc.html
```diff
@@ -1521,9 +1521,10 @@ <h4>Set the RTCSessionDescription</h4>
                           </li>
                           <li>
                             <p>Fire an event named <code><a data-link-for=
-                            "RTCDataChannel">error</a></code > with an
-                            <code>OperationError</code> exception at
-                            <var>channel</var>.</p>
+                            "RTCDataChannel">error</a></code > using the
+                            <code><a>RTCErrorEvent</a></code> interface with its
```
jan-ivar: nit: s/its/the/ for consistency

stefhak:
Rebased.

PR #1980, opened 2018-09-04
Issue 1619 patch

stefhak:
***
<a href="https://pr-preview.s3.amazonaws.com/stefhak/webrtc-pc/pull/1980.html" title="Last updated on Sep 11, 2018, 8:31 AM GMT (27fd692)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1980/06345b0...stefhak:27fd692.html" title="Last updated on Sep 11, 2018, 8:31 AM GMT (27fd692)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9866,10 +9866,9 @@ interface RTCDataChannelEvent : Event {
         nor <code>"sendonly"</code> return <code>false</code> and abort these steps.</li>
         <li>If <var>sender</var>'s <a>[[\SendEncodings]]</a><code>[0].active</code> is
         <code>false</code> return <code>false</code> and abort these steps.</li>
-        <li>If no codec with mimetype <code>"audio/telephone-event"</code> has been
-        negotiated for sending, and which the user agent is currently
-        capable of sending, for <var>sender</var>, return <code>false</code>
-        and abort these steps.</li>
+        <li>If, for <var>sender</var>, no codec (which the user agent is currently capable
+        of sending) with mimetype <code>"audio/telephone-event"</code> has been
+        negotiated for sending, return <code>false</code> and abort these steps.</li>
         <li>Return <code>true</code>.</li>
       </ol>
     </section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -9868,7 +9868,7 @@ interface RTCDataChannelEvent : Event {
         <code>false</code> return <code>false</code> and abort these steps.</li>
         <li>If no codec with mimetype <code>"audio/telephone-event"</code> has been
         negotiated for sending, and which the user agent is currently
-        capable of sending, return <code>false</code>
+        capable of sending, for <var>sender</var>, return <code>false</code>
         and abort these steps.</li>
         <li>Return <code>true</code>.</li>
       </ol>
--- a/webrtc.html
+++ b/webrtc.html
@@ -9859,17 +9859,13 @@ interface RTCDataChannelEvent : Event {
         <li>Let <var>connection</var> be the <code><a>RTCPeerConnection</a></code> associated with
         <var>transceiver</var>.</li>
         <li>If <var>connection</var>'s <code><a>RTCPeerConnectionState</a></code> is not <code>"connected"</code>
-       return <code>false</code> and abort these steps.</li>
-        <li>If <code><var>sender</var>.track</code> is <code>null</code> return <code>false</code>
-        and abort these steps.</li>
-        <li>If <code><var>sender</var>.track.kind</code> is not <code>"audio"</code> return
-        <code>false</code> and abort these steps.</li>
-        <li>If <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot is not set return
-        <code>false</code> and abort these steps.</li>
+        return <code>false</code> and abort these steps.</li>
+        <li>If <var>sender</var>'s <a>[[\SenderTrack]]</a> is <code>null</code>
+        return <code>false</code> and abort these steps.</li>
         <li>If <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> is neither <code>"sendrecv"</code>
         nor <code>"sendonly"</code> return <code>false</code> and abort these steps.</li>
         <li>If there is no value of <var>j</var> ranging from 0 to the number of
-        codecs, for which <code><var>sender</var>.getParameters().codecs[<var>j</var>].mimetype</code>
+        codecs, for which <var>sender</var><code>.getParameters().codecs[<var>j</var>].mimetype</code>
         has the value <code>"audio/telephone-event"</code>, return <code>false</code>
         and abort these steps.</li>
         <li>Return <code>true</code>.</li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -9849,6 +9847,34 @@ <h2>Methods</h2>
...
+        associated with <var>sender</var>.</li>
+        <li>Let <var>connection</var> be the <code><a>RTCPeerConnection</a></code> associated with
+        <var>transceiver</var>.</li>
+        <li>If <var>connection</var>'s <code><a>RTCPeerConnectionState</a></code> is not <code>"connected"</code>
+       return <code>false</code> and abort these steps.</li>
+        <li>If <code><var>sender</var>.track</code> is <code>null</code> return <code>false</code>
+        and abort these steps.</li>
+        <li>If <code><var>sender</var>.track.kind</code> is not <code>"audio"</code> return
+        <code>false</code> and abort these steps.</li>
+        <li>If <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> slot is not set return
```
stefhak: Do we need this check given that next thing done is to check if it is sendrecv or sendonly?

webrtc.html
```diff
@@ -9849,6 +9847,34 @@ <h2>Methods</h2>
...
+        <code>false</code> and abort these steps.</li>
+        <li>If <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> is neither <code>"sendrecv"</code>
+        nor <code>"sendonly"</code> return <code>false</code> and abort these steps.</li>
+        <li>If there is no value of <var>j</var> ranging from 0 to the number of
+        codecs, for which <code><var>sender</var>.getParameters().codecs[<var>j</var>].mimetype</code>
+        has the value <code>"audio/telephone-event"</code>, return <code>false</code>
```
jan-ivar: This to me implies normatively calling `getParameters()`, which is bad, because of side-effects (increments `transactionId`).

We generally try to avoid calling JS-facing APIs from algorithms, for this very reason, favoring split-out algorithms.

Is there another way to describe this? Do we need a **[[SendCodecs]]** similar to **[[SendEncodings]]**?

aboba: Yes, we probably do need [[SendCodecs]].  Ugh.

webrtc.html
```diff
@@ -9849,6 +9847,34 @@ <h2>Methods</h2>
...
+        <var>dtmfSender</var>.</li>
+        <li>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
+        associated with <var>sender</var>.</li>
+        <li>Let <var>connection</var> be the <code><a>RTCPeerConnection</a></code> associated with
+        <var>transceiver</var>.</li>
+        <li>If <var>connection</var>'s <code><a>RTCPeerConnectionState</a></code> is not <code>"connected"</code>
+       return <code>false</code> and abort these steps.</li>
+        <li>If <code><var>sender</var>.track</code> is <code>null</code> return <code>false</code>
+        and abort these steps.</li>
+        <li>If <code><var>sender</var>.track.kind</code> is not <code>"audio"</code> return
```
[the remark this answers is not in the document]
aboba: Yes.

stefhak:
I did the easy part of what @jan-ivar proposed, and also removed a check not needed (there will only be a DTMF sender if kind is audio, so no need to check if kind is audio).

webrtc.html
stefhak: This is getting complex. I guess that [[SendCodecs]] would be updated in the process of `set an RTCSessionDescription`so we would need to put in language there as well.

stefhak:
I changed to check via `[[SendEncodings]]` if a) the first encoding is active and b) it contains the telephony-events coded. Would this work? PTAL @aboba and @jan-ivar

aboba:
Elsewhere we throw if `encodings[].codec.mimeType` is "audio/telephone-event".  So I don't think this would work.

stefhak:
Alternative for determining if a dtmf codec has been negotiated. Modeled after https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-getparameters. @aboba @jan-ivar PTAL.

webrtc.html
```diff
@@ -9866,10 +9866,9 @@ <h3>canInsertDTMF algorithm</h3>
         nor <code>"sendonly"</code> return <code>false</code> and abort these steps.</li>
         <li>If <var>sender</var>'s <a>[[\SendEncodings]]</a><code>[0].active</code> is
         <code>false</code> return <code>false</code> and abort these steps.</li>
-        <li>If there is no value of <var>j</var> ranging from 0 to the number of
-        codecs, for which <var>sender</var>'s 
-        <a>[[\SendEncodings]]</a><code>[0].codecs[<var>j</var>].mimetype</code>
-        has the value <code>"audio/telephone-event"</code>, return <code>false</code>
+        <li>If no codec with mimetype <code>"audio/telephone-event"</code> has been
+        negotiated for sending, and which the user agent is currently
+        capable of sending, return <code>false</code>
```
jan-ivar: Nit: I find it hard to parse the negated *"and"* here. Do we mean *"or the user agent is not currently capable of sending"*? Not sure how to phrase better.

stefhak: I made an attempt with https://github.com/w3c/webrtc-pc/pull/1980/commits/33281cdb3207a57a3d10a6ea9264f6db977d536e.

webrtc.html
```diff
@@ -9849,6 +9847,31 @@ <h2>Methods</h2>
...
+        return <code>false</code> and abort these steps.</li>
+        <li>If <var>sender</var>'s <a>[[\SenderTrack]]</a> is <code>null</code>
+        return <code>false</code> and abort these steps.</li>
+        <li>If <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> is neither <code>"sendrecv"</code>
+        nor <code>"sendonly"</code> return <code>false</code> and abort these steps.</li>
+        <li>If <var>sender</var>'s <a>[[\SendEncodings]]</a><code>[0].active</code> is
+        <code>false</code> return <code>false</code> and abort these steps.</li>
+        <li>If, for <var>sender</var>, no codec (which the user agent is currently capable
+        of sending) with mimetype <code>"audio/telephone-event"</code> has been
+        negotiated for sending, return <code>false</code> and abort these steps.</li>
```
jan-ivar: How about: *"Of the codecs the user agent is currently capable of sending, if no codec with mimetype `"audio/telephone-event"` has been negotiated for sending with this* sender *, return <code>false</code>"* ?

webrtc.html
```diff
@@ -9849,6 +9847,31 @@ <h2>Methods</h2>
...
+        <li>Return <code>true</code>.</li>
```
jan-ivar: If  we said *"Otherwise, return <code>true</code>."* here, could we remove *"and abort these steps"* from all preceeding steps?

stefhak: Fixed in https://github.com/w3c/webrtc-pc/pull/1980/commits/27fd6925776df0d655589bbf8ab06110e6763f88.

webrtc.html
stefhak: I took an easy way out in https://github.com/w3c/webrtc-pc/pull/1980/commits/27fd6925776df0d655589bbf8ab06110e6763f88. I removed the `is capable of sending` from the equation since it sort of falls into the category of "if there is any other reason why the UA can't send" which we discussed in https://github.com/w3c/webrtc-pc/pull/1957#discussion_r212790055 and said we would not have. Would https://github.com/w3c/webrtc-pc/pull/1980/commits/27fd6925776df0d655589bbf8ab06110e6763f88 work for you @aboba and @jan-ivar ?

Issue #1983, opened 2018-09-12
getSynchronizationSources (and getContriburingSources) should work for video too


henbos:
Multiple things in the spec hint that this is only for audio RTP streams, e.g.

- "When the first audio frame contained in an RTP packet is delivered to the RTCRtpReceiver's MediaStreamTrack for playout, the user agent MUST queue a task to update the relevant information..."
- audioLevel says that audioLevel must be calculated based on the audio data even if the header extension containing this value is missing, concluding that the member must never be absent for SSRCs. This would not be a valid conclusion if getSynchronizationSources() was applicable to video RTP streams too.

The spec should be updated to support both audio and video cases. The "source" (SSRC or CSRC) and timestamp are both valuable pieces of information for a receiver, whether receiving audio or video. For example, they can be used to figure out whose ("source" used as an identifier) stream is currently being received or if a stream is "active" (if timestamp is recent). These APIs are designed to work in real-time. Trying to get this information through other means such as getStats() is not performance-viable.

Any reason this was written for only audio in mind?

henbos:
@fippo  was wondering why you don't just look at MediaStreamTrack.muted to figure out if the stream is active or not but while unmuting happens as a result of an RTP packet arriving, muting only happens as a result of negotiation (setRemoteDescription's muteTracks list).

If you want timestamps for RTP packets to determine how active your stream is or what its "source" is there already is an appropriate API, so let's enable this for video too. It already takes care of the the fact that we can only update this info per task execution cycle or else we expose the fact that it is not single-threaded.

fippo:
> but while unmuting happens as a result of an RTP packet arriving, muting only happens as a result of negotiation (setRemoteDescription's muteTracks list).
I do not think that is true, see https://w3c.github.io/webrtc-pc/#mediastreamtrack-network-use
And I am not happy with muted flipflopping for both network and api events.

getSynchronizationSources() and getContributingSources() was built for audio-levels mainly (which is also the reason it does not make much sense for video -- i do not think any video rtp mixer uses csrcs...). If we want a timestamp for last-packet-received we should not use this API but expose something on the receiver.

fippo:
what getContributingSources was designed for is this (jitsi, probably from around 2011):
![csrc](https://user-images.githubusercontent.com/289731/45491114-2bdaaa80-b769-11e8-8c49-48b60e0437ca.jpg)
(from wikipedia via @emcho)
[...]

PR #1986, opened 2018-09-13
getSynchronizationSources/getContributingSources for video as well

henbos:
Fix for #1983.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/1986.html" title="Last updated on Sep 13, 2018, 11:27 AM GMT (c9e37cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1986/ae0d095...henbos:c9e37cc.html" title="Last updated on Sep 13, 2018, 11:27 AM GMT (c9e37cc)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6907,10 +6907,11 @@ async function updateParameters() {
       respectively, including the most recent time a
       packet that the source contributed to was played out. The browser MUST
       keep information from RTP packets received in the previous 10 seconds.
-      When the first audio frame contained in an RTP packet is delivered to the
-      <code><a>RTCRtpReceiver</a></code>'s <code><a>MediaStreamTrack</a></code>
-      for playout, the user agent MUST queue a task to update the relevant
-      information for the <code><a>RTCRtpContributingSource</a></code> and
+      When the first audio or video frame contained in an RTP packet is
+      delivered to the <code><a>RTCRtpReceiver</a></code>'s
+      <code><a>MediaStreamTrack</a></code> for playout, the user agent MUST
+      queue a task to update the relevant information for the
+      <code><a>RTCRtpContributingSource</a></code> and
       <code><a>RTCRtpSynchronizationSource</a></code> dictionaries based on the
       contents of the packet. The information relevant to the
       <code><a>RTCRtpSynchronizationSource</a></code> dictionary corresponding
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6907,10 +6907,11 @@ <h2>Methods</h2>
       respectively, including the most recent time a
       packet that the source contributed to was played out. The browser MUST
       keep information from RTP packets received in the previous 10 seconds.
-      When the first audio frame contained in an RTP packet is delivered to the
-      <code><a>RTCRtpReceiver</a></code>'s <code><a>MediaStreamTrack</a></code>
-      for playout, the user agent MUST queue a task to update the relevant
-      information for the <code><a>RTCRtpContributingSource</a></code> and
+      When the first audio or video frame contained in an RTP packet is
+      delivered to the <code><a>RTCRtpReceiver</a></code>'s
+      <code><a>MediaStreamTrack</a></code> for playout, the user agent MUST
```
jan-ivar: This would mandate browsers MUST implement this for video as well as audio. Wouldn't this require list discussion?

Firefox currently implements this API for audio only. cc @na-g

aboba: getContributingSources() was first implemented in ORTC both for audio and video.  So not sure how this was interpretted as "audio only" in WebRTC 1.0.

stefhak: FWIW I'd agree to @jan-ivar here. I _think_ we made a conscious decision to only do audio once, and should discuss if we revisit.

henbos: This is basically just exposing RTP timestamps, which is not inherently audio-only, and is useful for group calls. I do propose we mandate browsers must implement it, I also think it would be low hanging fruit to do so (in third_party/webrtc it's mostly a few lines of copy-paste). The "source" and "timestamp" is enough to:
- Know if stream is active or not in in a way that _muted_ inherently does not tell you.
- Timestamps in real-time in a way that getStats() cannot give you (too heavy-weight).
- For large meetings where streams are not always active (whether SSRCs or CSRCs), this allows you to know which streams should be prioritized e.g. shown in the UI.

stefhak: I agree on the usage, and I personally have no opinion either way. If implementors doesn't object it sounds like a good idea to add IMHO. But I think it's a decision for @aboba and @alvestrand whether this should just be added, or if any discussion is needed.

jan-ivar: > This is basically just exposing RTP timestamps

It's not the RTP timestamp, it's playout time. See https://github.com/w3c/webrtc-pc/issues/1983#issuecomment-425080297

henbos:
To be updated to say timestamp is receive timestamp for video, not playout timestamp. Will update PR, maybe add an example

webrtc.html
henbos: My bad, it's playout timestamp (not RTP timestamp). This PR already reflects this based on existing language.

Based on follow-up discussion at https://github.com/w3c/webrtc-pc/issues/1983, is there anything that needs to change with the current PR or should we merge it?

jan-ivar: Lgtm.

jan-ivar:
To clarify (since comments show out of order), we went with playout timestamp for video as well.

henbos:
I filed https://github.com/web-platform-tests/wpt/issues/13364

Issue #1961, opened 2018-08-13
Create separate SVC document


aboba:
At the face-to-face meeting in Stockholm, the WG agreed to move simulcast functionality to a separate document.  This is a placeholder issue for that work item.

aboba:
A snapshot of a potential separate simulcast document is here: 
https://rawgit.com/aboba/webrtc-sim/master/simulcast.html

There are several downsides to this approach:

- Simulcast implementers would need to read both webrtc-pc and this document, and refer back and forth between them. 
- Simulcast functionality would lose its CR status until the scalable video coding (SVC) material was ready for CR. 

For these reasons, it may make more sense to reorganize the simulcast material in the existing document and create a separate extension focusing exclusively on SVC.

aboba:
Document exists: https://github.com/w3c/webrtc-svc

PR #1991, opened 2018-09-21
Extension specification for SVC support

aboba:
Work-in-progress, do not merge. 

Fix for Issue https://github.com/w3c/webrtc-pc/issues/1961

Spec (recently updated to utilize scalability modes instead of encoding dependencies) is available for inspection here:
https://rawgit.com/aboba/webrtc-sim/master/svc.html

### DIFF
--- a/svc.html
+++ b/svc.html
@@ -144,17 +144,18 @@
               <code><a>RTCRtpEncodingParameters</a></code> for any given
               <code><a>RTCRtpParameters</a></code> object. Values MUST be composed only
               of alphanumeric characters (a-z, A-Z, 0-9) up to a maximum
-              of 16 characters.</p>
+              of 16 characters. While the <code>encodingId</code> attribute has the same syntax
+              as the <code>rid</code> attribute, these attributes are distinct, since the
+              <code>encodingId</code> attribute does not imply inclusion of a RID header extension
+              within RTP packets.</p>
             </dd>
             <dt><dfn data-idl><code>dependencyEncodingIds</code></dfn> of type <span class=
             "idlMemberType">sequence&lt;<a>DOMString</a>&gt;</span></dt>
             <dd>
               <p>The <code><a>encodingId</a></code>s on which this layer depends. Within
               this specification <code><a>encodingId</a></code>s are permitted only
-              within the same <code><a>RTCRtpEncodingParameters</a></code> sequence. In
-              the future if <a>MST</a> were to be supported, then if searching within an
-              <code>encodings[]</code> sequence did not produce a match, then a global
-              search would be carried out. In order to send scalable video coding
+              within the same <code><a>RTCRtpEncodingParameters</a></code> sequence.
+              In order to send scalable video coding
               (<a>SVC</a>), both the <code>encodingId</code> and
               <code>dependencyEncodingIds</code> are required.</p>
             </dd>
@@ -204,10 +205,10 @@ var encodings = [
 // High resolution base layer has half the input framerate, full resolution
 // Temporal enhancement layers have full input framerate
 var encodings = [
-  {encodingId: 'H0', scaleResolutionDownBy: 2.0, scaleFramerateDownBy: 2.0},
-  {encodingId: 'F0', scaleFramerateDownBy: 2.0}, 
-  {encodingId: 'H1', scaleResolutionDownBy: 2.0, dependencyEncodingIds: ['H0']},
-  {encodingId: 'F1', dependencyEncodingIds: ['F0']}
+  {rid: 'H', encodingId: 'H0', scaleResolutionDownBy: 2.0, scaleFramerateDownBy: 2.0},
+  {rid: 'F', encodingId: 'F0', scaleFramerateDownBy: 2.0}, 
+  {rid: 'H', encodingId: 'H1', scaleResolutionDownBy: 2.0, dependencyEncodingIds: ['H0']},
+  {rid: 'F', encodingId: 'F1', dependencyEncodingIds: ['F0']}
 ];
                     </pre>
           <p>Below is a representation of 2-layer temporal scalability combined with


murillo128:
@aboba I have one doubt as this doc doesn't reflect any of the ideas that were discussed in the mailing list. 

Is this document a placeholder for starting the discussion or is it the final proposal?

svc.html
```diff
@@ -0,0 +1,347 @@
...
+    multiple RTP streams with a distinct SSRC for each layer. This specification
+    only supports <a>SRST</a> transport.</p>
+    <p>Within this specification it is assumed that scalable video coding operation is primarily
+    controlled utilizing the <code>addTransceiver</code> method of the <code><a>RTCPeerConnection</a></code>
+    and the <code>setParameters</code> method of the <code><a>RTCRtpSender</a></code> object.</p>
+    <p>The <code>addTransceiver</code> method establishes the <dfn>SVC envelope</dfn> which includes
+    the maximum number of scalable video coding layers that can be sent, as well as the ordering of the <code>encodings</code>.
+    While characteristics of individual scalable video coding layers can be modified using the <code>setParameters</code>
+    method, the <a>SVC envelope</a> cannot be changed.  One of the implications of this model is that
+    the <code>addTrack</code> method is not suitable for use with scalable video coding since this method
```
henbos: Tangent: Huh, I guess you're right. Wouldn't it make sense for addTrack to initialize with a default encoding parameters dictionary? I would not expect setParameters() to fail for non-simulcast cases if the sender was created by addTrack and not addTransceiver.

svc.html
```diff
@@ -0,0 +1,347 @@
...
+    configuration of an <code><a>RTCRtpTransceiver</a></code> with an <a>SVC envelope</a>.</p>
+    <p>Note that the inability to modify the <a>SVC envelope</a> using <code>setParameters</code> does not restrict
+    application flexibility in controlling the number of scalable video coding layers that are sent, or the characteristics of
+    those layers. For example, using <code>setParameters</code> individual SVC layers can be made inactive by setting the <code>active</code>
+    attribute to <code>false</code>, or reactivated by setting the <code>active</code>
+    attribute to <code>true</code>.</p> 
+    <p>This specification does not define how to configure <code>createOffer</code> to
+    receive scalable video coding. Codecs such as VP8, VP9 or AV1 require a compliant decoder
+    to be able to decode anything that an encoder can send so that configuration is not required
+    to enable an <code><a>RTCRtpReceiver</a></code> to decode scalable video coding.</p>
```
henbos: Will this be revisited?

svc.html
```diff
@@ -0,0 +1,347 @@
...
+            </dd>
+            <dt><dfn data-idl><code>dependencyEncodingIds</code></dfn> of type <span class=
+            "idlMemberType">sequence&lt;<a>DOMString</a>&gt;</span></dt>
+            <dd>
+              <p>The <code><a>encodingId</a></code>s on which this layer depends. Within
+              this specification <code><a>encodingId</a></code>s are permitted only
+              within the same <code><a>RTCRtpEncodingParameters</a></code> sequence.
+              In order to send scalable video coding
+              (<a>SVC</a>), both the <code>encodingId</code> and
+              <code>dependencyEncodingIds</code> are required.</p>
```
henbos: What does it mean that one layer depends on another? Can you create invalid dependencies?

svc.html
```diff
@@ -0,0 +1,347 @@
...
+          <dl data-link-for="RTCRtpEncodingParameters" data-dfn-for=
+          "RTCRtpEncodingParameters" class="dictionary-members">
+            <dt><dfn data-idl><code>scaleFramerateDownBy</code></dfn> of type <span class=
+            "idlMemberType"><a>double</a></span></dt>
+            <dd>
+              <p>Inverse of the input framerate fraction to be encoded. Example: 1.0 =
+              full framerate, 2.0 = one half of the full framerate. For scalable video
+              coding, <code>scaleFramerateDownBy</code> refers to the inverse of the aggregate
+              fraction of input framerate achieved by this layer when combined with all
+              dependent layers.</p>
```
henbos: What happens if you use weird values like 0.1 or 42.33?

svc.html
```diff
@@ -0,0 +1,347 @@
...
+  {encodingId: 'T1', scaleFramerateDownBy: 2.0, dependencyEncodingIds: ['T0']},
+  {encodingId: 'T2', dependencyEncodingIds: ['T0', 'T1']} 
+];
+
+// Example of 3-layer temporal scalability with all but the base layer disabled
+var encodings = [
+  {encodingId: 'T0', scaleFramerateDownBy: 4.0, active: true}, 
+  {encodingId: 'T1', scaleFramerateDownBy: 2.0, dependencyEncodingIds: ['T0'], active: false},
+  {encodingId: 'T2', dependencyEncodingIds: ['T0', 'T1'], active: false}
+];
```
henbos: Would it be equally valid to list these in the reverse order (scaleFramerateDownBy increasing instead of decreasing)?

svc.html
```diff
@@ -0,0 +1,347 @@
...
+  {encodingId: 'T0', scaleFramerateDownBy: 4.0},
+  {encodingId: 'T1', scaleFramerateDownBy: 2.0, dependencyEncodingIds: ['T0']},
+  {encodingId: 'T2', dependencyEncodingIds: ['T0', 'T1']} 
+];
+
+// Example of 3-layer temporal scalability with all but the base layer disabled
+var encodings = [
+  {encodingId: 'T0', scaleFramerateDownBy: 4.0, active: true}, 
+  {encodingId: 'T1', scaleFramerateDownBy: 2.0, dependencyEncodingIds: ['T0'], active: false},
+  {encodingId: 'T2', dependencyEncodingIds: ['T0', 'T1'], active: false}
```
henbos: What happens if I only specify a dependency on T1? Or only T0?
What happens if I create a circular dependency?
What happens if I don't specify the dependency?
What if I specify a dependency T0 -> T2 instead of T2 -> T0?
Etc.

svc.html
```diff
@@ -0,0 +1,347 @@
...
+    to be able to decode anything that an encoder can send so that configuration is not required
+    to enable an <code><a>RTCRtpReceiver</a></code> to decode scalable video coding.</p> 
+  </section>
+  <section id="rtpdictionaries">
+    <h2>Dictionary extensions</h2>
+    <section id="rtcrtpencodingparameters">
+      <h3>RTCRtpEncodingParameters Dictionary Extensions</h3>
+      <div>
+        <pre class="idl">partial dictionary RTCRtpEncodingParameters {
+             double              scaleFramerateDownBy;
```
henbos: Does scaleResolutionDownBy belong here too?

svc.html
```diff
@@ -0,0 +1,347 @@
...
+// Example of 2-layer spatial scalability combined with 2-layer temporal scalability
+// Base layer has one half input framerate and half resolution
+// Temporal enhancement layer has full input framerate, half resolution
+// Spatial enhancement to the base layer has half input framerate, full resolution
+// Temporal enhancement to the spatial enhancement layer has full framerate and resolution
+var encodings = [
+  {encodingId: 'H0', scaleResolutionDownBy: 2.0, scaleFramerateDownBy: 2.0}, 
+  {encodingId: 'H1', scaleResolutionDownBy: 2.0, dependencyEncodingIds: ['H0']},
+  {encodingId: 'F0', scaleFramerateDownBy: 2.0, dependencyEncodingIds: ['H0']},
+  {encodingId: 'F1', dependencyEnodingIds: ['F0', 'H1']}
```
henbos: Even more combinations of dependency and weird values, what are valid and what are invalid setups? What happens if I do setParameters() with weird values?

svc.html
```diff
@@ -0,0 +1,347 @@
...
+            <figcaption>
+              2-layer spatial and temporal scalability encoding
+            </figcaption>
+          </figure>
+        </section>
+   </section>
+   </section>
+    <section id="privacy-security">
+    <h2>Privacy and Security Considerations</h2>
+    <p>This section is non-normative; it specifies no new behaviour, but
```
henbos: nit: "behavior" instead of "behaviour"? (webrtc-pc spelling is 12 vs 2)

svc.html
```diff
@@ -0,0 +1,347 @@
...
+    <section>
+      <h2>Impact on same origin policy</h2>
+      <p>This API enables data to be communicated between
+      browsers and other devices, including other browsers.</p>
+      <p>This means that data can be shared between applications
+      running in different browsers, or between an application running in the
+      same browser and something that is not a browser.  This is an extension
+      to the Web model which has had barriers against sending data
+      between entities with different origins.</p>
+      <p>This specification provides no user prompts or chrome indicators
```
henbos: What does "chrome indicators" mean?

svc.html
murillo128: Yes, quite easily. VP9 has a fixed dependency structure where each temporal or spatial layer depends only on the immediate lower one.

```
   Within a picture, a frame with spatial layer ID equal to SID, where
   SID > 0, can depend on a frame of the same picture with a lower
   spatial layer ID.  This "inter-layer" dependency can result in
   additional coding gain compared to the case where only traditional
   "inter-picture" dependency is used, where a frame depends on
   previously coded frame in time.  For simplicity, this payload format
   assumes that, within a picture and if inter-layer dependency is used,
   a spatial layer SID frame can depend only on the immediately previous
   spatial layer SID-1 frame, when S > 0.  Additionally, if inter-
   picture dependency is used, a spatial layer SID frame is assumed to
   only depend on a previously coded spatial layer SID frame
```

svc.html
```diff
@@ -0,0 +1,347 @@
...
+    attribute to <code>false</code>, or reactivated by setting the <code>active</code>
+    attribute to <code>true</code>.</p> 
+    <p>This specification does not define how to configure <code>createOffer</code> to
+    receive scalable video coding. Codecs such as VP8, VP9 or AV1 require a compliant decoder
+    to be able to decode anything that an encoder can send so that configuration is not required
+    to enable an <code><a>RTCRtpReceiver</a></code> to decode scalable video coding.</p> 
+  </section>
+  <section id="rtpdictionaries">
+    <h2>Dictionary extensions</h2>
+    <section id="rtcrtpencodingparameters">
```
murillo128: If SVC parameters extend the RTCRtpEncodingParameters, how do the following attributes apply to SVC:

- codecPayloadType;
- priority
- rid

svc.html
aboba: @pthatcher An alternative is to enable a choice between “canned” scalability structures, such as those described in AV1 Section 6.7.5. This limits the ability to specify invalid structures and could also support K-SVC.

youennf:
As I see it, the current proposal empowers applications to precisely tune the configuration.
I am wondering whether something less powerful but simpler could not cover most use cases.
Having less expressiveness has a few advantages, a user agent might better react if something goes wrong like CPU overload.

Taking example 6.1 of a temporal-only scalable codec as an example, do we need to support both examples, or could we just decide to only support the first one for instance?

I haven't looked at what current scalable codec implementations expose in terms of API.
Do they all expose the level of granularity required by the current proposal?

SVC has quality scalability. Is that something we should consider to expose?

svc.html
aboba: The assumption that a "decoder can decode anything an encoder can send" is valid for VP8 and VP9. It also appears to be valid for AV1, since support for SVC tools is required at all profile levels.

svc.html
aboba: @henbos The proposal has been updated to utilize "scalability modes".  This limits the ability to specify invalid scalability dependencies.

aboba:
> @aboba I have one doubt

The original proposal has been replaced by a new one which utilizes "scalability modes". This simplifies configuration considerably while still potentially supporting all popular configurations.

aboba:
Closing this PR, since it has been replaced by a new proposal.

Issue #1888, opened 2018-06-11
RTCPriorityType is not documented for simulcast


ibc:
When using simulcast, N encodings must be provided to `RtpSender.send()`. Each encoding may have a different `priority` field with value "very-low", "low", "medium" or "high".

As per https://w3c.github.io/webrtc-pc/#dom-rtcprioritytype:

> low
> See [RTCWEB-TRANSPORT], Section 4. Corresponds to "normal" as defined in [RTCWEB-DATA]

In https://tools.ietf.org/html/draft-ietf-rtcweb-transports-17#section-4:

> For media, a "media flow", which can be an "audio flow" or a "video flow", is what [RFC7656] calls a "media source", which results in a "source RTP stream" and one or more "redundancy RTP streams".  This specification does not describe prioritization between the RTP streams that come from a single "media source".

But it happens that simulcast is about RTP streams coming from a **single media source**.

So, to be perfectly clear, the meaning of `RTCPriorityType` is not documented at all. AFAIS people usually assigns "low" priority to the minor bitrate stream and "high" priority" to the largest bitrate stream without too much rationale behind.

alvestrand:
The meaning (what you do if you get this value) is documented in sections 4.1 and 4.2. That's why they are subsections of section 4.

alvestrand:
What people SHOULD be doing in the simulcast case when congestion is expected is to assign high priority to the minor bitrate stream (so that it will always be sent) and low priority to the largest stream (so that its packets will be dropped first).

ibc:
> The meaning (what you do if you get this value) is documented in sections 4.1 and 4.2. That's why they are subsections of section 4.

Sure, but those subsections do not even mention "simulcast" use case.

> What people SHOULD be doing in the simulcast case when congestion is expected is to assign high priority to the minor bitrate stream (so that it will always be sent) and low priority to the largest stream (so that its packets will be dropped first).
The effect of wrong settings is probably mitigated by the fact that nobody does prioritization well (if at all) yet.

Thanks.

taylor-b:
> What people SHOULD be doing in the simulcast case when congestion is expected is to assign high priority to the minor bitrate stream (so that it will always be sent) and low priority to the largest stream (so that its packets will be dropped first).

I don't think I understand the definition of `RTCPriorityType`, then. The two example implementation strategies are:

>    o  When the available bandwidth is known from the congestion control
>       algorithm, configure each codec and each data channel with a
[...]

PR #1995, opened 2018-09-27
Move priority from RTCRtpEncodingParameters to RTCRtpSendParameters

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1888

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/1995.html" title="Last updated on Sep 27, 2018, 10:58 PM GMT (2000e24)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/1995/8878a9c...2000e24.html" title="Last updated on Sep 27, 2018, 10:58 PM GMT (2000e24)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6180,6 +6180,7 @@ async function updateParameters() {
              required DOMString             transactionId;
              required sequence&lt;RTCRtpEncodingParameters&gt;  encodings;
              RTCDegradationPreference       degradationPreference = "balanced";
+             RTCPriorityType                priority = "low";
 };</pre>
         <section>
           <h2>Dictionary <code><a>RTCRtpSendParameters</a></code> Members</h2>
@@ -6208,6 +6209,13 @@ async function updateParameters() {
               resolution or degrading framerate,
               <code>degradationPreference</code> indicates which is
               preferred.</p>
+            </dd>
+            <dt><dfn data-idl><code>priority</code></dfn> of type <span class=
+            "idlMemberType"><a>RTCPriorityType</a></span>, defaulting to
+            <code>"low"</code></dt>
+            <dd>
+              <p>Indicates the priority of this encoding. It is specified in
+              [[!RTCWEB-TRANSPORT]], Section 4.</p>
             </dd>
           </dl>
         </section>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6209,6 +6210,13 @@ <h2>Dictionary <code><a>RTCRtpSendParameters</a></code> Members</h2>
               <code>degradationPreference</code> indicates which is
               preferred.</p>
             </dd>
+            <dt><dfn data-idl><code>priority</code></dfn> of type <span class=
+            "idlMemberType"><a>RTCPriorityType</a></span>, defaulting to
+            <code>"low"</code></dt>
+            <dd>
+              <p>Indicates the priority of this encoding. It is specified in
+              [[!RTCWEB-TRANSPORT]], Section 4.</p>
```
henbos: It is now clear the bitrate of the entire sender, but it is still not specified how much each individual simulcast layer should occupy. I think this is fine, not imposing too much restrictions on the encoders, or how to prioritize between layers.

murillo128: I believe having the priority on the RTCRtpSendParameters is a good thing so browser can decide how much bandwidth it allocates this sender vs the other ones.

But if we are providing a fine grained control for simulcast, I think we should keep the priority on the encodings too to be able to prioritize between the different layers of the simulcast stream and assign bandwidth, drop layers, or apply different protection schemes. 

For example, Chrome always start dropping the highest layer  if there is not enough 
 bandwidth, while an app could decide to maintain the highest and drop the mid layer instead.

alvestrand: I think that if the app wants control over this, then calling it "priority" is the wrong thing to do. Let's design a mechanism for that when we have requirements.

murillo128: Not sure why priority is not the correct mechanism for this. According to RTCWEB-TRANSPORT the priority attribute is used for local prioritization and if bwe is know from the congestion control algorithm, configure each codec and each data channel with a target send rate that is appropriate to its share of the available bandwidth.

Once the target bitrate is set for a sender, same mechanism based on the priority attribute can be used to distribute the available bitrate for the sender within the simulcast streams.

murillo128: I just copied what is in the RTCWEB-TRANSPORT doc already. 

Note that this is not about taking priority on congestion control or bandwidth estimation, it is about how to divide the available bandwidth gathered from the bandwidth estimation between the different senders and withing the different simulcast streams of each sender.

PR #2000, opened 2018-10-04
Define meaning of a zero frame rate

### LINKED ISSUE #1992: Define what happens if maxFramerate is 0?
I don't think we've defined what should happen. OTOH, it seems that the options for the UA are not that many. The only case I can think of when different interpretations could be made would be for the case the UA is about to transmit, or has started to transmit, a frame when `setParameters` is called with 'maxFramerate` being `0.0`. This could be a big one (I-frame), and the options would be to either transmit that frame or abort before finalizing the transmission. Being a corner case means that perhaps we can just leave it as is?

alvestrand:
Fixes #1992

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2000.html" title="Last updated on Oct 9, 2018, 2:32 PM GMT (4dc929f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2000/8878a9c...4dc929f.html" title="Last updated on Oct 9, 2018, 2:32 PM GMT (4dc929f)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6366,7 +6366,7 @@ async function updateParameters() {
               <p>
                 If changed with <code>setParameters</code>, the new framerate
                 takes effect after the current
-                picture is completed; setting the framerate to zero thus has
+                picture is completed; setting the max framerate to zero thus has
                 the effect of freezing the video on the next frame.
               </p>
             </dd>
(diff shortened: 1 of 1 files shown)


aboba:
@alvestrand Should it be "setting the maximum framerate"?

alvestrand:
@aboba yes. Fixed. Ready to merge on Thursday?

PR #2001, opened 2018-10-04
Describing which media description is offerer tagged.

### LINKED ISSUE #1919: Investigate stopping SRD and createAnswer

henbos:
Fixes #1919.

This PR...
- Makes sure that rejected transceivers are marked as stopped in setLocalDescription, as is necessary because the offerer tagged m= section might have been rejected between setRemoteDescription(offer) and setLocalDescription(answer).
- Mandates that the first non-stopped transceiver is picked as the "offerer tagged" m= section. (The BUNDLE spec does not say which one you should pick, it just suggests that you pick one that you think is unlikely be rejected.)
- Clarifies the existing NOTE in RTCRtpTransceiver.stop() that rejected m= sections causes their transceivers to be stopped when applying the answer at setLocalDescription and that this can be avoided by predicting which transceiver is offerer tagged per how it must have been picked.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2001.html" title="Last updated on Oct 4, 2018, 1:59 PM GMT (460bfe2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2001/8878a9c...henbos:460bfe2.html" title="Last updated on Oct 4, 2018, 1:59 PM GMT (460bfe2)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1679,6 +1679,11 @@
                               </li>
                             </ol>
                           </li>
+                          <li>
+                            <p>If the <a>media description</a> is rejected, and
+                            <var>transceiver</var> is not already stopped, <a>stop
+                            the RTCRtpTransceiver</a> <var>transceiver</var>.</p>
+                          </li>
                         </ol>
                       </li>
                     </ol>
@@ -2500,6 +2505,19 @@ interface RTCPeerConnection : EventTarget  {
                     identity assertion from <var>provider</var> (if non-null),
                     generate an SDP offer, <var>sdpString</var>, as described
                     in <span data-jsep="create-offer">[[!JSEP]]</span>.</p>
+                    <p>As described in [[!BUNDLE]] (Section 7), if bundling is
+                    used (see <a>RTCBundlePolicy</a>) an offerer tagged m=
+                    section must be selected in order to negotiate a BUNDLE
+                    group. The user agent MUST choose the m= section that
+                    correspond to the first non-stopped transciever in the
+                    <a>set of transceivers</a> as the offerer tagged m=
+                    section. This allows the remote endpoint to predict which
+                    transciever is the offerer tagged m= section without having
+                    to parse the SDP. This is relevant when rejecting an m=
+                    section by stopping an
+                    <code><a>RTCRtpTransceiver</a></code>, see the NOTE in
+                    <code><a>RTCRtpTransceiver</a>.stop()</code> for more
+                    information.</p>
                   </li>
                   <li>
                     <p>Let <var>offer</var> be a newly created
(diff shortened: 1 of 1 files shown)


henbos:
TODO: Split up

alvestrand:
@jan-ivar to review a bit more.
@henbos will split this up a bit.

webrtc.html
```diff
@@ -1679,6 +1679,11 @@ <h4>Set the RTCSessionDescription</h4>
                               </li>
                             </ol>
                           </li>
+                          <li>
+                            <p>If the <a>media description</a> is rejected, and
+                            <var>transceiver</var> is not already stopped, <a>stop
+                            the RTCRtpTransceiver</a> <var>transceiver</var>.</p>
```
jan-ivar: Can you point me to the normative support for executing this line which is in both in SLD(offer) and SLD(answer)?

Especially for SLD(offer), it seems inconsistent with other local effects of calling [stop()](http://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-stop), which are that receiving tracks are ended immediately on calling *stop()*, not deferred until next *setLocalDescription(offer)*.

This seems inconsistent with the existing note under the stop method, which only talks about calls to stop() between SRD(offer) and SLD(answer), though I confess I'm having a hard time understanding that note as well, and need to read up more on the BUNDLE spec.

henbos: > Can you point me to the normative support for executing this line which is in both in SLD(offer) and SLD(answer)?
> 
> Especially for SLD(offer), it seems inconsistent with other local effects of calling stop(), which are that receiving tracks are ended immediately on calling stop(), not deferred until next setLocalDescription(offer).

See https://tools.ietf.org/html/draft-ietf-mmusic-sdp-bundle-negotiation-52#section-7.3.3:

>    When an answerer wants to reject a bundled "m=" section in an answer,
>    it MUST first check the following criterion:
> 
>    o  In the corresponding offer, the "m=" section is the offerer tagged
>       "m=" section.
> 
>    If the criterium above is fulfilled the answerer can not reject the
>    "m=" section in the answer.  The answerer can either reject the whole
>    offer, reject each bundled "m=" section within the BUNDLE group, or
>    keep the "m=" section within the BUNDLE group in the answer and later
>    create an offer where the "m=" section is disabled within the BUNDLE
>    group [Section 7.5.3].

This establishes that we have to reject all-bundled-or-nothing.

JSEP does not seem to know about this, https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-4.2.1:

> 4.2.1.  stop
> 
>    The stop method stops an RtpTransceiver.  This will cause future
>    calls to createOffer to generate a zero port for the associated m=
>    section.  See below for more details.
> 
> 4.2.2.  stopped
> 
>    The stopped property indicates whether the transceiver has been
>    stopped, either by a call to stopTransceiver or by applying an answer
>    that rejects the associated m= section.  In either of these cases, it
>    is set to "true", and otherwise will be set to "false".
> 
>    A stopped RtpTransceiver does not send any outgoing RTP or RTCP or
>    process any incoming RTP or RTCP.  It cannot be restarted.

It only talks about stopping transceivers per-stop() or per-SRD(answer). It mentions setting ports and additional attributes, but does not have a notion of offerer tagged m= sections or this type of rejection.

It references the bundle spec in several places, but never addresses this specifically. I think it is under-specified what to do: the BUNDLE spec says what you are allowed to do, but not what the JSEP APIs should do; JSEP says what the APIs should do, but seems blissfully ignorant about this issue. We have to address it, JSEP should probably be updated too but maybe that is harder?

So what can we do? Here are some options.
1. Reject the whole offer?
2. View this as generating answer-logic: Surface the fact that we just implicitly rejected the other m= sections in association with generating the answer, (this is akin to how we already surface other state changes like setting the transceivers' mids). (This PR.)
3. Modify stop() to say something like if we're in signaling state "have-remote-offer" and the transceiver is offerer tagged, also stop any transceiver that would are bundled and as such would have been stopped according to answerer logic.
4. If we are in the signaling state "have-remote-offer", always throw an exception to avoid the foot-gun, but then you can no longer use stop() between offer and answer.
5. Remove stop() because it's horrid.

I can see how 3) might be a better idea. I'm a fan of 5) but I think I am alone on that one :)

"Which m= section is the offerer tagged one?" is still a can of worms.

alvestrand: I think 3) is what we agreed to do earlier.

alvestrand:
@henbos to clarify that the invasive stop can only happen in SLD(answer) in have-remote-offer state.

henbos:
And to split the PR up

henbos:
Ping @henbos !

henbos:
Filing separate issues corresponding to splitting this up...
#2047

henbos:
#2049

henbos:
I think we're covered

PR #2002, opened 2018-10-04
Add link to "run-to-completion" in terminology

### LINKED ISSUE #1111: Describe update strategy on variables
We should put in wording somewhere that says "all variables are conceptually read when the UA is in a stable state" or something similar - that they dont' change within the frame of a single JS execution.

alvestrand:
Fixes #1111

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2002.html" title="Last updated on Oct 9, 2018, 2:29 PM GMT (6d888b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2002/8878a9c...6d888b1.html" title="Last updated on Oct 9, 2018, 2:29 PM GMT (6d888b1)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -128,6 +128,16 @@
       <dfn>peeridentity</dfn>, <dfn>request an identity assertion</dfn> and
       <dfn>validate the identity</dfn> are defined in [[!WEBRTC-IDENTITY]].
     </p>
+    <p class="note">
+      The general principles for Javascript APIs apply, including the principle
+      of <a href="https://w3ctag.github.io/design-principles/#js-rtc">
+        run-to-completion</a> and no-data-races as defined in [[API-DESIGN-PRINCIPLES]].
+      That is, between microtask checkpoints, external events do
+      not influence what's visible to the Javascript application. For example,
+      the amount of data buffered on a data channel will increase due to
+      "send" calls while Javascript is executing, but will only decrease due
+      to packets being sent at a microtask checkpoint.
+    </p>
   </section>
   <section>
     <h2>Peer-to-peer connections</h2>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -128,6 +128,16 @@ <h2>Terminology</h2>
       <dfn>peeridentity</dfn>, <dfn>request an identity assertion</dfn> and
       <dfn>validate the identity</dfn> are defined in [[!WEBRTC-IDENTITY]].
     </p>
+    <p class="note">
+      The general principles for Javascript APIs apply, including the principle
+      of <a href="https://w3ctag.github.io/design-principles/#js-rtc">
+        run-to-completion</a> and no-data-races as defined in [[API-DESIGN-PRINCIPLES]].
+      That is, between microtask checkpoints, external events do
+      not influence what's visible to the Javascript application. For example,
```
jan-ivar: Actually, "between (micro)task checkpoints" is the only time it's allowed, so this is backwards. Also, not sure we should limit it to microtasks. Microtask checkpoints are usually specific to promise-resolutions, whereas most UA changes come in the form of "queued tasks" which are regular (macro) tasks, so it's both.

If we need to paraphrase at all here, we could use Taylor's suggestion in https://github.com/w3c/webrtc-pc/issues/1111#issuecomment-292211247 seems good:
*"... all readonly attributes must behave as though they return the value of an internal slot, which is only updated inside a queued microtask."*

I think it's worth adding his second part as well, https://github.com/w3c/webrtc-pc/issues/1111#issuecomment-292597563.

webrtc.html
```diff
@@ -128,6 +128,16 @@ <h2>Terminology</h2>
...
+      the amount of data buffered on a data channel will increase due to
+      "send" calls while Javascript is executing, but will only decrease due
+      to packets being sent at a microtask checkpoint.
```
jan-ivar: Technically, "task checkpoint".

webrtc.html
alvestrand: Now I see what's confusing here .... I can't figure out if "between (micro)task checkpoints" should mean "while a (micro)task is running" or "while a (micro)task is NOT running". I don't think I want to use that term.

alvestrand:
@jan-ivar more review requested - does this make sense to you?

PR #2030, opened 2018-11-08
Specify how codec preferences is affected by direction.

### LINKED ISSUE #2006: setCodecPreferences and transceiver direction change
In #1998 , it was made clear that the codec preferences is dependent on the transceiver direction.
What should happen when there are codec preferences and a change of direction on a transceiver?

If the transceiver was marked sendonly and had a codec preference to use a codec that can only be sent, not received and is later changed to recvonly, it is clear the preference won't work anymore.

henbos:
Fixes #2006 

This clarifies what otherwise has to be guessed at from reading this spec or the JSEP spec.
In clarifying what to do, this also adds steps to the setCodecPreferences algorithm, including throwing an exception on invalid parameters.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2030.html" title="Last updated on Nov 29, 2018, 11:57 AM GMT (ab85855)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2030/a3c1358...henbos:ab85855.html" title="Last updated on Nov 29, 2018, 11:57 AM GMT (ab85855)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2368,7 +2368,7 @@ interface RTCPeerConnection : EventTarget  {
                 <p>Creating the SDP MUST follow the appropriate process for
                 generating an offer described in [[!JSEP]].
                 As an offer, the generated SDP will contain the full set of
-                codec/RTP/RTCP capabilities supported by the session (as
+                codec/RTP/RTCP capabilities supported or preferred by the session (as
                 opposed to an answer, which will include only a specific
                 negotiated subset to use). In the event
                 <code>createOffer</code> is called after the session is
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7374,6 +7443,72 @@ <h2>Methods</h2>
...
+                  <p>Let <var>kind</var> be the <var>transceiver</var>'s
+                  <a>transceiver kind</a>.</p>
+                </li>
+                <li>
+                  <p>If the intersection between <var>codecs</var> and
+                  <code>RTCRtpSender.getCapabilities(kind).codecs</code> or the intersection
+                  between <var>codecs</var> and
+                  <code>RTCRtpReceiver.getCapabilities(kind).codecs</code> is an empty set,
+                  throw <code>InvalidModificationError</code>. This ensures that we always have
+                  something to offer, regardless of <code><a>transceiver</a><a>direction</a></code>.</p>
```
henbos: If we don't do this then we would have to consider throwing an exception when you set "direction" or when doing createOffer() or createAnswer(); or we would have to have a fallback-plan at createOffer/Answer to say that if filtering based on direction results in an empty list of preferences, act as if no preferences was set.

webrtc.html
```diff
@@ -7374,6 +7443,72 @@ <h2>Methods</h2>
...
+              <ol>
+                <li>
+                  <p>Let <var>transceiver</var> be the <code><a>RTCRtpTransceiver</a></code>
+                  object this method was invoked on.</p>
+                </li>
+                <li>
+                  <p>Let <var>codecs</var> be the first argument.</p>
+                </li>
+                <li>
+                  <p>If <var>codecs</var> is an empty list, set set <var>transceiver</var>'s
```
Orphis: Typo: set set

henbos: Done.

henbos:
Please take a look, @aboba

Issue #2031, opened 2018-11-14
Partially connected peerconnection might have "new" connectionstate.


jonasesolsson:
Assume we have a peerconnection with a since ice transport and a dtls transport running on top of it. 
During setup we might have these steps:

ice transport | dtls transport | connectionState
--- | --- | ---
new | new | new
checking | new | connecting
connected | new | new <---
connected | connecting | connecting
connected | connected | connected

This seems really unintuitive, I'd expect the connectionState to stay as "connecting". In https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectionstate would it make sense to move the "some connected, some new" case from "new" to "connecting"?

A similar thing might happen to the ice connection state if there's multiple ice transports and one connects while another remains "new", but that seems like more of an edge case.

alvestrand:
Yes, this state should map to "connecting".
The text for "new" is:

Any of the RTCIceTransports or RTCDtlsTransports are in the "new" state and none of the transports are in the "connecting", "checking", "failed" or "disconnected" state, or all transports are in the "closed" state, or there are no transports.

Suggested fix: Add "connected" to the list of states after "none of the transports are".

fippo:
I think this highly depends on when the dtls transport changes from new to connecting. See the definition [here](https://w3c.github.io/webrtc-pc/#dom-rtcdtlstransportstate)
If it starts the process of negotiating at the same time as the ice transport this can not happen.
If it waits for the ice transport to finish and only moves to connecting once the first dtls hello is sent/received then this situation can happen.

@pthatcherg did i forget something? I lost the hand-written note unfortunately :-)

alvestrand:
My interpretation of https://w3c.github.io/webrtc-pc/#dom-rtcdtlstransportstate is that it's "new" until it sends the ClientHello / receives the ClientHello, and remains in "connecting" until the Finished message is received. (RFC 5246 section 7).
If this is true, the question becomes "is it possible that sending ClientHello can be delayed until after the ICE Transport is "connected"?
of course, if you don't have a buffer to store the ClientHello in while you wait for ICE to become connected, the answer is trivially "yes".

alvestrand:
@jonasesolsson I dont' understand why connected ICE + closed DTLS should be "connected" - I would think it would be "closed".
The table is just text - if we can make the text clearer by calling out things before or after the table, we should do so!

jonasesolsson:
[...]

PR #2036, opened 2018-12-03
Cleanup connection states.

jonasesolsson:
Here's the changes suggested in https://github.com/w3c/webrtc-pc/issues/2031.

These are the effective changes:

- In RTCPeerConnectionState we now prefer "disconnected" over "connecting", that is to say, we no longer do the 'and none of them are in the "connecting" or "checking"' thing. This was done to ensure that the two connection states behave similarly, unlike now.
- Neither of the "new" states allow transports to be "connected" or "completed" any more. The states that would have previously ended up covered here are now "connected" instead.

***
<a href="https://pr-preview.s3.amazonaws.com/jonasesolsson/webrtc-pc/pull/2036.html" title="Last updated on Dec 3, 2018, 3:20 PM GMT (7f406d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2036/3b4dcc4...jonasesolsson:7f406d3.html" title="Last updated on Dec 3, 2018, 3:20 PM GMT (7f406d3)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -826,12 +826,12 @@
         <h4><dfn>RTCPeerConnectionState</dfn> Enum</h4>
         <div>
           <pre class="idl">enum RTCPeerConnectionState {
+    "closed",
+    "failed",
+    "disconnected",
     "new",
     "connecting",
-    "connected",
-    "disconnected",
-    "failed",
-    "closed"
+    "connected"
 };</pre>
           <table data-link-for="RTCPeerConnectionState" data-dfn-for=
           "RTCPeerConnectionState" class="simple">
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -840,50 +840,48 @@ <h4><dfn>RTCPeerConnectionState</dfn> Enum</h4>
...
-                <td><dfn data-idl><code>connected</code></dfn></td>
-                <td>All <code><a>RTCIceTransport</a></code>s and
+                <td><dfn data-idl><code>disconnected</code></dfn></td>
+                <td>None of the previous states apply and any
+                <code><a>RTCIceTransport</a></code>s or
                 <code><a>RTCDtlsTransport</a></code>s are in the
-                <code>"connected"</code>, <code>"completed"</code> or
-                <code>"closed"</code> state and at least one of them is in the
-                <code>"connected"</code> or <code>"completed"</code> state.</td>
+                <code>"disconnected"</code> state.</td>
```
henbos: Me thinking out loud:

So if you have two transports, one connected and one failed or disconnected, you would be failed or disconnected even if you do have a connection to the other side?

It sounds weird to be in "failed/disconnected" and still have data flowing to the other endpoint. But if "connecting" is prioritized over "connected" because SOMETHING is connecting, then it makes sense to say that we are "failed/disconnected" because SOMETHING is "failed/disconnected". Maybe it's better to highlight that "something is wrong" than to give a happy state - the caller can inspect individual transports for a full picture.

Otherwise we could have a "partially-connected" state, but I bet people don't want to add more enum values.

henbos: Ok I think I am happy with this. :) I was thinking "what if one of the transports gets stuck in a 'bad' state?" but if that is the case then you want to know about it. And if a transport should be discarded, then we should make sure that the transport is thrown away rather than add hacks to make the enum produce something more helpful.

PR #2037, opened 2018-12-06
Make codecPayloadType read-only

### LINKED ISSUE #2008: Using codecPayloadType with addTransceiver()
The `codecPayloadType` values come from a sender's `getParameters()` call. It seems to me that it isn't then possible to have a `addTransceiver()` call with `sendEncodings` specifying a codec to use (and possibly a scalability mode later).

To force a single codec to be used, I imagine it's possible to use `setCodecPreferences()` with a single entry, but it would prevent negotiation of other codecs, which is less than ideal.

Would the recommended usage be then to:
- Create a generic transceiver with no forced codec in sendEncodings
- Do the negotiation
- Call getParameters() to retrieve the codecPayloadType values
- Call setParameters() to set the codecPayloadType (and possibly the scalability mode)
- Call replaceTrack() to finally start the transmission with the right settings

henbos:
Fixes #2008

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2037.html" title="Last updated on Dec 6, 2018, 12:08 PM GMT (b8841bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2037/3b4dcc4...henbos:b8841bc.html" title="Last updated on Dec 6, 2018, 12:08 PM GMT (b8841bc)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5779,21 +5779,6 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                    <a data-link-for="exception" data-lt="create">created</a>
                    <code>RangeError</code>.</p>
                  </li>
-                 <li>
-                   <p>For each value of <var>i</var> from 0 to the number of encodings,
-                   check whether
-                   <code><var>encodings</var>[<var>i</var>].codecPayloadType</code>
-                   (if set) corresponds to a value of
-                   <code><var>codecs</var>[<var>j</var>].payloadType</code> where
-                   <var>j</var> goes from 0 to the number of codecs. If there is no
-                   correspondence, or if the MIME subtype portion of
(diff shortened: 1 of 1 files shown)


henbos:
The side-effect of this CL is "if you want to decide codec, make sure this is adequately negotiated"

Orphis:
You should probably also update the algorithm to validate `sendEncodings` in `addTransceiver()` and return an error if it is set there.

henbos:
@Orphis This is already covered by "If any RTCRtpEncodingParameters dictionary in sendEncodings contains a read-only parameter other than rid, throw an InvalidAccessError." so there is no need to add any reference to codecPayloadType.

PR #2041, opened 2018-12-11
Refer to whatwg same-origin and current settings to define the certificate origin check

youennf:
Split the checks in substeps to try improving readability.

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/webrtc-pc/pull/2041.html" title="Last updated on Dec 11, 2018, 6:01 AM GMT (45978f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2041/66678f6...youennf:45978f4.html" title="Last updated on Dec 11, 2018, 6:01 AM GMT (45978f4)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1012,14 +1012,30 @@
           </li>
           <li>
             <p>If the <code>certificates</code> value in
-            <var>configuration</var> is non-empty, check that
-            the <code>expires</code> on each value is in the future. If a
-            certificate has expired or a the [[\Origin]] internal slot of
-            the certificate does not match the current origin, <a>throw</a> an
-            <code>InvalidAccessError</code>; otherwise, store the certificates.
-            If no <code>certificates</code> value was specified, one or more
-            new <code>RTCCertificate</code> instances are generated for use
-            with this <code>RTCPeerConnection</code> instance. This MAY happen
+            <var>configuration</var> is non-empty, run the following steps for
+            each certificate in certificates:
+            <ol>
+              <li>
+                <p>If the value of the certificate <code>expires</code>
+                attribute is not in the future, <a>throw</a> an
+                <code>InvalidAccessError</code>.</p> </li>
+              <li>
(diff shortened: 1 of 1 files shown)


youennf:
Two notes:
- The concept of 'in the future' is not well defined. Not a big deal, but it would be nice to link to something more precise
- Store the certificate(s) is not clear either. This may have a potential effect on whether setting the configuration will throw if the certificates are not the same: is order of the certificates important, is it equality of certificate object that is tested or equality of certificate objects internal slots (say I serialize/deserialize a certificate, can I use the new one instead of the old one).

webrtc.html
```diff
@@ -1012,14 +1012,30 @@ <h4>Constructor</h4>
...
-            <code>InvalidAccessError</code>; otherwise, store the certificates.
-            If no <code>certificates</code> value was specified, one or more
-            new <code>RTCCertificate</code> instances are generated for use
-            with this <code>RTCPeerConnection</code> instance. This MAY happen
+            <var>configuration</var> is non-empty, run the following steps for
+            each certificate in certificates:
+            <ol>
+              <li>
+                <p>If the value of the certificate <code>expires</code>
+                attribute is not in the future, <a>throw</a> an
```
alvestrand: specify that the "future" is "larger than performance.timeOrigin + performance.now()"

alvestrand: I'm not so worried about int vs double - what I want to make clear is that nobody expects the browser to somehow magically invent yet another concept of "now" different from the ones it's already using (for instance, performance.now() is a monotonic clock, while performance.timeOrigin is tied to real-world clock, so the sum will drift a little bit away from "true" over time; by using the reference, we're saying that that's OK.) I have the same lack of concern with precisely when we check as long as it's within the current run of the main loop.

youennf: Agreed it should hopefully not matter much.
I was hoping to find some notion of timing related to HTTP cache or certificates, which seem closer to me than performance timing. But performance timing will do the trick too

PR #2064, opened 2019-01-17
Add description of DTLS transport error handling.

### LINKED ISSUE #2044: DTLSTransport: No description of when to fire the error event
It seems that we have no description of when the error event in DTLSTransport is fired.
I believe it should be fired:
- when the fingerprint fails to verify the remote certificate
- when the DTLS connection is closed due to an error at the DTLS layer

I believe it should fire before the state change event, but could argue this either way; I think it's important for compatibility that we say whether it is fired before or after the state change event.

I believe it should NOT fire when there's an error in the ICE transport. All ICE transport errors are theoretically recoverable with an ICE restart, and might result in the DTLS transport being able to continue.

alvestrand:
Fixes #2044

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2064.html" title="Last updated on Feb 1, 2019, 8:28 AM UTC (eb696ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2064/3455fd8...eb696ab.html" title="Last updated on Feb 1, 2019, 8:28 AM UTC (eb696ab)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7738,7 +7738,37 @@ async function onOffHold() {
           at <var>transport</var>.</p>
         </li>
       </ol>
-
+      <p>When the underlying DTLS transport experiences an error, such as
+        a certificate validation failure, or a fatal alert (see RFC 5246
+        section 7.2), the user agent MUST queue a task that runs the following
+        steps:
+      </p>
+      <ol>
+        <li>
+          <p>Let <var>transport</var> be the <code>
+          <a>RTCDtlsTransport</a></code> object to receive the state update.</p>
+        </li>
+        <li>
+          <p>Set <var>transport</var>'s <a>[[\DtlsTransportState]]</a> slot
+            to <code>failed</code>.</p>
+        </li>
+        <li>
+          <p>
+            <a>Fire an event</a> named
+            <code><a data-lt="RTCDtlsTransport error">error</a></code>
+            using the <a>RTCErrorEvent</a> interface with its errorDetail
+            attribute set to either "dtls-failure" or "fingerprint-failure",
+            as appropriate, and
+            other fields set as described under the <a>RTCErrorDetailType</a>
+            enum description, at <var>transport</var>.
+          </p>
+        </li>
+        <li>
+          <p><a>Fire an event</a> named <code><a data-lt="RTCDtlsTransport state change">statechange</a></code>
+          at <var>transport</var>.</p>
+        </li>
+      </ol>
+      
       <div>
         <pre class="idl">[Exposed=Window] interface RTCDtlsTransport : EventTarget {
     [SameObject]
--- a/webrtc.html
+++ b/webrtc.html
@@ -7709,9 +7709,47 @@ async function onOffHold() {
       <dfn>[[\DtlsTransportState]]</dfn> internal slot initialized to <code>
       <a data-link-for="RTCDtlsTransportState">new</a></code> and a
       <dfn>[[\RemoteCertificates]]</dfn> slot initialized to an empty list.</p>
+      <p>When the underlying DTLS transport experiences an error, such as
+        a certificate validation failure, or a fatal alert (see [[RFC5246]]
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7699,7 +7699,37 @@ <h3><dfn>RTCDtlsTransport</dfn> Interface</h3>
           at <var>transport</var>.</p>
         </li>
       </ol>
-
+      <p>When the underlying DTLS transport experiences an error, such as
+        a certificate validation failure, or a fatal alert (see RFC 5246
+        section 7.2), the user agent MUST queue a task that runs the following
+        steps:
+      </p>
```
henbos: nit: Here and elsewhere - There shouldn't be an extra indent here and &lt;/p&gt; should be on the line above

webrtc.html
```diff
@@ -7699,7 +7699,37 @@ <h3><dfn>RTCDtlsTransport</dfn> Interface</h3>
...
+          <p>Set <var>transport</var>'s <a>[[\DtlsTransportState]]</a> slot
+            to <code>failed</code>.</p>
+        </li>
+        <li>
+          <p>
+            <a>Fire an event</a> named
+            <code><a data-lt="RTCDtlsTransport error">error</a></code>
+            using the <a>RTCErrorEvent</a> interface with its errorDetail
+            attribute set to either "dtls-failure" or "fingerprint-failure",
+            as appropriate, and
```
henbos: nit: unnecessary newline

webrtc.html
```diff
@@ -7699,7 +7699,37 @@ <h3><dfn>RTCDtlsTransport</dfn> Interface</h3>
...
+            other fields set as described under the <a>RTCErrorDetailType</a>
+            enum description, at <var>transport</var>.
+          </p>
+        </li>
+        <li>
+          <p><a>Fire an event</a> named <code><a data-lt="RTCDtlsTransport state change">statechange</a></code>
+          at <var>transport</var>.</p>
```
henbos: It's good to get the order of these things well-defined. Note that we already have "When the underlying DTLS transport needs to update the state of the corresponding RTCDtlsTransport object, the user agent MUST queue a task that runs the following steps" though so a stupid reading of the spec might attempt to fire the event twice.

Perhaps the other "statechange" steps should be updated to say "needs to update the state for any other reason than experiencing an error"?

alvestrand: Agreed. Will swap the two sections too.

webrtc.html
```diff
@@ -7699,7 +7699,37 @@ <h3><dfn>RTCDtlsTransport</dfn> Interface</h3>
...
+            <code><a data-lt="RTCDtlsTransport error">error</a></code>
+            using the <a>RTCErrorEvent</a> interface with its errorDetail
+            attribute set to either "dtls-failure" or "fingerprint-failure",
+            as appropriate, and
+            other fields set as described under the <a>RTCErrorDetailType</a>
+            enum description, at <var>transport</var>.
+          </p>
+        </li>
+        <li>
+          <p><a>Fire an event</a> named <code><a data-lt="RTCDtlsTransport state change">statechange</a></code>
```
alvestrand: If the state is not already "failed".....

webrtc.html
henbos: Don't fire the event if you were already failed.

webrtc.html
```diff
@@ -7699,7 +7699,37 @@ <h3><dfn>RTCDtlsTransport</dfn> Interface</h3>
...
-
+      <p>When the underlying DTLS transport experiences an error, such as
+        a certificate validation failure, or a fatal alert (see RFC 5246
+        section 7.2), the user agent MUST queue a task that runs the following
+        steps:
+      </p>
+      <ol>
+        <li>
+          <p>Let <var>transport</var> be the <code>
+          <a>RTCDtlsTransport</a></code> object to receive the state update.</p>
```
alvestrand: s/state update/error notification/

PR #2073, opened 2019-01-21
Effect of RTCRtpSendParameters on Simulcast

### LINKED ISSUE #1964: Effect of RTCRtpSendParameters on Simulcast
The effect of `RTCRtpSendParameters` on simulcast is under-specified.  Some questions: 

1. Is congestion control affected by the value of `RTCRtpSendParameters.degradationPreference`? For example, if `degradationPreference` = `maintain-resolution` does this imply that framerate is lowered before higher resolution simulcast layers are dropped? Does the opposite happen if `degradationPreference` has the value `maintain-framerate`?

2. What effect does `parameters.encodings[].priority` have?  Do the values influence which simulcast layers are dropped? 

3. What effect does `parameters.encodings[].maxBitrate` have?  Do the values of `parameters.encodings[].maxBitrate` influence which simulcast layers are dropped?  If there is insufficient bandwidth, is bandwidth allocated to the layers in proportion to the values of `maxBitrate`? 

4. What effect does `parameters.encodings[].maxFramerate` have? Do the values of `parameters.encodings[].maxFramerate` influence which simulcast layers are dropped?  If there is insufficient bandwidth, is bandwidth allocated to the layers in proportion to the values of `maxFramerate`?

aboba:
Fixes #1964

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2073.html" title="Last updated on Jan 24, 2019, 7:31 PM UTC (f735be8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2073/1084b12...f735be8.html" title="Last updated on Jan 24, 2019, 7:31 PM UTC (f735be8)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7541,6 +7541,30 @@ async function updateParameters() {
         therefore need to be able to correctly order the received packets, recognize potential loss events and
         react to them. Correct operation in this scenario is non-trivial and therefore is optional for
         implementations of this specification.</p>
+        <section id="encodingbehavior">
+           <h3>Parameter behavior</h3>
+           <section id="priorityinsimulcast">
+             <h3>priority</h3>
+             <p><code><a>RTCRtpSendParameters</a></code> dictionary member <code>priority</code>
+             can be utilized to set the priority of an <code><a>RTCRtpSender</a></code>.
+             This enables <code>priority</code> to be used to influence the bandwidth allocation
+             among <code><a>RTCRtpSender</a></code> objects.</p>
+           </section>
+           <section id="maxbitrateinsimulcast">
+             <h3>maxBitrate</h3>
+             <p><code><a>RTCRtpEncodingParameters</a></code> dictionary member <code>maxBitrate</code>
+             can be utilized to set the maximum bitrate of a simulcast stream. The user agent is free
+             to allocate bandwidth between the encodings, as long as the <code>maxBitrate</code>
+             value is not exceeded.</p>
+           </section>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7541,6 +7541,29 @@ <h3>Simulcast functionality</h3>
         therefore need to be able to correctly order the received packets, recognize potential loss events and
         react to them. Correct operation in this scenario is non-trivial and therefore is optional for
         implementations of this specification.</p>
+        <section id="encodingbehavior">
+           <h3>Parameter behavior</h3>
+           <section id="priorityinsimulcast">
+             <h3>priority</h3>
+             <p><code><a>RTCRtpSendParameters</a></code> dictionary member <code>priority</code>
+             sets the priority of an <code><a>RTCRtpSender</a></code>, which influences the
+             bandwidth allocation among <code><a>RTCRtpSender</a></code> objects.</p>
```
henbos: The "priority" is already per-send parameters, not per encodings, so I'm not sure what this subsection is trying to say clarify that is specific to simulcast.

If the UA is free to decide how to allocate between different streams on the same sender, it should say that.
If the UA has to prioritize all streams with the same priority, it should say that.

Right now I'm not sure what the difference is between two senders with one encoding each and one sender with two encodings. Can you clarify?

henbos:
@aboba to move the sections to the parameters sections

Issue #2014, opened 2018-10-24
Not clear how to set # of layers when answering an offer with a track


alvestrand:
The process for reusing an m-line is only referenced in the spec under addTrack:

http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtrack

In particular, it's not referenced under addTransceiver, which is our only API point for saying how many layers we wish to send.

This means that I don't know what decides the number of layers available for sending on a transceiver that was created as a result of an incoming offer.

fippo:
see also https://lists.w3.org/Archives/Public/public-webrtc/2017Apr/0040.html ?

aboba:
@fippo @alvestrand  This seems like quite an important question to answer. In a typical conference the SFU sends the Offer because it alone knows how many participants are in the conference. The SFU will typically not have simulcast configured in its Offer m-lines (because the browser typically won't support receiving simulcast).  Yet the browser may be sending simulcast to the SFU, so it needs to be able to configure how many simulcast streams it will send in the Answer. 

In such a situation, the application will receive a track event providing the transceiver that has been created as a result applying the SFU's Offer in setRemoteDescription(). To configure that transceiver to send simulcast, the only viable mechanism is transceiver.sender.setParameters(). In such a situation, I wouldn't expect that transceiver.sender.getParameters() would return much information of interest since the sender hasn't been configured and transceiver.receiver.getParameters() provides minimal information on what was Offered (only the number of encodings to be received and the rid). 

The part that makes me uncomfortable is that while an Offer to send simulcast can be responded to by an Answer with an indication of willingness to receive, an Answer specifying simulcast sending cannot. So what happens if the browser want to send simulcast to the the SFU, but the SFU doesn't want it to do so? 

Section 5.4.1 says "This specification does not define how to configure createOffer to receive multiple RTP encodings."  I believe this statement should be expanded to apply to createAnswer as well, since there is no receiver.setParameters() method.

alvestrand:
adding a=rid:x lines in the offer from the SFU, and reflecting the number of RIDs in the transceiver configuration presented, would make sense. This requires modifying the setRemoteDescription() step that creates transceivers, I believe - this is a minimal change, and doesn't impact API surface. But someone still needs to write a test to go along with it.

fippo:
We currently send offers including a=simulcast from our SFU to Firefox:
```
a=rid:hi recv
a=rid:mid recv
a=rid:lo recv
[...]

PR #2081, opened 2019-01-22
Clarifying how the simulcast envelope is created.

amithilbuch:
Added text to articulate the difference between the offerer who
calls addTransceiver to define the simulcast envelope and the
answerer who is passive and has the envelope set up for them in a
call to setRemoteDescription.
Fixes: #2014.

***
<a href="https://pr-preview.s3.amazonaws.com/amithilbuch/webrtc-pc/pull/2081.html" title="Last updated on Feb 7, 2019, 8:55 PM UTC (8d76314)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2081/288752e...amithilbuch:8d76314.html" title="Last updated on Feb 7, 2019, 8:55 PM UTC (8d76314)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1785,19 +1785,14 @@
                             <ol>
                               <li>
                                 <p>If <var>description</var> indicates that simulcast is
-                                not supported or desired, then set the
-                                <code><a data-link-for="RTCRtpEncodingParameters">active</a></code>
-                                member to <code>false</code> on all dictionaries in
+                                not supported or desired, then remove all dictionaries in
                                 <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>
                                 except the first one and abort these sub steps.</p>
                               </li>
                               <li>
                                 <p>If <var>description</var> rejects any of the offered layers,
-                                then set the
-                                <code><a data-link-for="RTCRtpEncodingParameters">active</a></code>
-                                member to <code>false</code> on the dictionaries in
-                                <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>
-                                that correspond to rejected layers.</p>
+                                then remove the dictionaries that correspond to rejected layers from
+                                <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>.</p>
                               </li>
                               <li>
                                 <p>Update the paused status as indicated by [[MMUSIC-SIMULCAST]] of
--- a/webrtc.html
+++ b/webrtc.html
@@ -1751,8 +1751,18 @@
                             steps:</p>
                             <ol>
                               <li>
+                                <p>If the <var>description</var> is of type <code>"offer"</code>
+                                and contains a request to receive simulcast, follow the order
+                                of the rid values specified in the simulcast attribute to create
+                                an <code><a>RTCRtpEncodingParameters</a></code> dictionary for
+                                each of the simulcast layers, populating the <a>rid</a> member
+                                according to the corresponding rid value, and let <var>sendEncodings</var>
+                                be the list containing the created dictionaries.
+                                Otherwise, let <var>sendEncodings</var> be an empty list.</p>
+                              </li>
+                              <li>
                                 <p><a>Create an RTCRtpSender</a>, <var>sender</var>,
-                                from the <a>media description</a>.</p>
+                                from the <a>media description</a> using <var>sendEncodings</var>.</p>
                               </li>
                               <li>
                                 <p><a>Create an RTCRtpReceiver</a>,
@@ -1764,6 +1774,38 @@
                                 an <code><a>RTCRtpTransceiverDirection</a></code>
                                 value of <code>"recvonly"</code>, and let
                                 <var>transceiver</var> be the result.</p>
+                              </li>
+                            </ol>
+                          </li>
+                          <li>
+                            <p>If <var>description</var> is of type <code>"answer"</code>
+                            or <code>"pranswer"</code>, and <var>transceiver</var>.
+                            <a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a> .length is
+                            greater than <code>1</code>, then run the following steps:</p>
+                            <ol>
+                              <li>
+                                <p>If <var>description</var> indicates that simulcast is
+                                not supported or desired, then set the
+                                <code><a data-link-for="RTCRtpEncodingParameters">active</a></code>
+                                member to <code>false</code> on all dictionaries in
+                                <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>
+                                except the first one and abort these sub steps.</p>
+                              </li>
+                              <li>
+                                <p>If <var>description</var> rejects any of the offered layers,
+                                then set the
+                                <code><a data-link-for="RTCRtpEncodingParameters">active</a></code>
+                                member to <code>false</code> on the dictionaries in
+                                <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>
+                                that correspond to rejected layers.</p>
+                              </li>
+                              <li>
+                                <p>Update the paused status as indicated by [[MMUSIC-SIMULCAST]] of
+                                each simulcast layer by setting the
+                                <code><a data-link-for="RTCRtpEncodingParameters">active</a></code>
+                                member on the corresponding dictionaries in
+                                <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>
+                                to <code>true</code> for unpaused or to <code>false</code> for paused.</p>
                               </li>
                             </ol>
                           </li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7520,6 +7520,12 @@ <h3>Simulcast functionality</h3>
         implications of this model is that the <code>addTrack</code> method cannot provide simulcast
         functionality since it does not take <code>sendEncodings</code> as an argument, and therefore cannot
         configure an <code><a>RTCRtpTransceiver</a></code> to send simulcast.</p>
+        <p>The answerer cannot set the <a>simulcast envelope</a> directly. Upon calling the 
+        <code>setRemoteDescription</code> method of the <code><a>RTCPeerConnection</a></code> object, 
+        the <a>simulcast envelope</a> should be configured on the <code><a>RTCRtpTransceiver</a></code>
+        based upon the specified <code><a>RTCSessionDescription</a></code>. While layers cannot be removed,
+        they can be marked as inactive by setting the <code>active</code> attribute to <code>false</code> 
+        effectively disabling the layer.</p>
```
henbos: Can you update the setRemoteDescription() steps to make it explicitly part of the algorithms how the encodings is set?

webrtc.html
```diff
@@ -7520,6 +7520,12 @@ <h3>Simulcast functionality</h3>
         implications of this model is that the <code>addTrack</code> method cannot provide simulcast
         functionality since it does not take <code>sendEncodings</code> as an argument, and therefore cannot
         configure an <code><a>RTCRtpTransceiver</a></code> to send simulcast.</p>
+        <p>The answerer cannot set the <a>simulcast envelope</a> directly. Upon calling the 
+        <code>setRemoteDescription</code> method of the <code><a>RTCPeerConnection</a></code> object, 
+        the <a>simulcast envelope</a> should be configured on the <code><a>RTCRtpTransceiver</a></code>
```
henbos: Can you change the language to say "this is what happens" rather than "should"? There is a difference between "SHOULD" and "MUST" and this can make you wonder if this is a requirement or if its a bit wishy-washy.

webrtc.html
henbos: I.e. clarify the "Create an RTCRtpSender, sender, from the media description." at https://w3c.github.io/webrtc-pc/#set-description

webrtc.html
jan-ivar: Ideally, I prefer avoiding passive language, and speak to what the implementer MUST do, instead of speaking to a user/observer. The spec uses both already quite a lot. Just my 2 cents, as I found it particularly hard to understand here.

amithilbuch: i have made changes to make it clearer.

webrtc.html
amithilbuch: Done. please let me know if the added steps are clear enough.

webrtc.html
```diff
@@ -1755,6 +1762,36 @@ <h4>Set the RTCSessionDescription</h4>
                               </li>
                             </ol>
                           </li>
+                          <li>
+                            <p>If <var>description</var> is of type <code>"answer"</code>
+                            or <code>"pranswer"</code>, then run the following steps:</p>
+                            <ol>
+                              <li>
+                                <p>If the offer contained a request to send simulcast and
+                                the answer indicates that simulcast is not supported or
```
jan-ivar: `s/the answer/<var>description</var>/`

webrtc.html
```diff
@@ -1755,6 +1762,36 @@ <h4>Set the RTCSessionDescription</h4>
...
+                                desired by the answerer, then set the
```
jan-ivar: I'd like to remove "by the answerer" here, since we've avoided mentioning offerer/answerer thus far in this algorithm, and thus any outside inferences. We're in a for-loop here where the operands are *description* and *transceiver*.

webrtc.html
```diff
@@ -1755,6 +1762,36 @@ <h4>Set the RTCSessionDescription</h4>
...
+                                <code><a data-link-for="RTCRtpEncodingParameters">active</a></code>
+                                flag to <code>false</code> on all the <var>sendEncodings</var>
```
jan-ivar: `s/flag/member/`

webrtc.html
```diff
@@ -1755,6 +1762,36 @@ <h4>Set the RTCSessionDescription</h4>
...
+                                except the first, in the <code><a>RTCRtpSender</a></code>
```
jan-ivar: "on all dictionaries in *transceiver*.[[Sender]].[[sendEncodings]] except the first one,"

webrtc.html
```diff
@@ -1755,6 +1762,36 @@ <h4>Set the RTCSessionDescription</h4>
...
+                                and abort these sub steps.</p>
+                              </li>
+                              <li>
+                                <p>If the offer contained a request to send simulcast and
```
jan-ivar: Same here. Though I wonder, can this "If" be lifted up one level? Only two of these three sub-steps have it, yet all three are about simulcast. That seems like a bug.

webrtc.html
```diff
@@ -1755,6 +1762,36 @@ <h4>Set the RTCSessionDescription</h4>
...
+                                the answer rejects any of the offered layers, then set the
+                                <code><a data-link-for="RTCRtpEncodingParameters">active</a></code>
+                                flag to <code>false</code> on the <var>sendEncodings</var>
+                                that correspond to rejected layers in the
+                                <code><a>RTCRtpSender</a></code>.</p>
+                              </li>
+                              <li>
+                                <p>Update the paused status of simulcast layers by setting the
+                                <code><a data-link-for="RTCRtpEncodingParameters">active</a></code>
+                                flag on the <var>sendEncodings</var> in the
```
jan-ivar: Setting it to what? I found no reference to "pause status" in this spec or in JSEP. Is there a section in JSEP we can reference?

webrtc.html
```diff
@@ -7520,6 +7557,12 @@ <h3>Simulcast functionality</h3>
         implications of this model is that the <code>addTrack</code> method cannot provide simulcast
         functionality since it does not take <code>sendEncodings</code> as an argument, and therefore cannot
         configure an <code><a>RTCRtpTransceiver</a></code> to send simulcast.</p>
+        <p>The answerer MUST NOT set the <a>simulcast envelope</a> directly. Upon calling the
```
jan-ivar: I apologize. Now that you've added normative steps in the algorithm itself (which I like much better!), this section should probably no longer use normative language, in order to not compete. What you had here before I meddled was probably fine, and gelled better with the paragraph directly above and below it. Can you put it back? Maybe with a bridge like:

"Another implication is that the answerer cannot set the <a>simulcast envelope</a> directly."

Some specs might even mark this whole section as *non-normative*, as it appears to merely summarize behavior dictated elsewhere, although I don't know if @aboba would go for that, in case there are other important bits mentioned only here?

alvestrand:
This PR seems to strongly overlap in intent (but not language) with #2071 - is it intended to be "the other half" (handling incoming)?

webrtc.html
```diff
@@ -1755,6 +1762,36 @@ <h4>Set the RTCSessionDescription</h4>
                               </li>
                             </ol>
                           </li>
+                          <li>
+                            <p>If <var>description</var> is of type <code>"answer"</code>
+                            or <code>"pranswer"</code>, then run the following steps:</p>
```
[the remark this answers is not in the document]
amithilbuch: I know that there is another if description == answer below, but due to having 'mutually exclusive' steps in this list (i.e. "and abort these substeps.") merging the two instruction sets might make it less readable and increase nesting.
Regardless, i think we should "refactor" some of these steps into separate algorithms, as was done with "Create and RTCRtpSender".

amithilbuch:
Simulcast and RIDs are seemingly two distinct concepts, however, with recent changes we made, they are converging to the same thing (other aspects of RIDs removed, making using RIDs without simulcast benign).
#2071 deals with explaining that the order of send_encodings is authoritative and should therefore be used to describe the layers in the simulcast attribute, etc.
#2081 describes how simulcast is set up, and focuses on what happens when the server acts as the caller and requests Simulcast from the client (callee).

i can understand how under the offer vs answer semantics this would appear as "the other half", but the details are on different granularity levels, because of how different this scenario is from the 'regular' offer vs answer setup.

webrtc.html
amithilbuch: pause status is mentioned in [I-D.ietf-mmusic-sdp-simulcast](https://tools.ietf.org/html/draft-ietf-mmusic-sdp-simulcast-13) and that is referenced by jsep:

> When using JSEP to transmit multiple encodings from a RtpSender, the
>  techniques from [I-D.ietf-mmusic-sdp-simulcast] and
>  [I-D.ietf-mmusic-rid] are used.

There is somewhat of a missing flag to match this. This doesn't cause any issues, as the active flag can be used to indicate paused as well.
i will clarify that paused = !active.

webrtc.html
amithilbuch: <var>description</var> covers it, so i will drop it.

amithilbuch:
ok. this is now the final version of my changes. please review / merge if appropriate.

amithilbuch:
@jan-ivar, i seem to have messed something up and cannot access any suggestions that were made. i still see the comments, and i resolved them.
please let me know your thoughts on the new version.

alvestrand:
Once Bernard's comment has been addressed, we can integrate this one.

aboba:
Line 1795:  It seems problematic to let layers remain (set to false) if the answer provides fewer layers.  I would suggest that the excess layers not be present at all.  Otherwise the application could set active to "true" again and cause more layers to be sent than can be handled on the Answerer.

amithilbuch:
@aboba 
I made changes to indicate that layers are removed when either:
Simulcast is rejected altogether or
specific layers are rejected.

aboba:
@amithilbuch An editorial comment.  Instead of `transceiver.[[\Sender]].[[\sendEncodings]]` we use prose, "the [[\sendEncodings]] internal slot of <var>transceiver's sender object."

PR #2084, opened 2019-01-24
RTCError extends DOMException

henbos:
Fixes #2082

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2084.html" title="Last updated on Jan 24, 2019, 11:25 AM UTC (1bc679e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2084/ce3e19f...henbos:1bc679e.html" title="Last updated on Jan 24, 2019, 11:25 AM UTC (1bc679e)">Diff</a>

PR #2089, opened 2019-01-31
RTCError extends DOMException

### LINKED ISSUE #2082: Make RTCError inherit from DOMException?
According to @domenic it is now possible to define IDL types that inherit from DOMException.
This would allow us to do a pure IDL specification of RTCError, which is a far simpler specification than the one we're doing now (and comports with the one in-progress implementation I know about).

DOMException spec:
https://heycam.github.io/webidl/#idl-DOMException

Should we do this?

henbos:
Fixes #2082 

This is a reupload of https://github.com/w3c/webrtc-pc/pull/2084 with respec issues addressed.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2089.html" title="Last updated on Jan 31, 2019, 10:44 AM UTC (658a4f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2089/3455fd8...henbos:658a4f3.html" title="Last updated on Jan 31, 2019, 10:44 AM UTC (658a4f3)">Diff</a>


henbos:
Comment from @aboba from the old version of this PR:

> Section 11.1.2 says that sdpLineNumber is only set if the errorDetail value is "sdp-syntax-error". Would it make sense for sdpLineNumber to be set for other errorDetail values such as "hardware-encoder-not-available"/"hardware-encoder-error" (to indicate the specific encoder that is not available)?

webrtc.html
```diff
@@ -11342,148 +11342,170 @@ <h3>Call Flow Browser to MCU</h3>
...
+    information.</p>
     <section>
-      <h2><dfn>RTCError</dfn> Object</h2>
+      <h3><dfn>RTCError</dfn> Interface</h3>
+      <div>
+        <pre class="idl">
+          [
+              Exposed=Window,
+              Constructor(DOMString message, RTCErrorInit init)
+          ] interface RTCError /*: DOMException*/ {
```
henbos: I get a respec error if I do ": DOMException" and I don't know why. DOMException is an interface and you should be able to extend it. Commented it out to make respec happy.

aboba: You get an error because DOMException is not defined in the document.

alvestrand: You probably have to add "DOMException is defined in <ref>" to the terminology section up front.

henbos:
Reply to previously unaddressed comment:

> Section 11.1.2 says that sdpLineNumber is only set if the errorDetail value is "sdp-syntax-error". Would it make sense for sdpLineNumber to be set for other errorDetail values such as "hardware-encoder-not-available"/"hardware-encoder-error" (to indicate the specific encoder that is not available)?

Not sure, this is something that can happen at setParameters(), now when you're parsing SDP. I think having more info on that is a separate issue than this "refactoring" of RTCError.

webrtc.html
henbos: In other places in the spec, including in WebIDL (example: `callback RTCPeerConnectionErrorCallback = void (DOMException error);`), it doesn't complain about it and the link produce takes you to the interface definition of another spec.

Our list of terminology has 
"When referring to exceptions, the terms throw and create are defined in [WEBIDL-1]."
which is referring to a document whose DOMException is old: https://www.w3.org/TR/2016/REC-WebIDL-1-20161215/#dfn-DOMException.

This is the reference we want: https://heycam.github.io/webidl/#idl-DOMException

henbos: Should we update the Terminology section in a separate PR or in this one? Our WebIDL reference might affect other references in the document, and I assume we don't want to have multiple references to different versions of WebIDL.

alvestrand: Definitely need to consult @dontcallmedom, and we should do it in a separate PR.

PR #2104, opened 2019-02-14
User agents MUST send audioLevel. If other endpoint is legacy, don't calculate it.

### LINKED ISSUE #2046: RTCRtpContributingSource.audioLevel should be present, but if it isn't due to legacy endpoint, don't calculate it
The [audioLevel](https://w3c.github.io/webrtc-pc/#dom-rtcrtpcontributingsource-audiolevel) is optional for CSRCs, but for SSRCs the spec says it MUST be present:

> For SSRCs, this MUST be converted from the level value defined in [RFC6464] if the RFC 6464 header extension is present, otherwise the user agent must compute the value from the audio data (the member must never be absent for audio receivers).

I think this implies that if you receive a packet, even if it arrives late and you intent to drop it, you have to waste cycles decoding it for the sake of calculating the audio level. I also wonder if there are performance considerations with having to calculate it.

Perhaps there should be some leeway about not always having to calculate this if the header extension is present? Chrome currently only surfaces this value from the header extension, not from calculating it.

henbos:
Fixes #2046

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2104.html" title="Last updated on Feb 19, 2019, 9:15 PM UTC (dcbe9ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2104/9b5bb5b...henbos:dcbe9ef.html" title="Last updated on Feb 19, 2019, 9:15 PM UTC (dcbe9ef)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7141,9 +7141,11 @@ async function updateParameters() {
               in [[!RFC6465]] if the RFC 6465 header extension is present,
               otherwise this member MUST be absent.</p>
               <p>For SSRCs, this MUST be converted from the level value defined
-              in [[!RFC6464]] if the RFC 6464 header extension is present,
-              otherwise the user agent must compute the value from the audio
-              data (the member must never be absent for audio receivers).</p>
+              in [[!RFC6464]]. This extension MUST be sent by user agents, and
+              as such should be present. But if the RFC 6464 header extension is
+              not present in the received packets (such as if the other endpoint
+              is not a user agent or is a legacy endpoint), this value SHOULD be
+              absent.</p>
 
               <p>Both RFCs define the level as an integral value from 0 to 127
               representing the audio level in negative decibels relative to the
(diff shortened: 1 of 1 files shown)


aboba:
@henbos  The extension isn't sent if it is not negotiated - it's only mandatory to implement, not use.

webrtc.html
```diff
@@ -7141,9 +7141,11 @@ <h2>Dictionary RTCRtpContributingSource Members</h2>
...
               otherwise this member MUST be absent.</p>
               <p>For SSRCs, this MUST be converted from the level value defined
-              in [[!RFC6464]] if the RFC 6464 header extension is present,
-              otherwise the user agent must compute the value from the audio
-              data (the member must never be absent for audio receivers).</p>
+              in [[!RFC6464]]. This extension MUST be sent by user agents, and
+              as such should be present. But if the RFC 6464 header extension is
+              not present in the received packets (such as if the other endpoint
+              is not a user agent or is a legacy endpoint), this value SHOULD be
+              absent.</p>
```
henbos: Keep it simple!

henbos: SHOULD NOT be present.

aboba: Recommend deleting "This extension MUST be sent by user agents, and as such should be present. But "

PR #2105, opened 2019-02-15
Define a step-by-step generateCertificate algorithm

youennf:
This allows to define the algorithm more precisely:
- the order of operation is clearly defined.
- each internal slot of a certificate are initialized explicitly.

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/webrtc-pc/pull/2105.html" title="Last updated on Mar 22, 2019, 2:18 PM UTC (22a144d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2105/76a6464...youennf:22a144d.html" title="Last updated on Mar 22, 2019, 2:18 PM UTC (22a144d)">Diff</a>


youennf:
I think PR needs some more work, any feedback appreciated though.

webrtc.html
```diff
@@ -4745,7 +4745,7 @@ <h2>Methods</h2>
             <dt><dfn data-idl><code>generateCertificate</code></dfn>, static</dt>
             <dd>
               <p>The <code>generateCertificate</code> function causes the
-              <a>user agent</a> to create and store an X.509 certificate
+              <a>user agent</a> to create an X.509 certificate
```
alvestrand: The generateCertificate function doesn't store a certificate (except into an internal slot). It creates a certificate that can be stored.

webrtc.html
```diff
@@ -4790,23 +4770,43 @@ <h2>Methods</h2>
...
-              namedCurve: "<a data-cite=
-              "!WebCryptoAPI#dfn-NamedCurve">P-256</a>"
-              }</code>.</p>
-              <p class="note">It is expected that a <a>user agent</a> will have
-              a small or even fixed set of values that it will accept.</p>
+
+              <p>When the method is called, the user agent MUST run the following steps:</p>
+              <ol>
+                <li>Let <var>keygenAlgorithm</var> be the first argument to <code>generateCertificate</code>.</li>
+                <li>Let <var>certificateExpiration</var> be the result of converting <var>keygenAlgorithm</var> to an <code>RTCCertificateExpiration</code> value.</li>
```
alvestrand: This makes no sense. You can't convert keygenAlgorithm to an expiration value. Wrong variable?

webrtc.html
```diff
@@ -4790,23 +4770,43 @@ <h2>Methods</h2>
...
+                <li>If the above step throws an <var>error</var>, return a promise that is <a>rejected</a> with this <var>error</var>.</li>
+                <!-- FIXME: Is there a better way to express try/catch. -->
```
alvestrand: Why not say "If the conversion fails", and leave it at that?

webrtc.html
```diff
@@ -4790,23 +4770,43 @@ <h2>Methods</h2>
...
+                <li>Let <var>normalizedAlgorithm</var> be the result of normalizing an algorithm according to the <a data-cite="!WebCryptoAPI#algorithm-normalization">
+                  algorithm normalization process</a> [[!WebCryptoAPI]]
+                  with an operation name of <code>generateKey</code> and a [[<a data-cite= "!WebCryptoAPI#dfn-supportedAlgorithms">supportedAlgorithms</a>]]
+                  value specific to production of certificates for <code>RTCPeerConnection</code>.</li>
```
alvestrand: I don't see what the "and a value" part adds, since it's not further restricted or specified.

webrtc.html
```diff
@@ -4790,23 +4770,43 @@ <h2>Methods</h2>
...
+                <li>If the above step produces an <var>error</var>, return a promise that is <a>rejected</a> with this <var>error</var>.</li>
+                <li>If the <var>normalizedAlgorithm</var> parameter identifies an algorithm that the <a>user agent</a> cannot or will not use to
+                  generate a certificate for <code>RTCPeerConnection</code>, return a promise that is <a>rejected</a> with a <code>DOMException</code> of type <code>NotSupportedError</code>.
+                  In particular, <var>normalizedAlgorithm</var> MUST be an asymmetric algorithm that can be used to produce a signature.</li>
```
alvestrand: s/signature/signature in an X.509 certificate/?

webrtc.html
```diff
@@ -4790,23 +4770,43 @@ <h2>Methods</h2>
...
+                <li>Let <var>p</var> be a new promise.
+                <li>Perform the remaining steps in parallel:
+                  <ol>
+                      <li>Perform the generate key operation specified by <var>normalizedAlgorithm</var> using <var>keygenAlgorithm</var>.</li>
+                      <!-- FIXME: What about extractable and usages parameters used in WebCrypto API spec. -->
+                      <li>Let <var>generatedKeyingMaterial</var> and <var>generatedKeyCertificate</var> be the private keying material and certificate generated by the above step.</li>
+                      <li>Let certificate be a new <code><a>RTCCertificate</a></code> object.</li>
```
alvestrand: < var>certificate</ var>?

youennf:
> Are there observable differences between the specified algorithm and the currently-specified algorithm?

The intent is to have the same behavior but the currently-specificied algorithm is not very precise in the ordering of the steps.

webrtc.html
youennf: This tries to implement what is being defined in https://w3c.github.io/webrtc-pc/#rtccertificateexpiration-dictionary.

I agree that this is not great. I thought about changing the IDL to use any instead of AlgorithmIdentifier there but kept it this way for now.

The idea behind the current description is that "keygenAlgorithm" being the first argument to generateCertificate, it is not yet typed as a AlgorithmIdentifier.

webrtc.html
youennf: Sounds good

webrtc.html
youennf: I tried to keep existing description here.

webrtc.html
youennf: How about "signature used to authenticate DTLS connections"?

webrtc.html
youennf: ok

webrtc.html
youennf: I changed the WebIDL to have generateCertificate take 'any' instead and do the conversion twice.

youennf:
Two remaining FIXMES:
- Should we define a value in case no expire value is provided? This could be done in WebIDL if needed.
- generate key action takes several parameters and we only define values for the first two parameters. I haven't checked yet whether this is fine or whether we can/should be more precise.

alvestrand:
@jan-ivar to give advice on how to avoid using "any".

youennf:
Checking Chrome, Firefox and Safari, it seems the default expires value is the same, equal to roughly 30 days. We could then use it as a default.

As of avoiding 'any', I wonder whether any generateCertificate implementation allows passing a String. If that is not the case, maybe all we need to do is moving from 'any' to 'object' in generateCertificate WebIDL.

Alternatively, since passing a String is a shortcut, we could only grab 'expires' in case of AlgorithmIdentifier being an object. I'll update the PR in this way.

webrtc.html
jan-ivar: @alvestrand @youennf  I think we should stick with [AlgorithmIdentifier](https://w3c.github.io/webcrypto/#dfn-AlgorithmIdentifier) in the WebIDL. It is defined as:
```js
typedef (object or DOMString) AlgorithmIdentifier;
```
So when the old language says *"When generateCertificate is called with an `object` argument, the user agent attempts to convert the object into an RTCCertificateExpiration."*, it's not unclear how it can be an `object`—it can.

The unclear part is perhaps how to "convert" said `object` to said dictionary. Linking to https://heycam.github.io/webidl/#dfn-convert-ecmascript-to-idl-value should clarify this.

We always want to do as much in WebIDL vs. prose as possible, so we should avoid `any`.

jan-ivar:
The specific "convert" algorithm for dictionaries is under https://heycam.github.io/webidl/#es-dictionary

webrtc.html
jan-ivar: I meant to reply here, but clicked "Review", so see below instead.

webrtc.html
```diff
@@ -4736,7 +4736,7 @@ <h2 id="sec.cert-mgmt">Certificate Management</h2>
       SHA-256 hash.</p>
       <div>
         <pre class="idl">partial interface RTCPeerConnection {
-    static Promise&lt;RTCCertificate&gt; generateCertificate (AlgorithmIdentifier keygenAlgorithm);
+    static Promise&lt;RTCCertificate&gt; generateCertificate (any parameters);
```
jan-ivar: Revert per above.

webrtc.html
```diff
@@ -4790,23 +4770,44 @@ <h2>Methods</h2>
...
-              a small or even fixed set of values that it will accept.</p>
+
+              <p>When the method is called, the user agent MUST run the following steps:</p>
+              <ol>
+                <li>Let <var>parameters</var> be the first argument to <code>generateCertificate</code>.</li>
+                <li>Let <var>certificateExpiration</var> be the result of converting <var>parameters</var> to an <code>RTCCertificateExpiration</code> value.</li>
+                <li>If the conversion fails with an <var>error</var>, return a promise that is <a>rejected</a> with this <var>error</var>.</li>
+                <li>Let <var>keygenAlgorithm</var> be the result of converting <var>parameters</var> to an <code>AlgorithmIdentifier</code> value.</li>
+                <li>If the conversion fails with an <var>error</var>, return a promise that is <a>rejected</a> with this <var>error</var>.</li>
+                <li>Let <var>normalizedAlgorithm</var> be the result of normalizing an algorithm according to the <a data-cite="!WebCryptoAPI#algorithm-normalization">
```
jan-ivar: Please word-wrap to make review easier.

Naming-wise I'd follow https://w3c.github.io/webcrypto/#SubtleCrypto-method-encrypt and use `algorithm` and `normalizedAlgorithm` respectively here.

Otherwise `keygenAlgorithm` and `normalizedKeygenAlgorithm` for scoping.

webrtc.html
```diff
@@ -4790,23 +4770,44 @@ <h2>Methods</h2>
...
-              namedCurve: "<a data-cite=
-              "!WebCryptoAPI#dfn-NamedCurve">P-256</a>"
-              }</code>.</p>
-              <p class="note">It is expected that a <a>user agent</a> will have
-              a small or even fixed set of values that it will accept.</p>
+
+              <p>When the method is called, the user agent MUST run the following steps:</p>
+              <ol>
+                <li>Let <var>parameters</var> be the first argument to <code>generateCertificate</code>.</li>
+                <li>Let <var>certificateExpiration</var> be the result of converting <var>parameters</var> to an <code>RTCCertificateExpiration</code> value.</li>
```
jan-ivar: s/value/dictionary/ and maybe with links?

Let *certificateExpiration* be the result of [converting](https://heycam.github.io/webidl/#dfn-convert-ecmascript-to-idl-value) *algorithm* to an `RTCCertificateExpiration` [dictionary](https://heycam.github.io/webidl/#es-dictionary).

webrtc.html
```diff
@@ -4790,23 +4770,44 @@ <h2>Methods</h2>
...
+                <li>If the conversion fails with an <var>error</var>, return a promise that is <a>rejected</a> with this <var>error</var>.</li>
+                <li>Let <var>keygenAlgorithm</var> be the result of converting <var>parameters</var> to an <code>AlgorithmIdentifier</code> value.</li>
+                <li>If the conversion fails with an <var>error</var>, return a promise that is <a>rejected</a> with this <var>error</var>.</li>
+                <li>Let <var>normalizedAlgorithm</var> be the result of normalizing an algorithm according to the <a data-cite="!WebCryptoAPI#algorithm-normalization">
+                  algorithm normalization process</a> [[!WebCryptoAPI]]
```
jan-ivar: Should this be "[normalizing an algorithm](https://w3c.github.io/webcrypto/#dfn-normalize-an-algorithm)" ?

webrtc.html
youennf: OK, I will revert the related changes back to the fist commit and will add an "if object" check to make it clear that conversion happens only in that case.

We can try to use 
https://heycam.github.io/webidl/#dfn-convert-ecmascript-to-idl-value but we would need to convert back the IDL value (which is an IDL object) in an ECMAScript value first (https://heycam.github.io/webidl/#dfn-convert-idl-to-ecmascript-value would do that I guess).
This is more or less matching what WebKit is currently doing so I am fine with that.

I wonder whether we can write a test for this. Is Chrome or Firefox supporting generateCertificate with a string value?

webrtc.html
```diff
@@ -4790,23 +4770,44 @@ <h2>Methods</h2>
...
-              "!WebCryptoAPI#dfn-NamedCurve">P-256</a>"
-              }</code>.</p>
-              <p class="note">It is expected that a <a>user agent</a> will have
-              a small or even fixed set of values that it will accept.</p>
+
+              <p>When the method is called, the user agent MUST run the following steps:</p>
+              <ol>
+                <li>Let <var>parameters</var> be the first argument to <code>generateCertificate</code>.</li>
+                <li>Let <var>certificateExpiration</var> be the result of converting <var>parameters</var> to an <code>RTCCertificateExpiration</code> value.</li>
+                <li>If the conversion fails with an <var>error</var>, return a promise that is <a>rejected</a> with this <var>error</var>.</li>
```
jan-ivar: This will cause *generateCertificate* to fail if passed a `DOMString`. Is this intentional? If so, I'd prefer we be more explicit about this and use `object` instead of `AlgorithmIdentifier` in the WebIDL.

If not, we need to add a conditional here to not fail over `DOMString`.

webrtc.html
```diff
@@ -4790,23 +4770,44 @@ <h2>Methods</h2>
...
+                  algorithm normalization process</a> [[!WebCryptoAPI]]
+                  with an operation name of <code>generateKey</code> and a [[<a data-cite= "!WebCryptoAPI#dfn-supportedAlgorithms">supportedAlgorithms</a>]]
+                  value specific to production of certificates for <code>RTCPeerConnection</code>.</li>
+                <li>If the above step produces an <var>error</var>, return a promise that is <a>rejected</a> with this <var>error</var>.</li>
+                <li>If the <var>normalizedAlgorithm</var> parameter identifies an algorithm that the <a>user agent</a> cannot or will not use to
+                  generate a certificate for <code>RTCPeerConnection</code>, return a promise that is <a>rejected</a> with a <code>DOMException</code> of type <code>NotSupportedError</code>.
+                  In particular, <var>normalizedAlgorithm</var> MUST be an asymmetric algorithm that can be used to produce a signature used to authenticate DTLS connections.</li>
+                <li>Let <var>p</var> be a new promise.
+                <li>Perform the remaining steps in parallel:
```
jan-ivar: s/remaining/following/ or use *"return* p *and perform the remaining steps in parallel."*

This spec tends to use the former.

webrtc.html
```diff
@@ -4790,23 +4770,44 @@ <h2>Methods</h2>
...
+                  <ol>
+                      <li>Perform the generate key operation specified by <var>normalizedAlgorithm</var> using <var>keygenAlgorithm</var>.</li>
+                      <!-- FIXME: What about extractable and usages parameters used in WebCrypto API spec. -->
+                      <li>Let <var>generatedKeyingMaterial</var> and <var>generatedKeyCertificate</var> be the private keying material and certificate generated by the above step.</li>
+                      <li>Let <var>certificate</var> be a new <code><a>RTCCertificate</a></code> object.</li>
+                      <li>If <var>certificateExpirations</var>'s <code>expires</code> value is undefined, set it to a default value.</li>
```
jan-ivar: You use dot notation below. I'm partial to it here as well: "If *certificateExpiration*.`expires` is `undefined`, ..."

webrtc.html
```diff
@@ -4790,23 +4770,44 @@ <h2>Methods</h2>
...
+                      <!-- FIXME: Can we do better than this? -->
```
jan-ivar: I hope so. A "default value" is real vague. The [old language](http://w3c.github.io/webrtc-pc/#idl-def-rtccertificateexpiration-expires) reads as if it returns already expired certificates.

"a reasonable amount of time"?

webrtc.html
jan-ivar: > we would need to convert back the IDL value (which is an IDL object)

Technically. Hopefully [scant difference](https://heycam.github.io/webidl/#es-object) in practice.

I see [crypto](https://w3c.github.io/webcrypto/#dfn-normalize-an-algorithm) uses: *"Let initialAlg be the result of converting the ECMAScript object represented by alg to the IDL dictionary type Algorithm, as defined by [WebIDL]. "*

> Is Chrome or Firefox supporting generateCertificate with a string value?

Both pass the [wpt tests](https://wpt.fyi/results/webrtc/RTCPeerConnection-generateCertificate.html?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned) but I only see a test for `'invalid-algo'`. Unsure what plain strings are valid, but I managed to get to `NotSupportedError` in both with:
```js
await RTCPeerConnection.generateCertificate('P-256');
```
I guess some extra input validation tests may be in order.

webrtc.html
```diff
@@ -4790,23 +4770,74 @@ <h2>Methods</h2>
...
-              "!WebCryptoAPI#dfn-NamedCurve">P-256</a>"
-              }</code>.</p>
-              <p class="note">It is expected that a <a>user agent</a> will have
-              a small or even fixed set of values that it will accept.</p>
+
+              <p>When the method is called, the user agent MUST run the following steps:</p>
+              <ol>
+                <li>Let <var>keygenAlgorithm</var> be the first argument to
+                  <code>generateCertificate</code>.</li>
+                <li>Let <var>expires</var> be a <code>DOMTimestamp</code> value
```
henbos: nit: DOMTimeStamp

webrtc.html
```diff
@@ -4790,23 +4770,74 @@ <h2>Methods</h2>
...
-              namedCurve: "<a data-cite=
-              "!WebCryptoAPI#dfn-NamedCurve">P-256</a>"
-              }</code>.</p>
-              <p class="note">It is expected that a <a>user agent</a> will have
-              a small or even fixed set of values that it will accept.</p>
+
+              <p>When the method is called, the user agent MUST run the following steps:</p>
+              <ol>
+                <li>Let <var>keygenAlgorithm</var> be the first argument to
+                  <code>generateCertificate</code>.</li>
```
henbos: nit: Here and elsewhere, skip the indent inside of the &lt;li&gt; for consistency with other parts of the document:
```
    <li>Let <var>keygenAlgorithm</var> be the first argument to
    <code>generateCertificate</code>.</li>
```

webrtc.html
```diff
@@ -4790,23 +4770,74 @@ <h2>Methods</h2>
...
+                <li>Let <var>expires</var> be a <code>DOMTimestamp</code> value
+                  of 2592000.</li>
```
henbos: If this was seconds it would be 30 days, but DOMTimeStamp is defined in terms of milliseconds. Make it 2592000000.

And perhaps?
```
<div class="note">
  <p>This means the certificate will expire in 30 days from the time of the
  <code>generateCertificate</code> call.</p>
</div>
```

webrtc.html
```diff
@@ -4790,23 +4770,74 @@ <h2>Methods</h2>
...
+                  <ol>
+                    <li>Let <var>certificateExpiration</var> be the result of
+                      <a href="https://heycam.github.io/webidl/#dfn-convert-ecmascript-to-idl-value">converting</a>
+                      the ECMAScript object represented by <var>keygenAlgorithm</var> to an
+                      <code>RTCCertificateExpiration</code> dictionary.</li>
+                    <li>If the conversion fails with an <var>error</var>,
+                      return a promise that is <a>rejected</a> with this <var>error</var>.</li>
+                    <li>If <var>certificateExpiration</var>.<code>expires</code>
+                      is not undefined, set <var>expires</var> to
+                      <var>certificateExpirations</var>.<code>expires</code>.</li>
```
henbos: nit: certificateExpiration

webrtc.html
```diff
@@ -4790,23 +4770,74 @@ <h2>Methods</h2>
...
+                        <var>normalizedKeygenAlgorithm</var> using
+                        <var>keygenAlgorithm</var>.</li>
+                      <li>Let <var>generatedKeyingMaterial</var> and
+                        <var>generatedKeyCertificate</var> be the private
+                        keying material and certificate generated by the
+                        above step.</li>
+                      <li>Let <var>certificate</var> be a new
+                        <code><a>RTCCertificate</a></code> object.</li>
+                      <li>Set <var>certificate</var>.[[\Expires]] to the
+                        current time plus <var>expires</var> value. A <a>user
```
henbos: Should "current time" be more precise? In other places we say `<a>performance.timeOrigin</a> + <a>performance.now()</a>`.

webrtc.html
```diff
@@ -4790,23 +4770,74 @@ <h2>Methods</h2>
...
+                        agent</a> MAY choose to limit the value of the
```
lgrahl: I apologise for the sidetrack but what was the rationale for this limitation?

webrtc.html
youennf: ok

webrtc.html
youennf: ok

webrtc.html
youennf: I wrapped everything with <p> as it seems to be done this way in setConfiguration.

webrtc.html
youennf: For now, I kept current time with is a DOMTimeStamp.
In other places, we are using DOMHighResTimeStamp while we are using DOMTimeStamp here.

webrtc.html
youennf: I was not able to track when this was added, but it has been there since a lot of time.

webrtc.html
```diff
@@ -4790,23 +4770,112 @@ <h2>Methods</h2>
...
+                    </li>
+                    <li>
+                      <p>Let <var>certificate</var> be a new
+                      <code><a>RTCCertificate</a></code> object.</p>
+                    </li>
+                    <li>
+                      <p>Set <var>certificate</var>.[[\Expires]] to the
+                      current time plus <var>expires</var> value. A <a>user
+                      agent</a> MAY choose to limit the value of the
+                      <var>certificate</var>.[[\Expires]] value.</p>
```
jan-ivar: This MAY is lost. Intentional? Without this, JS could create a certificate that never expires. Is that wise?

webrtc.html
```diff
@@ -4757,27 +4757,7 @@ <h2>Methods</h2>
               <var>keygenAlgorithm</var> argument uses the WebCrypto
               [[!WebCryptoAPI]] <a data-cite=
               "!WebCryptoAPI#dfn-AlgorithmIdentifier">
-              <code>AlgorithmIdentifier</code></a> type. The
-              <var>keygenAlgorithm</var> value MUST be a valid argument to
-              <a data-cite=
-              "!WebCryptoAPI#SubtleCrypto-method-generateKey">
-              <code>window.crypto.subtle.generateKey</code></a>; that is, the
-              value MUST produce a non-error result when normalized according
-              to the WebCrypto <a data-cite=
```
jan-ivar: This requirement is lost. Is there equivalent language in the new PR to limit valid values here?

webrtc.html
lgrahl: Perhaps @martinthomson knows? :slightly_smiling_face:

martinthomson: Which limitation?

lgrahl: That the user agent (which I read as *browser*) MAY choose to limit the *expires* value.

martinthomson: Because browsers (or the users they act for) might not be well served by long-lived keys. And an expiration encourages/forces sites to deal with key changes. Maybe one days we will want to cleanly change requirements for keys and algorithms. Having keys expire prepares the ecosystem for that being something less than a massive disruption.

youennf: Isn't the question related to why it is a MAY instead of a SHOULD or event a MUST?
It seems that Chrome, Firefox and Safari cap expires by a 1 year.
Maybe we should engrave that in the spec.

martinthomson: I'd be OK with that, but I wouldn't want a hard requirement to support a maximum that long.  I would suggest making a separate issue for tracking that though.

lgrahl: Making a suggestion for the range is something I would also appreciate. Otherwise, it's not really clear what's *acceptable*, not even for browser implementations. Also agree we should create a new issue and, once again, apologise for the sidetrack. :slightly_smiling_face:

webrtc.html
youennf: This should be covered by step 5

webrtc.html
```diff
@@ -4790,23 +4770,112 @@ <h2>Methods</h2>
...
               for distinguished name and serial number SHOULD be used.</p>
-              <p>A <a>user agent</a> MUST reject a call to
-              <code>generateCertificate()</code> with a
-              <code>DOMException</code> of type <code>NotSupportedError</code>
-              if the <var>keygenAlgorithm</var> parameter identifies an
-              algorithm that the <a>user agent</a> cannot or will not use to
-              generate a certificate for <code>RTCPeerConnection</code>.</p>
-              <p>The following values MUST be supported by a <a>user agent</a>:
-              <code>{ name: "<a data-cite=
-              "!WebCryptoAPI#rsassa-pkcs1">RSASSA-PKCS1-v1_5</a>",
```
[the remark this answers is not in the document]
youennf: Will add it back

webrtc.html
```diff
@@ -4790,23 +4770,112 @@ <h2>Methods</h2>
...
-              modulusLength: 2048, publicExponent: new Uint8Array([1, 0, 1]),
-              hash: "SHA-256" }</code>, and <code>{ name: "<a data-cite=
-              "!WebCryptoAPI#ecdsa">ECDSA</a>",
-              namedCurve: "<a data-cite=
-              "!WebCryptoAPI#dfn-NamedCurve">P-256</a>"
-              }</code>.</p>
-              <p class="note">It is expected that a <a>user agent</a> will have
-              a small or even fixed set of values that it will accept.</p>
```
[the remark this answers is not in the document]
youennf: Ditto

webrtc.html
youennf: I do not think the PR is changing anything there.
As discussed elsewhere, the idea is to strengthen this MAY, given several implementations seem to cap by 1 year.

webrtc.html
youennf: Filed https://github.com/w3c/webrtc-pc/issues/2137

Issue #2106, opened 2019-02-16
transceiver.stop() shouldn't block in-progress replaceTrack from succeeding


jan-ivar:
Most WebRTC methods' final queued task say:

   1. *If connection's [[IsClosed]] slot is true, then abort these steps.*

But [sender.replaceTrack](http://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack)'s says:

   1. *If transceiver's [[Stopped]] slot is true, abort these steps.*

That may seem like the sender-scoped equivalent, but has the side-effect that `transceiver.stop()` causes an already in-progress `replaceTrack` to never resolve. That seems unintended.

I propose we make replaceTrack like the other methods.

One question then is should the promise resolve or reject with an error? I think resolving is fine, since it's a glorified setter at that point; no harm will come of it.

youennf:
I know this is an existing practice for RTCPeerConnection but it does not seem like a good pattern for an API to return a promise and never reject/resolve it. That may leave code to be hanging forever due to some timing issues.
I would just remove step 6.4.4.1.

jan-ivar:
@youennf I believe you're arguing for less blocking. This issue and PR result in less blocking, so I'm not hearing an argument against merging this.

We should be consistent, and this PR merely aligns with our existing pattern (removing unintended blocking). This isn't theoretical, as we're fixing our Firefox code right now, and would like to merge this spec support to avoid blocking on `stop()`. I suggest opening a separate issue to challenge the overall pattern.

> That may leave code to be hanging forever due to some timing issues.

FWIW I think it's inaccurate to characterize the absence of a callback as "hanging".

aboba:
Agree that the promise should be rejected or resolved. However, we don't seem to have been consistent with respect to this throughout the specification. 

For example, several of the "If connection's [[IsClosed]] slot is true, then abort these steps." instances would appear to result in a promise never being rejected or resolved.  An example is `addIceCandidate`.  Was this what we really intended? 

Note that Step 6 sub-step 1 already tests if `transceiver` has been stopped, and rejects the promise if it has. Would it make more sense for `replaceTrack` to reject the promise if it discovers a closed `RTCPeerConnection` or `transceiver` later?

jan-ivar:
> Agree that the promise should be rejected or resolved.

That's not actually a thing in JS.

> However, we don't seem to have been consistent with respect to this throughout the specification.
>
[...]

PR #2107, opened 2019-02-16
Block replaceTrack on pc.close() like the others, not tc.stop().

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2106.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2107.html" title="Last updated on Feb 16, 2019, 1:04 AM UTC (7ab0e11)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2107/76a6464...jan-ivar:7ab0e11.html" title="Last updated on Feb 16, 2019, 1:04 AM UTC (7ab0e11)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6196,8 +6196,8 @@ async function updateParameters() {
                           <p>Queue a task that runs the following steps:</p>
                           <ol>
                             <li>
-                              <p>If <var>transceiver</var>'s <a>[[\Stopped]]</a> slot is
-                              <code>true</code>, abort these steps.</p>
+                              <p>If <var>connection</var>'s <a>[[\IsClosed]]</a>
+                              slot is <code>true</code>, abort these steps.</p>
                             </li>
                             <li>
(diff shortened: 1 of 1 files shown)


henbos:
> Only question is whether the `replaceTrack` promise should resolve or be rejected.

I think the PR is correct; the promise should be resolved because the track was in-fact replaced. Transmission of that track is a different operation, and even a successful replaceTrack does not guarantee that the new frames reaches the other endpoint. You could be pedantic and update "have the sender switch seamlessly to transmitting withTrack instead of the sender's existing track" to clarify that transmission only occurs if the link is still open, but I don't think it's necessary.

henbos:
@jan-ivar Any reason to abort `replaceTrack` _even if_ the connection is closed? close() is sort of a "hard stop", but by the time we abort, the "lower layer" has already replaced the track. I view replaceTrack() as a glorified setter.

Issue #2043, opened 2018-12-11
set-pc-configuration is unclear on how to match certificates


youennf:
https://w3c.github.io/webrtc-pc/#set-pc-configuration step 4 is checking certificates as follows: "the set of certificates differs from the ones used when connection was constructed".

It is not clear whether that means whether certificate order is significant or not. 
The term 'set' is used, which would tend to make the order insignificant, but certificates are passed as a sequence. WebKit is implementing sequence-matching.

For a RTCCertificate to 'differ' from another, it can mean that the RTCCertificate objects are not the same. WebKit is implementing this interpretation.
Another interpretation might be that a RTCCertificate differ from another if one of its slots (origin, expires, keying material, certificate) is different.

Additional WPT tests might be good in that area.

alvestrand:
With the normal case of getConfiguration/modify/setConfiguration, the certificates will be object-equal. This is simplest to specify and simplest to implement.

youennf:
WPT issue https://github.com/web-platform-tests/wpt/issues/15840

youennf:
Fixed by https://github.com/w3c/webrtc-pc/pull/2113

PR #2113, opened 2019-03-04
Clarifying how setConfiguration checks certificate values

youennf:
Certificate comparison is based on JS ===.
The order of the certificates in the sequence is insignificant.

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/webrtc-pc/pull/2113.html" title="Last updated on Mar 14, 2019, 4:32 AM UTC (9a7217a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2113/76a6464...youennf:9a7217a.html" title="Last updated on Mar 14, 2019, 4:32 AM UTC (9a7217a)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2061,9 +2061,21 @@
             identity</a>, <a>throw</a> an <code>InvalidModificationError</code>.</p>
             </li>
             <li><p>If <code><var>configuration</var>.certificates</code> is
-            set and the set of certificates differs from the ones used
-            when <var>connection</var> was constructed, <a>throw</a> an
-            <code>InvalidModificationError</code>.</p></li>
+              set, run the following steps:</p>
+              <ol>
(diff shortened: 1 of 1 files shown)


youennf:
Related issue is https://github.com/w3c/webrtc-pc/issues/2043

lgrahl:
It may make sense to also replace the term *set* with *sequence*.

youennf:
> It may make sense to also replace the term _set_ with _sequence_.

Agreed it is strange that iceServers is defined as an array and certificates as a set while they are both WebIDL sequences. 'list' seems to be used a lot in the spec, so maybe that should be the term we use for both.

henbos:
Check the order

aboba:
@youennf What about checking the order?

PR #2138, opened 2019-03-22
Add setTargetJitterBufferDelay method to RTCRtpReceiver

### LINKED ISSUE #2139: Add setTargetJitterBufferDelay method to RTCRtpReceiver
Following discussion [here](https://github.com/w3c/webrtc-pc/issues/2109), can we add a new method to the `RTCRtpReceiver`, called `setTargetJitterBufferDelay`. It would allow to provide a preference for the user agent for the target jitter buffer delay, but depending on the congestion control and network condition actual delay may differ.

The justification is essentially the same as in previous [issue](https://github.com/w3c/webrtc-pc/issues/2109). This extensions would open a room for quality improvements for both audio and video media sourced by an `RTCPeerConnection` as you will have additional time until the media rendered. For example, in the case of network issues you will have more video samples to play and recover. Or if the packets are lost then you will have more time for packets retransmission.

kuddai:
Fixes #2139

***
<a href="https://pr-preview.s3.amazonaws.com/kuddai/webrtc-pc/pull/2138.html" title="Last updated on Mar 22, 2019, 2:35 PM UTC (96c64c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2138/589e451...kuddai:96c64c7.html" title="Last updated on Mar 22, 2019, 2:35 PM UTC (96c64c7)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6990,6 +6990,7 @@ async function updateParameters() {
     sequence&lt;RTCRtpContributingSource&gt;    getContributingSources ();
     sequence&lt;RTCRtpSynchronizationSource&gt; getSynchronizationSources ();
     Promise&lt;RTCStatsReport&gt;   getStats();
+    Promise&lt;void&gt; setTargetJitterBufferDelay (double delay);
 };</pre>
         <section>
           <h2>Attributes</h2>
@@ -7147,6 +7148,31 @@ async function updateParameters() {
                 </li>
                 <li>
                   <p>Return <var>p</var>.</p>
+                </li>
+              </ol>
+            </dd>
+            <dt><code>setTargetJitterBufferDelay</code></dt>
+            <dd>
+              <p>Asynchronously provides a preference for the target jitter
+              buffer delay in seconds. The user agent SHOULD aim to match this
+              preference but due to congestion control and other network related
+              reasons the actual delay MAY be different.</p>
+              <p>When the <dfn data-idl id=
+              "widl-RTCRtpReceiver-setTargetJitterBufferDelay-Promise-void">
+              <code>setTargetJitterBufferDelay()</code></dfn> method is called:
+              </p>
+              <ol>
+                <li>
+                  <p>Let <var>delay</var> be the target jitter buffer delay in
+                  seconds</p>
+                </li>
+                <li>
+                  <p>Let <var>p</var> be a new promise which is <a>resolved</a>
+                  with <code>undefined</code> unless <var>delay</var> is a
+                  negative number. If the <var>delay</var> is a negative number
+                  then <a> reject</a> <var>p</var> with a newly <a
+                  data-link-for="exception" data-lt="create">created</a> <code>
+                  OperationError</code>.</p>
                 </li>
               </ol>
             </dd>


kuddai:
Fixes [2139](https://github.com/w3c/webrtc-pc/issues/2139)

henbos:
I have some language-nits but that can wait (I'm in Prague for IETF), @jan-ivar what do you think about this type of method? Without specifying things we don't want to specify, I think this type of "SHOULD" and "MAY" language is what you would end up with. The elephant in the room being that we don't know how flexible different implementations would be in controlling this, so different APIs might have different strategies or care different amounts about abiding by the desired target.

Questions like "How should targeting 10^-10 or 10^10 be interpreted?" might arise, and that would be implementation-specific. But congestion control is implementation-specific, and we like it that way. Is a vague API in this area OK?

I'd argue that it is OK because it is merely a "target". There is already an example of a target, at least in stats: [targetBitrate](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-targetbitrate). Encoders have targets, but different codecs/implementations would abide by the target differently, and the actual bitrate may be different, and different under different conditions.

In this case we are asking to set the target though, not merely observe it.

A concern with this API is that if you set a "bad" target the application may need to diagnose itself (e.g. with getStats) not only to find out how well the jitter buffer obeys, but what quality level the application is willing to tolerate when the buffer gets too small or too big for the optimal experience.

Should you be allowed to shoot yourself in the foot? The amount of foot-gunning would currently be up to the implementation. Maybe targeting 0 delay in browser A leads to acceptable performance, but in browser B the experience becomes horrible? We could add implementation recommendations, but it would all be very vague.

I do see the use case for wanting to control the delay though, just like "latency" makes sense for some applications that use getUserMedia().

jan-ivar:
This sounds like an interesting new feature, but a new feature nonetheless.

I worry it's a bit late to be considering new features in 1.0. New features take time, and we've mostly been focused on finishing 1.0 over the last year, which I think means saying no to new features at this point.

I also have some concerns with this API. It's essentially an API for JS to make browsers stutter.

How web compatible is it to outsource control of our jitter buffer length? As a non-dominant vendor, we sometimes see web sites not testing their site with every browser or platform, so giving this tuning control to someone who may never test its effect on our browser, isn't terribly appealing. When things stutter people blame the browser.

I'm not immune to its benefits, just making a point that it might take a while to work out the details, which might be better done in an extension spec?

fippo:
I think the use-case for this is pretty clear (livestreaming, not talking heads) and relevant. Its basically the equivalent of Flashs netconnection buffertime.

I am not sure that a API to cover *just this* is better than a generic getParameters/setParameters like we have on the sender. That is *much* more effort to specify though.

henbos:
Would it be less problematic if instead of targeting X seconds, you had a enum for the delay mode? E.g. "realtime", "conferencing". Sort of like a content hint, rather than a control. But maybe that's different than what @kuddai has in mind?

webrtc.html
```diff
@@ -7150,6 +7151,31 @@ <h2>Methods</h2>
...
+              preference but due to congestion control and other network related
+              reasons the actual delay MAY be different.</p>
+              <p>When the <dfn data-idl id=
+              "widl-RTCRtpReceiver-setTargetJitterBufferDelay-Promise-void">
+              <code>setTargetJitterBufferDelay()</code></dfn> method is called:
+              </p>
+              <ol>
+                <li>
+                  <p>Let <var>delay</var> be the target jitter buffer delay in
+                  seconds</p>
```
jan-ivar: This algorithm doesn't do anything. As written there's also no need to return a promise, since it's synchronous.

I see nothing here that wouldn't work as an attribute, e.g. `sender.latency = 0.5;` That would also address reading back the set value.

jan-ivar: Compare to:
```js
const parameters = receiver.getParameters();
parameters.latency = 0.5; 
await receiver.setParameters(parameters);
```
🤔

kuddai: What should happen if user specify negative delay? Should be rejected by setParameters then?

henbos: Presumably the network stack and whatever controls the jitter buffer would live on a different thread than the main one, so specifying this as promise-based seemed like a good idea to me, if only the steps were updated to say stuff like "in parallel" and "update the underlying congestion control's preferred jitter buffer delay" or something like that. But you could make is synchronous and have the preference live in JS thread, but that would make it an implementation-detail how the preference is communicated to the network stack.

With regards to using get/setParameters(): These reflect negotiated parameters, and for the receiver there is no setParameters() method. The jitter buffer on the other hand is a local preference that is not negotiated. So it sounds like a different API to me, but I'm not against expressing this as a "receiver parameter" if we want to go that route.

jan-ivar: The only reason to return a promise would be if there's a step that must be done off main-thread that can fail. I don't see that here.

henbos:
Closing this in favor of https://github.com/w3c/webrtc-pc/pull/2149 which will be moved to an extension spec.

PR #2149, opened 2019-03-28
RTCRtpReceiver.jitterBufferDelayHint attribute added.

### LINKED ISSUE #2139: Add setTargetJitterBufferDelay method to RTCRtpReceiver

henbos:
Fixes #2139.

This is another stab at what #2138 is trying to achieve.
This reformulates and clarifies things, but does not necessarily address all concerns of this API. Notably:
- The API is not well specified because congestion control is outside the scope of webrtc-pc and I don't think anyone wants to change that. More SHOULD language is added to advice implementations, but there are no hard requirements. As such there is a risk that different implementations behave differently, but note that this statement is already true for any congestion control in place already.
- This may be an "NV" feature, not "1.0", but if not here, where does this go? Extension specs is an option, but lack of support becomes a bigger risk.

Is this enough to continue the discussion, or do we (Chrome) need to investigate alternatives?

@kuddai @alvestrand @jan-ivar @foolip

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6977,6 +6977,10 @@ async function updateParameters() {
           initialized to an empty list.</p>
         </li>
         <li>
+          <p>Let <var>receiver</var> have a <dfn>[[\ReceiverJitterBufferDelayHint]]</dfn>
+          internal slot initialized to <code>null</code>.</p>
+        </li>
+        <li>
           <p>Return <var>receiver</var>.</p>
         </li>
       </ol>
@@ -6985,6 +6989,7 @@ async function updateParameters() {
     readonly        attribute MediaStreamTrack  track;
     readonly        attribute RTCDtlsTransport? transport;
     readonly        attribute RTCDtlsTransport? rtcpTransport;
+                    attribute double? jitterBufferDelayHint;
     static RTCRtpCapabilities?         getCapabilities (DOMString kind);
     RTCRtpReceiveParameters            getParameters ();
     sequence&lt;RTCRtpContributingSource&gt;    getContributingSources ();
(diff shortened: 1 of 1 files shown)


henbos:
There is examples of similar things in Web Audio API already: https://webaudio.github.io/web-audio-api/#AudioContextOptions

henbos:
And content hint: https://www.w3.org/TR/mst-content-hint/

webrtc.html
```diff
@@ -7039,6 +7044,63 @@ <h2>Attributes</h2>
...
+            length of the jitter buffer using a <i>hint</i>. This allows the
+            application to influence the trafeoffs between having a longer or
+            shorter jitter buffer. The longer the buffer, the more chance there
+            is that packets arrive in time for playout, but the higher the delay
+            is. The shorter the buffer, the higher the risk of stuttering, but
+            the shorter the playout delay. For interactive applications, low
+            delay may be important, wheras for passive applications a
+            stutter-free experience may be more important.</p>
+            <p>Note that the delay is only a hint; a soft target. The underlying
+            congestion control mechanisms employed by user agensts are outside
```
henbos: Congestion control is not the same as jitter buffer delay.

henbos:
Next step: Extension spec, don't take away time from 1.0.

webrtc.html
fippo: playout buffering mechanisms? (stolen from the perkins book)

foolip:
@henbos do you want review of this PR, or will you go straight to a standalone spec?

henbos:
I'll make a standalone one and ask for your input on it over there :) after that I'll close all the PRs and issues related to this on webrtc-pc.

henbos:
Marking as Icebox for now, will move to an extension spec and close this.

henbos:
I moved this here: https://github.com/henbos/webrtc-timing/pull/2
New extension spec, currently in a "hello world"-like state.

Closing the PR on this spec.

PR #2152, opened 2019-04-01
Remove redundant steps in setCodecParameters

dontcallmedom:
Could not detect a difference these 2 additional steps make beyond what the previous step does.

(I found this during the webrtc testing hackathon and asked @henbos who seemed to agree this was redundant on a quick look)

***
<a href="https://pr-preview.s3.amazonaws.com/dontcallmedom/webrtc-pc/pull/2152.html" title="Last updated on Apr 1, 2019, 3:32 PM UTC (6af5614)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2152/943ebd1...dontcallmedom:6af5614.html" title="Last updated on Apr 1, 2019, 3:32 PM UTC (6af5614)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7717,18 +7717,6 @@ async function updateParameters() {
                   something to offer, regardless of <code><var>transceiver</var>.<a data-link-for="RTCRTPTransceiver">direction</a></code>.</p>
                 </li>
                 <li>
-                  <p>Let <var>codecCapabilities</var> be the union of
-                  <code>RTCRtpSender.getCapabilities(kind).codecs</code> and
-                  <code>RTCRtpReceiver.getCapabilities(kind).codecs</code>.</p>
-                </li>
-                <li>
-                  <p>For each <var>codec</var> in <var>codecs</var>,</p>
-                  <ol>
-                    <li>If <var>codec</var> is not in <var>codecCapabilities</var>, throw
-                    <code>InvalidModificationError</code>.</li>
-                  </ol>
-                </li>
-                <li>
                   <p>Set <var>transceiver</var>'s <a>[[\PreferredCodecs]]</a> slot to
                   <var>codecs</var>.</p>
                 </li>


webrtc.html
```diff
@@ -7716,18 +7716,6 @@ <h2>Methods</h2>
                   throw <code>InvalidModificationError</code>. This ensures that we always have
                   something to offer, regardless of <code><var>transceiver</var>.<a data-link-for="RTCRTPTransceiver">direction</a></code>.</p>
```
henbos: This part ensures that we have a "real" codec that we can use for sending, and we have a "real" codec that we can use for receiving. By "real" I mean a codec (e.g. VP8, VP9, H264 or profile thereof), not a line that says we support RTX/RED/FEC.

This means that it will work if direction is sendonly or the direction is recvonly.

What about sendrecv?
Let's assume you have the capability to send VP8 and H264 and receive VP9 and H264. If H264 is not in your list of codecs and the direction is sendrecv, what should be in the SDP? Neither VP8 or VP9 is valid in both directions.

webrtc.html
```diff
@@ -7716,18 +7716,6 @@ <h2>Methods</h2>
...
-                <li>
-                  <p>Let <var>codecCapabilities</var> be the union of
-                  <code>RTCRtpSender.getCapabilities(kind).codecs</code> and
-                  <code>RTCRtpReceiver.getCapabilities(kind).codecs</code>.</p>
-                </li>
-                <li>
-                  <p>For each <var>codec</var> in <var>codecs</var>,</p>
-                  <ol>
-                    <li>If <var>codec</var> is not in <var>codecCapabilities</var>, throw
-                    <code>InvalidModificationError</code>.</li>
```
henbos: I don't think this section is redundant, however I think it needs to say the same thing as above that if the intersection "only contains RTX, RED or FEC codecs or is an empty set" we should throw, otherwise you might succeed because codec contains "FEC" even though you don't have a codec for both sending and receiving.

henbos: If this section is updated to say that I do think that makes the first section redundant.

But we also need to check that there is no element in codecs that is not in either capability set, to make sure you cannot prefer a codec that you do not support.

dontcallmedom:
@henbos - any chance you could update the PR? I'm about to disappear for a couple of weeks, and you would be much better positioned not to make a mess of it anyway :)

henbos:
I'll close this one and continue at #2183

Issue #2147, opened 2019-03-27
simulcast offer doesn't work with addTrack. An omission


jan-ivar:
https://github.com/w3c/webrtc-pc/pull/2081 put all simulcast processing of offers from SFUs under:

 *"3. If no suitable transceiver is found (transceiver is unset), run the following steps:"*

This means that when a suitable transceiver from *addTrack* **is** found—[section 5.10](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-25#section-5.10) says they're suitable—then we succeed without processing the simulcast rids, a mistake.

To solve this, either we modify JSEP to say it's unsuitable, or add language to set the existing transceiver's [[SendEncodings]].

I see no reason why we can't do the latter. We've seen real-world interest in having *addTrack* work with simulcast.

We may need to bump the [transactionId](http://w3c.github.io/webrtc-pc/#dom-rtcrtpsendparameters-transactionid), but that's a whole other issue.

fippo:
Thanks @jan-ivar! I've run into a couple of bugs in Chrome [here](https://bugs.chromium.org/p/chromium/issues/detail?id=944821)

I do not think that `SRD, then use replaceTrack on a sender` is an API we should promote. Also there seem to be some cases where addTrack works (when called after setRemoteDescription; this is covered by the reuse in the addTrack algorithm afaics).

Quite oddly, calling addTransceiver and then having a remote description which matches the SendEncodings did not seem to match the existing transceiver either. Not an API I would like to see but its a case where we need to decide what to do.

jan-ivar:
> remote description which matches the SendEncodings did not seem to match 

[suitable](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-25#section-5.10) = *"If the m= section is sendrecv or recvonly, and there are RtpTransceivers of the same type that were added to the PeerConnection by addTrack and are not associated with any m= section and are not stopped, find the first (according to the canonical order described in Section 5.2.1) such RtpTransceiver."*

The WG decided that *addTransceiver* should be explicit, not an abstraction. See https://github.com/w3c/webrtc-pc/issues/2108

PR #2155, opened 2019-04-01
Attach simulcast offer sendEncodings to suitable (addTrack) transceiver.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2147.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2155.html" title="Last updated on Apr 4, 2019, 12:50 PM UTC (31161d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2155/e38ba0c...jan-ivar:31161d5.html" title="Last updated on Apr 4, 2019, 12:50 PM UTC (31161d5)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1764,7 +1764,9 @@
                             <p>If a suitable transceiver is found (<var>transceiver</var>
                             is set) and <var>sendEncodings</var> is non-empty, set
                             <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>
-                            to <var>sendEncodings</var>.</p>
+                            to <var>sendEncodings</var>, and set
+                            <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\LastReturnedParameters]]</a>
+                            to <code>null</code>.</p>
                           </li>
                           <li>
                             <p>If no suitable transceiver is found
--- a/webrtc.html
+++ b/webrtc.html
@@ -1751,20 +1751,26 @@
                             description</a>.</p>
                           </li>
                           <li>
+                            <p>If the <var>description</var> is of type <code>"offer"</code>
+                            and contains a request to receive simulcast, use the order
+                            of the rid values specified in the simulcast attribute to create
+                            an <code><a>RTCRtpEncodingParameters</a></code> dictionary for
+                            each of the simulcast layers, populating the <code>rid</code> member
+                            according to the corresponding rid value, and let <var>sendEncodings</var>
+                            be the list containing the created dictionaries. Otherwise,
+                            let <var>sendEncodings</var> be an empty list.</p>
+                          </li>
+                          <li>
+                            <p>If a suitable transceiver is found (<var>transceiver</var>
+                            is set) and <var>sendEncodings</var> is non-empty, set
+                            <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>
+                            to <var>sendEncodings</var>.</p>
+                          </li>
+                          <li>
                             <p>If no suitable transceiver is found
                             (<var>transceiver</var> is unset), run the following
                             steps:</p>
                             <ol>
-                              <li>
-                                <p>If the <var>description</var> is of type <code>"offer"</code>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1750,21 +1750,29 @@ <h4>Set the RTCSessionDescription</h4>
...
                             description</a>.</p>
                           </li>
+                          <li>
+                            <p>If the <var>description</var> is of type <code>"offer"</code>
+                            and contains a request to receive simulcast, use the order
+                            of the rid values specified in the simulcast attribute to create
+                            an <code><a>RTCRtpEncodingParameters</a></code> dictionary for
+                            each of the simulcast layers, populating the <code>rid</code> member
+                            according to the corresponding rid value, and let <var>sendEncodings</var>
+                            be the list containing the created dictionaries. Otherwise,
```
alvestrand: Suggest adding:
Let "supportedEncodings" be the maximum number of encodings that the implementation can support.
If "supportedEncodings" is less than the length of sendEncodings, truncate sendEncodings so that its length is "supportedEncodings".

webrtc.html
```diff
@@ -1750,21 +1750,29 @@ <h4>Set the RTCSessionDescription</h4>
                             object, <var>transceiver</var>, to represent the <a>media
                             description</a>.</p>
                           </li>
+                          <li>
+                            <p>If the <var>description</var> is of type <code>"offer"</code>
```
alvestrand: Suggest moving this paragraph before the "find a suitable transceiver" step, to make it clear that the sendEncodings may factor into the decision on whether a transceiver is suitable or not.
(I'm not sure what we should do if the existing transceiver has sendEncodings longer than this one; both truncating the list and failing to match seems to be defensible options. We should pick one.)

alvestrand:
I like where this is going.

webrtc.html
```diff
@@ -1750,21 +1750,29 @@ <h4>Set the RTCSessionDescription</h4>
...
+                            and contains a request to receive simulcast, use the order
+                            of the rid values specified in the simulcast attribute to create
+                            an <code><a>RTCRtpEncodingParameters</a></code> dictionary for
+                            each of the simulcast layers, populating the <code>rid</code> member
+                            according to the corresponding rid value, and let <var>sendEncodings</var>
+                            be the list containing the created dictionaries. Otherwise,
+                            let <var>sendEncodings</var> be an empty list.</p>
+                          </li>
+                          <li>
+                            <p>If a suitable transceiver is found (<var>transceiver</var>
```
henbos: nit: "was found", since this refers to a previous step. The same goes for the step below which you didn't touch but might as well in the drive-by.

webrtc.html
henbos: Good idea.

webrtc.html
jan-ivar: Will move it. Transceivers from *addTrack* will never have *sendEncodings* longer than one, so they should already fail to match.

jan-ivar:
Updated. Thanks for the feedback!

webrtc.html
```diff
@@ -1751,20 +1768,18 @@ <h4>Set the RTCSessionDescription</h4>
                             description</a>.</p>
                           </li>
                           <li>
-                            <p>If no suitable transceiver is found
+                            <p>If a suitable transceiver was found (<var>transceiver</var>
+                            is set) and <var>sendEncodings</var> is non-empty, set
+                            <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\sendEncodings]]</a>
+                            to <var>sendEncodings</var>, and set
+                            <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\LastReturnedParameters]]</a>
+                            to <code>null</code>.</p>
```
henbos: Does this mean we allow modifying sendEncodings in subsequent offers for transceivers that already exist?

jan-ivar: No, suitable = unused. See https://github.com/w3c/webrtc-pc/issues/2147#issuecomment-477355180.

henbos: Ack!

PR #2156, opened 2019-04-01
Add in-line references to relevant test cases

dontcallmedom:
To help measure and track test suite coverage.

Done during the WebRTC Hackathon in IETF 104.

I'm submitting this as is to avoid rot, and am working separately on additional tools to measure coverage based on that data, and to nudge us toward keeping the data up to date (when new tests are created, or when normative test is added/modified).

***
<a href="https://pr-preview.s3.amazonaws.com/dontcallmedom/webrtc-pc/pull/2156.html" title="Last updated on Apr 4, 2019, 2:58 PM UTC (b7d7657)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2156/42b2757...dontcallmedom:b7d7657.html" title="Last updated on Apr 4, 2019, 2:58 PM UTC (b7d7657)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -181,20 +181,20 @@
             <h2>Dictionary <a class="idlType">RTCConfiguration</a> Members</h2>
             <dl data-link-for="RTCConfiguration" data-dfn-for=
             "RTCConfiguration" class="dictionary-members">
-              <dt><dfn data-idl><code>iceServers</code></dfn> of type <span class=
+              <dt data-tests="RTCConfiguration-iceServers.html"><dfn data-idl><code>iceServers</code></dfn> of type <span class=
               "idlMemberType">sequence&lt;<a>RTCIceServer</a>&gt;</span></dt>
               <dd>
                 <p>An array of objects describing servers available to be used
                 by ICE, such as STUN and TURN servers.</p>
               </dd>
-              <dt><dfn data-idl><code>iceTransportPolicy</code></dfn> of type
+              <dt data-tests="RTCConfiguration-iceTransportPolicy.html"><dfn data-idl><code>iceTransportPolicy</code></dfn> of type
               <span class="idlMemberType"><a>RTCIceTransportPolicy</a></span>,
               defaulting to <code>"all"</code></dt>
               <dd>
                 <p>Indicates which candidates the <a>ICE Agent</a> is allowed
                 to use.</p>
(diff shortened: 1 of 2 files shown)


alvestrand:
Do we have a recommendation written up somewhere on how to use these attributes?
ie should we attach them to any paragraph defining any behavior that is tested by the relevant test?
Is there an attribute somewhere that says which directory the test is residing in (webrtc in this case)?
If a test tests behaviors for multiple sections, should we hoist it as far up as possible, or push it down to the lowest level paragraph that makes sense to tag?

If I'm going to try to maintain this, I'd like to understand the shape of the goal first.

dontcallmedom:
The property which states where tests reside is `testSuiteURI` in the ReSpec configuration object.

My guiding principle for where and when to use the `data-tests` attribute have been:
* apply it only to content with normative language (I have had to ignore that rule in a number of cases because we have not been very consistent in our usage of normative language)
* put it as high in the DOM as possible
* when set on an element, you assert that the said test case provides reasonable coverage of the entirety of the element

One of the expected outcome (for which I already have a proof of concept) is to get a view of what parts of the specs should be tested but aren't.

Does that help? Once this guidance is clear enough, it probably should be brought into the `README.md` file

webrtc.html
```diff
@@ -5274,25 +5274,25 @@ <h2>Methods</h2>
...
                   was called with a single argument.</p>
                 </li>
-                <li>
+                <li data-tests="RTCPeerConnection-addTrack.https.html">
                   <p>If <var>connection</var>'s <a>[[\IsClosed]]</a> slot is
                   <code>true</code>, <a>throw</a> an
                   <code>InvalidStateError</code>.</p>
                 </li>
-                <li>
+                <li data-tests="RTCPeerConnection-addTrack.https.html,">
```
henbos: Comma!

alvestrand:
Can you incorporate the README changes in this PR too?
Can we make a tag (data-tests-missing?) that we can incorporate as part of spec changes, before the tests are written, when we know we need some?
Please merge this once it's rebased and comments addressed.

dontcallmedom:
Missing tests can be tagged with an empty `data-tests` attribute (I've marked a few pieces of the spec that way).

I'll proceed with the updates and merge, thanks.

Issue #2150, opened 2019-03-28
transceiver.stop() needs more work (avoid BUNDLE footgun)


jan-ivar:
Most API complaints I hear are from people negotiating manually. It's hard to get right, and steeped in historical reasons (needing to munge SDP, Chrome's `negotiationneeded` not working). So let's check in on `negotiationneeded`:

```js
button.onclick = () => pc.getTransceivers()[0].stop();

pc.onnegotiationneeded = async e => {
  await pc1.setLocalDescription(await pc1.createOffer());
  await pc2.setRemoteDescription(pc1.localDescription);
  await pc2.setLocalDescription(await pc2.createAnswer());
  await pc1.setRemoteDescription(pc2.localDescription);
}
```
Clicking the button will most likely stop only the first transceiver, but once in a blue moon it will stop all of them. That's broken. 😞

Quick reason: `stop()` on the first transceiver marks this BUNDLE transport as a goner on next negotiation; hardly noticeable in `"stable"` state, since a new one pops up on next negotiation. But in `"have-remote-offer"` state, there's no allowance for creating or finding a replacement transport in the answer. The BUNDLE spec has painted us in a corner here.

I think we must address this.

### Proposal

~`transceiver.stop()` is already a bit weird, in that it's both instant *and* a negotiated property. I think we can solve this by making `transceiver.[[Stopped]]` more like `transceiver.[[Direction]]`. That is: postpone the effects—and more importantly, the expectations—of stop until negotiation. This should gives us the leeway to avoid rejecting the BUNDLE transport and instead leave the negotiationneeded flag set for next round. This might mean needing a new method—perhaps `transceiver.reject()`—if we're interested in the old behavior.~

jan-ivar:
Due to https://github.com/w3c/webrtc-pc/issues/2176 I no longer think there's a way to make `stop()` work safely with the offerer-tagged m-line being one of the exposed transceivers.

The BUNDLE spec [says](https://tools.ietf.org/html/draft-ietf-mmusic-sdp-bundle-negotiation-54#section-7.2.1): *"It is RECOMMENDED that the suggested offerer tagged "m=" section is a bundled "m=" section that the offerer believes it is unlikely that the answerer will reject, or move out of the BUNDLE group."*

**I see a couple of remaining options**:
 1. Live with it. Ironically, the problem seems easily worked around in JS, like this:
```js
const pc = new RTCPeerConnection(config);
[...]

PR #2168, opened 2019-04-09
Add stopping state (transceiver.stop() queues to stable). Rename old unsafe stop to reject()

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2150.

I had to scrap my initial thoughts of modeling a `currentlyStopped` state after `currentDirection`.

After reading JSEP, I realized we have no shot of altering its definition of ***stopped*** without major pain. So I left it alone and added a *webrtc-pc*-only ***stopping*** one instead:

stopping

The other difference I realized from `currentDirection` is that `stopped` is an *input* to negotiation, not a result from it, so it needs to happen ahead of it. I put it in the queued task that fires *negotiationneeded*. This ends up working whether people rely on the event to negotiate or not.

This PR is best explained though a mental detour:
 1. First define a new `stableStop()` method that sets this new `stopping` attribute.
 2. Then rename that `stableStop()` to `stop()`, and rename the old hazardous `stop()` to `reject()`.

Net result: A `stop()` safe to call regardless of state, designed to miss the danger window of `"have-remote-offer"`—instead landing in "stable", where a subsequent negotiation will pick it up, handling BUNDLE optimally.

Pros with safety glasses may use `reject()` to reject m-lines in `"have-remote-offer"`.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3981,6 +3981,15 @@ interface RTCSessionDescription {
                 <p>If <var>connection</var>'s <a>[[\NegotiationNeeded]]</a>
                 slot is <code>false</code>, abort these steps.</p>
               </li>
+              <li>
+                <p>For each <var>transceiver</var> in <var>connection</var>'s
+                <a>set of transceivers</a>, if
+                <var>transceiver</var>.<a>[[\Stopping]]</a> is <code>true</code> and
+                <var>transceiver</var>.<a>[[\Stopped]]</a> is <code>false</code>,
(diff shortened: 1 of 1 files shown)


jan-ivar:
I'm abandoning this approach because of  https://github.com/w3c/webrtc-pc/issues/2176. 

There's no guarantee an SRD(offer) won't have to contend with a `stop()` from stable that hasn't been negotiated yet. This is still too brittle.

Issue #2167, opened 2019-04-07
{iceRestart: true} works poorly with negotiationneeded


jan-ivar:
The currrent ICE restart API doesn't work well with `negotiationneeded`. Say you have this:
```js
pc.onnegotiationneeded = async () => {
  await pc.setLocalDescription(await pc.createOffer());
  io.send({desc: pc.localDescription});
}
```
How is the application supposed to restart ICE? Break that abstraction and call the methods directly, duplicating whatever logic the app may have put in there—after of course first checking we're in `"stable"` state and handling glare? That seems error-prone.

Performing an ICE restart is an application choice, often made for external network reasons. I think it deserves a first-class API that works with `negotiationneeded` and survives rollback:
```js
pc.restartIce();
```
This new method would set an internal slot that would cause negotiation to be needed, would invert the default value of the `{iceRestart}` offer option, and would be cleared in SRD(answer).

fippo:
What keeps the app from doing this:
```
async function negotiate(offerOptions) {
  await pc.setLocalDescription(await pc.createOffer(offerOptions));
  io.send({desc: pc.localDescription});
}
pc.onnegotiationneeded = () => negotiate();
pc.oniceconnectionstatechange = (e) => {
  if (pc.icecConnectonState == 'failed') {
    negotiate({iceRestart: true});
  }
};
```
Adding a restartIce method leads down the slippery slope of triggering it based on peerconnection logic, not application logic.

jan-ivar:
@fippo `iceconnectionstatechange` does not wait for `"stable"` state, so that code will fail occasionally. I think safer APIs will avoid these races.

> Adding a restartIce method leads down the slippery slope of triggering it based on peerconnection logic, not application logic.

I don't see that, but feel free to clarify. I see quite the opposite. It's today's API that seems buried in the details of peer connection's negotiation handling.

`negotiationneeded` is meant to help people encapsulate the detail-task of negotiating. If we split application logic up between what goes above that abstraction and what goes below it—with up being higher-level, closer to the app—I think ICE restart belongs above it.
```js
pc.oniceconnectionstatechange = () => pc.icecConnectonState == 'failed' && pc.restartIce();
```
Users shouldn't have to worry about `signalingState` at this level IMHO.

alvestrand:
On general principles, I object to adding new API surface to WebRTC 1.0.

It's not clear to me that the original question is well formed; what in the quoted code indicates that the application ever wants to restart ICE?

[...]

PR #2169, opened 2019-04-10
Add pc.restartIce() method.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2167. Provided for discussion.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2169.html" title="Last updated on Jul 11, 2019, 1:52 PM UTC (1ec7d9f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2169/230a0b2...jan-ivar:1ec7d9f.html" title="Last updated on Jul 11, 2019, 1:52 PM UTC (1ec7d9f)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1115,6 +1115,10 @@
             <dfn>[[\CurrentRemoteDescription]]</dfn> internal slot, initialized
             to <code>null</code>.</p>
           </li>
+          <li data-tests="RTCPeerConnection-constructor.html">
+            <p>Let <var>connection</var> have a <dfn>[[\RestartIce]]</dfn>
+            internal slot, initialized to <code>false</code>.</p>
+          </li>
           <li>
             <p>Return <var>connection</var>.</p>
           </li>
@@ -7422,10 +7426,6 @@ async function updateParameters() {
         <li>
           <p>Let <var>transceiver</var> have a <dfn>[[\PreferredCodecs]]</dfn> internal slot,
           initialized to an empty list.</p>
-        </li>
-        <li>
-          <p>Let <var>transceiver</var> have a <dfn>[[\RestartIce]]</dfn> internal slot,
-          initialized to <code>false</code>.</p>
         </li>
         <li>
           <p>Return <var>transceiver</var>.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -656,25 +656,28 @@
         </div>
         <div>
           <pre class="idl">dictionary RTCOfferOptions : RTCOfferAnswerOptions {
-             boolean iceRestart = false;
+             boolean iceRestart;
 };</pre>
           <section>
             <h2>Dictionary <dfn>RTCOfferOptions</dfn> Members</h2>
             <dl data-link-for="RTCOfferOptions" data-dfn-for="RTCOfferOptions"
             class="dictionary-members">
               <dt><dfn data-idl><code>iceRestart</code></dfn> of type <span class=
-              "idlMemberType">boolean</span>, defaulting to
-              <code>false</code></dt>
+              "idlMemberType">boolean</span></dt>
               <dd>
-                <p>When the value of this dictionary member is true, the
-                generated description will have ICE credentials that are
+                <p>When this dictionary member is not present, treat it as
+                having a default value set to the relevant
+                <code><a>RTCPeerConnection</a></code> object's
+                <a>[[\RestartIce]]</a> slot.</p>
+                <p>When the value of this dictionary member is <code>true</code>,
+                the generated description will have ICE credentials that are
                 different from the current credentials (as visible in the
                 <code><a data-link-for=
                 "RTCPeerConnection">localDescription</a></code> attribute's
                 SDP). Applying the generated description will restart ICE, as
                 described in section 9.1.1.1 of [[!ICE]].</p>
-                <p>When the value of this dictionary member is false, and the
-                <code><a data-link-for=
+                <p>When the value of this dictionary member is <code>false</code>,
+                and the <code><a data-link-for=
                 "RTCPeerConnection">localDescription</a></code> attribute has
                 valid ICE credentials, the generated description will have the
                 same ICE credentials as the current value from the
@@ -1488,7 +1491,9 @@
                         to <var>connection</var>.<a>[[\PendingLocalDescription]]</a>.
                         Set both <var>connection</var>.<a>[[\PendingRemoteDescription]]</a>
                         and <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
-                        to <code>null</code>. Finally set <var>connection</var>'s
+                        to <code>null</code>. Set
+                        <var>connection</var>.<a>[[\RestartIce]]</a> to
+                        <code>false</code>, and finally set <var>connection</var>'s
                         <a>signaling state</a> to <code>"stable"</code>.</p>
                       </li>
                       <!-- E) transition rollback to haveRemoteOffer to stable -->
@@ -2289,6 +2294,7 @@ interface RTCPeerConnection : EventTarget  {
     readonly        attribute RTCIceConnectionState     iceConnectionState;
     readonly        attribute RTCPeerConnectionState    connectionState;
     readonly        attribute boolean?                  canTrickleIceCandidates;
+    boolean                            restartIce ();
     static sequence&lt;RTCIceServer&gt;      getDefaultIceServers ();
     RTCConfiguration                   getConfiguration ();
     void                               setConfiguration (RTCConfiguration configuration);
@@ -3215,6 +3221,19 @@ interface RTCPeerConnection : EventTarget  {
                 above algorithm, indicates end-of-candidates for all media
                 descriptions and ICE candidate generation. This is by design for
                 legacy reasons.</p>
+              </dd>
+              <dt><dfn data-idl><code>restartIce</code></dfn></dt>
+              <dd>
+                <p>The <code>restartIce</code> method tells the
+                <code><a>RTCPeerConnection</a></code> that ICE should be
+                restarted. Subsequent calls to <code>createOffer</code> and
+                <code>createAnswer</code> will create descriptions that will
+                restart ICE, as described in section 9.1.1.1 of [[!ICE]].</p>
+                <p>When this method is called, the user agent MUST set the
+                <code><a>RTCPeerConnection</a></code> object's
+                <a>[[\RestartIce]]</a> internal slot to <code>true</code>, and
+                return <code>true</code> if it was previously <code>false</code>;
+                otherwise return <code>false</code>.</p>
               </dd>
               <dt data-tests="RTCPeerConnection-getDefaultIceServers.html"><dfn data-idl><code>getDefaultIceServers</code></dfn></dt>
               <dd>
@@ -7397,6 +7416,10 @@ async function updateParameters() {
         <li>
           <p>Let <var>transceiver</var> have a <dfn>[[\PreferredCodecs]]</dfn> internal slot,
           initialized to an empty list.</p>
+        </li>
+        <li>
+          <p>Let <var>transceiver</var> have a <dfn>[[\RestartIce]]</dfn> internal slot,
+          initialized to <code>false</code>.</p>
         </li>
         <li>
           <p>Return <var>transceiver</var>.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -3261,7 +3261,7 @@ interface RTCPeerConnection : EventTarget  {
                     <p><a data-lt="update the negotiation-needed flag">Update the
                     negotiation-needed flag</a> for <var>connection</var>.</p>
                   </li>
-                <ol>
+                </ol>
               </dd>
               <dt data-tests="RTCPeerConnection-getDefaultIceServers.html"><dfn data-idl><code>getDefaultIceServers</code></dfn></dt>
               <dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -666,24 +666,23 @@
               "idlMemberType">boolean</span></dt>
               <dd>
                 <p>When the value of this dictionary member is <code>true</code>,
+                or the relevant <code><a>RTCPeerConnection</a></code> object's
+                <a>[[\LocalUfragsToReplace]]</a> slot is not empty, then
                 the generated description will have ICE credentials that are
                 different from the current credentials (as visible in the
                 <code><a data-link-for=
-                "RTCPeerConnection">localDescription</a></code> attribute's
+                "RTCPeerConnection">currentLocalDescription</a></code> attribute's
                 SDP). Applying the generated description will restart ICE, as
                 described in section 9.1.1.1 of [[!ICE]].</p>
                 <p>When the value of this dictionary member is <code>false</code>,
+                and the relevant <code><a>RTCPeerConnection</a></code> object's
+                <a>[[\LocalUfragsToReplace]]</a> slot is empty,
                 and the <code><a data-link-for=
-                "RTCPeerConnection">localDescription</a></code> attribute has
-                valid ICE credentials, the generated description will have the
+                "RTCPeerConnection">currentLocalDescription</a></code> attribute has
+                valid ICE credentials, then the generated description will have the
                 same ICE credentials as the current value from the
                 <code><a data-link-for=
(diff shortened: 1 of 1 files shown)


henbos:
Should the "update negotiation-needed flag" steps be updated to say true if iceRestart is true, and calling restartIce() trigger updating negotiation-needed flag?

jan-ivar:
@henbos Yes! 👍 Thanks for catching that. That's what I get for working late. Will fix!

jan-ivar:
Fixed.

dontcallmedom:
It feels pretty late to add new API surface to WebRTC 1.0 - do we think implementations would implement this in the next few weeks?

webrtc.html
```diff
@@ -7398,6 +7423,10 @@ <h3><dfn>RTCRtpTransceiver</dfn> Interface</h3>
           <p>Let <var>transceiver</var> have a <dfn>[[\PreferredCodecs]]</dfn> internal slot,
           initialized to an empty list.</p>
         </li>
+        <li>
+          <p>Let <var>transceiver</var> have a <dfn>[[\RestartIce]]</dfn> internal slot,
+          initialized to <code>false</code>.</p>
```
henbos: This belongs to the PC, not the transceiver.

henbos:
If it weren't for the "update negotiation needed" steps this would probably be easy to shim.

henbos:
> It feels pretty late to add new API surface to WebRTC 1.0 - do we think implementations would implement this in the next few weeks?

I couldn't promise it would be prioritized in "the next few weeks". That is the danger with new APIs, even if they sound like relatively small work efforts...

jan-ivar:
> do we think implementations would implement this in the next few weeks?

I'll commit to having this in Firefox within a few weeks, if that will help.

As @henbos mentions, this isn't adding new functionality so much it's providing an API bug fix to an existing feature.

henbos:
If the remote offer restarts ICE you could clear the flag about restarting ICE. Requires checking the m= sections if the ufrag/pw has been updated. See negotiation needed algorithm

jan-ivar:
Firefox will have an implementation ready soon, behind a pref. WPT tests are available [here](https://phabricator.services.mozilla.com/D30978#C975484NL1).

webrtc.html
```diff
@@ -1471,39 +1484,42 @@ <h4>Set the RTCSessionDescription</h4>
...
+                        <var>connection</var>'s <a>[[\CurrentRemoteDescription]]</a>
+                        internal slot to a new <code><a>RTCSessionDescription</a></code>
+                        object constructed from <var>description</var>, and set
+                        <var>connection</var>'s <a>[[\CurrentLocalDescription]]</a>
+                        internal slot to <var>connection</var>'s
+                        <a>[[\PendingLocalDescription]]</a> internal slot.
+                        Set <var>connection</var>'s <a>[[\PendingRemoteDescription]]</a>
+                        and <a>[[\PendingLocalDescription]]</a> internal slots
+                        to <code>null</code>. Set <var>connection</var>'s
+                        <a>[[\RestartIce]]</a> internal slot to <code>false</code>,
```
henbos: Above, at SLD(answer) there is a condition on restart: "if none of the current ICE credentials are present in `description`, then set `connection`'s `[[RestartIce]]` internal slot to `false`".
But here, at SRD(answer), it unconditionally clears the [[RestartIce]] slot. Was that an optimization? Shouldn't we check if ICE was restarted here too?

Otherwise I think the following would not cause an ICE restart:
```
await pc.setLocalDescription(await pc.createOffer());  // old credentials
pc.restartIce();
await pc.setRemoteDescription(answer from the previous offer);
```
Whereas I would expect this to trigger a follow-up negotiation.

In both SRD(answer) and SLD(answer) I think we should be explicit, rather than saying "if the current ICE credentials are different" we should compare the [[Current...]] description slots with the description before we update current to match.

henbos: I'm implementing this right now... Question: in order to check if "current ICE credentials" are present in the description, does that mean I have to check both [[CurrentLocalDescription]] and [[CurrentRemoteDescription]] against description to be sure I don't miss any ufrags/pws?

henbos: If I understand things correctly, this is what should be done at both SLD(answer) and SRD(answer):

Prior to updating the [[CurrentLocalDescription]] slot: if [[CurrentLocalDescription]] is null or none of its ICE credentials are present in _description_, set [[RestartIce]] to false.

henbos: And ignoring rejected _description_ content in this comparison

henbos:
I have a working webrtc.org implementation of what I think this PR intends to do, [code here](https://webrtc-review.googlesource.com/c/src/+/144941). @jan-ivar is this correct?
1. In both SLD(answer) and SRD(answer) I am comparing the set description with the current ICE credentials, based on looking at [[CurrentLocalDescription]].
2. Because restartIce() does not take signalingState into account, this causes the [[RestartIce]] slot to reset when the O/A finishes, even if restartIce() was called after createOffer(). I am wondering [if sometimes we need to restart ICE twice](https://webrtc-review.googlesource.com/c/src/+/144941/1/pc/peer_connection_ice_unittest.cc#887)?

jan-ivar:
@henbos Great! At this rate you might beat me to it (I held off since Firefox merge day was yesterday)!

> In both SLD(answer) and SRD(answer) I am comparing the set description with the current ICE credentials, based on looking at [[CurrentLocalDescription]].

That sounds right. Note though that in [our implementation](https://phabricator.services.mozilla.com/D30976#C975916NL1170) we compare against currentLocalDescription from when `pc.restartIce()` was called. I think I was concerned about partial restarts from the other side moving the goalpost (but note bug I found from thinking about this again, thanks!)

> I am wondering [if sometimes we need to restart ICE twice?](https://webrtc-review.googlesource.com/c/src/+/144941/1/pc/peer_connection_ice_unittest.cc#887)

I answered [here](https://webrtc-review.googlesource.com/c/src/+/144941/1#message-a5cb61ee5f4a02097f00edea898e40824449c9e6). To me, this mirrors how e.g. `direction` works: we don't trigger on every set, only on value change. That seems most conservative to me. The alternative would be to have calling `pc.restartIce()` 10 times in a row trigger 10 renegotiations restarting ICE 10 times, which seems undesirable.

henbos:
> To me, this mirrors how e.g. direction works: we don't trigger on every set, only on value change. That seems most conservative to me. The alternative would be to have calling pc.restartIce() 10 times in a row trigger 10 renegotiations restarting ICE 10 times, which seems undesirable.

Hmm that is true. There is one difference though: with direction, you're happy as long as you end up with the most up-to-date direction having been negotiated. With restartIce(), are you happy that the ICE credentials are relatively new, even if they are older than the call to restartIce()? (Maybe? I don't understand some details here!)

Example:
1. createOffer() such that new ICE credentials are generated. This could be due to {iceRestart:true} or because it was the initial offer or new m= sections or whatever. Or because of restartIce().
2. restartIce(). Let's say that the network interface changed and all existing candidates are invalid, so you call this now. (Depending on what caused new ICE credentials at 1), this function may return true or it may return false!)
3. setLocalDescription() with the offer from 1); i.e. the ICE credentials are new compared to before this O/A exchange, but they are old compared to when the restartIce() at 2) was called. Perform the rest of the O/A, and at setRemoteDescription(answer) we get [[RestartIce]] = false.

In this example, ICE was in restarted, but it was not restarted as a consequence of 2), and no onnegotiationneeded event was ever fired because of restartIce().

My question is this: Did we just restart ICE with obsolete credentials or obsolete candidates, or does the relevant candidate gathering happen at 3) and we are happy anyway?

[Here](https://webrtc-review.googlesource.com/c/src/+/144941/1#message-a5cb61ee5f4a02097f00edea898e40824449c9e6) you say:

> Web sites can use the boolean return value from pc.restartIce() to learn whether it triggered anything. E.g. to force an additional ICE restart even if a restart is in progress:
> 
>     (async () => {
>       while (!pc.restartIce()) await new Promise(r => setTimeout(r, 100));
>     })();
> Or to avoid polling, replace setTimeout with
> 
>     pc.onsignalingstatechange = () => pc.signalingState == "stable" && r()

Firstly, if the ICE restart was not caused by restartIce() but by being the first offer, then restartIce() would still return true even though the ICE-restart-in-progress did not correspond to the restartIce() call.

Secondly, it is not intuitive that you would have to wrap your restartIce() calls in logic like this. This looks similar to me to the way that setLocalDescription() is not glare-proof today so you have to add additional logic around it for it to work (e.g. [slide](https://docs.google.com/presentation/d/1xcvf0udNeSH7s1FOY7RRqr1dEFvokZjn-MZPjwy3iXQ/edit#slide=id.g5c2f3df65b_11_66)).

It would be nice if our API "_just works_", no ifs-and-buts.

If this is a real problem, one way to solve it would be to say that [[RestartIce]] is not a boolean, but an enum with value "not-needed", "needed-when-stable" and "needed". restartIce() would cause its value to change to "needed" except if an O/A is already pending, in which case it becomes "needed-when-stable". If "needed-when-stable", whenever we reach the stable signaling state we set it to "needed". Is this needed? Does it fix the problem?

henbos:
> If this is a real problem, one way to solve it would be to say that [[RestartIce]] is not a boolean, but an enum with value "not-needed", "needed-when-stable" and "needed". restartIce() would cause its value to change to "needed" except if an O/A is already pending, in which case it becomes "needed-when-stable". If "needed-when-stable", whenever we reach the stable signaling state we set it to "needed".

Note that in this proposal, calling restartIce() 10 times in a row does not cause 10 onnegotiationneeded; it does however ensure that there will be a negotiation needed as a consequence of the call, even if we are currently negotiating old credentials. It does not queue up multiple calls, but it says that currently negotiating is not good enough for our restart.

jan-ivar:
> My question is this: Did we just restart ICE with obsolete credentials or obsolete candidates, or does the relevant candidate gathering happen at 3) and we are happy anyway?

It happens in 3). Does it matter that the credentials predate `restartIce` or are we happy? 🙂

Also, `createOffer`/`createAnswer` have [special logic](https://w3c.github.io/webrtc-pc/#dfn-final-steps-to-create-an-offer) to rerun to pick up state changes like this. 

But I think what you're touching on is what happens if `restartIce()` is called outside of stable state?

> With restartIce(), are you happy that the ICE credentials are relatively new, even if they are older than the call to restartIce()?

As written, yes: Success = no ICE credentials match those *current* when `restartIce()` was called.¹

This way, a remote restart = success as well.

> Let's say that the network interface changed and all existing candidates are invalid, so you call [restartIce()]

This would require a different definition: Success = candidates *generated* more recently than when `restartIce` was called. Maybe:

 Success = no ICE credentials match those *current* **or** ***pending*** when `restartIce()` was called?

Does that work? A remote offer would then satisfy the restart ONLY if it comes in *after* the call to `restartIce` (because otherwise it is the *pendingLocalDescription*).

> an enum with value "not-needed", "needed-when-stable" and "needed"

That might work, but I think looking at *pendingDescription* would have the same effect, and handle remote offers more efficiently.

---
<sub>1) I plan to update the PR to match Firefox as mentioned in https://github.com/w3c/webrtc-pc/pull/2169#issuecomment-509681629 to cache *currentLocalDescription* (as well as *pendingLocalDescription* if we decide to go that way) at time of call in [[RestartIce]].</sub>

webrtc.html
jan-ivar: Sorry for the late response!

> Otherwise I think the following would not cause an ICE restart:

If we look at both *currentLocalDescription* and *pendingLocalDescription*, then I think this should work. I'll update the PR. 

Additionally, I'll update the PR to instead cache the descriptions at time of restartIce, and compare against those, to match Firefox.

If a local answer is being set, then compare against [[CurrentLocalDescription]].

If a remote answer is being set, then in the PR I had us always clear the [[RestartIce]] slot, thinking that this was the culmination of a negotiation round originating from our end. But as you point out, this does the wrong thing if `restartIce()` is called outside of stable. I think this means we want to compare against *currentRemoteDescription* and *pendingRemoteDescription*.

> And ignoring rejected description content in this comparison

I don't know that we need to skip rejected ones. Isn't rejection orthogonal?

jan-ivar: Discussed with @docfaraday. We can ignore remote ufrags thanks to JSEP, focusing on local ones and comparing them against the newly set currentDescription in SRD(answer).

jan-ivar:
@henbos PTAL. I think this should do it. The method now returns `void`.

jan-ivar:
Last push clobbered @aboba's formatting. I've opened https://github.com/w3c/webrtc-pc/issues/2226 to deal formatting separately.

webrtc.html
```diff
@@ -656,30 +656,34 @@ <h2>Dictionary <dfn>RTCOfferAnswerOptions</dfn>
...
+                the generated description will have ICE credentials that are
                 different from the current credentials (as visible in the
                 <code><a data-link-for=
                 "RTCPeerConnection">localDescription</a></code> attribute's
                 SDP). Applying the generated description will restart ICE, as
                 described in section 9.1.1.1 of [[!ICE]].</p>
-                <p>When the value of this dictionary member is false, and the
-                <code><a data-link-for=
+                <p>When the value of this dictionary member is <code>false</code>,
+                and the <code><a data-link-for=
```
henbos: Do we want "iceRestart = false" to ignore restartIce()? It makes sense in terms of meaning, and onnegotiationneedede would still fire when getting back to stable, but is there a reason not to treat "= false" the same as missing? It adds complexity and means there are effectively three values of iceRestart, not two.

I'm not convinced that we should change this, but I think it's worth discussing before merging this.

webrtc.html
```diff
@@ -3216,6 +3235,34 @@ <h2>Methods</h2>
...
+                    <p>Let <var>connection</var> be the
+                    <code><a>RTCPeerConnection</a></code> on which the method
+                    was invoked.</p>
+                  </li>
+                  <li>
+                    <p>Empty <var>connection</var>.<a>[[\LocalUfragsToReplace]]</a>,
+                    and populate it with all ufrags found in
+                    <var>connection</var>.<a>[[\CurrentLocalDescription]]</a>,
+                    as well as all ufrags found in
+                    <var>connection</var>.<a>[[\PendingLocalDescription]]</a>.</p>
```
henbos: Populating and comparing ufrags is not explicit here (e.g. is it a list of strings or a map of content section names and ufrags, etc) but I think this makes sense to an implementer.

webrtc.html
henbos: (I'm also not convinced we shouldn't change it ;) )

jan-ivar: I still need to change the WPT tests. With the latest change where we no longer clear the internal slot always in SRD(answer), the effect of `{iceRestart: false}` will merely be to produce one non-restarting offer. It won't terminally cancel the restart the way it used to (meaning the engine will still fire `negotiationneeded` until the restart job is done.

That seems reasonable and benign to me. But at the same time it now seems kinda useless. There's really no way to cancel the restart now. Perhaps that is fine, but then perhaps we should also get it over with ASAP, and not leave these loopholes where one can drag it out.

We could also change it to the way you mention. It just seemed odd to me to restart ICE in an explicit`createOffer({iceRestart: false})` call. But we can do that.

henbos: For what it's worth, it would be easier to implement iceRestart: false as the same as undefined, because then we don't have to introduce a third value in all the layers in chromium and webrtc.org. There's also less to test. And as you say, you can't abort a restart, so why would you generate an offer that doesn't restart - you're just wasting time.

In the future, setLocalDescription() will generate the offer for you anyway, so I don't see the need to maintain this level of control.

henbos: If you did restartIce() + createOffer({iceRestart: false}) it was likely not intentional.

jan-ivar: Fixed. PTAL

henbos: Looks good, just make sure to revert the dictionary change, i.e. so that false is still the default value for it!

webrtc.html
jan-ivar: In Firefox we use a Set. We're populating here, not comparing. LMK if you want a change.

jan-ivar: It's also initialized to an "empty set" in this PR.

jan-ivar:
Tests in https://github.com/web-platform-tests/wpt/pull/17897.

Issue #2172, opened 2019-04-15
Clarify the definition of "playout" for `RTCRtpContributingSource`.


chxg:
This is related to [WebRTC Issue 10545](https://bugs.chromium.org/p/webrtc/issues/detail?id=10545).

There seem to be some confusion around the exact definition of "playout" for [RTCRtpContributingSource](https://w3c.github.io/webrtc-pc/#dom-rtcrtpcontributingsource):

1. **After receiving a packet. Before it's added to the jitter buffer.**

This interpretation is supported by:

> The browser MUST keep information from RTP packets _received_ in the previous 10 seconds.

2. **After a packet is removed from the jitter buffer. Before the samples/frames are delivered to `MediaStreamTrack`.**

This interpretation is supported by:

> When the first audio or video frame contained in an RTP packet is _delivered_ to the RTCRtpReceiver's MediaStreamTrack for playout...

3. **Exactly when the packet payload is physically played out on the monitor/speaker.**

This interpretation is supported by:

> The timestamp of type DOMHighResTimeStamp [HIGHRES-TIME], indicating the most recent time of _playout of media_ that arrived in an RTP packet originating from this source.

WebRTC currently implements (1). It updates `RTCRtpContributingSource` when a packet is received from the network. This is problematic. There are, for example, no guarantees that received packets are ever going to be played out (e.g. audio having already done concealment beyond the RTP timestamp of the packet payload). Packet reordering may also cause packets with lower sequence numbers to overwrite the information of packets with a higher sequence number.

I believe that (3) would be ideal in terms of usefulness, and ideal in terms of how a normal person would interpret the spec, but it's unfortunately not implementable in practice. An implementation of the spec will rarely have access to frame-perfect `onPhysicallyPlayedOut()` callbacks.

This leaves us with (2), which should be both implementable and useful.

jan-ivar:
"Playout" is 3. Isn't that clear? This is what Firefox implements. The primary goal of this info is audio bars visually synced with playing video.

na-g:
As @chxg says, physical device playout time isn't knowable in practice. Firefox records the levels into a 10 second rolling history when the packet is received and adjusts the playout time based on the jitter buffer at time of reception. This sounds like what (lib)WebRTC does. In practice this seems to yield good results for the use cases that @jan-ivar mentions.

chxg:
Thanks for the information!

[...]

PR #2178, opened 2019-04-24
Add rtpTimestamp to RTCRtpContributingSource

### LINKED ISSUE #2177: Add RTP timestamp to RTCRtpContributingSource
One use case for this would be to synchronize external events to the corresponding RTP packets.

drkron:
Fixes #2177

***
<a href="https://pr-preview.s3.amazonaws.com/drkron/webrtc-pc/pull/2178.html" title="Last updated on May 9, 2019, 4:09 AM UTC (5a49f49)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2178/0d55754...drkron:5a49f49.html" title="Last updated on May 9, 2019, 4:09 AM UTC (5a49f49)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7340,8 +7340,8 @@ async function updateParameters() {
               127 is converted to 0, and all other values are converted using
               the equation: <code>10^(-rfc_level/20)</code>.</p>
             </dd>
-            <dfn data-idl><code>rtpTimestamp</code></dfn> of type <span class=
-            "idlMemberType">unsigned long</span>
+            <dt><dfn data-idl><code>rtpTimestamp</code></dfn> of type
+                <span class="idlMemberType">unsigned long</span></dt>
             <dd>
               <p>The RTP timestamp of the latest played out media that arrived
               in an RTP packet originating from this source.</p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -7295,6 +7295,7 @@ async function updateParameters() {
     required DOMHighResTimeStamp timestamp;
     required unsigned long       source;
              double              audioLevel;
+             unsigned long       rtpTimestamp;
 };</pre>
         <section>
           <h2>Dictionary RTCRtpContributingSource Members</h2>
@@ -7338,6 +7339,12 @@ async function updateParameters() {
               <p>To convert these values to the linear 0..1 range, a value of
               127 is converted to 0, and all other values are converted using
               the equation: <code>10^(-rfc_level/20)</code>.</p>
+            </dd>
+            <dfn data-idl><code>rtpTimestamp</code></dfn> of type <span class=
+            "idlMemberType">unsigned long</span>
+            <dd>
+              <p>The RTP timestamp of the latest played out media that arrived
+              in an RTP packet originating from this source.</p>
             </dd>
           </dl>
         </section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -7343,8 +7343,8 @@ async function updateParameters() {
             <dt><dfn data-idl><code>rtpTimestamp</code></dfn> of type
                 <span class="idlMemberType">unsigned long</span></dt>
             <dd>
-              <p>The RTP timestamp of the latest played out media that arrived
-              in an RTP packet originating from this source.</p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7339,6 +7340,12 @@ <h2>Dictionary RTCRtpContributingSource Members</h2>
               127 is converted to 0, and all other values are converted using
               the equation: <code>10^(-rfc_level/20)</code>.</p>
             </dd>
+            <dfn data-idl><code>rtpTimestamp</code></dfn> of type <span class=
+            "idlMemberType">unsigned long</span>
+            <dd>
+              <p>The RTP timestamp of the latest played out media that arrived
+              in an RTP packet originating from this source.</p>
```
alvestrand: Can we refer this to the variable above, as "RTP timestamp of the media played out at "timestamp", to make it clear that the two timestamps are required to be from the same packet?

webrtc.html
```diff
@@ -7295,6 +7295,7 @@ <h2>Methods</h2>
     required DOMHighResTimeStamp timestamp;
     required unsigned long       source;
              double              audioLevel;
+             unsigned long       rtpTimestamp;
```
jan-ivar: Note that *RTCRtpSynchronizationSource* is derived from *RTCRtpContributingSource*. Did you mean to add *rtpTimestamp* to both? Just making sure it's not done by accident.

henbos:
@jan-ivar to get implementers to comment

jan-ivar:
@na-g can you comment on how this fits with our implementation? My concern is that this API is geared toward being "the audio bar API" in many respects, principally in that, at the time you poll the method, you get data related to playout *right now*. This means you won't get *rtpTimestamp* until playout time. Is this sufficient for the use-cases suggested, or does this require a new API?

webrtc.html
jan-ivar: I agree with @alvestrand that this needs to correlate with the same packet, not the "latest".

See also https://github.com/w3c/webrtc-pc/issues/2172#issuecomment-486740788 wrt this API—at least in Firefox—returns data no sooner than at time of playout.

webrtc.html
```diff
@@ -7339,6 +7340,12 @@ <h2>Dictionary RTCRtpContributingSource Members</h2>
               127 is converted to 0, and all other values are converted using
               the equation: <code>10^(-rfc_level/20)</code>.</p>
             </dd>
+            <dt><dfn data-idl><code>rtpTimestamp</code></dfn> of type
+                <span class="idlMemberType">unsigned long</span></dt>
+            <dd>
+              <p>The RTP timestamp of the latest played out media that arrived
```
jan-ivar: This RTP timestamp will be from a remote clock, right? Do we need language here defining this similar to [remoteTimestamp](https://w3c.github.io/webrtc-stats/#dom-rtcremoteoutboundrtpstreamstats-remotetimestamp)?

webrtc.html
drkron: Yes, that makes sense. I think this change should be coordinated with #2172 .

webrtc.html
drkron: I don't know the details of this well enough to tell which is the correct place to put it. To me it makes  sense put it where you have the other timestamp in contributing source though.

webrtc.html
henbos: Let's reference [RFC3550] section 5.1

webrtc.html
henbos: I think it belongs here. Both the SSRC and CSRC dictionary contains information from a particular RTP packet. This is more information about that RTP packet, so it can be added to both. (For the intended use case it would be enough to have it in the SSRC dictionary, but it doesn't hurt to have it here too.)

webrtc.html
drkron: Done!

webrtc.html
drkron: Done.

drkron:
@jan-ivar @aboba @alvestrand Please have a look at the updated PR.

Issue #2171, opened 2019-04-13
Clarify "multiple sources of media stitched together" at Note describing replaceTrack


guest271314:
1) Am consider composing a PR for another W3C repository (MediaRecorder) which will prospectively use language contained in this W3C repository; the description of replaceTrack() https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack, particularly 

> 6.4.3 If sending is true, and withTrack is not null, have the sender switch seamlessly to transmitting withTrack instead of the sender's existing track.

2) When reading content in this specification relating to `replaceTrack()` there is a note at https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack#issue-container-generatedID-29

> NOTE
> 
> There is not an exact 1:1 correspondence between tracks sent by one RTCPeerConnection and received by the other. For one, IDs of tracks sent have no mapping to the IDs of tracks received. Also, replaceTrack changes the track sent by an RTCRtpSender without creating a new track on the receiver side; the corresponding RTCRtpReceiver will only have a single track, potentially representing **multiple sources of media stitched together**. Both addTransceiver and replaceTrack can be used to cause the same track to be sent multiple times, which will be observed on the receiver side as multiple receivers each with its own separate track. Thus it's more accurate to think of a 1:1 relationship between an RTCRtpSender on one side and an RTCRtpReceiver's track on the other side, matching senders and receivers using the RTCRtpTransceiver's mid if necessary. (emphasis added)

where the term "multiple sources of media stitched together" is used. The specification does not clarify that procedure. Is using some of the above language (read https://www.w3.org/Consortium/Legal/2015/doc-license) in this specification in a PR for another W3C repository/specification ok; if so, does this specification need to be cited?

For the purpose of avoiding confusion as to precisely what the PR is describing can this specification clarify what "stitched together" technically means and if the specification defines the procedure to stitch together "multiple sources of media"; or, if the procedure for _multiple sources of media stitched together_ is left to implementers?

fippo:
"multiple sources of media stitched together" is probably describing the time behaviour, i.e. that RTCRtpSender is supposed to generate a continuous series of timestamps.

guest271314:
@fippo The code at Chromium source appears to use 

[...]

PR #2180, opened 2019-04-24
Clarify "multiple sources of media stitched together"

aboba:
When `sender.replaceTrack` is executed, the sender's track is replaced but the `RTCRtpReceiver`still only sees a single RTP stream with a continuous sequence number space.  So in this case, the `RTCRtpReceiver` does not "stitch together" two distinct RTP streams, as would be required when receiving simulcast.  Therefore, the text is wrong and should be removed.

Fix for Issue https://github.com/w3c/webrtc-pc/issues/2171

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2180.html" title="Last updated on Apr 24, 2019, 10:14 PM UTC (6c360cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2180/0d55754...6c360cb.html" title="Last updated on Apr 24, 2019, 10:14 PM UTC (6c360cb)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5121,8 +5121,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
     <code><a data-link-for="RTCRtpSender">replaceTrack</a></code> changes the
     track sent by an <code><a>RTCRtpSender</a></code> without creating a new
     track on the receiver side; the corresponding
-    <code><a>RTCRtpReceiver</a></code> will only have a single track,
-    potentially representing multiple sources of media stitched together. Both
+    <code><a>RTCRtpReceiver</a></code> will only have a single track. Both
     <code><a data-link-for="RTCPeerConnection">addTransceiver</a></code> and
     <code><a data-link-for="RTCRtpSender">replaceTrack</a></code> can be used to
     cause the same track to be sent multiple times, which will be observed on


webrtc.html
```diff
@@ -5121,8 +5121,7 @@ <h2>RTP Media API</h2>
     <code><a data-link-for="RTCRtpSender">replaceTrack</a></code> changes the
     track sent by an <code><a>RTCRtpSender</a></code> without creating a new
     track on the receiver side; the corresponding
-    <code><a>RTCRtpReceiver</a></code> will only have a single track,
-    potentially representing multiple sources of media stitched together. Both
+    <code><a>RTCRtpReceiver</a></code> will only have a single track. Both
```
jan-ivar: I think we should explain what happens. To me, this is a high-level note explaining the model, not instructions to an implementer. I was never confused by it, but if "stitched together" is confusing people, we should use other words to describe the net result.

How about referring to the observable settings just being added in https://github.com/w3c/webrtc-pc/pull/2188?

*"will only have a single track, composed of the RTP data received. This means that properties like* `width`, `height` *and* `frameRate` *may change as a result."*?

alvestrand: "Properties like width, height and frameRate may change over the lifetime of the track, based on the RTP data received."

guest271314: @jan-ivar "stitched together" is not _confusing_. You are correct in your analysis that the issue filed https://github.com/w3c/webrtc-pc/issues/2171 is asking for how the procedure happens. That is, how to reproduce the effect of `replaceTrack` without necessarily using `RTCRtpSender.replaceTrack`. Ultimately, the requirement for own code is to not use `RTCPeerConnection` at all while still having the ability to replace a `MediaStreamTrack` that is defined within a `MediaStream` passed to `MediaRecorder` while the `MediaRecorder` instance is active - at both Firefox and Chromium. Browsed the source code at Firefox and Chromium though not sure how to replicate the functionality of `replaceTrack`; perhaps create a `Proxy` object for each `MediaStreamTrack` which returns the same `id` for each respective track that is expected to be "stitched together"?

guest271314: @jan-ivar E.g., what would be necessary for `replaceTrack` to be implemented as a global method that can be used for any `MediaStream` or `MediaRecorder` instance to replace a given `MediaStreamTrack` - without having to use `RTCPeerConnection` code simply for the purpose of using `RTCRtpSender.replaceTrack()`?

guest271314: @aboba Should make it clear here and now that the language "multiple sources of media stitched together" and "have the sender switch seamlessly to transmitting withTrack instead of the sender's existing track" is not confusing - the issue did not state that the language should be removed. That language is what lead this user to experiment with `RTCPeerConnection` within the scope of attempting to record multiple media fragments to a single media file using `MediaRecorder` (specifically, not using any other third-party API than what is shipped with the ostensibly FOSS browsers Chromium and Firefox). The issue was filed to ask for the complete procedure used for `replaceTrack`, for the purpose of "reverse engineering" the procedure, and/or for the ability to file a PR at media-capture-record (or other appropriate repository) which uses the outline of how `replaceTrack` works to have no confusion as to what am requesting at media-capture-record, to address the long-standing feature requests of being able to record multiple media fragments or complete tracks (specifically, video tracks) using `MediaRecorder` (https://github.com/w3c/mediacapture-record/issues/4; https://github.com/w3c/mediacapture-record/issues/147; https://github.com/w3c/mediacapture-record/issues/166; https://bugs.chromium.org/p/chromium/issues/detail?id=528523; https://bugzilla.mozilla.org/show_bug.cgi?id=1276928).

PR #2186, opened 2019-04-25
Adds a note clarifying removeTrack()

### LINKED ISSUE #2024: Differences between pc.removeTrack(sender) and sender.replaceTrack(null)
The spec does not make clear whether a `transceiver` can be reused for sending a new track (or the previous one) after having called `pc.removeTrack(transceiver.sender)`.

Assuming a simple scenario with a single `transceiver` for just **sending** audio (don't receive anything).

* Q1: Is `sender.replaceTrack(track)` allowed after `pc.removeTrack(sender)`?
   * Chrome and Firefox allow it. Safari fails, but it seems they are fixing it [1][2].

* Q2: Should `sender.replaceTrack(null)` automatically set `a=inactive`?
   * Browsers don't do this and `direction` remains "sendonly".

* Q3: Is there any constraints regarding when `transceiver.direction` can be changed by the app? According to the spec there is not, and both Chrome and Firefox never complain, but Safari 12.1 produces an exception when setting the `direction` in certain circumstances (hard to explain them since it depends on different factors and probably it's just a bug).

* Q4: Should the browser keep sending RTCP SDES and RTCP SenderReports for the corresponding ssrc(s) after calling `sender.replaceTrack(null)`?
   * Chrome still sends both SDES and SenderReports.
   * If `transceiver.direction` is manually set to "inactive", then Chrome just sends SDES.

* Q5: Should the browser keep sending RTCP SDES for the corresponding ssrc(s) after calling `pc.replaceTrack(sender)`?
   * Chrome does. Firefox does not.

* Q6: If the answer to Q1 is yes, what's the purpose of `pc.removeTrack(sender)` having already `sender.replaceTrack(null)`? Just those items come to my mind:
    * `pc.removeTrack(sender)` cleans `RtpSender` settings (such as configured sending parameters, used SSRCs, stats, etc).
    * `sender.replaceTrack(null)` does not, so attaching a new track to the `sender` will send it with previous sender configuration/encodings.

[1] https://bugs.webkit.org/show_bug.cgi?id=191202
[2] https://bugs.webkit.org/attachment.cgi?id=353842&action=diff

henbos:
Fixes #2024.
Clarifies that the effects of removeTrack() can be achieved with transceiver.direction and replaceTrack.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2186.html" title="Last updated on May 2, 2019, 1:37 PM UTC (98cc6c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2186/0d55754...henbos:98cc6c3.html" title="Last updated on May 2, 2019, 1:37 PM UTC (98cc6c3)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5473,6 +5473,13 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               if the <code><a>MediaStreamTrack</a></code> is not already muted,
               a <code title="event-MediaStreamTrack-muted">muted</code> event is
               fired at the track.</p>
+              <div class="note">The same effect as <code>removeTrack()</code>
+              can be achieved setting the
+              <code>RTCRtpTransceiver.direction</code> attribute of the
+              corresponding transceiver and invoking
+              <code>RTCRtpSender.replaceTrack(null)</code> on the sender. One
+              minor difference is that <code>replaceTrack()</code> is
+              asynchronous and <code>removeTrack()</code> is synchronous.</div>
               <p>When the <dfn data-idl><code>removeTrack</code></dfn> method is
               invoked, the user agent MUST run the following steps:</p>
               <ol>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5474,7 +5474,7 @@ interface RTCPeerConnectionIceErrorEvent : Event {
               a <code title="event-MediaStreamTrack-muted">muted</code> event is
               fired at the track.</p>
               <div class="note">The same effect as <code>removeTrack()</code>
-              can be achieved setting the
+              can be achieved by setting the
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -5473,6 +5473,13 @@ <h2>Methods</h2>
               if the <code><a>MediaStreamTrack</a></code> is not already muted,
               a <code title="event-MediaStreamTrack-muted">muted</code> event is
               fired at the track.</p>
+              <div class="note">The same effect as <code>removeTrack()</code>
+              can be achieved setting the
```
alvestrand: "can be achieved setting" -> "can be achieved by setting"

PR #2188, opened 2019-04-25
Define remote track constraints (only applicable as settings)

### LINKED ISSUE #2121: Constrainable properties on remote tracks are under-specified
What behaviors specific to remote peer connection tracks are available through the constraints APIs? There's a [single paragraph](https://w3c.github.io/webrtc-pc/#mediatracksupportedconstraints-mediatrackcapabilities-mediatrackconstraints-and-mediatracksettings) on the topic, and it doesn't say.

From @guidou in https://github.com/w3c/webrtc-pc/issues/2109#issuecomment-470566522, Chrome appears to have implemented something, so we should probably specify it, like we did in https://github.com/w3c/mediacapture-screen-share/issues/31. Otherwise, the next question might be why don't we support [whiteBalanceMode](https://w3c.github.io/mediacapture-image/#dom-mediatracksettings-whitebalancemode)?

Discuss.

henbos:
Fixes #2121.

This PR contains only the constraints that have already been shipped in Chrome.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2188.html" title="Last updated on Oct 3, 2019, 2:23 PM UTC (e2cd748)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2188/5a8ff03...henbos:e2cd748.html" title="Last updated on Oct 3, 2019, 2:23 PM UTC (e2cd748)">Diff</a>


webrtc.html
```diff
@@ -11090,19 +11090,73 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+            </tr>
+            <tr id="def-constraint-resizeMode">
+              <td><dfn>resizeMode</dfn></td>
+              <td><code><a>ConstrainDOMString</a></code></td>
+              <td>Downscaling can happen if constraints are applied to width or
+              height. The resize mode affects how the User Agent downscales
+              frames, as defined in [[!GETUSERMEDIA]].</td>
+            </tr>
+          </tbody>
+        </table>
```
jan-ivar: I think we need a chapter on *how* downscaling and frame decimation MUST be done, and how the *applyConstraints* is to be implemented, like we do in https://w3c.github.io/mediacapture-screen-share/#downscaling-and-frame-decimation

henbos: Can I simply reference that section?

jan-ivar: Editors said they don't want webrtc-pc to normatively reference screen-capture.

Also, screen-capture disallows cropping. Doesn't Chrome support cropping? If we want that here we should perhaps provide our own section.

Sadly, there may not be one downscaling algorithm to rule them all. (Yes I know the remote track may be a screen sharing track, but we can't have it all I guess).

henbos: Taking another look, resizeMode is underspecified in gUM (I filed https://github.com/w3c/mediacapture-main/issues/584). In terms of remote track, "none" doesn't make any sense. I think resizeMode shouldn't be applicable to remote tracks. But we need to do make sure it is specified how to downscale it. I'll check with @guidou what Chrome does.

youennf:
I think the support of these constraints should be made optional, not mandatory. If we want to further beef up the support of constraints, a compelling use case would be useful. 

Also, if the implementation is generic enough to be applicable to all video tracks, I am not sure it makes sense to special case webrtc-pc remote tracks.

henbos:
> I think the support of these constraints should be made optional, not mandatory. If we want to further beef up the support of constraints, a compelling use case would be useful.
> 
> Also, if the implementation is generic enough to be applicable to all video tracks, I am not sure it makes sense to special case webrtc-pc remote tracks.

The intent is not to do more "special-casing" of webrtc tracks but less.
I am not proposing we add new constraints to remote tracks unless there is a strong use case for it that has to be argued for separately.

The constraints listed in this PR are ones generic enough to be applicable to any video track, which is why Chrome shipped them for webrtc tracks when the spec was unclear about what is or is not applicable, and our implementors though that they were because they were generic.

I think we should either say "constraints X, Y and Z are applicable and here's how: (description here that works with a generic implementation for any video track, so that implementations can do the same thing that they are already doing for gUM tracks)" or we should say "no constraints are applicable to remote tracks".

I'm trying to get the spec to match Chrome. It is also valid that Chrome conforms by removing its constraints support for remote tracks.

webrtc.html
```diff
@@ -11090,19 +11090,73 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+              <th>Values</th>
+              <th>Notes</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr id="def-constraint-width">
+              <td><dfn>width</dfn></td>
+              <td><code><a>ConstrainULong</a></code></td>
+              <td>The width or width range, in pixels. The constrainable
+              property allows downscaling. As a setting, this is the width of
```
henbos: Need to update PR to say how to downscale. Should be to "first downscale the original (as little as possible) while maintaining the aspect ratio until the target width or height is matched, and then crop to match the other dimension if necessary" - from discussion here: https://github.com/w3c/mediacapture-main/issues/584.

webrtc.html
```diff
@@ -11090,19 +11090,73 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+              <td><dfn>aspectRatio</dfn></td>
+              <td><code><a>ConstrainDouble</a></code></td>
+              <td>The exact aspect ratio (width in pixels divided by height in
+              pixels, represented as a double rounded to the tenth decimal
+              place) or aspect ratio range. The constrainable property allows
+              downscaling. As a setting, this is the aspect ratio of the latest
+              frame.</td>
+            </tr>
+            <tr id="def-constraint-resizeMode">
+              <td><dfn>resizeMode</dfn></td>
```
henbos: Need to update PR to remove resizeMode.

webrtc.html
```diff
@@ -11090,19 +11090,73 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+              <td>The width or width range, in pixels. The constrainable
+              property allows downscaling. As a setting, this is the width of
+              the latest frame.</td>
+            </tr>
+            <tr id="def-constraint-height">
+              <td><dfn>height</dfn></td>
+              <td><code><a>ConstrainULong</a></code></td>
+              <td>The height or height range, in pixels. The constrainable
+              property allows downscaling. As a setting, this is the height of
+              the latest frame.</td>
```
henbos: Clarification needed: If latest frame had height 800, but we are downscaling to 600, should getSettings() say 600 or 800?

henbos:
This PR is not ready. Self-review above.

Up for discussion: Define downscaling the way Chrome has shipped this, or say that no constraints are applicable (even if they are generic enough for any video source)?

webrtc.html
```diff
@@ -11498,19 +11498,70 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+              downscaling. As a setting, this is the aspect ratio of the latest
+              frame after any downscaling.</td>
+            </tr>
+          </tbody>
+        </table>
+        <p>When downscaling occurs due to applied constraints, the user agent
+        MUST downscale as little as possible while maintaining the original
+        aspect ratio until the target width or height is matched, and then
+        cropping the other dimension if necessary. Scaling occurs around the
+        center of the image.</p>
```
jan-ivar: I still think we need a normative section here like https://w3c.github.io/mediacapture-screen-share/#downscaling-and-frame-decimation even if it's shorter than what screencapture has.

youennf:
There are several questions that we should ask ourselves:
- Is it useful to easily query width, height and frameRate for a remote video track?
- Is it useful to easily query width, height and frameRate for any video track?
- Is it useful to downsample a remove video track?
- Is it useful to upsample/forbid upsampling a remove video track?

The current API is highly flexible and makes it hard to be fully interoperable across implementations.
This is a bit concerning to me. At least for simple cases, I would hope we can get good consistency though.

If a web application sets a width constraint of 640 for a remote track, I would hope all browsers to exhibit the same behaviour if the network decoded frames go from 1000 to 100 and then back to 1000 a minute after.

If we forbid upsampling, I guess calling getCapabilities would for instance give the current allowed downsampling range which would go from [1, 1000] to [1, 100] and then back to [1, 1000].

The same principle would seem to apply for window screen share tracks.

It seems that what we might want to enable here is to allow a web application to express something like 'provide me a video with width being at most XX'. If the call is successful, the web application should expect a consistent behaviour across browsers.

jan-ivar:
(With chair hat off) I'm not personally ready to concede (yet) that all types of tracks should have constrainable properties—`element.captureStream()`?

Constraints are an interoperability cop-out: a license for browsers to act differently, as long as they state their [capabilities](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-constrainablepattern-getcapabilities) (which are allowed by specs to vary astronomically). Even its original purview of camera device discovery seems largely moot thanks to modern down-scaling.

That said, it seems hard to argue against the four points you make being useful, except perhaps that these properties are [already observable using clever JS](https://jsfiddle.net/jib1/7df5e0ts/) through e.g. `element.videoWidth` and `videoHeight`.

It seems too late to argue for as simpler surface—`track.width` and `track.height` anyone?

So maybe we're stuck with the vestigial surface, but we use only part of it, like screen capture did. 

> If a web application sets a width constraint of 640 for a remote track, I would hope all browsers to exhibit the same behaviour if the network decoded frames go from 1000 to 100 and then back to 1000 a minute after.
> 
> If we forbid upsampling, I guess calling getCapabilities would for instance give the current allowed downsampling range which would go from [1, 1000] to [1, 100] and then back to [1, 1000].
> 
> The same principle would seem to apply for window screen share tracks.

A desirable goal, but would require stronger normative language than constraints imply. 

Screen-capture probably has the [strongest language](https://w3c.github.io/mediacapture-screen-share/#downscaling-and-frame-decimation) so far on this, and even it uses SHOULD:

 *"For the purposes of the SelectSettings algorithm, the user agent SHOULD consider all possible combinations of downscaled dimensions that preserve the aspect ratio of the original display surface (to the nearest pixel), and frame rates available through frame decimation, as available settings dictionaries."*

It also says: *"The user agent MUST NOT upscale the captured output"* which I think is good.

But then it says: *"The user agent MUST NOT crop the captured output."*, which makes perfect sense for screen sharing, but differs from camera. So sources behave differently. A remote track could be either, so I'd say MUST NOT crop.

youennf:
> It seems too late to argue for as simpler surface—`track.width` and `track.height` anyone?

If we add something like that, we might want to expose the fact that these values are changing so that the web app could react upon. This seems late for v1.0.

I am not convinced yet constraints/capabilities/settings mechanism is the best tool for the use cases we are trying to solve here. I am not clear of which real world issue we are trying to solve either.
Given that and the fact that this mechanism is difficult to get good interop, I am kind of reluctant to further increase the use of this mechanism.

> Constraints are an interoperability cop-out: a license for browsers to act differently, as long as they state their [capabilities](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-constrainablepattern-getcapabilities)

In the long run, I hope all browsers can converge though.
Also, if two browsers express supporting the same capability values, there is the expectation to get the same result when applying the same constraint values.
In this particular case, if we add precise wording in webrtc-pc for how remote tracks constraints are supposed to work, we need two webrtc-pc implementations to show interoperability.

Looking at what Chrome is exposing, resizeMode is also potentially in scope.
Remote audio tracks should also be described somehow, Chrome seems to expose some settings and capabilities.

I am fine saying in v1 something like:
- In mediacapture-main spec: each spec defining a new type of source is expected to define the exact list of supported constrainable properties. By default, there is none inherited from mediacapture-main.
- In webrtc-pc spec: no constrainable property is mandatory for any remote source. A user agent MAY expose constrainable properties defined in media capture-main to remote sources.

> _"The user agent MUST NOT crop the captured output."_, which makes perfect sense for screen sharing, but differs from camera. So sources behave differently. A remote track could be either, so I'd say MUST NOT crop.

These seem like good default behaviours. Ultimately, this seems application dependent though.
It is also weird that screen capture is stating to support crop-and-scale but actually says that cropping is forbidden. This might call for adding a new constrainable property value, letterbox.

jan-ivar:
> Looking at what Chrome is exposing, resizeMode is also potentially in scope.

This is a good example of something that would require specification. Presumable `"none"` might mean the RTPeerConnection has some indication or reason to believe the video has not downscaled for bandwidth reasons (but how can it know? Is it in the SDP?)

> By default, there is none inherited from mediacapture-main.

We already have a [note](https://w3c.github.io/mediacapture-main/getusermedia.html#media-track-supported-constraints) for that.

> In webrtc-pc spec: no constrainable property is mandatory for any remote source. A user agent MAY expose constrainable properties defined in media capture-main to remote sources.

I think I'd rather not say anything than have a MAY. It's hard to test a MAY. By the mediacapture note, I think we merely need to omit mentioning any constraints here. Then we can leave it to an extension spec to define them, maybe?

youennf:
> I think we merely need to omit mentioning any constraints here. Then we can leave it to an extension spec to define them, maybe?

An extension spec looks good to me here.
This would allow a much better description than what we might end up doing now and some time to define precisely what we are trying to enable here.

> > Looking at what Chrome is exposing, resizeMode is also potentially in scope.
> 
> This is a good example of something that would require specification. Presumable `"none"` might mean the RTPeerConnection has some indication or reason to believe the video has not downscaled for bandwidth reasons (but how can it know? Is it in the SDP?)

'none' to me means no post-processing past the video decoder and the video decoder decoding at max possible resolution. resizeMode would then apply in principle a post-processing based on these highest resolution frames.
Implementations would be free to use whatever means to achieve better performances of scaling/cropping for instance with scalable contents.
In that context, the web application might want to decide whether cropping or not, potentially based on whether content is screen-sharing vs. camera.
Of course, the cropping/scaling would best be done at sender side...

webrtc.html
```diff
@@ -11498,19 +11498,70 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+        (<code>MediaStreamTrack.getSettings()</code>) are outlined in
+        [[!GETUSERMEDIA]]. However, the constrainable properties of tracks
+        sourced from an <code>RTCPeerConnection</code> are different than those
+        sourced by <code>getUserMedia()</code>. The constraints and settings
+        applicable in WebRTC are defined here. The settings of a track sourced
+        from an <code>RTCPeerConnection</code> represent the currently received
+        frames/samples after downsampling based on applied constraints, if
+        applied. <code>MediaStreamTrack.getCapabilities()</code> MUST always
+        return the empty set.</p>
+        <p>The following constrainable properties are defined to apply only to
```
alvestrand: Grammar : "are defined to apply only" -> "are defined to apply to... ", and then add a sentence saying "The effect of all other constraints is not defined by this specification".

dontcallmedom:
> [RESOLUTION](https://www.w3.org/2019/09/18-webrtc-minutes.html#item06): define retrieval of constraint properties for getSettings but make them not modifiable

webrtc.html
```diff
@@ -11498,19 +11498,64 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+        (<code>MediaStreamTrack.getSettings()</code>) are outlined in
+        [[!GETUSERMEDIA]]. However, the constrainable properties of tracks
+        sourced from an <code>RTCPeerConnection</code> are different than those
+        sourced by <code>getUserMedia()</code>. The constraints and settings
+        applicable in WebRTC are defined here. The settings of a track sourced
+        from an <code>RTCPeerConnection</code> represent the currently received
+        frames/samples after downsampling based on applied constraints, if
+        applied. <code>MediaStreamTrack.getCapabilities()</code> MUST always
+        return the empty set. <code>MediaStreamTrack.applyConstraints()</code>
+        MUST always reject with <code>OverconstrainedError</code> fore remote
```
henbos: for remote tracks

youennf: s/fore/for

webrtc.html
```diff
@@ -11498,19 +11498,64 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+        (<code>MediaStreamTrack.getConstraints()</code>,
+        <code>MediaStreamTrack.applyConstraints()</code>), and
+        <code>MediaTrackSettings</code>
+        (<code>MediaStreamTrack.getSettings()</code>) are outlined in
+        [[!GETUSERMEDIA]]. However, the constrainable properties of tracks
+        sourced from an <code>RTCPeerConnection</code> are different than those
+        sourced by <code>getUserMedia()</code>. The constraints and settings
+        applicable in WebRTC are defined here. The settings of a track sourced
+        from an <code>RTCPeerConnection</code> represent the currently received
+        frames/samples after downsampling based on applied constraints, if
```
henbos: Remove "after downsampling"

webrtc.html
```diff
@@ -11498,19 +11498,64 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+              <th>Values</th>
+              <th>Notes</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr id="def-constraint-width">
+              <td><dfn>width</dfn></td>
+              <td><code><a>ConstrainULong</a></code></td>
+              <td>The width or width range, in pixels. As a setting, this is the
+              width of the latest frame.</td>
```
henbos: Just say width of the latest frame

webrtc.html
```diff
@@ -11498,19 +11498,64 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+        from an <code>RTCPeerConnection</code> represent the currently received
+        frames/samples after downsampling based on applied constraints, if
+        applied. <code>MediaStreamTrack.getCapabilities()</code> MUST always
+        return the empty set. <code>MediaStreamTrack.applyConstraints()</code>
+        MUST always reject with <code>OverconstrainedError</code> for remote
+        tracks.</p>
+        <p>The following constrainable properties are defined to apply to remote
+        video <code><a>MediaStreamTrack</a></code> objects generated by an
+        <code>RTCPeerConnection</code>. The effect of all other constraints is
+        not defined by this specification.</p>
```
henbos: Remove this

webrtc.html
```diff
@@ -11498,19 +11498,64 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+        <p>The concept of constraints and constrainable properties, including
+        <code>MediaTrackConstraints</code>
+        (<code>MediaStreamTrack.getConstraints()</code>,
+        <code>MediaStreamTrack.applyConstraints()</code>), and
+        <code>MediaTrackSettings</code>
+        (<code>MediaStreamTrack.getSettings()</code>) are outlined in
+        [[!GETUSERMEDIA]]. However, the constrainable properties of tracks
+        sourced from an <code>RTCPeerConnection</code> are different than those
+        sourced by <code>getUserMedia()</code>. The constraints and settings
+        applicable in WebRTC are defined here. The settings of a track sourced
```
henbos: "applicable to remote tracks"

webrtc.html
```diff
@@ -11090,19 +11090,73 @@ <h3 id="mediastreamtrack-network-use">MediaStreamTrack</h3>
...
+              property allows downscaling. As a setting, this is the height of
+              the latest frame.</td>
+            </tr>
+            <tr id="def-constraint-frameRate">
+              <td><dfn>frameRate</dfn></td>
+              <td><code><a>ConstrainDouble</a></code></td>
+              <td>The exact frame rate (frames per second) or frame rate range.
+              The constrainable property allows downsampling by dropping frames.
+              As a setting, this is an estimate of the frame rate based on
+              recently received frames.</td>
```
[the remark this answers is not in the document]
henbos: Can we do that separately? I prefer specificity but it seems like a minor concern and I don't want to block this PR on discussions about how frame rate is best approximated. I'd look in Guido/Marina for that and how Chrome does it today.

PR #2190, opened 2019-04-26
Clarify `RTCRtpContributingSource` and `RTCRtpSynchronizationSource`.

chxg:
Fixes: #2172

***
<a href="https://pr-preview.s3.amazonaws.com/chxg/webrtc-pc/pull/2190.html" title="Last updated on May 7, 2019, 10:53 AM UTC (7e7b8c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2190/f6ae67e...chxg:7e7b8c5.html" title="Last updated on May 7, 2019, 10:53 AM UTC (7e7b8c5)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7283,25 +7283,26 @@ async function updateParameters() {
       <p>The <dfn>RTCRtpContributingSource</dfn> and
       <dfn>RTCRtpSynchronizationSource</dfn> dictionaries contain information
       about a given contributing source (CSRC) or synchronization source (SSRC)
-      respectively. When an audio or video frame from an RTP packet is delivered
-      to the <code><a>RTCRtpReceiver</a></code>'s
-      <code><a>MediaStreamTrack</a></code>, the user agent SHOULD queue a task
-      to update the relevant information for the
+      respectively. When an audio or video frame from one or more RTP packets
+      is delivered to the <code><a>RTCRtpReceiver</a></code>'s
+      <code><a>MediaStreamTrack</a></code>, the user agent MUST queue a task to
+      update the relevant information for the
       <code>RTCRtpContributingSource</code> and
       <code>RTCRtpSynchronizationSource</code> dictionaries based on the
-      content of the packet. The information relevant to the
+      content of those packets. The information relevant to the
       <code>RTCRtpSynchronizationSource</code> dictionary corresponding to the
-      SSRC identifier, is updated each time, and if the RTP packet contains CSRC
+      SSRC identifier, is updated each time, and if an RTP packet contains CSRC
       identifiers, then the information relevant to the
       <code>RTCRtpContributingSource</code> dictionaries corresponding to those
-      CSRC identifiers is also updated. The user agent MUST keep information
+      CSRC identifiers is also updated. The user agent MUST process RTP packets
+      in order of ascending RTP timestamps. The user agent MUST keep information
       from RTP packets delivered to the <code><a>RTCRtpReceiver</a></code>'s
       <code><a>MediaStreamTrack</a></code> in the previous 10 seconds.</p>
       <div class="note">As stated in the <a href="#conformance">conformance
       section</a>, requirements phrased as algorithms may be implemented in
       any manner so long as the end result is equivalent. So, an
-      implementation does not need to literally queue a task for every
-      packet, as long as the end result is that within a single event loop task
+      implementation does not need to literally queue a task for every frame,
+      as long as the end result is that within a single event loop task
       execution, all returned <code><a>RTCRtpSynchronizationSource</a></code>
       and <code><a>RTCRtpContributingSource</a></code> dictionaries for a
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7283,25 +7283,24 @@ <h2>Methods</h2>
...
-      <code><a>RTCRtpSynchronizationSource</a></code> dictionaries based on the
-      contents of the packet. The information relevant to the
-      <code><a>RTCRtpSynchronizationSource</a></code> dictionary corresponding
-      to the SSRC identifier, is updated each time, and if the RTP packet
-      contains CSRC identifiers, then the information relevant to the
-      <code><a>RTCRtpContributingSource</a></code> dictionaries corresponding to
-      those CSRC identifiers is also updated.</p>
+      respectively. When an audio or video frame from an RTP packet is delivered
+      to the <code><a>RTCRtpReceiver</a></code>'s
+      <code><a>MediaStreamTrack</a></code>, the user agent SHOULD queue a task
```
alvestrand: Don't ever use SHOULD if you can avoid it. It's impossible to write a good test for a SHOULD, because the conditions under which the UA can skip it are not precisely defined. (If they had been, it would have been a "MUST if....")

webrtc.html
```diff
@@ -7318,13 +7317,14 @@ <h2>Dictionary RTCRtpContributingSource Members</h2>
...
-              indicating the most recent time of playout of media that arrived
-              in an RTP packet originating from this source. The timestamp is
-              defined as <a>performance.timeOrigin</a> +
-              <a>performance.now()</a> at the time of playout.</p>
+              <p>The <code>timestamp</code> indicating the most recent time a
+              frame from an RTP packet, originating from this source, was
+              delivered to the <code><a>RTCRtpReceiver</a></code>'s
+              <code><a>MediaStreamTrack</a></code>. The <code>timestamp</code>
+              is defined as <a>performance.timeOrigin</a> +
+              <a>performance.now()</a> at the time of delivery.</p>
```
alvestrand: Here you want to change the timestamp from "playout time" to "delivery time". What does "delivery" mean, and when does it happen?
(In particular, whether it happens before or after the jitter buffer.)

webrtc.html
```diff
@@ -7283,25 +7283,24 @@ <h2>Methods</h2>
...
-      queue a task to update the relevant information for the
-      <code><a>RTCRtpContributingSource</a></code> and
-      <code><a>RTCRtpSynchronizationSource</a></code> dictionaries based on the
-      contents of the packet. The information relevant to the
-      <code><a>RTCRtpSynchronizationSource</a></code> dictionary corresponding
-      to the SSRC identifier, is updated each time, and if the RTP packet
-      contains CSRC identifiers, then the information relevant to the
-      <code><a>RTCRtpContributingSource</a></code> dictionaries corresponding to
-      those CSRC identifiers is also updated.</p>
+      respectively. When an audio or video frame from an RTP packet is delivered
```
alvestrand: Note - this text (and the previous) assumes that a frame fits within a single RTP packet. This is true part of the time, and false part of the time (I-frames in particular). It's good if you can change the language to encompass both situations.

webrtc.html
chxg: "Delivered" is defined in the immediately preceding sentence. Would it make sense to just change the text from "at the time of delivery" to "at that time"?

The instant it's referring to is when a frame is passed from `RTCRtpReceiver` to `MediaStreamTrack`. That would be after the jitter buffer, decoder, sync buffer, and all the other WebRTC bits.

webrtc.html
chxg: Ack on that. I'll change the SHOULD here to MUST. It just felt strange to have a MUST followed by a note that immediately contradicts the usual RFC definition of MUST.

webrtc.html
chxg: It seems tricky to define it precisely. There are a couple of cases here:

1. A single frame built by multiple RTP packets with different RTP timestamps. This may happen for audio. It makes sense to process them in ascending RTP timestamp order so that the last packet "wins" for fields like `audioLevel`.

2. A single frame built by multiple RTP packets with the exact same RTP timestamps. This may, as you say, happen for video. Certain values may only be found in a RTP header extension that is only sent with either the first or last packet in a frame. So an implementation would likely want to process the RTP packets as a "group of packets" without any particular order.

3. Hybrid case where both (1) and (2) happens for a single frame that is delivered to `MediaStreamTrack`. Doesn't happen in practice today.

I've updated this PR with "one or more RTP packets" and "MUST process RTP packets in order of ascending RTP timestamps".

alvestrand: RTP timestamps are the timestamp of the frame, so if you have "a single (video) frame built by multiple RTP packets with different RTP timestamps", you have a bogus encoder.
The term "frame" is sometimes used with audio to describe a block of samples (20 ms or so), in which case you could imagine those coming from packets with different RTP timestamps.

Issue #2199, opened 2019-05-30
Clarify/strengthen timing invariants of candidates not firing before SLD(offer)


jan-ivar:
Based on a [discussion in this crbug](https://bugs.chromium.org/p/chromium/issues/detail?id=960718#c15), we've found that Firefox's behavior to avoid ice candidates racing with offer/answer is desirable, and seems to be following the broad intent of the spec and the desired API properties espoused, even if there's no explicit prose to guarantee it. Since it's not clear that the spec mandates this behavior, so we're opening an issue to enshrine it.

Firefox guarantees icecandidate always fires AFTER the SLD(offer) success callback, so even if users do:

    pc.onicecandidate = e => io.send(e.candidate);

    await pc.setLocalDescription(await pc.createOffer());
    io.send(pc.localDescription);

We guarantee the following signaling order:
```
    offer candidate candidate candidate
```
and in the other direction:
```
    answer candidate candidate candidate
```
That is, candidates always follow the offer/answer they belong to, a nice invariant. You'll never see this in Firefox:
```
    candidate offer candidate candidate
```
Which is a good thing, as it would cause addIceCandidate to throw InvalidStateError, unless you queued the candidates, which defeats the purpose of trickle ICE in the first place.

From my limited tests with https://jsfiddle.net/jib1/vfnc26mL/ it would appear Chrome has this invariant as well? At least I was not able to provoke otherwise. I get:
```
SLD(offer) success
pc1 candidate
pc1 candidate
pc1 candidate
pc1 candidate
pc1 candidate
SLD(answer) success
pc2 candidate
pc2 candidate
pc2 candidate
```

alvestrand:
If you can add a WPT test that verifies the behavior, I'm happy to make this change. Offhand, I can't think of a case where it would be beneficial to generate a candidate before SLD() resolves.

fippo:
while that is mostly guaranteed by timing (gathering candidates takes time), iceCandidatePoolSize is interesting. The intent in [JSEP is pretty clear](http://rtcweb-wg.github.io/jsep/#rfc.section.3.5.4), gather but surface only via onicecandidate which avoids duplicate candidates and the need to filter them as mentioned [here](https://github.com/feross/simple-peer/issues/469)

PR #2203, opened 2019-06-06
Ensure icecandidate events happen after SLD success.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2199.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2203.html" title="Last updated on Jul 18, 2019, 6:28 PM UTC (f32d48b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2203/8b172c5...jan-ivar:f32d48b.html" title="Last updated on Jul 18, 2019, 6:28 PM UTC (f32d48b)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8263,8 +8263,11 @@ async function onOffHold() {
       </ol>
 
       <p>When the <a>ICE Agent</a> indicates that it finished gathering a
-      <a>generation</a> of candidates for an <code><a>RTCIceTransport</a></code>, the
-      user agent MUST queue a task that runs the following steps:</p>
+      <a>generation</a> of candidates for an <code><a>RTCIceTransport</a></code>,
+      the user agent MUST wait until all candidates it has indicated as
+      available for that <code><a>RTCIceTransport</a></code> from that
+      <a>generation</a> have been surfaced to the application, and then queue a
+      task that runs the following steps:</p>
       <ol>
         <li>
           <p>Let <var>connection</var> be the
@@ -8324,7 +8327,11 @@ async function onOffHold() {
       <p>When the <a>ICE Agent</a> indicates that a new ICE candidate is
       available for an <code><a>RTCIceTransport</a></code>, either by taking one
       from the <a data-lt="ICE candidate pool size">ICE candidate pool</a> or
-      gathering it from scratch, the user agent MUST queue a task that runs the
+      gathering it from scratch, the user agent MUST wait until either the
+      [[\PendingLocalDescription]] or [[\CurrentLocalDescription]] of the
+      <code><a>RTCPeerConnection</a></code> object associated with this
+      <a>ICE Agent</a> represents the ICE <a>generation</a> for which the new
+      ICE candidate was gathered, and then queue a task that runs the
       following steps:</p>
       <ol>
         <li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1426,7 +1426,6 @@
                     <code>true</code>, then abort these steps.</p>
                   </li>
                   <li>
-                    
                     <p>If <var>description</var> is set as a local description,
                     then run one of the following steps:</p>
                     <ol>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1435,7 +1435,10 @@
                         <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
                         to a new <code><a>RTCSessionDescription</a></code> object
                         constructed from <var>description</var>, and set
-                        <a>signaling state</a> to <code>"have-local-offer"</code>.</p>
+                        <a>signaling state</a> to <code>"have-local-offer"</code>.
+                        If <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
+                        was previously <code>null</code>, <a>release early
+                        candidates</a>.</p>
                       </li>
                       <!-- C) transition haveRemoteOffer or haveLocalProvAnswer to
                       stable -->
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -8263,8 +8263,11 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
       </ol>
 
       <p>When the <a>ICE Agent</a> indicates that it finished gathering a
-      <a>generation</a> of candidates for an <code><a>RTCIceTransport</a></code>, the
-      user agent MUST queue a task that runs the following steps:</p>
+      <a>generation</a> of candidates for an <code><a>RTCIceTransport</a></code>,
+      the user agent MUST wait until all candidates it has indicated as
```
alvestrand: Rather than "must wait", why not say "when the ICE agent is finished gathering, and all candidates for that ICETransport have been surfaced to the application, the UA MUST queue a task...."

webrtc.html
```diff
@@ -8324,7 +8327,11 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
       <p>When the <a>ICE Agent</a> indicates that a new ICE candidate is
       available for an <code><a>RTCIceTransport</a></code>, either by taking one
       from the <a data-lt="ICE candidate pool size">ICE candidate pool</a> or
-      gathering it from scratch, the user agent MUST queue a task that runs the
+      gathering it from scratch, the user agent MUST wait until either the
```
alvestrand: the user agent MUST put the candidate into a processing queue [[queue]] for that transport, and do the check below.

When the currentLocalDescription or PendingLocalDescription indicate that the ICE generation is teh one for which candidates in the processing queue are intended, it MUST queue a task that...

alvestrand: Also add to SLD: "If a new SLD call is initiated when there are candidates in [[queue]], clear the queue."

webrtc.html
```diff
@@ -8335,6 +8362,28 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
           <code><a>RTCPeerConnection</a></code> object associated with this
           <a>ICE Agent</a>.</p>
         </li>
+        <li>
+          <p>If <var>connection</var>'s <a>[[\IsClosed]]</a> slot is
+          <code>true</code>, abort these steps.</p>
+        </li>
+        <li>
+          <p>If either <var>connection</var>'s <a>[[\PendingLocalDescription]]</a>
+          or <a>[\CurrentLocalDescription]]</a> internal slots are not
```
henbos: This would apply to either SLD(answer) or SRD(answer), I think you only want SLD(answer)?

webrtc.html
```diff
@@ -2035,6 +2039,28 @@ <h4>Set the RTCSessionDescription</h4>
                       </li>
                     </ol>
                   </li>
+                  <li>
+                    <p>If <var>connection</var>'s
+                    <a>[[\PendingLocalDescription]]</a>
+                    or <a>[[\CurrentLocalDescription]]</a> changed above, then
```
jan-ivar: Oof, this can also happen in SRD() as written. I need to fix that.

jan-ivar:
@henbos I now check in SLD() only. I broke out a "release early candidates" algorithm, which I think reads better. PTAL. My push blew away formatting again, which is https://github.com/w3c/webrtc-pc/issues/2226.

webrtc.html
```diff
@@ -1448,8 +1455,11 @@ <h4>Set the RTCSessionDescription</h4>
                         to <code>null</code>. Set both
                         <var>connection</var>.<a>[[\LastCreatedOffer]]</a> and
                         <var>connection</var>.<a>[[\LastCreatedAnswer]]</a> to
-                        <code>""</code>. Finally set <var>connection</var>'s
-                        <a>signaling state</a> to <code>"stable"</code>.</p>
+                        <code>""</code>, and finally set <var>connection</var>'s
+                        <a>signaling state</a> to <code>"stable"</code>.
+                        If <var>connection</var>.<a>[[\CurrentLocalDescription]]</a>
+                        was previously <code>null</code>, <a>release early
+                          candidates</a>.</p>
```
henbos: We could have early candidates after an ICE restart too, right? In that case [[CurrentLocalDescription]] wouldn't be null even if the candidates are early?
Can we always "release early candidates"? If there aren't any early candidates this would be a NO-OP anyway.

webrtc.html
```diff
@@ -8335,6 +8349,45 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
...
+      <ol>
+        <li>
+          <p>For each candidate, <var>candidate</var>, in
+          <var>connection</var>.<a>[[\EarlyCandidates]]</a>, if
+          either <var>connection</var>.<a>[[\PendingLocalDescription]]</a> or
+          <var>connection</var>.<a>[[\CurrentLocalDescription]]</a> are not
+          <code>null</code>, and represent the ICE <a>generation</a>
+          of <var>candidate</var>, then queue a task to <a
+            href="#dfn-surface-a-candidate">surface the candidate</a>
+          with <var>candidate</var> and <var>connection</var>.</p>
```
henbos: An early candidate is one we haven't fired an event for yet, so why are we filtering out candidates here? Is this in case of an ICE restart where they have become invalid already?

Otherwise - speaking of ICE restart: should a successful ICE restart cause [[EarlyCandidates]] to be cleared?

If this is something we always do in stable state - should we perhaps only look at [[CurrentLocalDescription]], or am I missing something?

webrtc.html
jan-ivar: @henbos good point! That should work. I think I was worried about rollback, but after the move that's not a problem anymore. I might also have been worried about SLD() failure, but at this point in the algorithm the description has already been successfully applied.

webrtc.html
jan-ivar: > why are we filtering out candidates here?

Some prose contortions led me to think this was needed. I think you're right it's not.

> speaking of ICE restart: should a successful ICE restart cause [[EarlyCandidates]] to be cleared?

Wouldn't [[EarlyCandidates]] already be emtpy? How can a restart be "successful" without SLD() having succeeded?

> If this is something we always do in stable state - should we perhaps only look at [[CurrentLocalDescription]], or am I missing something?

You're right. I'll remove the filtering outright.

jan-ivar:
Thanks @henbos for the good feedback! I think I got everything.

Issue #2208, opened 2019-06-14
SLD without {sdp} reads [[LastCreatedOffer/Answer]] at the wrong time


jan-ivar:
The current prose on [setLocalDescription](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-setlocaldescription) reads **[[LastCreatedOffer]]** and **[[LastCreatedAnswer]]** on method entry. This is the wrong time, and may produce the wrong result if other methods are enqueued. For example, it would fail in the [FIFO peer](https://docs.google.com/presentation/d/1UrXARLbAfwmfK686rX_9W_-FGIc62ZDwVd079wIoX7I/edit?ts=5cdd966f#slide=id.g5b048ace1b_21_24) case:
```js
io.onmessage = async ({data: {description, candidate}}) => {
  if (description) {
    await Promise.all([                           // ←- avoids race!
      pc.setRemoteDescription(description),       // ←- 
      pc.createAnswer(),                          // ←- 
      pc.setLocalDescription({type: "answer"})    // ←- unusual, works today
    ]);                                           // ←- 
```
Here, all three promise-returning negotiation methods are invoked at the same time, which means `setLocalDescription` would read `[[LastCreatedAnswer]]` before the *createAnswer* has succeeded (or even begun the in-parallel part of its algorithm). This would cause SLD to either fail (if `[[LastCreatedAnswer]]` was `null`, or apply an outdated answer.

Instead, we need to read these internal slots in the enqueued steps.

PR #2209, opened 2019-06-14
Read [[LastCreatedOffer/Answer]] at the right time in SLD

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2208.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2209.html" title="Last updated on Jul 11, 2019, 1:16 PM UTC (e326fc9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2209/3e20f80...jan-ivar:e326fc9.html" title="Last updated on Jul 11, 2019, 1:16 PM UTC (e326fc9)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1325,9 +1325,7 @@
         <p>To <dfn id="set-description" data-lt=
         "set the RTCSessionDescription">set an RTCSessionDescription</dfn>
         <var>description</var> on an <code><a>RTCPeerConnection</a></code>
-        object <var>connection</var>, <a href=
-        "#enqueue-an-operation">enqueue</a> the following steps to
-        <var>connection</var>'s operation queue:</p>
+        object <var>connection</var>, run the following steps:</p>
         <ol>
           <li data-tests="RTCPeerConnection-setLocalDescription-offer.html">
             <p>If <var>description</var> is set as a local description,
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3023,27 +3021,48 @@ <h2>Methods</h2>
...
+                        then set <var>description.sdp</var> to the value of
+                        <var>connection</var>'s <a>[[\LastCreatedOffer]]</a>
+                        slot.</p>
+                      </li>
+                      <li>
+                        <p>Return the result of <a
+                        href="#set-description">setting the RTCSessionDescription</a>
+                        indicated by <code><var>description</var></code>.</p>
+                      </li>
+                    <ol>
```
dontcallmedom: this should be `</ol>` (hence the Travis CI error on this and #2212)

webrtc.html
```diff
@@ -3023,27 +3021,48 @@ <h2>Methods</h2>
...
+                  <li>
+                    <p>Let <var>connection</var> be the
+                    <code><a>RTCPeerConnection</a></code> object on which the
+                    method was invoked.</p>
+                  </li>
+                  <li>
+                    <p>Return the result of <a href=
+                    "#enqueue-an-operation">enqueuing</a> the following steps to
+                    <var>connection</var>'s operation queue:</p>
+                    <ol>
```
henbos: We are enqueuing a task that sets the session description (= creates a promise and then enqueues a task to resolve it). It doesn't make sense to me to enqueue a task to enqueue a task.
1. It adds an additional delay, that sounds like a bug to me. While it probably rarely creates problems, it doesn't seem to add any value to do so.
2. Language-wise, it doesn't make sense to return the result of enqueuing an operation. The promise we want to return hasn't been created at this point. We would have to create another promise and wrap the other promise in it. But see 1), we don't want to do that.

webrtc.html
```diff
@@ -3059,10 +3078,29 @@ <h2>Methods</h2>
...
+                  </li>
+                  <li>
+                    <p>Let <var>connection</var> be the
+                    <code><a>RTCPeerConnection</a></code> object on which the
+                    method was invoked.</p>
+                  </li>
+                  <li>
+                    <p>Return the result of <a href=
+                    "#enqueue-an-operation">enqueuing</a> the following steps
+                    to <var>connection</var>'s operation queue:</p>
```
henbos: Same as above: Why are we enqueuing a task to enqueue a task? We should just say to return the result of setting the RTCSessionDescription immediately.

webrtc.html
```diff
@@ -3087,6 +3125,7 @@ <h2>Methods</h2>
                 <code>setRemoteDescription</code> does not await the completion
                 of identity validation.</p>
               </dd>
+
```
henbos: Accidental newline?

henbos:
> I think we moved the problem instead of fixing it

Correction: The race of the referenced bug is fixed here, but still enqueuing at a later stage is probably unnecessary and possibly problematic (correct me if I'm wrong)

webrtc.html
jan-ivar: Resolved offline. [Enqueueing an operation](http://w3c.github.io/webrtc-pc/#enqueue-an-operation) is not the same as [queue a task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task). The former chains an asynchronous operation on the `RTCPeerConnection`'s internal serialization queue. The latter takes us back to main thread, adding an event to the JS event loop.

jan-ivar: The former also returns a promise, so we should be good here.

PR #2211, opened 2019-06-14
Define an "unknown" value for RTCIceRole

### LINKED ISSUE #2210: RTCIceTransport role before start() called
This was first noticed in the webrtc-ice repository: https://github.com/w3c/webrtc-ice/issues/4

Before a role is assigned, it seems misleading to give "role" either of the values "controlling" or "controlled".

I suggest we create a new enum value called "unknown", which is the initial value.
(This name also corresponds to the implementation inside the WebRTC library. No need to make new names when we don't have to.)

alvestrand:
Fixes #2210

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2211.html" title="Last updated on Jun 14, 2019, 11:58 AM UTC (4b0fb6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2211/08c35ea...4b0fb6a.html" title="Last updated on Jun 14, 2019, 11:58 AM UTC (4b0fb6a)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8812,6 +8812,7 @@ async function onOffHold() {
       <h3><dfn>RTCIceRole</dfn> Enum</h3>
       <div>
         <pre class="idl">enum RTCIceRole {
+    "unknown",
     "controlling",
     "controlled"
 };</pre>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -8821,6 +8822,12 @@ <h3><dfn>RTCIceRole</dfn> Enum</h3>
             <tr>
               <th colspan="2"><code><a>RTCIceRole</a></code> Enumeration description</th>
             </tr>
+            <tr>
+              <td><dfn data-idl><code>unknown</code></dfn></td>
+              <td>An agent whose role as defined by [[!ICE]], Section 3,
+                has not yet been determined.
```
henbos: (Is this the right way to format the td?)

henbos: I wonder if "The role attribute MUST return the ICE role of the transport." is specified enough; if the role ever changes there needs to be an update step. E.g. as part of setting a description there could be a step to update the ice role internal slot. Care to file an issue for that? Non-blocker for this PR

Issue #2166, opened 2019-04-07
A simpler non-racy rollback


jan-ivar:
In [my latest blog](https://blog.mozilla.org/webrtc/perfect-negotiation-in-webrtc/) on *"Perfect negotiation in WebRTC"*, I cover using `negotiationneeded` and `"rollback"` in combination to abstract away negotiation. I describe implementing something I call "the polite peer" (a peer that rolls back in the face of an incoming offer):
```js
io.onmessage = async ({data: {description, candidate}}) => {
  if (description) {
    if (description.type == "offer" && pc.signalingState == "have-local-offer") {
      if (!polite) return;
      await Promise.all([
        pc.setLocalDescription({type: "rollback"}),
        pc.setRemoteDescription(description)
      ]);
    } else {
      await pc.setRemoteDescription(description);
    }
    if (description.type == "offer") {
      await pc.setLocalDescription(await pc.createAnswer());
      send({description: pc.localDescription});
    }
  } else if (candidate) await pc.addIceCandidate(candidate);
}
```
In the blog I cover in depth why the `Promise.all` is needed, but in short, it's there to cover a timing-window where we'd otherwise be vulnerable to *addIceCandidate* getting queued ahead of *setRemoteDescription* and failing.

To make this less of a footgun I think we should come up with a safer rollback API that does the rollback and sets the new description on the same run of the JS event loop. E.g.
```js
pc.allowRollback();
await pc.setRemoteDescription(description);
```
or perhaps
```js
await pc.setRemoteDescription(description, {rollback: true});
```
Open to other ideas.

solmaks:
@jan-ivar , great post , and finally a summary of annoying real world problems with current WebRTC media signalling APIs! A couple of extra thoughts about the proposed API:

- If intent is to simplify life for application developers, then I would suggest going step further and removing the need to provide the {rollback:true} altogether. SRD and current PC state should have all the information to decide whether it should trigger a rollback, or apply a remote answer (current signalling state + description type).

-  It would be great if support for rolling back through SRD would remove the need to have the SLD rollback variant altogether. However, it does not seem to be possible to cover all scenarios with it. Namely, there is still a need to handle timeouts (offer/answer never reached remote endpoint) and errors (remote endpoint rejected offer for some reason), and SLD seems to be a better fit for this.

jan-ivar:
[...]

PR #2212, opened 2019-06-14
Add implicit rollback to setRemoteDescription(description)

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2166. For consideration.

Please ignore first commit, as this patch sits on top of https://github.com/w3c/webrtc-pc/pull/2209.

If you got to "Files changed" you'll see both. [This](https://github.com/w3c/webrtc-pc/pull/2212/commits/32c45e9b931e417e4b0e46fb35f7f5100a658dfb) is the patch for review (commit 2).

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2212.html" title="Last updated on Jul 11, 2019, 2:37 PM UTC (c639289)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2212/513513b...jan-ivar:c639289.html" title="Last updated on Jul 11, 2019, 2:37 PM UTC (c639289)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3113,6 +3113,32 @@ interface RTCPeerConnection : EventTarget  {
                     to <var>connection</var>'s operation queue:</p>
                     <ol>
                       <li>
+                        <p>If the <var>description</var>'s <code><a data-link-for=
+                        "RTCSessionDescription">type</a></code> is
+                        <code>"offer"</code> and is invalid for the
+                        current <a>signaling state</a> of <var>connection</var>
+                        as described in <span data-jsep=
+                        "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>,
+                        then run the following sub-steps:</p>
+                        <ol>
+                          <li>
+                            <p>Let <var>p</var> be the result of <a
+                            href="#set-description">setting the RTCSessionDescription</a>
+                            to a new description with a <code><a
+                            data-link-for="RTCSessionDescription">type</a></code> of
+                            <code>"rollback"</code>.</p>
+                          </li>
+                          <li>
+                            <p>Return the result of <a
+                            href="https://w3ctag.github.io/promises-guide/#transforming-by">
+                            transforming</a> <var>p</var> with a fulfillment handler
+                            that returns the result of
+                            <a href="#set-description">setting the RTCSessionDescription</a>
+                            to <var>description</var>, and abort these steps.</p>
+                          </li>
+                        </ol>
+                      </li>
+                      <li>
                         <p>Return the result of <a
                           href="#set-description">setting the RTCSessionDescription</a>
                         to <var>description</var>.</p>


henbos:
Let's rebase, I can review offline

aboba:
Needs to be rebased now that 2209 is merged.

webrtc.html
```diff
@@ -3112,6 +3112,32 @@ <h2>Methods</h2>
...
+                            data-link-for="RTCSessionDescription">type</a></code> of
+                            <code>"rollback"</code>.</p>
+                          </li>
+                          <li>
+                            <p>Return the result of <a
+                            href="https://w3ctag.github.io/promises-guide/#transforming-by">
+                            transforming</a> <var>p</var> with a fulfillment handler
+                            that returns the result of
+                            <a href="#set-description">setting the RTCSessionDescription</a>
+                            to <var>description</var>, and abort these steps.</p>
```
henbos: We need to create both promises first to ensure that both things are placed on the operation queue first and then chain the promises

jan-ivar: @henbos Wait, I remember why I did it this way now!

We cannot know to enqueue two operations immediately, because we cannot check the "current signaling state" immediately (for the reasons explained in https://github.com/w3c/webrtc-pc/issues/2208).

But that's fine. We don't need to enqueue two operations, because we're already in an operation—our new opaque "setRemoteDescription-with-rollback" operation—so we can take as long as we want and do whatever we want before we resolve our promise which returns our result and causes the next operation on the peer connection's [operations chain](https://w3c.github.io/webrtc-pc/#dfn-operations-queue) to be considered.

Said differently: inside an enqueued (asynchronous) operation, we're serialized with all the other asynchronous negotiation methods that use the same operations chain, so we're safe against races with them.

jan-ivar: So I think the current PR is correct. Please let me know if you agree, or have questions.

henbos: We're on the operation queue, and _setting an RTCSessionDescription_ creates the promise and does things _in parallel_ that resolves it, so before we return here we have both SLD(rollback) and SRD(desc) to run in parallel. Sounds right, but, "in parallel" isn't a queue...?

Inside of an enqueued task, we are creating two promises, yes, but we're also spawning two "in parallel, start the process...". We've spawned two separate "in parallel" processes. Unless this means that they are queued ([doesn't sound like it](https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism)), doing "in parallel, SLD(rollback), and in parallel, SRD(desc)" could cause races.

If the only thing that ensures that in parallel processes don't collide is the peer connection operation queue, then what we want to achieve is:

> When setRemoteDescription is invoked, run the following steps:
> - Let p1 be the result of enqueuing on connection's operation queue to run the following steps: SLD(rollback);
> - Let p2 be the result of enqueuing on connection's operation queue to run the following steps: SRD(desc);
> - Return the result of transforming p1 with a fulfillment handler p2.

The problem is that we want to be inside of the operation queue to inspect the signaling state!

[WhatWG on parallelism](https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism) says "To avoid race conditions between different in parallel algorithms that operate on the same data, a parallel queue can be used."

So we could either use a **parallel queue** (which if used sounds like we might not need a peer connection operation queue?) or we could make sure that the "in parallel" steps of both SLD(rollback) and SRD(desc) are executed inside of the same "in parallel" steps.

henbos: How about we create a **parallel queue** that is only used for the implicit rollback, and pass in the queue as an argument to _setting an RTCSessionDescription_, saying that this algorithm will queue onto the parallel queue argument if there was one, and if no argument was provided it will do the steps "in parallel" instead.

That way we can write the steps the way you wrote them here and still ensure that the parallel processes aren't racing, and we don't have to rewrite the connection's operation queue logic.

jan-ivar: > Inside of an enqueued task, 

@henbos We're inside an enqueued *operation*, not a task. See https://github.com/w3c/webrtc-pc/issues/2227.

> We've spawned two separate "in parallel" processes.

These "in parallel" steps all begin and end within the same opaque asynchronous [operation](http://w3c.github.io/webrtc-pc/#enqueue-an-operation), as defined by webrtc-pc. The serializing of said operations happens solely on main thread. We can jump back and forth between threads as much as we want within a single such "operation".

I'm fairly sure this PR works as is. I'll add a PR to https://github.com/w3c/webrtc-pc/issues/2227 to fix our terminology, which I think will help. Also happy to jump on a call to discuss.

jan-ivar:
Tests in https://github.com/web-platform-tests/wpt/pull/18010.

Issue #2176, opened 2019-04-17
Spec steps on stop() underestimate BUNDLE problem.


jan-ivar:
Our [BUNDLE warning](http://w3c.github.io/webrtc-pc/#methods-8) says:
> Note
> 
> If this method is called in between applying a remote offer and creating an answer, and the transceiver is associated with the "offerer tagged" media description as defined in [BUNDLE], this will cause all other transceivers in the bundle group to be stopped as well.

This suggests the problem is limited to calling `stop()` in `"have-remote-offer"`, but the same problem exists if `stop()` has been called *before* SRD(offer).

> To avoid this, one could instead stop the transceiver when signalingState is "stable" and perform a subsequent offer/answer exchange.

This is true only if we replace "a subsequent" with "an immediate", and assume an SRD(offer) cannot come in beforehand. It also assumes our offer isn't rolled back.

henbos:
Stop()... our old nemesis is back. Jan-Ivar I think we need to burn the body once we are done with it for it not to come back alive.

jan-ivar:
Actually, it's more than just the note. Step 6 underestimates the problem as well:

> 6. If connection's signaling state is have-remote-offer and

In the case where `stop()` is called *before* SRD(offer) then step 6 will not stop all bundled transceivcers, leaving them unstopped, even though their underlying transport will be stopped/nonworking.

dontcallmedom:
@jan-ivar can this be closed now that that  #2220 has been merged?

aboba:
PR has been merged, so am closing this issue.

PR #2220, opened 2019-06-28
Make stop() safe and affect createOffer only. Replace attributes with "stopped" direction.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2150 and https://github.com/w3c/webrtc-pc/issues/2176. For consideration.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2220.html" title="Last updated on Jul 18, 2019, 2:00 PM UTC (538d576)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2220/dd4a8ec...jan-ivar:538d576.html" title="Last updated on Jul 18, 2019, 2:00 PM UTC (538d576)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1964,8 +1964,10 @@
                           </li>
                           <li>
                             <p>If the <a>media description</a> is rejected, and
-                            <var>transceiver</var> is not already stopped, <a>stop
-                            the RTCRtpTransceiver</a> <var>transceiver</var>.</p>
+                            <var>transceiver</var>'s <a>[[\Stopped]]</a> slot is
+                            <code>false</code>, then
+                            <a>stop the RTCRtpTransceiver</a>
+                            <var>transceiver</var>.</p>
                           </li>
                         </ol>
                       </li>
@@ -2540,8 +2542,13 @@ interface RTCPeerConnection : EventTarget  {
                 remain usable by <code>setLocalDescription</code> without
                 causing an error until at least the end of the <a>fulfillment</a>
                 callback of the returned promise.</p>
-                <p data-tests="sdp-offer.html">Creating the SDP MUST follow the appropriate process for
-                generating an offer described in [[!JSEP]].
+                <p data-tests="RTCRtpTransceiver-stop.html">Creating the SDP
+                MUST follow the appropriate process for generating an offer
+                described in [[!JSEP]], except the user agent MUST treat a
+                <a href="#dfn-stopping-0">stopping</a> transceiver as
+                <a href="#dfn-stopped-0">stopped</a> for the purposes of JSEP in
+                this case.</p>
+                <p data-tests="sdp-offer.html">
                 As an offer, the generated SDP will contain the full set of
                 codec/RTP/RTCP capabilities supported or preferred by the session (as
                 opposed to an answer, which will include only a specific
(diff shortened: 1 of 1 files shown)


jan-ivar:
@henbos This is based on what we discussed. PTAL. 

In short, the PR creates a new `tc.stopping` sibling to `tc.stopped`. The former only affects `createOffer`, not `createAnswer`, sidestepping https://github.com/w3c/webrtc-pc/issues/2150. The `stop()` method now sets `stopping` instead of `stopped`, but otherwise works as before locally, and `setRemoteDescription(answer)` transitions it to `stopped`, basically.

Renamed old dangerous stop to `reject()`; made it work only in `"have-remote-offer"`, solving https://github.com/w3c/webrtc-pc/issues/2176.

I had to audit existing places the spec reads `[[Stopped]]`, to see which should be `[[Stopping]]` instead. Very few it turned out, mostly high-level modification APIs like `addTrack`.

Stop reading now if you don't care about edge cases...

The good news is that the [setting an RTCSessionDescription](http://w3c.github.io/webrtc-pc/#set-description) algorithm did not look at `[[Stopped]]` much. I found two places, but they seem like optimizations, and one may even be a bug. I'll file a separate issue.

This means most of our negotiation event firing code already runs on stopped transceivers today, something I verified in Firefox with [this fiddle](https://jsfiddle.net/jib1/gzn9f20k/). That may sound bad, but is actually quite orthogonal and seems benign to me, since it's basically shuffling around `ended` tracks at that point. This was informative to figure out how a peer should react to incoming offers/answers in the new `stopping` state:

For example: an edge-case is where a fixed-role answerer may end up staying in `stopping` state for a while, because things like `tc.stop()` and `pc.addTrack` won't automatically be negotiated (requires out-of-band-signaling to the fixed-role offerer). So it sits until negotiation comes from the other end, because it's never the offerer. In this case, events may still fire from incoming non-stopped offers, except on the `transceiver.receiver.track` itself which has been `ended` synchronously already by `stop()`, like today.

henbos:
> Stop reading now if you don't care about edge cases...

Of course I care ;) I'll review carefully tomorrow

jan-ivar:
[Slides](https://docs.google.com/presentation/d/1xcvf0udNeSH7s1FOY7RRqr1dEFvokZjn-MZPjwy3iXQ/edit#slide=id.g5c2f3df65b_11_442)

aboba:
Need to remove the reject() portion.

jan-ivar:
Removes `reject()`, and incorporates @henbos's changes proposed at the tail of our last VI, which simplifies by removing the **stopped** and **stopping** attributes in favor of a single `"stopped"` direction.

So instead of reading a **stopping** attribute, use:
```js
if (tc.direction == "stopped") {
```
Instead of reading the old **stopped** attribute (which now only happens from SRD) use:
```js
if (tc.currentDirection == "stopped") {
```

Keeps the old `stop()` method, unless there's appetite for `direction = "stopped"`? (throws now)
```js
tc.stop();
console.log(tc.direction);        // stopped
console.log(tc.currentDirection); // sendrecv
doSignalingHandshake();
console.log(tc.direction);        // stopped
console.log(tc.currentDirection); // stopped
```
Note how the final value of `currentDirection` is now `"stopped"` rather than `null`. 🤘

jan-ivar:
Travis seems unhappy. Some kind of [respec](https://travis-ci.com/w3c/webrtc-pc/builds/119223169#L817) issue? I can't make out the problem here.

caribouW3:
> Travis seems unhappy. Some kind of [respec](https://travis-ci.com/w3c/webrtc-pc/builds/119223169#L817) issue? I can't make out the problem here.

Apparently a transient problem, it works OK now.

webrtc.html
```diff
@@ -7739,9 +7776,23 @@ <h2>Methods</h2>
                   <a data-cite="!GETUSERMEDIA#track-ended">ended</a>.</p>
                 </li>
                 <li>
-                  <p>Set <var>transceiver</var>'s <a>[[\Stopped]]</a>
+                  <p>Set <var>transceiver</var>'s <a>[[\Stopping]]</a>
                   slot to <code>true</code>.</p>
```
henbos: Should the "stop sending and receiving" algorithm also set the [[Direction]] to inactive?
- Generally speaking, if we are calling stop() then we have to initiate a negotiation, so it doesn't hurt to update [[Direction]].
- More specifically, if we are calling stop() in have-remote-offer, it makes more sense to answer with a=inactive than a=whatever considering we are not sending or receiving with the accepted m= section.

jan-ivar: I think I agree here. Will update PR.

PR #2222, opened 2019-07-08
Add initialization of DataChannelId on connection

### LINKED ISSUE #1818: RTCDataChannel.id assignment
I'm currently writing a bunch of WPT tests and found that channel ID assignment in particular trips all implementations in several places. I found out that channel id assignment is extremely hard to get right since it can fail in at least three different places even though the spec only speaks of one (the second case):

1. Before the DTLS role is known when creating a data channel (this is not defined in the spec since no ID is being assigned at that point but people aren't silly and do test for the maximum number of streams they support).
2. As soon as the DTLS role is known and already created data channels will be assigned an ID or closed followed by an error event (in case no ID can be assigned).
3. As soon as the SCTP association is established and the number of streams has been negotiated. The negotiated number may be lower than what has been assumed earlier.

I'm not sure if we need to update the spec but we probably should because at least the third case is not covered.
The second case is covered by the spec but DTLS role changes in a subsequent remote description after the channels have been assigned an ID is not covered... I don't even know what to suggest for that... just panic? :wink: 

(The Cpt. Hindsight in me says: A promise for `createDataChannel` would have been much less prone to error because we could have delayed all of this until the SCTP association had been established.)

Edit: My suggestion would be to specify that IDs are to be assigned after the SCTP association has been established and before the channel is opened. Before that, the ID should be `null` unless it's a negotiated channel.

alvestrand:
Fixes #1818

Note that this initializes DataChannelId on connecting
the SctpTransport, not when the DTLS role is determined.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2222.html" title="Last updated on Oct 3, 2019, 10:24 AM UTC (e8d2a4f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2222/14326c3...e8d2a4f.html" title="Last updated on Oct 3, 2019, 10:24 AM UTC (e8d2a4f)">Diff</a>


webrtc.html
```diff
@@ -9287,7 +9287,15 @@ <h4 id="sctp-transport-connected">Connected procedure</h4>
                   <p>Let <var>channel</var> be the <code><a>RTCDataChannel</a></code> object.</p>
                 </li>
                 <li>
-                  <p>If the <var>channel</var>'s <a>[[\DataChannelId]]</a> slot is greater or
+                  <p>If <var>channel</var>'s <a>[[\DataChannelId]]</a> slot is
+                    <code>null</code>, initialize <a>[[\DataChannelId]]</a>
+                    to a value generated by the
+                    user agent, according to [[!RTCWEB-DATA-PROTOCOL]].
+                    If no available ID could be generated,
+                    <a>throw</a> an <code>OperationError</code> exception.
```
henbos: You can't throw an exception here, we're in a queued task, we're not in a function

henbos: We should fail like we do below?

webrtc.html
```diff
@@ -9287,7 +9287,15 @@ <h4 id="sctp-transport-connected">Connected procedure</h4>
                   <p>Let <var>channel</var> be the <code><a>RTCDataChannel</a></code> object.</p>
                 </li>
                 <li>
-                  <p>If the <var>channel</var>'s <a>[[\DataChannelId]]</a> slot is greater or
+                  <p>If <var>channel</var>'s <a>[[\DataChannelId]]</a> slot is
+                    <code>null</code>, initialize <a>[[\DataChannelId]]</a>
+                    to a value generated by the
+                    user agent, according to [[!RTCWEB-DATA-PROTOCOL]].
```
henbos: It's important that the underlying data transport's ID matches this label. I think we either need to say that this generation took place as part of the underlying connection procedure and is surfaced here, or we have to generate the ID here and then "in parallel" update the underlying data transport to match this ID.

alvestrand: I took "generated by the user agent" to be so obvioiusly the underlying data transport's ID that it didn't need to be called out.

henbos: Clarification:

Each RTCDataChannel has an associated underlying data transport.
Whenever something happens to the underlying transport, like connecting or closing which could happen in parallel on a network thread, we queue a task to update the RTCDataChannel to match. We need to be careful about keeping these two objects (the JS version and the actual SCTP implementation) in sync.

For example, when creating an RTCDataChannel with createDataChannel(), there's a step to "Create channel's associated underlying data transport and configure it according to the relevant properties of channel." meaning we pass down the ID to the underlying transport. Likewise when a data channel is created by the underlying sctp transport, we queue a task to to create an RTCDataChannel to match the configuration received from the underlying data transport.

So if we modify the [DataChannelId] internal slot, we need to inform the underlying data transport that this is the ID that it should use.

Do we need to add "In parallel, update the underlying data transport's id"?

Of can we say "If [DataChannelId] is null, update it to match the ID received from the underlying sctp data channel that was generated as specified in [RTCWEB-DATA-PROTOCOL]"?

webrtc.html
```diff
@@ -9287,7 +9287,15 @@ <h4 id="sctp-transport-connected">Connected procedure</h4>
...
+                    If no available ID could be generated,
+                    <a>throw</a> an <code>OperationError</code> exception.
+                  </p>
+                <li>
+                  <p>If <var>channel</var>'s <a>[[\DataChannelId]]</a> slot is greater or
                   equal to <var>transport</var>'s <a>[[\MaxChannels]]</a> slot,
                   <a data-lt="data-channel-failure"> close the <var>channel</var> due to a failure
                   </a>. Otherwise, <a data-lt="announce the rtcdatachannel as open">announce the
```
henbos: Not new in this PR: We are inside of steps invoking "announce the rtcdatachannel as open" and "close the channel due to failure". The referenced algorithm queues a task. We should refactor this so that we can invoke the algorithms without queuing a task; the queuing of a task is in response to something happening in parallel.

webrtc.html
henbos: > I took "generated by the user agent" to be so obvioiusly the underlying data transport's ID that it didn't need to be called out.

The concern isn't so much that they mismatch since that clearly is the intent but that it is unclear if the ID is generated before or after the queuing of a task.

alvestrand: The id is not a *modifiable* property. It's a *settable* property. Once it's set to a non-null value, it can't be changed. So we don't have to deal with synchronization of changes - either it's set at creation (in JS, communicated to lower layers) or it's set in the protocol machine (in lower layers, communicated to JS). This makes thinking about synchronization somewhat easier.

alvestrand:
I think it has a test. Will look to check that it actually tests.

webrtc.html
```diff
@@ -9147,9 +9147,9 @@ <h2>Methods</h2>
                   <div class="note">
                     If the <a>[[\DataChannelId]]</a>
                     slot is <code>null</code> after this step, it will be
-                    populated once the DTLS role is determined during the
-                    process of <a data-lt="set the RTCSessionDescription">
-                    setting an <code>RTCSessionDescription</code></a>.
+                    populated during the process of
```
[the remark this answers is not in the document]
alvestrand: The section is called "connected procedure". I guess I should call it that.

webrtc.html
```diff
@@ -9287,7 +9287,15 @@ <h4 id="sctp-transport-connected">Connected procedure</h4>
                   <p>Let <var>channel</var> be the <code><a>RTCDataChannel</a></code> object.</p>
                 </li>
                 <li>
-                  <p>If the <var>channel</var>'s <a>[[\DataChannelId]]</a> slot is greater or
+                  <p>If <var>channel</var>'s <a>[[\DataChannelId]]</a> slot is
```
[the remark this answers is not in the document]
alvestrand: Not sure. I moved the statechange event to after the ID assignment, but that causes "announce the channel as open" to move before the statechange event. Does that matter?
I don't think so, because the "Announcing a data channel as open" section starts off with "MUST queue a task".

webrtc.html
alvestrand: I changed this to "initialize [[DataChannelId]] to the value generated by the underlying sctp data channel, according to..."

webrtc.html
alvestrand: Not addressed in this PR.

webrtc.html
alvestrand: Done.

alvestrand:
Please re-review. I changed text.

webrtc.html
```diff
@@ -9684,26 +9684,34 @@ <h4 id="sctp-transport-connected">Connected procedure</h4>
...
+                    or the previous step failed to assign an id,
                   <a data-lt="data-channel-failure"> close the <var>channel</var> due to a failure
                   </a>. Otherwise, <a data-lt="announce the rtcdatachannel as open">announce the
                   <var>channel</var> as open</a>.</p>
                 </li>
               </ol>
             </li>
+            <li>
+              <p><a>Fire an event</a> named <code><a data-lt="
+              RTCSctpTransport state change">statechange</a></code> at <var>
```
lgrahl: The event fire order can now be misinterpreted easily. The individual `open` events fire after the `statechange` event because the `open` event is scheduled in yet another task. When we refactor that as suggested by @henbos, we need to ensure not to break that order: `statechange`, **then** for each data channel, `open`.

alvestrand: Adding a note seems reasonable.

alvestrand:
Updated with note (and a drive-by grammar edit).

lgrahl:
LGTM now

PR #2223, opened 2019-07-08
Add the "closing" event to datachannel

### LINKED ISSUE #1827: RTCDataChannel.send during 'closing' state
Let's look at the following scenario: Peer A and B have an `open` data channel. If peer A closes the channel, peer B's channel will go into the `closing` state until all pending messages of both peers have been sent. If peer B calls the `.send` method, it will throw an exception since the channel is not in the `open` state any more. But peer B has no way of knowing that unless it polls `.readyState`.

Do we need to address that with an... (I'm reluctant to say that) event?

alvestrand:
As per October 2018 TPAC decision.
Fixes #1827

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2223.html" title="Last updated on Jul 8, 2019, 8:01 AM UTC (8e9cb19)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2223/dd4a8ec...8e9cb19.html" title="Last updated on Jul 8, 2019, 8:01 AM UTC (8e9cb19)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9765,6 +9765,7 @@ interface RTCTrackEvent : Event {
                     attribute EventHandler        onopen;
                     attribute EventHandler        onbufferedamountlow;
                     attribute EventHandler        onerror;
+                    attribute EventHandler        onclosing;
                     attribute EventHandler        onclose;
     void close ();
                     attribute EventHandler        onmessage;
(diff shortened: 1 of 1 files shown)


dontcallmedom:
tested in https://github.com/web-platform-tests/wpt/blob/master/webrtc/RTCDataChannel-send.html removing "needs test" label

foolip:
I've filed https://bugs.webkit.org/show_bug.cgi?id=206056 for this.

@jan-ivar do you have a Gecko bug filed already?

Tests are coming in https://github.com/web-platform-tests/wpt/pull/21117, https://github.com/web-platform-tests/wpt/blob/master/webrtc/RTCDataChannel-send.html doesn't currently test this.

jan-ivar:
@foolip Looks like bots created one https://bugzilla.mozilla.org/show_bug.cgi?id=1608212 🙂

foolip:
@jan-ivar I think that bug is just for importing the change and will be automatically closed. But @jgraham can say for sure how this work.

Issue #2165, opened 2019-04-07
A simpler glare-proof setLocalDescription()


jan-ivar:
In [my latest blog](https://blog.mozilla.org/webrtc/perfect-negotiation-in-webrtc/) on *"Perfect negotiation in WebRTC"*, I discovered that most `negotiationneeded` code is glare-prone. E.g.
```js
pc.onnegotiationneeded = async () => {
  await pc.setLocalDescription(await pc.createOffer());
  io.send({desc: pc.localDescription});
}
```
In short, *createOffer* is asynchronous and takes time, so a remote offer may race us here and sometimes take us out of `"stable"` to `"have-remote-offer"`, beating our call to *setLocalDescription*, causing it to fail. To safeguard against this requires testing against `"stable"` right beforehand:
```js
pc.onnegotiationneeded = async () => {
  const offer = await pc.createOffer();
  if (pc.signalingState != "stable") return;
  await pc.setLocalDescription(offer);
  io.send({desc: pc.localDescription});
}
```
That's a bit much, so I'm proposing we allow a simpler version that's glare-safe:
```js
pc.onnegotiationneeded = async () => io.send({desc: await pc.setLocalDescription()});
```

Basically, allow *setLocalDescription* without argument, and treat it as follows:
 1. Default *type* to `pc.signalingState.includes("offer")? "answer" : "offer"`.
 2. if [[LastOffer]](https://w3c.github.io/webrtc-pc/#dfn-lastoffer)/[[LastAnswer](https://w3c.github.io/webrtc-pc/#dfn-lastanswer)] is `""`, implicitly invoke *createOffer()*/*createAnswer()* respectively.
 3. Resolve with `pc.localDescription`.

This is glare-safe because the peer connection's [operations queue](https://w3c.github.io/webrtc-pc/#dfn-operations-queue) does not allow an opportunity for JS to insert any method calls before the signaling state is changed (on success).

Fun fact: the sdp passed to *setLocalDescription()* is already unused; a ritual:
```js
await pc.setLocalDescription(await pc.createOffer());
```
...is already equivalent to (and no more flexible than):
```js
await pc.createOffer(); await pc.setLocalDescription({type: "offer"});
```
So this change is smaller than it might seem.

This would be fully backwards compatible; *setRemoteDescription()* would remain unchanged.

jan-ivar:
[...]

PR #2234, opened 2019-07-19
Support setLocalDescription() without arguments.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2165.

I had to refactor *createOffer* and *createAnswer* slightly for this. A readability win I hope.

Otherwise, all changes in SLD. @henbos PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2234.html" title="Last updated on Jul 25, 2019, 1:48 PM UTC (e45e986)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2234/5a8ff03...jan-ivar:e45e986.html" title="Last updated on Jul 25, 2019, 1:48 PM UTC (e45e986)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2331,11 +2331,11 @@
 interface RTCPeerConnection : EventTarget  {
   Promise&lt;RTCSessionDescriptionInit&gt; createOffer(optional RTCOfferOptions options);
   Promise&lt;RTCSessionDescriptionInit&gt; createAnswer(optional RTCAnswerOptions options);
-  Promise&lt;void&gt; setLocalDescription(RTCLocalSessionDescriptionInit description);
+  Promise&lt;void&gt; setLocalDescription(optional RTCSessionDescriptionInit description);
   readonly attribute RTCSessionDescription? localDescription;
   readonly attribute RTCSessionDescription? currentLocalDescription;
   readonly attribute RTCSessionDescription? pendingLocalDescription;
-  Promise&lt;void&gt; setRemoteDescription(RTCSessionDescriptionInit description);
+  Promise&lt;void&gt; setRemoteDescription(optional RTCSessionDescriptionInit description);
   readonly attribute RTCSessionDescription? remoteDescription;
   readonly attribute RTCSessionDescription? currentRemoteDescription;
   readonly attribute RTCSessionDescription? pendingRemoteDescription;
@@ -3051,9 +3051,8 @@ interface RTCPeerConnection : EventTarget  {
                 <p>The <dfn id=
                 "dom-peerconnection-setlocaldescription"><code>setLocalDescription</code></dfn>
                 method instructs the <code><a>RTCPeerConnection</a></code> to
-                apply the supplied
-                <code><a>RTCLocalSessionDescriptionInit</a></code> as the local
-                description.</p>
+                apply the supplied <code><a>RTCSessionDescriptionInit</a></code>
+                as the local description.</p>
                 <p>This API changes the local media state. In order to
                 successfully handle scenarios where the application wants to
                 offer to change from one media format to a different,
@@ -3657,12 +3656,12 @@ interface RTCPeerConnection : EventTarget  {
   Promise&lt;void&gt; createOffer(RTCSessionDescriptionCallback successCallback,
                             RTCPeerConnectionErrorCallback failureCallback,
                             optional RTCOfferOptions options);
-  Promise&lt;void&gt; setLocalDescription(RTCSessionDescriptionInit description,
+  Promise&lt;void&gt; setLocalDescription(optional RTCSessionDescriptionInit description,
                                     VoidFunction successCallback,
                                     RTCPeerConnectionErrorCallback failureCallback);
   Promise&lt;void&gt; createAnswer(RTCSessionDescriptionCallback successCallback,
                              RTCPeerConnectionErrorCallback failureCallback);
-  Promise&lt;void&gt; setRemoteDescription(RTCSessionDescriptionInit description,
+  Promise&lt;void&gt; setRemoteDescription(optional RTCSessionDescriptionInit description,
                                      VoidFunction successCallback,
                                      RTCPeerConnectionErrorCallback failureCallback);
   Promise&lt;void&gt; addIceCandidate(RTCIceCandidateInit candidate,
@@ -4113,7 +4112,7 @@ interface RTCPeerConnection : EventTarget  {
         session descriptions.</p>
         <div>
           <pre class="idl" data-tests="idlharness.https.window.js"
->[Exposed=Window, Constructor(RTCSessionDescriptionInit descriptionInitDict)]
+>[Exposed=Window, Constructor(optional RTCSessionDescriptionInit descriptionInitDict)]
 interface RTCSessionDescription {
   readonly attribute RTCSdpType type;
   readonly attribute DOMString sdp;
@@ -4128,10 +4127,12 @@ interface RTCSessionDescription {
                 The <dfn id=
                 "dom-sessiondescription"><code>RTCSessionDescription()</code></dfn>
                 constructor takes a dictionary argument,
-                <var>descriptionInitDict</var>, whose content is used to
+                <var>description</var>, whose content is used to
                 initialize the new <code><a>RTCSessionDescription</a></code>
-                object. This constructor is deprecated; it exists for legacy
-                compatibility reasons only.
+                object. This constructor is deprecated; it exists for
+                legacy compatibility reasons only. The constructor MUST
+                <a>throw</a> a <code>TypeError</code> if
+                <var>description</var>.<code>type</code> is not present.
               </dd>
             </dl>
           </section>
@@ -4162,7 +4163,7 @@ interface RTCSessionDescription {
         <div>
           <pre class="idl"
 >dictionary RTCSessionDescriptionInit {
-  required RTCSdpType type;
+  RTCSdpType type;
   DOMString sdp = "";
 };</pre>
           <section>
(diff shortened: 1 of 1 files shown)


jan-ivar:
Github preview seems to be slow to appear. If there's no preview button above, try rawgit: [Preview](https://rawgit.com/jan-ivar/webrtc-pc/glareproof/webrtc.html#dom-peerconnection-setlocaldescription).

webrtc.html
```diff
@@ -2638,40 +2619,45 @@ <h2>Methods</h2>
...
+                  </li>
                   <li>
                     <p>If <var>connection</var> is configured with an identity
-                    provider, then begin <a data-cite="!WEBRTC-IDENTITY#sec.identity-verify-assertion">the identity
-                    assertion request process</a> if it has not already
-                    begun.</p>
+                    provider, then begin <a
+                    data-cite="!WEBRTC-IDENTITY#sec.identity-verify-assertion">
+                    the identity assertion request process</a> if it has not
+                    already begun.</p>
```
henbos: In both createOffer and createAnswer: Shouldn't the identity assertion request process begin as part of the refactored algorithm? I think we need a part that includes both this bit and the "in parallel, begin the in-parallel steps..."
Otherwise setLocalDescription() is a way to bypass identity assertion.

henbos: Although _7. Verifying Identity Assertions_ says "Identity assertion validation happens when setRemoteDescription is invoked on RTCPeerConnection.", I have a feeling this all should to be clarified, but for now I don't want to introduce any accidental changes in this PR.

webrtc.html
```diff
@@ -3088,25 +3101,96 @@ <h2>Methods</h2>
...
+                          </li>
+                          <li data-tests="RTCPeerConnection-setLocalDescription-offer.html">
+                            <p>If <var>sdp</var> is the empty string, or if it
+                            no longer accurately represents the system state of
+                            <var>connection</var>, or if <var>connection</var>'s
+                            configured indentity provider has changed since
+                            <var>connection</var>'s <a>[[\LastCreatedOffer]]</a>
+                            was created, then let <var>p</var> be the result of
+                            <a href="#dfn-create-an-offer">creating an offer</a>
+                            with <var>connection</var>, and return the result of
```
henbos: Can you verify that the following is true?
Because these steps are chained in the connection's operation queue, there is no risk that the offer created here becomes invalid by the time that SLD() is performed. Thus no validation is required of the result of creating the offer.

webrtc.html
```diff
@@ -3088,25 +3101,96 @@ <h2>Methods</h2>
...
+                            <a href="https://w3ctag.github.io/promises-guide/#transforming-by">
+                            transforming</a> <var>p</var> with a fulfillment
+                            handler that returns the result of
+                            <a href="#set-description">setting the RTCSessionDescription</a>
+                            indicated by its first argument.</p>
```
henbos: "indicated by its first argument"? I have trouble parsing this sentence.
Should it say "setting the RTCSessionDescription with the created offer"?

webrtc.html
```diff
@@ -3088,25 +3101,96 @@ <h2>Methods</h2>
...
+                              <li>
+                                <p>Let <var>answer</var> be the handler's first
+                                argument.</p>
+                              </li>
+                              <li>
+                                <p>Return the result of <a
+                                href="#set-description">setting the RTCSessionDescription</a>
+                                indicated by
+                                <code>{<var>type</var>, <var>answer</var>.sdp}</code>.
+                                </p>
```
henbos: Can we phrase the create-offer-and-SLD and create-answer-and-SLD with the same language? (Why are sub-steps needed here but not above?)

webrtc.html
```diff
@@ -4088,7 +4172,33 @@ <h2>Dictionary <dfn>RTCSessionDescriptionInit</dfn>
...
+              <dd>The string representation of the SDP [[!SDP]]; if
+              <code>type</code> is <code>"rollback"</code>, this member is unused.
+              </dd>
+            </dl>
+          </section>
+        </div>
+        <div>
+          <pre class="idl"
+>dictionary RTCLocalSessionDescriptionInit {
+  RTCSdpType type;
```
henbos: The difference between RTCSessionDescriptionInit and RTCLocalSessionDescriptionInit is that "type" is not required? Can we remove the "required" bit of the original dictionary? If there is an implicit TypeError being lost when performing setRemoteDescription() or RTCSessionDescription constructor, I think I would rather add an "if type is undefined, throw TypeError" step than to introduce a near-identical dictionary for almost the same purpose.

Introducing a new dictionary to navigate around implicit IDL stuff seems less straight-forward to the reader.

webrtc.html
```diff
@@ -2350,7 +2331,7 @@ <h3>Interface Definition</h3>
 interface RTCPeerConnection : EventTarget  {
   Promise&lt;RTCSessionDescriptionInit&gt; createOffer(optional RTCOfferOptions options);
   Promise&lt;RTCSessionDescriptionInit&gt; createAnswer(optional RTCAnswerOptions options);
-  Promise&lt;void&gt; setLocalDescription(RTCSessionDescriptionInit description);
+  Promise&lt;void&gt; setLocalDescription(RTCLocalSessionDescriptionInit description);
```
henbos: Would this be more readable if we said "optional RTCSessionDescription description" instead?

jan-ivar: @henbos It should say `optional` for sure, thanks for catching that.

We don't require an `RTCSessionDescription` interface however—which would require a browser object as input—because we wanna support people doing tricks like
```js
pc.setLocalDescription({type: "pranswer", sdp: answer.sdp});
```

jan-ivar: And also
```js
pc.setLocalDescription({type: "rollback"});
```

webrtc.html
jan-ivar: Note you're commenting on a word-wrap change. The surrounding changes, however, move this step from the outer *createOffer* steps to inside the new [**create an offer**](https://rawgit.com/jan-ivar/webrtc-pc/glareproof/webrtc.html#dfn-create-an-offer) algorithm, *specifically* so *setLocalDescription()* won't bypass it.

This means identity initialization may happen a tad later in some esoteric edge cases on the answerer side. e.g:
```js
await Promise.all([pc.setRemoteDescription(offer), pc.createAnswer()]);
```
but I think that's ok.

The " the identity assertion request process" is asynchronous, and is merely triggered to start here, not end.

webrtc.html
jan-ivar: No, the operation chain isolates us from the other negotiation methods that also use the chain—*createOffer*, *createAnswer*, *setLocalDescription*, *setRemoteDescription*, *addIceCandidate* and *replaceTrack*—but not from other modification methods that *require* negotiation like e.g. *addTransceiver*, *addTrack*, *createDataChannel*, *setStreams*.

This PR text mirrors existing text in the [final steps to create an offer](https://w3c.github.io/webrtc-pc/#dfn-final-steps-to-create-an-offer) (in fact I almost linked to it, but it read a bit odd and didn't work right).

webrtc.html
jan-ivar: This language predates this PR. I also find it a bit odd, but the word "RTCSessionDescription" here refers to the input, not the set methods of a similar name.

I'm working on another PR that looks to clean this up a bit FWIW.

webrtc.html
jan-ivar: It's different, because here we need to use the passed-in *type*, which in the case of pranswer is `"pranswer"`, ***not*** `"answer"` which is what the [**create an answer**](https://rawgit.com/jan-ivar/webrtc-pc/glareproof/webrtc.html#dfn-create-an-answer) algorithm returns.

webrtc.html
jan-ivar: Counter-point: This shows clearly that type is required in `setRemoteDescription` but not `setLocalDescription`.

But I can try that, if people prefer it.

jan-ivar:
@henbos thanks for the feedback! I hope I've addressed your concerns with either answers or code changes. Can you take another look?

webrtc.html
henbos: Ack

webrtc.html
```diff
@@ -2722,18 +2708,18 @@ <h2>Methods</h2>
                     <p>If <var>connection</var> was modified in such a way that
                     additional inspection of the system state is necessary, or
                     if its configured indentity provider is no longer
```
henbos: (existing typo: indentity)

webrtc.html
henbos: Ack

webrtc.html
```diff
@@ -4029,7 +4112,7 @@ <h4><dfn>RTCSessionDescription</dfn> Class</h4>
         session descriptions.</p>
         <div>
           <pre class="idl" data-tests="idlharness.https.window.js"
->[Exposed=Window, Constructor(RTCSessionDescriptionInit descriptionInitDict)]
+>[Exposed=Window, Constructor(optional RTCSessionDescriptionInit descriptionInitDict)]
```
henbos: Editorial nit: Does "optional" make sense here? If it's missing, then implicitly type is missing, so we are throwing TypeError anyway. Would removing "optional" here have the same effect, only it would not lead the reader to believe it is optional until further reading?

webrtc.html
```diff
@@ -2350,11 +2331,11 @@ <h3>Interface Definition</h3>
 interface RTCPeerConnection : EventTarget  {
   Promise&lt;RTCSessionDescriptionInit&gt; createOffer(optional RTCOfferOptions options);
   Promise&lt;RTCSessionDescriptionInit&gt; createAnswer(optional RTCAnswerOptions options);
-  Promise&lt;void&gt; setLocalDescription(RTCSessionDescriptionInit description);
+  Promise&lt;void&gt; setLocalDescription(optional RTCSessionDescriptionInit description);
   readonly attribute RTCSessionDescription? localDescription;
   readonly attribute RTCSessionDescription? currentLocalDescription;
   readonly attribute RTCSessionDescription? pendingLocalDescription;
-  Promise&lt;void&gt; setRemoteDescription(RTCSessionDescriptionInit description);
+  Promise&lt;void&gt; setRemoteDescription(optional RTCSessionDescriptionInit description);
```
henbos: setRemoteDescription() still requires an argument, it should not be optional here.

webrtc.html
```diff
@@ -3147,7 +3230,7 @@ <h2>Methods</h2>
                         current <a>signaling state</a> of <var>connection</var>
                         as described in <span data-jsep=
                         "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>,
-                        then run the following sub-steps:</p>
+                        then run the following sub steps:</p>
```
henbos: setRemoteDescription(): Don't we need to explicitly guard against description.type being missing now, since it is no longer required?
Otherwise we would end up at "Return the result of setting the RTCSessionDescription to description" with a missing description.

webrtc.html
jan-ivar: It MUST be optional according to [WebIDL](https://heycam.github.io/webidl/#dfn-optional-argument-default-value): *"If the type of an argument is a dictionary type ..., and that dictionary type and its ancestors have no required members, and the argument is either the final argument or is followed only by optional arguments, then the argument must be specified as optional ..."*

webrtc.html
jan-ivar: Same thing: WebIDL requires it.

webrtc.html
jan-ivar: Hum yes. I must have been asleep. Will fix. Good catch!

dontcallmedom:
This is tested in https://github.com/web-platform-tests/wpt/blob/master/webrtc/RTCPeerConnection-setLocalDescription-parameterless.https.html removing "needs test"

Issue #2238, opened 2019-07-25
Nulling mid when m-line is recycled is surprising.


jan-ivar:
Right now [mid](https://w3c.github.io/webrtc-pc/#dom-rtptransceiver-mid) says *"After rollbacks, the value may change from a non-null value to null."*—for good reason: if we kept a `mid` from a remote offer, we'd risk sharing that mid with the transceiver the remote offer ultimately associates with (not unique).

But there's another place it's set to `null` in [JSEP](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-5.10): *"If the m= section is being recycled (see Section 5.2.2), dissociate the currently associated RtpTransceiver by setting its mid property to null"*

Disassociating here may be important to JSEP's logic, but `null`ing the JS observable `transceiver.mid`  after `transceiver` has been stopped and removed from `pc.getTransceivers()` seems surprising and pointless (the mid can't be reused anyway).

In general, removing ids on users seems a bit like neener-neener. Let's not.

We can't touch JSEP, but I think we can work around it with an internal slot we don't touch past stopped.

cc @docfaraday

alvestrand:
My intuition says that clearing the mid property was designed to maintain the property that "all transceivers have a mid that is either unique or null".

In particular, the offer/answer exchange:

-> offer: m=video port=5 / a=mid=a

creates a transceiver with mid=a

<- answer: m=video port=0

disassociates the transceiver from the m-line.
If a new offer comes in:

<- offer: m=video port=97 / a=mid=a

will create a new transceiver with mid=a

If we don't clear, we now have two transceivers with mid=a, one of them stopped.

My point of confusion may be the statement "the mid can't be reused anyway", which I can't remember offhand that we have an explicit agreement to (and since mid can come in from non-webrtc endpoints, that agreement needs to be in the sdp spec, not in webrtc spec).

jan-ivar:
@alvestrand said he like the invariant that `mid` tracked association. I'm going to abandon this then, unless there's interest. cc @docfaraday

PR #2239, opened 2019-07-25
Don't clear mid property if stopped.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2238. @docfaraday @henbos PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2239.html" title="Last updated on Jul 26, 2019, 6:59 PM UTC (e9357bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2239/72ed617...jan-ivar:e9357bc.html" title="Last updated on Jul 26, 2019, 6:59 PM UTC (e9357bc)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7873,7 +7873,7 @@ interface RTCRtpTransceiver {
                   <p>Set <var>transceiver</var>'s <a>[[\Mid]]</a> slot to the
                   new mid value.</p>
                 </li>
-              </li>
+              </ol>
             </dd>
             <dt data-tests="RTCPeerConnection-transceivers.https.html,RTCRtpTransceiver.https.html"><dfn data-idl><code>sender</code></dfn> of type <span class=
             "idlAttrType"><a>RTCRtpSender</a></span>, readonly</dt>
--- a/webrtc.html
+++ b/webrtc.html
@@ -7752,10 +7752,17 @@ interface RTCRtpReceiver {
       <a><code>RTCRtpReceiver</code></a> that share a common
       <code>mid</code>. As defined in <span data-jsep="rtptransceivers">[[!JSEP]]</span>,
       an <a><code>RTCRtpTransceiver</code></a> is said to be <dfn>associated</dfn> with
-      a <a>media description</a> if its <code><a data-link-for=RTCRtpTransceiver>mid</a></code>
-      property is non-null; otherwise it is said to be disassociated. Conceptually, an
-      <a>associated</a> transceiver is one that's represented in the last applied session
-      description.</p>
+      a <a>media description</a> if its "mid property" is non-null; otherwise it is
+      said to be <dfn>disassociated</dfn>.
+      Conceptually, a transceiver becomes <a>associated</a> once it has been
+      successfully represented in a session description. Once associated, a
+      transceiver only becomes <a>disassociated</a> if it is immediately rolled
+      back, or if the transceiver has been <a>stopped</a> and its m-line has
+      been recycled, as defined in <span data-jsep=
+      "subsequent-offers subsequent-answers">[[!JSEP]]</span>. In the latter
+      case, the <code><a>RTCRtpReceiver</a></code>'s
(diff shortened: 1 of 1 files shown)


jan-ivar:
This is a JSEP mistake we're fixing in webrtc-pc basically.

webrtc.html
```diff
@@ -7752,10 +7752,17 @@ <h3><dfn>RTCRtpTransceiver</dfn> Interface</h3>
...
-      description.</p>
+      a <a>media description</a> if its "mid property" is non-null; otherwise it is
+      said to be <dfn>disassociated</dfn>.
+      Conceptually, a transceiver becomes <a>associated</a> once it has been
+      successfully represented in a session description. Once associated, a
+      transceiver only becomes <a>disassociated</a> if it is immediately rolled
+      back, or if the transceiver has been <a>stopped</a> and its m-line has
+      been recycled, as defined in <span data-jsep=
+      "subsequent-offers subsequent-answers">[[!JSEP]]</span>. In the latter
+      case, the <code><a>RTCRtpReceiver</a></code>'s
```
henbos: the RTCRtpTransceiver's mid value

webrtc.html
```diff
@@ -7778,6 +7785,10 @@ <h3><dfn>RTCRtpTransceiver</dfn> Interface</h3>
           <p>Let <var>transceiver</var> have a <dfn>[[\Receiver]]</dfn> internal
           slot, initialized to <var>receiver</var>.</p>
         </li>
+        <li data-tests="RTCPeerConnection-addTransceiver.https.html">
+          <p>Let <var>transceiver</var> have a <dfn>[[\Mid]]</dfn> internal
+          slot, initialized to <var>null</var>.</p>
+        </li>
```
henbos: An internal slot is good. But we have existing language (i.e. when setting an RTCSessionDescription) that says to "Set transceiver's mid value to the mid of the media description" and similar sentences in several places. Update those to modify the internal slot.

webrtc.html
```diff
@@ -7841,6 +7852,28 @@ <h2>Attributes</h2>
               negotiation is complete, the <code>mid</code> value may be null.
               After rollbacks, the value may change from a non-null value
               to null.</p>
+              <p>On getting, the attribute MUST return the value of the
+              <a>[[\Mid]]</a> slot.</p>
+              <p>Whenever <span data-jsep=
+              "subsequent-offers subsequent-answers">[[!JSEP]]</span> says to
+              modify the "mid property", the user agent MUST queue a task that
+              runs the following steps:</p>
```
henbos: I think this is the wrong place to put this algorithm. This decouples it from setting an RTCSessionDescription and it becomes unclear whether to update the mid before or after firing events etc. But is this algorithm even needed, we're already updating the mid inside of the task that was queued for SLD/SRD. Let's not queue two tasks.

webrtc.html
```diff
@@ -7752,10 +7752,17 @@ <h3><dfn>RTCRtpTransceiver</dfn> Interface</h3>
...
-      property is non-null; otherwise it is said to be disassociated. Conceptually, an
-      <a>associated</a> transceiver is one that's represented in the last applied session
-      description.</p>
+      a <a>media description</a> if its "mid property" is non-null; otherwise it is
+      said to be <dfn>disassociated</dfn>.
+      Conceptually, a transceiver becomes <a>associated</a> once it has been
+      successfully represented in a session description. Once associated, a
+      transceiver only becomes <a>disassociated</a> if it is immediately rolled
+      back, or if the transceiver has been <a>stopped</a> and its m-line has
+      been recycled, as defined in <span data-jsep=
```
henbos: I think the "if the transceiver has been stopped and its m-line has been recycled" should be part of the SLD/SRD steps.

aboba: Why remove the link from "mid"?  Also, "successfully represented in a session description" is not as specific as "last applied session description".

Issue #2173, opened 2019-04-15
Timing of setRemoteDescription's identity validation is unclear


jan-ivar:
The steps under [setRemoteDescription](https://w3c.github.io/webrtc-pc/#dom-peerconnection-setremotedescription) talk about two things:
  1. calling the [setting an RTCSessionDescription](http://w3c.github.io/webrtc-pc/#set-description) algorithm, and
  2. validating any identity assertion.

But it doesn't nail down the timing between the two. This must be clarified.

I believe the [setting an RTCSessionDescription](http://w3c.github.io/webrtc-pc/#set-description) algorithm needs to go last, because it fires a ton of events, and queues a task to [update the negotiationneeded flag](http://w3c.github.io/webrtc-pc/#dfn-update-the-negotiation-needed-flag), something that is timed to happen after the JS success callback.

Blocking on identitiy validation beyond that point in time would ruin invariants that callers of the API rely on.

Likewise, in the case where SRD success is blocked on verifying an identity assertion, holding off on firing a ton of media events until the identity has been positively verified, seems wise.

henbos:
I think we've assumed that ontrack fires and the promise resolves, allowing you to do things in then() as soon as possible. I'm not sure that resolves all issues, but more guarantees is certainly a good thing.

jan-ivar:
> I think we've assumed that ontrack fires and the promise resolves, allowing you to do things in then() as soon as possible.

@henbos Most events (`track`, `mute`, `unmute`, `addTrack` and `removeTrack`) fire *right before*¹ the success callback, which is fine. The exception is `negotiationneeded` which is designed to fire *after* for obvious reasons (super confusing to initiate a new round of negotiation before `SRD(answer)` has resolved).

https://github.com/w3c/webrtc-pc/pull/2251 ensures `negotiationneeded` fires *after* even when `peerIdentity` is used.

---
<sub>1. Substitute *"right before"* with *"before"* if identity validation takes a while, something https://github.com/w3c/webrtc-pc/pull/2251 also fixes.</sub>

henbos:
Provided we [move identity to the identity spec](https://github.com/w3c/webrtc-pc/issues/2364) we can move this issue as well when we do that.

PR #2251, opened 2019-08-03
Don't fire SRD() events until target peer identity has been verified.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2173.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2251.html" title="Last updated on Aug 3, 2019, 12:47 AM UTC (bdf95fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2251/f665b4d...jan-ivar:bdf95fd.html" title="Last updated on Aug 3, 2019, 12:47 AM UTC (bdf95fd)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1418,6 +1418,17 @@
                 </ol>
               </li>
               <li>
+                <p>If <var>description</var> is set as a remote description and
+                is applied successfully, and a <a>target peer identity</a> is
+                set, then the user agent MUST wait for
+                <a href="#dfn-validate-the-identity">identity validation</a>
+                to complete before proceeding to the next step. If
+                <a href="#dfn-validate-the-identity">identity validation</a>
+                fails, <a>reject</a> <var>p</var> with a newly
+                <a data-link-for="exception" data-lt="create">created</a>
+                <code>OperationError</code>, and abort these steps.</p>
+              </li>
+              <li>
                 <p>If <var>description</var> is applied successfully, the user
                 agent MUST queue a task that runs the following steps:</p>
                 <ol>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1417,6 +1417,17 @@ <h4>Set the RTCSessionDescription</h4>
...
+              <li>
+                <p>If <var>description</var> is set as a remote description and
+                is applied successfully, and a <a>target peer identity</a> is
+                set, then the user agent MUST wait for
+                <a href="#dfn-validate-the-identity">identity validation</a>
+                to complete before proceeding to the next step. If
+                <a href="#dfn-validate-the-identity">identity validation</a>
+                fails, <a>reject</a> <var>p</var> with a newly
+                <a data-link-for="exception" data-lt="create">created</a>
+                <code>OperationError</code>, and abort these steps.</p>
```
henbos: If the description was applied successfully, but we reject the promise and abort the steps, won't that mean that the JavaScript layer and the JSEP layer are out-of-sync? I.e. there may be JSEP transceivers or m= sections or whatnot, but no JavaScript equivalent RTCRtpTransceivers?

henbos: Do we need to tell JSEP to roll back?

webrtc.html
```diff
@@ -1417,6 +1417,17 @@ <h4>Set the RTCSessionDescription</h4>
                   </li>
                 </ol>
               </li>
+              <li>
+                <p>If <var>description</var> is set as a remote description and
```
alvestrand: It doesn't make sense to apply the description before you have verified the identity, I think.

jan-ivar: @alvestrand Good point. @martinthomson WDYT?

webrtc.html
jan-ivar: You and @alvestrand make the same observation.

In practice if your peer connection has a target peerIdentity (cannot be changed), and you get a bad-identity remote description, you're most likely toast anyway and will want to discard the peer connection, unless you there's hope you'll get a valid offer reasonably soon thereafter, in which case you apply SRD again (which is legal).

A corner case may be a previously established connection without a target peerIdentity, then you set one, and then you get a bad-identity remote description. In this case I don't know whether having JSEP apply successfully is any worse than not applying it.

To be safe we should probably just hold off applying anything until the identity is validated. Right now Firefox does it in parallel with SRD, an optimization with little benefit, I suspect.

henbos:
Start the process before applying the description

alvestrand:
This is OBE based on the TPAC 2019 plan of moving the normative language into the identity spec.

PR #1926, opened 2018-07-10
Delete Example 14

aboba:
Fix for Issue  https://github.com/w3c/webrtc-pc/issues/1709

Issue #1709, opened 2018-01-02
example 14: render before verifying the remote fingerprint?


fippo:
[example 14](https://w3c.github.io/webrtc-pc/#peer-to-peer-example-with-media-before-signaling) says
> the offerer may wish to render the media before the answer arrives.

For this to happen, the offerer would have to accept unverified media as the remote fingerprint is not known  before SRD (unless we are talking about renegotiation, thanks @alvestrand ). I assume showing unverified media is prohibited *somewhere*.

When SRD is called this can be done with a track/stream which means there is virtually no advantage in doing "early" media as shown in this example compared to a normal negotiation.

If this example is intended to show a renegotiation that is completly missing.

alvestrand:
https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24#section-7.3 is probably the most worked example.

alvestrand:
https://tools.ietf.org/html/draft-ietf-mmusic-msid-16#section-3.1 talks about waiting until we're in a stable negotiation state, but that's for a different reason.
I remember discussing the unverified early media thing and agreeing on the language for it (buffer or drop, but do NOT handle), but I'm unable to remember where we put it. @aboba might remember.

nils-ohlmeier:
Another little thing which appears broken to me in example 14 is that it calls gUM and only ever attaches the stream to the local preview video 'selfView.srcObject'. But since I don't see it ever getting attached to the PeerConnection 'onnegotiationneeded' will never fire.

And yes @alvestrand I remember the same outcome of an earlier discussion.

stefhak:
>Another little thing which appears broken to me in example 14 is that it calls gUM and only ever attaches the stream to the local preview video 'selfView.srcObject'. But since I don't see it ever getting attached to the PeerConnection 'onnegotiationneeded' will never fire.

I opened #1708 for this!

aboba:
@alvestrand Peter demonstrated that in webrtc-pc media could not arrive prior to the Answer (containing both the remote fingerprint and the remote ufrag/password), so the question was moot.  Note that the proposed RTCIceTransport extension could change this if it permits ICE and DTLS signaling to be decoupled.

alvestrand:
The case where it can happen is if a PR-Answer arrives with the fingerprint and ufrag/password but no media, and then an answer arrives with the media configuration.

aboba:
It's not clear to me what value Example 14 provides.  Can we delete it?

[...]

PR #2266, opened 2019-08-12
Clarify example on example with media before signaling

### LINKED ISSUE #1708: Example 14 never sends any media
https://w3c.github.io/webrtc-pc/#peer-to-peer-example-with-media-before-signaling

There is no `addTrack` or `replaceTrack`.

dontcallmedom:
As per discussion in #1708:
* Change the title to 'with incoming media before signaling'
* Change the intro to add 'This code shows the receiving side'
* Remove the 'selfView' code

close #1708

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2266.html" title="Last updated on Aug 14, 2019, 7:37 AM UTC (4c97e90)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2266/f665b4d...4c97e90.html" title="Last updated on Aug 14, 2019, 7:37 AM UTC (4c97e90)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -11810,11 +11810,11 @@ signaling.onmessage = async (event) =&gt; {
       </div>
     </section>
     <section>
-      <h3>Peer-to-peer Example with media before signaling</h3>
+      <h3>Peer-to-peer Example with incoming media before signaling</h3>
       <div>
         <p>The answerer may wish to send media in parallel with sending the
         answer, and the offerer may wish to render the media before the answer
-        arrives.</p>
+        arrives. This code shows the receiving side.</p>
         <pre class="example highlight">
 const signaling = new SignalingChannel();
 const configuration = {iceServers: [{urls: 'stuns:stun.example.org'}]};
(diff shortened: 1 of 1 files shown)


aboba:
By removing the call to `getUserMedia`, where will _stream_ come from?

dontcallmedom:
`stream` is no longer used in the example, so it doesn't need to come from anywhere as far as I can tell; I have reverted my removal of the insertion of the `remoteView` though.

fippo:
You're a bit meta by clarifying the example on the example, no? :-)

The example remains wrong though, see also #1709:
```
remoteView.srcObject = new MediaStream(pc.getReceivers().map((r) => r.track));
```
is called in start() which is immediately after the peerconnection creation. Since there are no calls to addTransceiver, `pc.getReceivers()` will return `[]` and the result is not useful.

I still think removing it (as proposed in #1926) is easier than coming up with something that is not confusing. Maybe @jan-ivar can be convinced to write a blog post instead?

dontcallmedom:
ok, I'll close this PR then and wait to see how #1926 gets resolved then :)

PR #2269, opened 2019-08-13
RTCIceTransport selected candidate pair behavior when changing state

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1981

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2269.html" title="Last updated on Aug 14, 2019, 7:09 PM UTC (745bd20)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2269/1a2a324...745bd20.html" title="Last updated on Aug 14, 2019, 7:09 PM UTC (745bd20)">Diff</a>

Issue #1981, opened 2018-09-07
RTCIceTransport selected candidate pair behavior when changing state


steveanton:
I have some questions about the behavior of getSelectedCandidatePair() and the selectedcandidatepairevent in the following cases:

1) When the RTCIceTransport is closed (due to either RTCPeerConnection close() or SLD/SRD removing it)? Proposal: getSelectedCandidatePair() returns null and the event does not fire.
2) When the RTCIceTransport transitions to 'disconnected'? Proposal: getSelectedCandidatePair() returns the current value the event does not fire.
3) When the RTCIceTransport transitions to 'failed'? Proposal: getSelectedCandidatePair() returns null and the event does not fire.

Also, when might the selectedcandidatepairevent fire with a newly-null selected candidate pair?

alvestrand:
Thoughts:
- Let getSelectedCandidatePair return null if the state is "new", "checking" or "closed"
- In the "failed" state, return whatever the last selected pair was (or null if there never was a selected pair)
- in "connected" or "completed" states, return the currently selected candidate
There's no way to "disconnected" that doesn't go at least via "connected" or "completed", so in "disconnected", the currently selected candidate should always have a value.

We should remove the "and null otherwise" from step 4 under "When the ICE Agent indicates that the selected candidate pair for an RTCIceTransport has changed". That way, there's no way to change [[SelectedCandidatePair]] from non-null to null, and also no way to fire the event with an empty candidate pair.

(There's no definition for when the ICE Agent indicates that the selected candidate pair has changed; RFC 5245 section 11.1.1 (full implementations) says that the selected pair is empty only if there is no previous selected pair, so this is consistent with the above.)

steveanton:
Harald's proposal sounds good to me. I can see the value in keeping the selected candidate pair set when transitioning to "failed".

aboba:
Proposal: 

When the RTCIceTransport is closed (due to either RTCPeerConnection close() or SLD/SRD removing it)? Proposal: getSelectedCandidatePair() returns null and the event does not fire.
When the RTCIceTransport transitions to 'disconnected'? Proposal: getSelectedCandidatePair() returns the current value the event does not fire.
When the RTCIceTransport transitions to 'failed'? Proposal: getSelectedCandidatePair() returns the current value and the event does not fire.

PR #2276, opened 2019-08-16
RTCIceTransport selected candidate pair behavior when changing state

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/1981 and https://github.com/w3c/webrtc-pc/issues/2283

Rebase of https://github.com/w3c/webrtc-pc/pull/2269

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2276.html" title="Last updated on Sep 3, 2019, 8:36 PM UTC (dea9be8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2276/90526ef...dea9be8.html" title="Last updated on Sep 3, 2019, 8:36 PM UTC (dea9be8)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8796,20 +8796,21 @@ interface RTCDtlsTransport : EventTarget {
           <code><a>RTCIceTransportState</a></code>.</p>
         </li>
         <li>
-          <p>If <var>newState</var> is <code>"new"</code> or <code>"closed"</code>,
-          set <var>transport.<a>[[\SelectedCandidatePair]]</a></var> to
-          <code>null</code>.</p>
+          <p>If <var>newState</var> is <code>"new"</code>,
+          <code>"checking"</code> or <code>"closed"</code>,
+          set <var>transport.<a>[[\SelectedCandidatePair]]</a></var>
+          to <code>null</code>.</p>
         </li>
         <li>
-          <p>If <var>newState</var> is <code>"checking"</code>, set
+          <p>If <var>newState</var> is <code>"connected"</code>, set
           <var>transport.<a>[[\SelectedCandidatePair]]</a></var> to the
-          candidate pair on which media is sent if one exists, otherwise
-          <code>null</code>.</p>
+          candidate pair on which media is sent.</p>
         </li>
         <li>
-          <p>If <var>newState</var> is <code>"disconnected"</code> or <code>"failed"</code>,
-          <var>transport.<a>[[\SelectedCandidatePair]]</a></var> retains its last
-          set value.</p>
+          <p>If <var>newState</var> is <code>"completed"</code>,
+          <code>"disconnected"</code> or <code>"failed"</code>,
+          <var>transport.<a>[[\SelectedCandidatePair]]</a></var> retains
+          its last set value.</p>
         </li>
         <li>
           <p>Set <var>transport</var>.<a>[[\IceTransportState]]</a> to
@@ -8817,6 +8818,12 @@ interface RTCDtlsTransport : EventTarget {
         </li>
         <li>
           <p><a>Fire an event</a> named <code><a>statechange</a></code> at
+          <var>transport</var>.</p>
+        </li>
+        <li>
+          <p>If <var>transport.<a>[[\SelectedCandidatePair]]</a></var>
+          has changed, <a>Fire an event</a> named
+          <code><a>selectedcandidatepairchange</a></code> at
           <var>transport</var>.</p>
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


henbos:
Move to "When the ICE Agent indicates that the RTCIceTransportState for an RTCIceTransport has changed, the user agent MUST queue a task that runs the following steps:"

aboba:
> Move to "When the ICE Agent indicates that the RTCIceTransportState for an RTCIceTransport has changed, the user agent MUST queue a task that runs the following steps:"

Done.

youennf:
PR iterates over all state values, except completed and connected.
completed and connected seem to fall in the same bucket as disconnected and failed, which is basically do nothing. We could explicitly state so for completeness.

webrtc.html
```diff
@@ -8793,7 +8793,23 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
...
+        <li>
+          <p>If <var>newState</var> is <code>"new"</code> or <code>"closed"</code>,
+          set <var>transport.<a>[[\SelectedCandidatePair]]</a></var> to
+          <code>null</code>.</p>
+        </li>
+        <li>
+          <p>If <var>newState</var> is <code>"checking"</code>, set
+          <var>transport.<a>[[\SelectedCandidatePair]]</a></var> to the
+          candidate pair on which media is sent if one exists, otherwise
+          <code>null</code>.</p>
```
henbos: When "checking" we don't have a candidate pair, see https://w3c.github.io/webrtc-pc/#dom-rtcicetransportstate

henbos: Follow-up issue #2283 ?

aboba: Right.  In "checking" the selected candidate pair is null.  In "connected" it should be set to the selected pair.

aboba: @henbos Combined the steps as requested in Issue https://github.com/w3c/webrtc-pc/issues/2283

webrtc.html
```diff
@@ -8793,57 +8793,49 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
...
+          <var>transport.<a>[[\SelectedCandidatePair]]</a></var> to the
+          best valid pair (prior to nomination) or the nominated candidate
+          pair.</p>
         </li>
         <li>
-          <p><a>Update the connection state</a> of <var>connection</var>.</p>
+          <p>If <var>newState</var> is <code>"completed"</code>,
+          <code>"disconnected"</code> or <code>"failed"</code>,
+          <var>transport.<a>[[\SelectedCandidatePair]]</a></var> retains
+          its last set value.</p>
```
henbos: This implies that we must go to "connected" before "completed", because only in "connected" do we update the candidate pair. Is this true? What if there is only a single pair to check, we check it, and it connects. There's no pairs to check, so we're completed. Would connected -> completed necessarily be two steps and not one?

webrtc.html
```diff
@@ -8793,57 +8793,49 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
         </li>
```
henbos: Now that we've merged the state and pair update as per #2283, we need to update the above text "When the ICE Agent indicates that the RTCIceTransportState for an RTCIceTransport has changed" to say "When the ICE Agent indicates that the RTCIceTransportState for an RTCIceTransport has changed _or the selected candidate pair has changed_". Otherwise changing pair while connected to another pair and still being connected would not update the pair.

webrtc.html
```diff
@@ -8793,57 +8793,49 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
...
         <li>
           <p>Let <var>newState</var> be the new indicated
-          <code><a>RTCIceTransportState</a></code>.
+          <code><a>RTCIceTransportState</a></code>.</p>
         </li>
```
henbos: Suggestion: Have a step here like the old way of phrasing it: "Let newCandidatePair be a newly created RTCIceCandidatePair representing the indicated pair if one is selected, and null otherwise."

Then in a note explain that this value would be null under such and such circumstances and not null in such and such circumstances. I think that is more readable and you can set [SelectedCandidatePair] to match the underlying pair without conditionals around it. I think having conditionals only increases the risk of not updating the pair sometimes (like going from connected to connected). An explanation is better as a note than as additional algorithm steps.

webrtc.html
henbos: Also saying "retains its last value" is a NO-OP, see suggestion for clarifying what the pair's value is as a note instead.

henbos:
If the state OR pair changes, queue a task that runs the following:
- Let newState be the new state
- Let newPair be the new pair. [Note: It's null in such and such case and not null in such and such case.]
- transport.state = newState, transport.pair = newPair
- If the state changed, fire onstatechanged
- If the pair changed, fire onselectedcandidatepairchanged

Issue #2215, opened 2019-06-21
Order of multiple RTCDataChannel.send calls in case of blobs


youennf:
https://w3c.github.io/webrtc-pc/#datachannel-send defines how a blob is to be sent.
It is not clear though how blob data is retrieved.

Reading the spec, it seems that blob data retrieval is expected to happen synchronously.
This guarantees the sending order of the messages.
Sending (blob1, arrayBuffer1, blob2) would lead to three messages in the same order (blob1, arrayBuffer, blob2) on the receiving side for ordered channels.

Browsers usually do retrieve blob data asynchronously. If done so, the order of the messages is not guaranteed.
Taking the previous example, the order might actually be (arrayBuffer, blob1, blob2) or (arrayBuffer, blob2, blob1). This might be surprising to web authors.

My understanding of the intent of the spec is to guarantee order of sending.
Implementation-wise, this makes it somehow harder to implement since a queue of messages might be needed.
Clarification might be good in any case.

youennf:
I did some testing.
Safari WebSocket implementation preserves the order of sending.
Firefox DataChannel implementation does not seem to preserve the order (https://jsfiddle.net/Lgmp2shu/).

lgrahl:
I don't think we need to say anything about order since an ordered channel should always reflect order to the API. But a clarification note is never wrong.

You will always need a send queue with the high level API we have. Firefox has one, libwebrtc has one. It's just a matter of ensuring the order. The current behaviour should be fixed by Mozilla. /cc @nils-ohlmeier

alvestrand:
Adding a note (making sure it's obvious that we intend ordered in -> ordered out) seems like the only reasonable spec change here.

jan-ivar:
(A bit late, sorry) I'd prefer stronger language than a note, but I welcome the note.

> I don't think we need to say anything about order since an ordered channel should always reflect order to the API.

@lgrahl That's the desired outcome, but these are the instructions on how to make that happen on the sending side. As such, I think they fall short by treating Blobs—borne out of Files with asynchronous accessors—as having synchronously available data.

As you and @youennf point out, we want this solved in a way that doesn't block main thread, which requires either an additional level of queuing of *all* send data, or some placeholder in any existing queue.

guest271314:
[...]

PR #2286, opened 2019-08-30
Clarify that blob send order should respect the send call order

youennf:
PR for https://github.com/w3c/webrtc-pc/issues/2215

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/webrtc-pc/pull/2286.html" title="Last updated on Aug 30, 2019, 1:30 PM UTC (90a54f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2286/c3acf00...youennf:90a54f2.html" title="Last updated on Aug 30, 2019, 1:30 PM UTC (90a54f2)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -10483,6 +10483,12 @@ interface RTCDataChannel : EventTarget {
               <p><code>Blob</code> object:</p>
               <p>Let <var>data</var> be the raw data represented by the
               <code>Blob</code> object.</p>
+              <div class="note" data-tests="RTCDataChannel-send-blob-order.html">
+              Although the actual retrieval of <var>data</var> from a <code>Blob</code> object
+              can happen asynchronously, the user agent will make sure to queue the data on
+              the <var>channel</var>'s <a>underlying data transport</a> in the same order
+              as the send method is called. The byte size of <var>data</var> needs to be known
+              synchronously.</div>
             </li>
             <li>
               <p><code>ArrayBuffer</code> object:</p>


youennf:
Related WPT PR is https://github.com/web-platform-tests/wpt/pull/18770

webrtc.html
```diff
@@ -10483,6 +10483,12 @@ <h2>Methods</h2>
               <p><code>Blob</code> object:</p>
               <p>Let <var>data</var> be the raw data represented by the
               <code>Blob</code> object.</p>
+              <div class="note" data-tests="RTCDataChannel-send-blob-order.html">
+              Although the actual retrieval of <var>data</var> from a <code>Blob</code> object
+              can happen asynchronously, the user agent will make sure to queue the data on
```
jan-ivar: I like this language, but I think we need to go further and use MUST here, since I don't see this enforced anywhere else.

If MUST in a note isn't ok then we should make it regular normative language (not a note). See my https://github.com/w3c/webrtc-pc/issues/2215#issuecomment-526667672.

lgrahl:
Didn't see the comment of @jan-ivar being addressed.

youennf:
> Didn't see the comment of @jan-ivar being addressed.

This PR was discussed during the editor's call.
Consensus is that the test and the spec note are sufficient.

It would be even better to add some more precise steps as suggested by @jan-ivar.
This should probably be done for both web socket and webrtc-pc specs.

PR #2298, opened 2019-09-12
Fire events and update state correctly for rollback.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/1706. Sits on top of https://github.com/w3c/webrtc-pc/pull/2295.


webrtc.html
```diff
@@ -1620,8 +1623,80 @@ <h4>Set the RTCSessionDescription</h4>
...
+                        <p>Set <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
+                        and <var>connection</var>.<a>[[\PendingRemoteDescription]]</a>
+                        to <code>null</code>, and set <a>signaling state</a> to
+                        <code>"stable"</code>.</p>
+                      </li>
+                    </ol>
+                  </li>
+                  <li>
+                    <p>If <var>remote</var> is <code>false</code>, then run the
+                    following steps:</p>
```
henbos: Now that rolling back the initial offer makes it possible for description to be null, I think this needs to say "If remote is false _and description is not null_, then run the following steps:

webrtc.html
```diff
@@ -1620,8 +1623,80 @@ <h4>Set the RTCSessionDescription</h4>
...
+                          "rollback">[[!JSEP]]</span>.</p>
+                      </li>
+                      <li data-tests="RTCPeerConnection-setLocalDescription-rollback.html,RTCPeerConnection-setLocalDescription-answer.html,RTCPeerConnection-setRemoteDescription-rollback.html">
+                        <p>If <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
+                        is not <code>null</code>, then set <var>description</var> to
+                        <var>connection</var>.<a>[[\CurrentLocalDescription]]</a>,
+                        and set <var>remote</var> to <code>false</code>. Otherwise,
+                        set <var>description</var> to
+                        <var>connection</var>.<a>[[\CurrentRemoteDescription]]</a>,
+                        and set <var>remote</var> to <code>true</code>.
```
henbos: In other words "if we are rolling back a local description, set _description_ to the current local description, and if we're rolling back a remote description, set _description_ to the current remote description." And then we try to update to match _description_.

Just to make sure: Is this correct? Descriptions, both local and remote ones, can modify transceivers. Should we always roll back to a description based on the type of description being rolled back, or should we always roll back to the most up-to-date answer that got us to "stable"?

For example:
```
...
setLocalDescription(offer);
...
setRemoteDescription(answer);
setRemoteDescription(offer);
setRemoteDescription(rollback);
```

In this case we are rolling back a remote offer, and the current remote description would represent the answer SDP that got us to the previous "stable" state. That sounds good... I think?

```
...
setRemoteDescription(offer);  // 1)
...
setLocalDescription(answer);  // 2)
setRemoteDescription(offer);
setRemoteDescription(rollback);
```

But in this case, while we are also rolling back a remote offer, the current remote description would be the offer... however what was offered could have been rejected in the answer, and this was set as a _local_ description.

Assume offer 1) had a=sendrecv, but 2) made it a=recvonly because we didn't have anything to send. Wouldn't rolling back to the current remote description imply rolling back to 1), i.e. now we're firing "ontrack" again, even though currentDirection should become recvonly again (local answer), not sendrecv (remote offer).

@jan-ivar Could you shed some light on this? I'm not sure how trying to apply a particular current description is guaranteed to capture all the relevant states that need to be rolled back. Or what would happen if we modify transceivers while an offer is pending.

I might imagine that _description_ should be set to the "latest description applied that was an answer", but I'm not sure if that is correct either, e.g. does it always contain all the relevant m= sections or not?

webrtc.html
```diff
@@ -1771,7 +1846,7 @@ <h4>Set the RTCSessionDescription</h4>
                     </ol>
                   </li>
                   <li>
-                    <p>If <var>description</var> is set as a remote description,
+                    <p>Otherwise, if <var>remote</var> is <code>true</code>,
```
henbos: The "otherwise" condition might also need to guard against description being null?

webrtc.html
henbos: Hang on a minute, from this step and onward, why are we running different steps based on if remote is true or false? Applying a rollback should have the same effect, regardless if we do SLD(rollback) or SRD(rollback).

webrtc.html
henbos: Ping @jan-ivar: Does the PR need updating or am I confused? (Or both :) )

webrtc.html
jan-ivar: > I think this needs to say "If remote is false and description is not null

The sole step is "Run the following steps for each media description in description", so that might be implied, but yes I'll clarify this (and save an indent)

> why are we running different steps based on if remote is true or false? Applying a rollback should have the same effect, 

It has. See line 1652: *"If connection.[[PendingLocalDescription]] is not null, then set description to connection.[[CurrentLocalDescription]], and set remote to false. Otherwise, set description to connection.[[CurrentRemoteDescription]], and set remote to true. "*

webrtc.html
jan-ivar: > Assume offer 1) had a=sendrecv, but 2) made it a=recvonly because we didn't have anything to send. Wouldn't rolling back to the current remote description imply rolling back to 1), i.e. now we're firing "ontrack" again, even though currentDirection should become recvonly again (local answer), not sendrecv (remote offer).

Thanks, yes you're right, I need to update the PR here. Our rejection API wort (expose in offer, revert in answer) is the gift that keeps on giving. 😕

We need to reconstruct not just from [[CurrentRemoteDescription]] in this case, but from [[CurrentDirection]] as well, to account for rejections.

We want to roll back correctly to recvonly in this case, and thankfully [[CurrentDirection]] has this value, thanks to existing prose in SLD(answer): *"Set transceiver.[[CurrentDirection]] and transceiver.[[FiredDirection]] to direction."*

As to your other questions, the scope of this PR is solely about firing the correct events, so I suggest we discuss those in the issue you opened.

henbos:
```
await pc.setLocalDescription(await pc.createOffer());
...
await pc setRemoteDescription(answer);

// [[CurrentLocalDescription]] has feature set "ABC"
// [[CurrentRemoteDescription]] has feature set "AB", some subset of "ABC".

if (rollbackLocalOffer) {
  await pc.setLocalDescription(await pc.createOffer());
  await pc.set[Local/Remote]Description({type:'rollback'});
  // We've rolled back to an offer.
  // Our starting point is feature set "ABC", though we try to make sure
  // direction is [[CurrentDirection]], i.e. a property of feature set "AB".
} else if (rollbackRemoteOffer) {
  await pc.setRemoteDescription(offer);
  await pc.set[Local/Remote]Description({type:'rollback'});
  // We've rolled back to an answer.
  // Our starting point is feature set "AB".
}
```
Here, "feature set" is a generic term meant to apply to anything that SDP allows. We've addressed the direction problem, but that's just one example of a feature set. Other examples may or may not include (depending on how confused I am):
- Header extensions
- Codecs
- ICE ufrags
- Bundling or not?

In terms of getting the right events to fire, this PR might be correct, but in terms of trying to surface lower layer states to JS, I wonder what happens.

For example, if transports are created in at the "offer" stage, then it might be incorrect to try to roll back to an "offer", but if transports are not created until "answer", then rolling back to an "offer" is a NO-OP in terms of transport states, in which case it is safe. I think.

Instead of "does this fire the right events?" I want to rephrase the problem as "which state are we trying to get to?"

jan-ivar:
> Why is a local offer trying to get back to a previous local description?

It's not. JSEP has already happened (restored to stable state e.g. from a snapshot). The webrtc-pc algorithm calls JSEP in step 3 to do this. The steps this PR modifies all run on JSEP success. The *description* auto variable this PR modifies is used solely to fire events + move tracks based on stream association changes. Nothing else. Unlike JSEP, the conditions for these decisions overlap 100% with what can be done through new incoming offers (once we accommodate our existing direction-rejection-in-answer kludge).

Local descriptions are uninteresting in this PR. Only remote descriptions fire events & move tracks around, so that's what we need to undo with more events/moves.

> What if codecs were "VP8, VP9" in the offer but in the answer they were "VP8" only?

I'd say this is out of scope for this PR, and not unique to rollback. Let's discuss in https://github.com/w3c/webrtc-pc/issues/2313#issuecomment-541164223?

This may not be the last PR needed to get rollback right—I'm looking at transport objects exposed by this same algorithm—but I suspect solutions to those will come in forms more closer to: *"Restore the value of connection.[[SctpTransport]] to its value at the last stable signaling state."*

> Instead of "does this fire the right events?" I want to rephrase the problem as "which state are we trying to get to?"

JSEP already does, and it's `"stable"`. Events are just not state that can be restored that way.

JS specs favor algorithms, to lock down timing, since so much is JS observable. This algorithm in particular—our biggest and most complex by far—already contains careful steps to avoid security issues inherent with firing multiple events from the same task, so I think being pedantic is necessary.

I suggest we merge this and iterate.

alvestrand:
@henbos will file an issue for which description is chosen to generate "the previous state" ("it seems arbitrary").

PR #2309, opened 2019-09-25
Add adaptivePtime to RTCRtpEncodingParameters

### LINKED ISSUE #2300: Prefer variable ptime (audio frame lengths)
We negotiate a maxptime (in SDP but there is no control knob for), and RTCRtpEncodingParameters.ptime allows preferring a specific ptime.

Questions:
- Today: Is an implementation allowed to vary frame lengths dynamically/autonomously during a call (e.g. depending on network conditions)?
- Do we think that enabling variation would cause backwards compatibility issues?

Possible proposal:
- Do we want an API to control whether or not we prefer varying frame lengths?

eladalon1983:
When adaptivePtime is "on", the browser may dynamically change the audio frame length during the call, either reducing bandwidth consumption at the cost of increased latency, or vice versa.

The full rationale is in [this document](https://docs.google.com/document/d/12sVXRog-Dl9c-hbMInz9obmODk_yft06z7GMgfpdjPw/edit?usp=sharing).

Fixes #2300.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/webrtc-pc/pull/2309.html" title="Last updated on Oct 4, 2019, 3:20 PM UTC (78cc80b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2309/39a91c5...eladalon1983:78cc80b.html" title="Last updated on Oct 4, 2019, 3:20 PM UTC (78cc80b)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7057,11 +7057,8 @@ async function updateParameters() {
               <p>The value for this configuration option cannot be changed
               from its initial value. If this is attempted,
               <code>InvalidModificationError</code> MUST be thrown.</p>
-              <p><code>adaptivePtime</code> MUST NOT be set to <code>true</code>,
-              if a value was ever set for <code>ptime</code>, even if that value
-              was later cleared. Similarly, if <code>adaptivePtime</code> is set
-              to <code>true</code>, <code>ptime</code> MUST NOT subsequently be
-              set. If either of these is done,
+              <p>If <code>adaptivePtime</code> is set to <code>true</code>,
+              <code>ptime</code> MUST NOT be set; otherwise,
               <code>InvalidModificationError</code> MUST be thrown.</p>
             </dd>
             <dt data-tests="RTCRtpParameters-encodings.html"><dfn data-idl><code>maxBitrate</code></dfn> of type <span class=
--- a/webrtc.html
+++ b/webrtc.html
@@ -188,7 +188,6 @@
   DOMString peerIdentity;
   sequence&lt;RTCCertificate&gt; certificates;
   [EnforceRange] octet iceCandidatePoolSize = 0;
-  boolean adaptivePtime = false;
 };</pre>
           <section>
             <h2>Dictionary <a class="idlType">RTCConfiguration</a> Members</h2>
@@ -261,21 +260,6 @@
                 <p>Size of the prefetched ICE pool as defined in
                 <span data-jsep=
                 "ice-candidate-pool constructor">[[!JSEP]]</span>.</p>
-              </dd>
-              <dt><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType">boolean</span>, defaulting to <code>false</code>.</dt>
-              <dd>
-                <p>Indicates whether audio RTCRtpSenders MAY dynamically change
-                the frame length. If the value is <code>true</code>,
-                the user agent MAY use any valid frame length for any of its
-                frames, and MAY change this at any time.</p>
-                <p class="note">Using a longer frame length reduces the
-                bandwidth consumption due to overhead, but does so at the cost
-                of increased latency. Allowing the user agent to change the
-                frame length dynamically, allows it to adapt its bandwidth
-                allocation strategy based on the current network conditions.</p>
-                <p>The value for this configuration option cannot be changed
-                from its initial value. If this is attempted,
-                InvalidModificationError MUST be thrown.</p>
               </dd>
             </dl>
           </section>
@@ -6993,6 +6977,7 @@ async function updateParameters() {
   RTCDtxStatus dtx;
   boolean active = true;
   unsigned long ptime;
+  boolean adaptivePtime = false;
   unsigned long maxBitrate;
   double maxFramerate;
   double scaleResolutionDownBy;
@@ -7055,6 +7040,29 @@ async function updateParameters() {
               data-jsep= "applying-a-remote-desc">[[!JSEP]]</span>. Note that
               the user agent MUST still respect the limit imposed by any
               "maxptime" attribute, as defined in [[!RFC4566]], Section 6.</p>
+            </dd>
+            <dt><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType">boolean</span>, defaulting to <code>false</code>.</dt>
+            <dd>
+              <p>Indicates whether this encoding MAY dynamically change
+              the frame length. If the value is <code>true</code>,
+              the user agent MAY use any valid frame length for any of its
+              frames, and MAY change this at any time. Valid values are positive
+              integers divisible by 10. If the "maxptime" attribute (defined in
+              [[!RFC4566]] Section 6) is specified, that maximum applies.</p>
+              <p class="note">Using a longer frame length reduces the
+              bandwidth consumption due to overhead, but does so at the cost
+              of increased latency. Allowing the user agent to change the
+              frame length dynamically, allows it to adapt its bandwidth
+              allocation strategy based on the current network conditions.</p>
+              <p>The value for this configuration option cannot be changed
+              from its initial value. If this is attempted,
+              <code>InvalidModificationError</code> MUST be thrown.</p>
+              <p><code>adaptivePtime</code> MUST NOT be set to <code>true</code>,
+              if a value was ever set for <code>ptime</code>, even if that value
+              was later cleared. Similarly, if <code>adaptivePtime</code> is set
+              to <code>true</code>, <code>ptime</code> MUST NOT subsequently be
+              set. If either of these is done,
+              <code>InvalidModificationError</code> MUST be thrown.</p>
             </dd>
             <dt data-tests="RTCRtpParameters-encodings.html"><dfn data-idl><code>maxBitrate</code></dfn> of type <span class=
             "idlMemberType">unsigned long</span></dt>
--- a/webrtc.html
+++ b/webrtc.html
@@ -188,6 +188,7 @@
   DOMString peerIdentity;
   sequence&lt;RTCCertificate&gt; certificates;
   [EnforceRange] octet iceCandidatePoolSize = 0;
+  RTCAdaptivePtime adaptivePtime;
 };</pre>
           <section>
             <h2>Dictionary <a class="idlType">RTCConfiguration</a> Members</h2>
@@ -260,6 +261,14 @@
                 <p>Size of the prefetched ICE pool as defined in
                 <span data-jsep=
                 "ice-candidate-pool constructor">[[!JSEP]]</span>.</p>
+              </dd>
+              <!-- TODO: !!! Add data-tests before pushing for full review. -->
+              <dt data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType"><a>RTCAdaptivePtime</a></span>, defaulting to <code>"default"</code>.</dt>
+              <dd>
+                <p>Indicates whether adaptive audio frame length should be used, should be suppressed, or whether this should be left up to the implementation to decide.</p>
+                <p>When adaptive audio frame length is used, the implementation should dynamically change the audio frame length
+                during the call, so as to adapt to changing network conditions. (Higher frame lengths reduce the bandwidth consumption
+                due to overhead, but do so at the cost of increased latency.)</p>
               </dd>
             </dl>
           </section>
@@ -646,6 +655,39 @@
                 <code><a>RTCRtpSender</a></code> and <code><a>RTCRtpReceiver</a></code>.</li>
             </ol>
           </div>
+        </div>
+      </section>
+      <section>
+        <h4><dfn>RTCAdaptivePtime</dfn> Enum</h4>
+        <div>
+          <pre class="idl">enum RTCAdaptivePtime {
+  "off",
+  "on",
+  "default"
+};</pre>
+          <table data-link-for="RTCAdaptivePtime" data-dfn-for="RTCAdaptivePtime"
+          class="simple">
+            <tbody>
+              <tr>
+                <th colspan="2">Enumeration description</th>
+              </tr>
+              <tr>
+                <!-- TODO: !!! Add data-tests before full review. -->
+                <td data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>off</code></dfn></td>
+                <td>The implementation is not allowed to dynamically change the audio frame length during the call.</td>
+              </tr>
+              <tr>
+                <!-- TODO: !!! Add data-tests before full review. -->
+                <td data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>on</code></dfn></td>
+                <td>The implementation should dynamically change the audio frame length during the call.</td>
+              </tr>
+              <tr>
+                <!-- TODO: !!! Add data-tests before full review. -->
+                <td data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>default</code></dfn></td>
+                <td>It's up to the implementation to decide whether the audio frame length adaptation should be dynamically changed during the call.</td>
+              </tr>
+            </tbody>
+          </table>
         </div>
       </section>
       <section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -264,9 +264,18 @@
               </dd>
               <dt><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType">boolean</span>, defaulting to <code>false</code>.</dt>
               <dd>
-                <p>Indicates whether audio senders may dynamically change the frame lengths so as to adapt to changing network conditions.
-                (Higher frame lengths reduce the bandwidth consumption due to overhead, but do so at the cost of increased latency.)</p>
-                <p>The value for this configuration option cannot change after its value is initially selected.</p>
+                <p>Indicates whether audio RTCRtpSenders MAY dynamically change
+                the frame length. If the value is <code>true</code>,
+                the user agent MAY use any valid frame length for any of its
+                frames, and MAY change this at any time.</p>
+                <p class="note">Using a longer frame length reduces the
+                bandwidth consumption due to overhead, but does so at the cost
+                of increased latency. Allowing the user agent to change the
+                frame length dynamically, allows it to adapt its bandwidth
+                allocation strategy based on the current network conditions.</p>
+                <p>The value for this configuration option cannot be changed
+                from its initial value. If this is attempted,
+                InvalidModificationError MUST be thrown.</p>
               </dd>
             </dl>
           </section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -266,6 +266,7 @@
               <dd>
                 <p>Indicates whether audio senders may dynamically change the frame lengths so as to adapt to changing network conditions.
                 (Higher frame lengths reduce the bandwidth consumption due to overhead, but do so at the cost of increased latency.)</p>
+                <p>The value for this configuration option cannot change after its value is initially selected.</p>
               </dd>
             </dl>
           </section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -264,7 +264,7 @@
               </dd>
               <dt><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType">boolean</span>, defaulting to <code>false</code>.</dt>
               <dd>
-                <p>Indicates whether audio senders may dynamically change the frame lengths so as to adapt adapt to changing network conditions.
+                <p>Indicates whether audio senders may dynamically change the frame lengths so as to adapt to changing network conditions.
                 (Higher frame lengths reduce the bandwidth consumption due to overhead, but do so at the cost of increased latency.)</p>
               </dd>
             </dl>
--- a/webrtc.html
+++ b/webrtc.html
@@ -188,7 +188,7 @@
   DOMString peerIdentity;
   sequence&lt;RTCCertificate&gt; certificates;
   [EnforceRange] octet iceCandidatePoolSize = 0;
-  RTCAdaptivePtime adaptivePtime;
+  boolean adaptivePtime = false;
 };</pre>
           <section>
             <h2>Dictionary <a class="idlType">RTCConfiguration</a> Members</h2>
@@ -262,13 +262,10 @@
                 <span data-jsep=
                 "ice-candidate-pool constructor">[[!JSEP]]</span>.</p>
               </dd>
-              <!-- TODO: !!! Add data-tests before actually pulling. -->
-              <dt data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType"><a>RTCAdaptivePtime</a></span>, defaulting to <code>"default"</code>.</dt>
+              <dt><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType">boolean</span>, defaulting to <code>false</code>.</dt>
               <dd>
-                <p>Indicates whether adaptive audio frame length should be used, should be suppressed, or whether this should be left up to the implementation to decide.</p>
-                <p>When adaptive audio frame length is used, the implementation should dynamically change the audio frame length
-                during the call, so as to adapt to changing network conditions. (Higher frame lengths reduce the bandwidth consumption
-                due to overhead, but do so at the cost of increased latency.)</p>
+                <p>Indicates whether audio senders may dynamically change the frame lengths so as to adapt adapt to changing network conditions.
+                (Higher frame lengths reduce the bandwidth consumption due to overhead, but do so at the cost of increased latency.)</p>
               </dd>
             </dl>
           </section>
@@ -655,39 +652,6 @@
                 <code><a>RTCRtpSender</a></code> and <code><a>RTCRtpReceiver</a></code>.</li>
             </ol>
           </div>
-        </div>
-      </section>
-      <section>
-        <h4><dfn>RTCAdaptivePtime</dfn> Enum</h4>
-        <div>
-          <pre class="idl">enum RTCAdaptivePtime {
-  "off",
-  "on",
-  "default"
-};</pre>
-          <table data-link-for="RTCAdaptivePtime" data-dfn-for="RTCAdaptivePtime"
-          class="simple">
-            <tbody>
-              <tr>
-                <th colspan="2">Enumeration description</th>
-              </tr>
-              <tr>
-                <!-- TODO: !!! Add data-tests before actually pulling. -->
-                <td data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>off</code></dfn></td>
-                <td>The implementation is not allowed to dynamically change the audio frame length during the call.</td>
-              </tr>
-              <tr>
-                <!-- TODO: !!! Add data-tests before actually pulling. -->
-                <td data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>on</code></dfn></td>
-                <td>The implementation should dynamically change the audio frame length during the call.</td>
-              </tr>
-              <tr>
-                <!-- TODO: !!! Add data-tests before actually pulling. -->
-                <td data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>default</code></dfn></td>
-                <td>It's up to the implementation to decide whether the audio frame length adaptation should be dynamically changed during the call.</td>
-              </tr>
-            </tbody>
-          </table>
         </div>
       </section>
       <section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -262,7 +262,7 @@
                 <span data-jsep=
                 "ice-candidate-pool constructor">[[!JSEP]]</span>.</p>
               </dd>
-              <!-- TODO: !!! Add data-tests before pushing for full review. -->
+              <!-- TODO: !!! Add data-tests before actually pulling. -->
               <dt data-tests="RTCPeerConnection-constructor.html"><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType"><a>RTCAdaptivePtime</a></span>, defaulting to <code>"default"</code>.</dt>
               <dd>
                 <p>Indicates whether adaptive audio frame length should be used, should be suppressed, or whether this should be left up to the implementation to decide.</p>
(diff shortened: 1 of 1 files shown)


eladalon1983:
(Please note that this is WIP. I intend to resolve the TODOs prior to submitting for final review, but after this has been approved in principle.)

rshpount:
The rational document is private and not readable.

eladalon1983:
> The rational document is private and not readable.

I did not **save** my change to the sharing setting. That is fixed now. Thank you for alerting me to the problem!

rshpount:
Is there any sort of event or statistics that will inform what audio frame length change or will allow to detect what audio frame length is currently being used?

How would minimum and maximum supported audio frame lengths will be set or negotiated?

What is the benefit of this approach vs generating renegotiation needed event and changing ptime using signaling exchange?

fippo:
as I just said on blink-dev I think RTCConfiguration is the wrong place for this, it belongs next to https://w3c.github.io/webrtc-pc/#dom-rtcrtpencodingparameters-ptime

@rshpount one would already not need renegotiation to change the ptime via encoding parameters, no?
This would simply automate the process instead of the consumer having to call getStats, decide the netwok conditions are "bad" and then change the ptime for a "better" tradeoff.

alvestrand:
Has anyone experimented with setting ptime at runtime? (Does it work in any existing browser?)
It would serve to validate (or invalidate) the concern about existing apps breaking when ptime varies.

webrtc.html
```diff
@@ -648,6 +657,39 @@ <h4><dfn>RTCRtcpMuxPolicy</dfn> Enum</h4>
           </div>
         </div>
       </section>
+      <section>
+        <h4><dfn>RTCAdaptivePtime</dfn> Enum</h4>
```
alvestrand: Why bother with an enum? Is there any chance we'll ever want to do something different from "true" and "false"?

alvestrand:
As part of formalities - can you file a bug for this issue, to point out what the pull request is solving?

webrtc.html
eladalon1983: > Why bother with an enum? Is there any chance we'll ever want to do something different from "true" and "false"?

I understood Philip Jägenstedt as wanting that. A boolean would serve my needs equally well.

eladalon1983:
> As part of formalities - can you file a bug for this issue, to point out what the pull request is solving?

Done.

eladalon1983:
> Is there any sort of event or statistics that will inform what audio frame length change or will allow to detect what audio frame length is currently being used?
No. We can discuss adding that separately, I think?

> How would minimum and maximum supported audio frame lengths will be set or negotiated?
This is up for discussion. I suggest as a first approach to leave it up to the implementation, as it mostly is at the moment. Another possibility is to change adaptivePtime  to a dictionary that holds the allowed range.

> What is the benefit of this approach vs generating renegotiation needed event and changing ptime using signaling exchange?
Signaling and renegotiation take time and bandwidth. When the available bandwidth drops, both are at a premium. We allow an implementation to vary the target bitrate for video and audio dynamically, without renegotiation. The same rationale goes for varying the frame length, and the overhead it consumes, I believe.

fippo:
> Another possibility is to change adaptivePtime to a dictionary that holds the allowed range.

Its not a range, its a set of discrete values (which actually applies to ptime)
`3, 5, 10, 20, 40, 60, or an arbitrary multiple of an Opus frame size rounded up to the next full integer value, up to a maximum value of 120`
(from RFC 7587).

The range adaptivePtime can use would be limited by maxptime as an upper bound, similar to how maxBitrate in the senders parameters is effectively limited by b= settings in the SDP. Oddly, while some of the opus drafts had a minptime fmtp that didn't make it to the RFC...

> The same rationale goes for varying the frame length, and the overhead it consumes, I believe.

Note that if that is the only parameter you're changing, doing a local renegotiation is perfectly possible. Been there, done that, used it for bandwidth adaption until setParameters became available. Not going to miss it :-)

> Yes. I'd add that the browser is likely to be able to react faster than the web-app to changes in bandwidth; the web-app can only call getStats so often.

True, but adapting within a 1s window should be fast enough? And its a trade-off, how does the application developer know this tradeoff fits the  use-case?

eladalon1983:
> Its not a range, its a set of discrete values (which actually applies to ptime) ...

fippo, [ptime ](https://w3c.github.io/webrtc-pc/#dom-rtcrtpencodingparameters-ptime) can be taken from a range, but it is itself a single value, so I don't think we can use that to determine the range of allowed adaptive values.

> The range adaptivePtime can use would be limited by maxptime as an upper bound ...

maxptime is an Opus-specific fmtp line, so I'm not sure it's useful for a codec-agnostic flag.

> doing a local renegotiation is perfectly possible

The web-app can only call getStats so often. The browser is at a better position to make quick decisions about the frame length.

> ... adapting within a 1s window should be fast enough?

1. Adapting within a 1s window sounds suboptimal to me.
2. Also, the logic for adapting the frame length would now to be baked into the web-app. It could be argued that this would not be making the Web better.

> ... how does the application developer know this tradeoff fits the use-case?

It's obvious in the extreme cases, and grey otherwise. E.g., if you're building a streaming service for high-end networks, requiring 50mbit/s as a barrier to entry, likely the trade-off is not for you.

henbos:
Discussion during meeting:
- Somewhat related: ptime should be [at risk](https://github.com/w3c/webrtc-pc/issues/2311), if it is removed then we don't have to specify which value takes priority if both are specified.
- The feeling was that this makes sense but that we should use a boolean.
- Considering WebRTC 1.0 is reaching PR and features not implemented by multiple browsers are at risk, this should be defined in an extension spec.
- To decide in the extension spec: Should this be a parameter (setParameters() on a per-sender basis) or an RTCConfiguration?

henbos:
With regards to spawning extension specs: We should consolidate them in a few months time?

webrtc.html
```diff
@@ -648,6 +657,39 @@ <h4><dfn>RTCRtcpMuxPolicy</dfn> Enum</h4>
...
+        <div>
+          <pre class="idl">enum RTCAdaptivePtime {
+  "off",
+  "on",
+  "default"
```
henbos: Suggestion for use of MUST/SHOULD which have specific meaning in spec and is upper-cased: When adaptivePtime is on, the User Agent SHOULD adapt the frame length based on network conditions and when it is off, the User Agent MUST NOT adapt the frame length based on network conditions.

fippo:
> ptime can be taken from a range, but it is itself a single value, so I don't think we can use that to determine the range of allowed adaptive values.

No. My point here is that the current spec has an issue that ptime (in opus) restricts valid values. When JS attempts to set an invalid value like 42 what should  happen? I might head over to the other issue.

> maxptime is an Opus-specific fmtp line, so I'm not sure it's useful for a codec-agnostic flag.

its not, its in RFC 4566. minptime is not defined anywhere but https://tools.ietf.org/html/draft-spittka-payload-rtp-opus-03 which didn't make it to the RFC.
Apart from that, what codecs beside Opus do support this even? I doubt G711 does, G722 may.
We do have a range defined by SDP. The big question is whether it is safe to vary within that range *without* negotiating that in the SDP. 

> The web-app can only call getStats so often. The browser is at a better position to make quick decisions about the frame length.

How dynamic are the adaptions required in practice? Also it might be useful to discuss scenarios... my favorite one is "swiss 2g" with a maximum bitrate of 200kbps, 200ms latency and 2% packet loss. Note that one of the key drivers for this is "we just lost wifi  and are now on 2g", i.e. the selectedcandidatepairchange event.

> It could be argued that this would not be making the Web better.

It could be argued that allowing more differentiation will make the web better. See a certain krankygeek talk last year.

One thing to keep in mind: when the network conditions are bad there are other things to consider such as turning off video or reducing the frequency of stun ping/pong (yes chrome, please be spec-compliant, there is an RFC for consent). Those conditions might require a more holistic approach.
Adapting ptime is part of the solution for sure!

webrtc.html
fippo: I'd suggest not overusing 2119 language. When adaptivePtime is off, the UA MUST use the ptime negotiated via SDP. When it is on, it may (lowercase!) vary within the envelope negotiated in the SDP

rshpount:
First of all, audio frame size generally is done by the sender based on something that happens on the stream receiver. It can be triggered by RTCP statistics received by the sender, but doing this through signaling is generally not much slower. In other words, stream receiver can send an RTCP report which shows packet loss (or higher delay) or it can initiate another signaling exchange which will cause ptime to change. Time-wiser this is almost the same.

Codecs like Opus support a finite set of audio frame lengths. Codecs like G.711 can theoretically support any audio frame length dividable by 1/8 ms increment. For instance 27.5 ms frame is technically legal. This being said, jitter buffers often assume specific audio frame size and will not adapt to variable audio frame size. Because of this, when variable audio frame size is used, both sides should negotiate what frame sizes are supported in advance (i.e. specify that 5 ms, 10 ms, 20 ms, 30 ms, and 60 ms frames are supported and others are not). One added benefit of having a negotiated set, is that it makes interop testing easier. Properly testing that all possible frame sizes are supported (3 to 120 ms with 1/8 ms interval in any order), is a nightmare. Not testing this properly would likely lead to buffer overflow bugs due to extremely large frames or large number of packets in jitter buffer, that can result in serious security issues.

For me, ideal interface for this would be some sort of event on the receiver side signaling that ideal audio frame size should be X ms and a method on the sender side to set the audio frame size to X ms. Connecting the two, would be something application specific and can be done via signaling or data channel. Doing something automatic and implementation specific is tempting and makes initial feature adoption faster but will make it less usable in the long run.

webrtc.html
eladalon1983: 1. I agree with fippo about using "may" rather than "must" in the on-case.
2. Why should "may" be lowercase in the on-case, fippo?
3. Do we want to explicitly say what behavior is expected when adaptivePtime is simply not provided? I.e. is it up to the UA to decide if it defaults to on/off, or must it default to off?

eladalon1983:
@rshpount, I apologize for not responding to you directly, but I think my comments to @fippo's message, which came first, cover your concerns as well. That is to say - the approach you propose has its own flaws, e.g. (1) the time required to respond to changes, or (2) the need to bake this important functionality into all JS applications. BWE and target-bitrate adaptation, for example, are done by the UA. It would be bad for the Web to require JS apps to bake that into their JS code.

webrtc.html
eladalon1983: I've made changes. Please take a look.

rshpount:
> We have tested and seen that existing versions of Chrome, Firefox and Safari already de-facto support receiving adaptive frame lengths. 

What exactly did you test?
That these browsers will receive variable frame rate and do not blow up?
That audio quality scores do not degrade with variable frame lengths under normal network conditions?
That audio quality would not degrade with variable frame lengths under adverse network conditions (delay spikes, packets going over two different routes with different delays, clock skew, high random jitter)?

You probably need a fairly extensive test using fuzzing plus recorded network delay profiles  and objective audio quality evaluation algorithm to verify this. I do not think you have done this even for your regular jitter buffer implementation with fixed frames.

>By default, an Opus frame is produced every 20ms. That's 50 times per second. We want to be able >to respond to network changes immediately when they are perceived. We cannot call getStats 50 >times per second and run client code that frequently (not to mention bake it into all JS apps). We >can do that in the UA quite easily, though. And we can optimize the code for that, and get all of the >Web to enjoy it, rather than force everyone to reinvent this wheel in JS.

Are you actually planning to update frame length 50 times per second? Bandwidth estimate does not change 50 times per second. Furthermore, the bandwidth estimate is actually obtained by receiver which needs to send it to sender which in turn will adapt the frame length. There is a natural delay to this.

>Imagine then that a change of frame-length from 20ms to 120ms is desired as soon as the BWE 
>drops below 60kbps.

First of all, most web application are quite useless at 60kbps.
Second, using 120 ms frame results in very high delay making communication feel more like broadcast then actual interactive conversation.

> (1) the time required to respond to changes,

Time is naturally required to do bandwidth estimate and transmit it to the other party. Making frame rate size changes too quickly results in higher quality degrade then waiting few hundred ms before doing this allowing connection to recover.

> (2) the need to bake this important functionality into all JS applications. BWE and target-bitrate adaptation, for example, are done by the UA. It would be bad for the Web to require JS apps to bake that into their JS code.

I am of the opposite opinion. Browser should do less. It is better for the web that if something can be implemented in JS, it should be implemented there vs dealing with 20 different versions of 4 different implementations in different browsers each doing something incomprehensible.

P.S. I have some experience building and operating a VoIP application that was designed to work under low bitrate under adverse network conditions. All we needed to do, in addition to codec bitrate adaption, was to change ptime using regular offer/answer between 20/30/40/60 ms roughly 1-2 seconds after major network change was detected. This network change often corresponded to IP address change for the client, so offer/answer came naturally.

webrtc.html
```diff
@@ -261,6 +262,12 @@ <h2>Dictionary <a class="idlType">RTCConfiguration</a> Members</h2>
                 <span data-jsep=
                 "ice-candidate-pool constructor">[[!JSEP]]</span>.</p>
               </dd>
+              <dt><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType">boolean</span>, defaulting to <code>false</code>.</dt>
+              <dd>
+                <p>Indicates whether audio senders may dynamically change the frame lengths so as to adapt to changing network conditions.
```
henbos: I'd like to clarify what the UA is or is not allowed to do, it sounds like if the value is true then changing the length is optional and up to the UA (is it a MAY or a SHOULD, would it be allowed to change the value for reasons other than "network conditions", how or when would it be allowed to change it, etc?), whereas if the value is false it sounds like it MUST NOT change. So I'd say that it MAY change on a per-packet basis, i.e. it's entirely up to the implementation, and that we don't mandate what the cause of the change is.

It should also be clear that it MUST use the same length if the value was false, otherwise there's no difference between on or off. Still, I don't think different codecs would necessarily use the same frame length, so I would tie deciding the "constant" frame length to be done when negotiating and say that it is the same per sender-and-codec. This is when we ensure resources are available as well, so it would makes sense there I guess.

I still want to mention network conditions since that is the motivation for the API, but considering it is entirely up to the UA on how it decides to change the length, I'd put this in a note rather than in a description about how the UA implements it.

Suggestion:

> Indicates whether audio RTCRtpSenders may dynamically change the frame lengths. If this value is true, the User Agent may use any valid frame length (less than the negotiated maxptime) for any of its packets, and MAY use a different frame length on a per-packet basis. If this value is false, the User Agent MUST decide the frame length to use for a particular RTCRtpSender and codec at time of negotiation, and MUST NOT change the frame length used for that RTCRtpSender-codec pair unless another neogitation is performed.
> [Note: Higher frame lengths reduce the bandwidth consumption due to overhead, but do so at the cost of increased latency. Allowing the User Agent to change the frame length dynamically allows it to adapt it's strategy, for example, based on network conditions.]

webrtc.html
henbos: OK I'm now convinced "MAY" is better than "SHOULD" considering how optional it is and totally up to the UA about its strategy. Don't know about "MAY" vs "may" though.

eladalon1983:
> What exactly did you test?

We are aware of the difficulties of testing this properly. A colleague of mine tested the happy-case, but a stressed version of it. The frame length was changed every 3s, cycling some select values between 20ms and 120ms (therefore "stressed"). Network conditions were kept constant, and favorable, throughout the test (therefore "happy-case"). We are aware that this is only a partial test. That's why we're not turning it on by default on Chrome, but rather leaving it up to the web-app. And the web-app is welcome to negotiate this capability on its own, using whatever method it chooses.

Repeat - one can negotiate this OOB if one is concerned about interoperability.

> Are you actually planning to update frame length 50 times per second?

I do not need to update 50 times per-second. What I want is to minimize response time. If you poll every second, you can have a delay of up to 1s from conditions changing, and your response to them. If you poll 50 times per second - which the UA easily can, but the web-app shouldn't - your max response time drops to 20ms. That's important for our usecase.

> First of all, most web application are quite useless at 60kbps.

I'm here to fix that. *micdrop*

Kidding aside, though (hope it will be taken in good spirit), this feature can save ~20kbps of overhead. If you're concerned about web-apps underperforming at lower bitrates, you're my target audience for this feature. I understand that you would be better served by a negotiated feature, but that would mean the feature is served to you later, or not at all (because I'd be forced to pick other, lower-hanging fruit, this season). Please negotiate this OOB if necessary.

> Second, using 120 ms frame results in very high delay making communication feel more like broadcast then actual interactive conversation.

We think our customers would prefer that over a dropped call. That's why we'll experiment with this feature. Others may either avoid this feature, or limit adaptation to something like 60ms using maxptime.

> P.S. I have some experience building and operating a VoIP application that was designed to work under low bitrate under adverse network conditions.

Thank you for sharing your experience. Your feedback is valued. Truly.

> All we needed to do, in addition to codec bitrate adaption, was to change ptime using regular offer/answer between 20/30/40/60 ms roughly 1-2 seconds after major network change was detected. This network change often corresponded to IP address change for the client, so offer/answer came naturally.

1. What about changes due to side-traffic? These can appear without a network change. They can persist for several seconds/minutes, then disappear as unexpectedly as they had come. This can happen frequently, too; perhaps more so than network changes.
2. I am curious about your solution - how it chose how long after a network change to set a new ptime, how it chose to change the ptime when the bandwidth estimation stabilized, etc. But this is probably not the right place to discuss this. Maybe face-to-face one day, or by email.

> I think @rshpount makes a good argument for possibly keeping ptime. If we end up supporting ptime I think we'd just need to say that if ptime is set than the UA is now allowed to use any other value. Don't think it's a blocker here.

Agreed.

> With regards to how or when the ptime dynamically changes, since we want the UA to be able to decide, my current suggestion is to not restrict how it changes, i.e. it's entirely up to the UA on each packet what length to use, as long as it is within valid values.

We're on the same page here.

webrtc.html
eladalon1983: I've taken part of your suggestion. PTAL.

webrtc.html
```diff
@@ -261,6 +262,12 @@ <h2>Dictionary <a class="idlType">RTCConfiguration</a> Members</h2>
...
+                (Higher frame lengths reduce the bandwidth consumption due to overhead, but do so at the cost of increased latency.)</p>
+                <p>The value for this configuration option cannot change after its value is initially selected.</p>
```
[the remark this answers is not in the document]
eladalon1983: I've taken part of your suggestion. PTAL.

henbos:
PTAL @alvestrand

rshpount:
My question is still -- why is this a boolean vs a set of allowed ptimes? For instance, I would see being able to set this to 20, 40, and 60 ms and let the UA pick one dynamically.

This is much easier to verify being handled correctly then support 3, 10, 18, 20, 46 ms frames that technically are possible, but troublesome when transcoding or designing jitter buffers.

eladalon1983:
> My question is still -- why is this a boolean vs a set of allowed ptimes? For instance, I would see being able to set this to 20, 40, and 60 ms and let the UA pick one dynamically.
> 
> This is much easier to verify being handled correctly then support 3, 10, 18, 20, 46 ms frames that technically are possible, but troublesome when transcoding or designing jitter buffers.

Your suggestion of a list of allowed ptimes sounds good. However, my internal devil's advocate came up with the following objections:
* What if the applications specifies values that the UA does not support? (Either all of the values, some, or none, could be supported by the UA.) Web developers would be confused if some magical values produced good results, but other values were ignored (or threw an error).
* Web developers would need to concern themselves with matters which are not their area of expertise, such as which sets of frame-lengths would produce the optimal behavior. This is not as easy as the alternative of simply opting in to a new feature. (Please remember that domain specialists like you can still vary the ptime directly from their app, using the methods you had listed earlier, rather than use this feature. Acceptance of adaptivePtime is orthogonal to the discussion over the deprecation of ptime.)

henbos:
My gut feeling is that specifying allowed ptimes is an expert use case and not what most people would want, and it adds more complexity in terms of how to use it correctly and which values the UA is capable of etc.

That said, a nice thing about a boolean like this is that it allows the UA to pick any _valid_ value. Which means it would be perfectly valid to add an extension API to allow the application to specify which ptimes should be valid, which could be useful in non-adaptive scenarios as well.

rshpount:
Typically the discussion about API would happen in w3c and network protocols would be discussed in IETF, but the RTCWEB group was recently closed so I am not sure what is the right place to talk about this. 

My main issue is that in case of adaptive audio frame, actual valid frame sizes are an implementation specific issue. What I would prefer is some sort of limitation of what is allowed (like anything that is dividable by 20 or 10 ms up to maxptime). Without some limitation on allowed audio frame values interop would be a nightmare due to very high number of use cases. This would effectively force me to disable adaptive frame size when using WebRTC with my media server, since I would not be able to effectively test it.

An additional API to set the valid audio frame size would be nice but might not be necessary. Some sort of API which informs about frame size used or when it is switched would definitely be useful.

rshpount:
> What if the applications specifies values that the UA does not support? (Either all of the values, some, or none, could be supported by the UA.) Web developers would be confused if some magical values produced good results, but other values were ignored (or threw an error).

Implementation should pick the nearest valid audio frame size no larger then specified frame size. This way things, at least, will not fail.

> Web developers would need to concern themselves with matters which are not their area of expertise, such as which sets of frame-lengths would produce the optimal behavior. This is not as easy as the alternative of simply opting in to a new feature. (Please remember that domain specialists like you can still vary the ptime directly from their app, using the methods you had listed earlier, rather than use this feature. Acceptance of adaptivePtime is orthogonal to the discussion over the deprecation of ptime.)

I think we should either limit the audio frame size allowed to be used by browsers or allow develop set this up, but provide a reasonable default. I think 20, 40, 60, 80, 100, 120 ms audio frame set seems reasonable. Having predictable frame sizes would make both implementation and testing a lot easier.

rshpount:
>We are aware of the difficulties of testing this properly. A colleague of mine tested the happy-case, but a stressed version of it. The frame length was changed every 3s, cycling some select values between 20ms and 120ms (therefore "stressed"). Network conditions were kept constant, and favorable, throughout the test (therefore "happy-case"). 

My concern is that some jitter buffers are designed to track number of queued packets not their total duration. Because of this, changing packet sizes dynamically might cause jitter buffers to switch to spike more and introduce excessive delay or cause excessive playback speed adaptation.

> What about changes due to side-traffic? These can appear without a network change. They can persist for several seconds/minutes, then disappear as unexpectedly as they had come. This can happen frequently, too; perhaps more so than network changes.

Curiously when dealing with low bit rate situation you are dealing with 3 distinct scenarios:

1. QOS is properly implemented or user learned or force to behave and no other traffic is sent or received during the real time conversation. This is not unusual for VoIP software in dedicated hardware. In such cases all you need to make sure is that bandwidth is correctly estimated and then audio frame size / bitrate is selected to provide the best possible quality for this connection

2. QOS is not used and there is side traffic. Router before the low bandwidth link is configured with large buffers. In this case you get significant delay spikes when TCP stream fills up the router buffer. What you see on the receive side is no packets for a period of time and then lots of packets received quickly. Jitter buffers have a special mode to deal with this by introducing extra delay to allow packets to be delivered, but even this has its limits since packet gaps can be up to tens of seconds long. Best option when such gap is detected is to provide feed back to user that connection was disrupted and application is trying to recover it. In WebRTC case, this means pulling statistics and informing the user.

3. Same as 2 but Router before the low bandwidth link is configured with small buffers or modern qdisc. In this case, in case of side TCP traffic you start seeing RTP packet loss. In such cases it makes more sense to continue sending packet every 20 ms but include 60 ms worth of audio in it (FEC for more then one previous packet). Application in this case can recover from fairly high random packet loss without disrupting the communication experience for the client.

> I am curious about your solution - how it chose how long after a network change to set a new ptime, how it chose to change the ptime when the bandwidth estimation stabilized, etc. But this is probably not the right place to discuss this. Maybe face-to-face one day, or by email.

I would be happy to discuss off line.

fippo:
playing devils advocate: how is this different from  a too simplified version of the webrtc.org [ANA](https://cs.chromium.org/chromium/src/third_party/webrtc/modules/audio_coding/audio_network_adaptor/)? The full ANA would be clearly out of scope for 1.0 obviously.

(also note that even ANA only supports a subset of the opus framesizes)

eladalon1983:
Roman, I think your suggestion of any-value-divisible-by-20-up-to-ptime is good. Modified slightly, it would be something along the lines of "the UA may use any value divisible by 20, but no more than maxptime (if specified)." If Henrik is also happy with this, I think we should go with it.

> Curiously when dealing with low bit rate situation you are dealing with 3 distinct scenarios

Good summary of some common cases. There are other cases out there, though I have to confess ignorance of their prevalence. Examples:
* Non-TCP side-traffic.
* Short bursts of TCP side-traffic.
* Varying bandwidth due to changing distance from a WAP, or due to varying interference.
* ISPs with an SLA which does not specify a minimum bandwidth, especially during peak hours.
* WAPs which divide bandwidth equally between currently connected users, at a place where the number of connected users is not very stable, such as airports, restaurant or cafés. 

I also wonder (perhaps you know?) what the network use patterns are for streaming services like Netflix. It might not be TCP. There are probably enough customers who would not be pleased if domestic disputes erupted whenever a member of the family sat down to watch TV, while another person was doing something else online. This should be enough to create an incentive for using a protocol which shares better than TCP.

> playing devils advocate: how is this different from a too simplified version of the webrtc.org ANA? The full ANA would be clearly out of scope for 1.0 obviously.

Chrome will indeed use a subset of ANA in its implementation of adaptivePtime.

> (also note that even ANA only supports a subset of the opus framesizes)

I know. I was involved with the ANA project. This PR expects any given UA would only implement support for a small finite number of frame lengths.

rshpount:
> I think your suggestion of any-value-divisible-by-20-up-to-ptime is good. Modified slightly, it would be something along the lines of "the UA may use any value divisible by 20, but no more than maxptime (if specified)." If Henrik is also happy with this, I think we should go with it.

I would be happy with current proposal with this limitation. I just don't want to figure out the implication of supporting any possible Opus or, much worse, G.711 audio frame length. Something sending 17.125 ms frames make me shiver.

> There are other cases out there, though I have to confess ignorance of their prevalence.

So far TCP in combination with mis-configured routers was the biggest offender. Things you commonly see disrupting connection are things like loading a modern web page resulting in load of ten 2 MB background images and five 1 MB font files. You also see somebody starting a torrent transfer and provider punishing the customer for this with high packet loss. Of cause, once HTTP/3 will be more widely deployed, non TCP related traffic patterns will likely emerge, but better queuing in HTTP/3 might help dealing with low bandwidth links more gracefully.

I did mention both short and long bursts of side TCP traffic. In case of large router buffers, short TCP burst results in short delay spike, which jitter buffer should be able to handle using "spike" mode. Long TCP burst results in longer delay spike which results in disrupted connection and requires some sort of recovery mechanism with some sort of UI. In case of smaller or smarter router buffers, long and short side TCP traffic results in short or long periods of RTP packet loss.

Some of your other scenarios do no result in bandwidth changes during the call. They simply result in calls running for its entire duration on the congested network.

> I also wonder (perhaps you know?) what the network use patterns are for streaming services like Netflix.

Most of video providers, like Netflix, use TCP based protocols such as DASH. This also changes with companies like Peer5 building delivery networks on top of DataChannel. Biggest issue with video providers is that they can affect not only home Internet connections, but entire service providers or backbone links. For instance, we saw quality of real time media served by Amazon cloud degrade on snow days due to more people staying home and streaming Netflix.

juberti:
@henbos: I would also prefer doing this on RTCRtpSendParameters.

aboba:
@juberti +1

henbos:
A few different issues here, I think for "do we want to adaptive ptime?" the answer is yes, for "where do we put it?" the overwhelming feeling is send parameters, as for "how is it allowed to change?" I get a sense that the UA should be able to decide, but the set of valid ptimes needs to be defined somewhere. So:
- Let's move this to RTCRtpSendParameters.
- Let's either define here that valid ptimes MUST be divisible by 10, or, if this is controversial and more discussion is needed, file a separate issue discussing valid frame lengths.

Are there any other concerns or issues? If so it might be good to file follow-up issues for the sake of focus and progress and unblocking rolling out adaptive ptime.

If we can reach the quality bar for "Editors can integrate" during the meeting today, I propose merging this into a "webrtc-extensions" repository to be created by one of us.

eladalon1983:
Discussed with Henrik offline. We'll go with RTCRtpEncodingParameters, not RTCRtpSendParameters.

eladalon1983:
PR updated; PTAL.

webrtc.html
fippo: the "may" does not limit behaviour, see section 6 in 2119. Use them sparingly ;-)

eladalon1983: Hi fippo@. You're commenting on an out of date revision.

webrtc.html
```diff
@@ -7040,6 +7041,26 @@ <h2>Dictionary <code><a>RTCRtpEncodingParameters</a></code>
...
+              integers divisible by 10. If the "maxptime" attribute (defined in
+              [[!RFC4566]] Section 6) is specified, that maximum applies.</p>
+              <p class="note">Using a longer frame length reduces the
+              bandwidth consumption due to overhead, but does so at the cost
+              of increased latency. Allowing the user agent to change the
+              frame length dynamically, allows it to adapt its bandwidth
+              allocation strategy based on the current network conditions.</p>
+              <p>The value for this configuration option cannot be changed
+              from its initial value. If this is attempted,
+              <code>InvalidModificationError</code> MUST be thrown.</p>
```
henbos: Note: The rationale for throwing if you try to turn off adaptivePtime after-the-fact is that it may be unclear which ptime to use at that point (should it change to some default ptime or use whatever the current ptime was, etc?). If we want to allow turning this off after-the-fact in order to allow e.g. reusing senders in non-adaptive cases we could revisit this in a follow-up issue.

webrtc.html
```diff
@@ -7040,6 +7041,26 @@ <h2>Dictionary <code><a>RTCRtpEncodingParameters</a></code>
...
+              <p>If <code>adaptivePtime</code> is set to <code>true</code>,
+              <code>ptime</code> MUST NOT be set; otherwise,
+              <code>InvalidModificationError</code> MUST be thrown.</p>
```
henbos: Note: While we would normally put information about when to throw exceptions under the respective algorithms (addTransceiver/setParameters) I think this is the right approach to enable merging this into an extension spec.

webrtc.html
```diff
@@ -7040,6 +7041,26 @@ <h2>Dictionary <code><a>RTCRtpEncodingParameters</a></code>
               the user agent MUST still respect the limit imposed by any
               "maxptime" attribute, as defined in [[!RFC4566]], Section 6.</p>
             </dd>
+            <dt><dfn data-idl><code>adaptivePtime</code></dfn> of type <span class="idlMemberType">boolean</span>, defaulting to <code>false</code>.</dt>
```
henbos: Update to say this is a read-only parameter.

henbos:
Takeaway from the meeting:
- Nobody is objecting to this feature. Not sure whether or not there is interest in it; there seem to be little opinion about it. Firefox/Safari would need to check internally.
- Agreement that setParameters() is the right place for it if we add it.
- The PR: Minor nit-pick editorial issues like it should be marked as a read-only attribute (see other parameters) which implies that it throws InvalidModificationError so we don't need to say that again.
- It seems to make sense to merge this (after fixing the nit) to the extension spec. I can do that.

jan-ivar:
Mozilla feedback: "seems fine".

Since `ptime` is already a feature at risk, having both in the same extension spec would make sense.

henbos:
Closing and will be merged on extension spec.

PR #2317, opened 2019-10-03
Add text describing *Description consistency

### LINKED ISSUE #1965: Describe consistency model between *{local|remote}*Description attributes
The spec doesn't enforce a particular consistency model between the localDescription, currentLocalDescription, pendingLocalDescription and their remote counterparts.
These descriptions can potentially change over the time of execution of Javascript, meaning that their values could be observed inconsistently by a script.

Suggestion: Specify that all the six attributes are observed from a single consistent snapshot. (In theory, we could specify that this is observed at the time of script startup, but I think that increases overhead for functions that don't observe these attributes, so I think we should leave the exact time of observation implementation-defined.)

CC @foolip

alvestrand:
Fixes #1965

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2317.html" title="Last updated on Oct 17, 2019, 11:50 AM UTC (3eb3760)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2317/53a2dd0...3eb3760.html" title="Last updated on Oct 17, 2019, 11:50 AM UTC (3eb3760)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1031,7 +1031,10 @@
         representation of an <code><a>RTCIceTransport</a></code> changes, as
         described in <a href= "#rtcicetransport"></a>.</p>
 
-        <p data-link-for="RTCPeerConnection">The state of the SDP
+        <p>The task source for the tasks listed in this section is the
+        <a>networking task source</a>.</p>
+
+        <p class="note" data-link-for="RTCPeerConnection">The state of the SDP
           negotiation is represented by
         the <a>connection state</a> and the internal
         variables <a>[[\CurrentLocalDescription]]</a>,
@@ -1043,10 +1046,7 @@
         the <a>surface a candidate</a> procedure. In each case, all
         the modifications to all the five variables are completed
         before the procedures fire any events or invoke any callbacks,
-        so the modifications are made visible as a single event.</p>
-
-        <p>The task source for the tasks listed in this section is the
-        <a>networking task source</a>.</p>
+        so the modifications are made visible at a single point in time.</p>
 
         <section>
         <h4>Constructor</h4>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1031,6 +1031,20 @@ <h4>Operation</h4>
...
+        variables <a>[[\CurrentLocalDescription]]</a>,
+        <a>[[\CurrentRemoteDescription]]</a>, <a>[[\PendingLocalDescription]]</a>
+        and <a>[[\PendingRemoteDescription]]</a>. These are only set
+        inside the <code><a>setLocalDescription</a></code>
+        and <code><a>setRemoteDescription</a></code> operations, and
+        modified by the <code><a>addIceCandidate</a></code> operation and
+        the <a>surface a candidate</a> procedure. In each case, all
+        the modifications to all the five variables are completed
+        before the procedures fire any events or invoke any callbacks,
+        so the modifications are made visible as a single event.</p>
```
henbos: Considering we use internal slots, and are clear about when these are updated, I'm not sure there is any ambiguity with these attributes. But it doesn't hurt to have a note about the intent in a place like this.

henbos:
Editors can integrate % rephrasing

webrtc.html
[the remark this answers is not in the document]
alvestrand: I'll go with "point in time". Another problem is that I didn't have the right word for the entity that it's made visible to (the Javascript context that owns the variable), but I hope implying what it is will be OK.

alvestrand: I don't think there's any ambiguity either, this text is merely saying what we are aiming for.

Issue #2257, opened 2019-08-06
Consider making RTCCertificate throw when serialized when _forStorage_ is false


annevk:
This would make `postMessage()` and friends throw and only allow same-origin usage of the object as seems to be intended. This would allow you to get rid of the `[[Origin]]` field and prevent inadvertent sharing of this object with untrusted origins (who might be able to Spectre-attack its contents at that point, unless the user agent has a more convoluted design whereby it only messages a pointer around and keeps the actual underlying data only in processes accessible to the origin that created it).

dontcallmedom:
[TPAC discussion](https://www.w3.org/2019/09/18-webrtc-minutes.html#item06) points toward keeping the capability to use `postMessage`, maybe limiting it to same-origin communication and calling attention to security risks.

jan-ivar:
@annevk Anything else we can do here to somehow limit postMessage use to same origin?

annevk:
I have an idea for a design inspired by you at https://github.com/whatwg/html/issues/4939. I'd like to know if others are somewhat okay with it before writing the text though. I can file a new issue once that's in place.

PR #2333, opened 2019-10-17
Strengthen secure module language around keying material & add security note

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2343. _(Updated. Previous description:)_

From TPAC decision, add security note for https://github.com/w3c/webrtc-pc/issues/2257. cc @annevk

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2333.html" title="Last updated on Dec 5, 2019, 2:02 PM UTC (9352a24)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2333/48c8250...jan-ivar:9352a24.html" title="Last updated on Dec 5, 2019, 2:02 PM UTC (9352a24)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -12936,6 +12936,11 @@ interface RTCErrorEvent : Event {
       <p>A mechanism, <code><a>peerIdentity</a></code>, is provided that gives
       Javascript the option of requesting media that the same javascript cannot
       access, but can only be sent to certain other entities.</p>
+      <p>Communication certificates may be opaquely shared with
+      <code>postMessage</code> in anticipation of future needs. User agents are
+      encouraged to not directly hold private keying material in these shared
+      objects, or limit sharing to the same origin, to reduce memory attack
+      surface.</p>
     </section>
     <section>
       <h2>Persistent information exposed by WebRTC</h2>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5333,8 +5333,13 @@ interface RTCPeerConnectionIceErrorEvent : Event {
                       <a href="https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-origin">origin</a>.</p>
                     </li>
                     <li>
-                      <p>Set <var>certificate</var>.[[\KeyingMaterial]] to
-                      <var>generatedKeyingMaterial</var>.</p>
+                      <p>Store the <var>generatedKeyingMaterial</var> in a
+                      secure module, and let <var>handle</var> be a reference
+                      identifier to it.</p>
+                    </li>
+                    <li>
+                      <p>Set <var>certificate</var>.[[\KeyingMaterialHandle]] to
+                      <var>handle</var>.</p>
                     </li>
                     <li>
                       <p>Set <var>certificate</var>.[[\Certificate]] to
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -12936,6 +12936,11 @@ <h2>Confidentiality of Communications</h2>
       <p>A mechanism, <code><a>peerIdentity</a></code>, is provided that gives
       Javascript the option of requesting media that the same javascript cannot
       access, but can only be sent to certain other entities.</p>
+      <p>Communication certificates may be opaquely shared with
+      <code>postMessage</code> in anticipation of future needs. User agents are
```
alvestrand: Nit: Shouldn't it be "using PostMessage"? I don't think we're sharing it with the message, we're using the message to share it.

webrtc.html
```diff
@@ -12936,6 +12936,11 @@ <h2>Confidentiality of Communications</h2>
...
+      encouraged to not directly hold private keying material in these shared
+      objects, or limit sharing to the same origin, to reduce memory attack
+      surface.</p>
```
alvestrand: If a memory-reading attack is possible, I'm not sure what an extra level of indirection gives you - pointers are made to be followed.
Won't an inconsistent limitation of sharing to the same origin mean that apps that want to share between origins will work in some browsers and not in others? Isn't that bad enough that we should either say MUST or drop the "limit sharing" language?

Should we mention that certificates can only be *used* in the origin context?

annevk:
Is this basically a note about the problem outlined at https://github.com/web-platform-tests/wpt/pull/19572#issue-325847359? That aside from serializing and deserializing we also need to have a hook when a serialization is about to cross a boundary?

webrtc.html
jan-ivar: > I'm not sure what an extra level of indirection gives you - pointers are made to be followed.

This is in reference to [existing language](http://w3c.github.io/webrtc-pc/#dom-rtccertificate-getfingerprints): *"In implementations where an RTCCertificate might not directly hold private keying material (it might be stored in a secure module), a reference to the private key can be held in the [[KeyingMaterial]] internal slot, allowing the private key to be stored and used."*

> Should we mention that certificates can only be used in the origin context?

We [already do](http://w3c.github.io/webrtc-pc/#h-note-31): *"the object cannot be used by any other origin than the one that originally created it."*

> Won't an inconsistent limitation of sharing to the same origin mean that apps that want to share between origins will work in some browsers and not in others? Isn't that bad enough that we should either say MUST or drop the "limit sharing" language?

I don't think it's that bad. You'd have to share it to a different origin and back again, basically. Who does that?

Yes I think we'd like to limit this normatively, but we're lacking a good way currently. I believe https://github.com/whatwg/html/issues/4939 was filed to fix that. Until then, we have this note. Happy to take suggestions.

henbos:
Ping! This has been inactive for a while

jan-ivar:
Updated prose to more explicitly suggest some form of secure module indirection (preferably a separate process, though I doubt we can do any stronger than SHOULD on that since some implementations may not have multiple processes).

@annevk does this look good to you? (sorry I'm not able to parse your last question in this thread).  I based this off your comment in https://github.com/w3c/webrtc-pc/issues/2343#issuecomment-549715469.

webrtc.html
jan-ivar: Clarified language to be about using a different process, and gave up trying to enforce sameorigin, based on https://github.com/w3c/webrtc-pc/issues/2343#issuecomment-549715469.

PR #2356, opened 2019-11-08
Add hardwareAccelerated field to RTCRtpCodecCapability

### LINKED ISSUE #2355: Add hardwareAccelerated field to RTCRtpCodecCapability
It is sometimes beneficial to be able to tell beforehand if a particular codec has support for hardware acceleration. Currently, the only way to determine if there is hardware acceleration is to start a stream that is using the particular codec and look at the implementation name in getStats().

By knowing beforehand if there is hardware support, a client could use this information to signal back to the server that it should avoid/prefer a particular codec. This is of special importance for high bitrate streams where it's likely that a client will not be able to decode the stream in full fps if there is no support for hardware acceleration.

drkron:
Fix #2355

***
<a href="https://pr-preview.s3.amazonaws.com/drkron/webrtc-pc/pull/2356.html" title="Last updated on Nov 8, 2019, 12:33 PM UTC (44266d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2356/9d6b85d...drkron:44266d7.html" title="Last updated on Nov 8, 2019, 12:33 PM UTC (44266d7)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7349,6 +7349,7 @@ async function updateParameters() {
   required unsigned long clockRate;
   unsigned short channels;
   DOMString sdpFmtpLine;
+  boolean hardwareAccelerated;
 };</pre>
         <section>
           <h2>Dictionary <code><a>RTCRtpCodecCapability</a></code> Members </h2>
@@ -7384,6 +7385,12 @@ async function updateParameters() {
             <dd>
               <p>The "format specific parameters" field from the "a=fmtp" line
               in the SDP corresponding to the codec, if one exists.</p>
+            </dd>
+            <dt data-tests="RTCRtpReceiver-getCapabilities.html"><dfn data-idl><code>hardwareAccelerated</code></dfn> of type <span class=
+            "idlMemberType">boolean</span></dt>
+            <dd>
+              <p>If present, indicates whether there is support for hardware
+              acceleration for the specified codec.</p>
             </dd>
           </dl>
         </section>


aboba:
@snyderp Can you review?

snyderp:
@aboba at first read, this would seem to be a way for sites to pierce the isolation between the browser, and device hardware, and so adds new fingerprinting surface.  Is this information (implicitly or otherwise) gated behind a permission somewhere earlier in the flow that I'm missing?

alvestrand:
It's returned from a static function: https://w3c.github.io/webrtc-pc/#dom-rtcrtpreceiver-getcapabilities

No gating on that function so far.

drkron:
> @aboba at first read, this would seem to be a way for sites to pierce the isolation between the browser, and device hardware, and so adds new fingerprinting surface. Is this information (implicitly or otherwise) gated behind a permission somewhere earlier in the flow that I'm missing?

I'm not sure if I see how this adds something that is not already implicitly known.

**In WebRTC**
In the Chromium implementation of WebRTC there are three included SW decoders - H264, VP8, and VP9. In addition to these codecs, there might be support for some additional codecs/profiles through external HW decoders. So already with the existing specification (without this field), any other codec than the three SW codecs would imply that this is a codec with only HW decoder support. The additional information that is added by this field would be if there's HW support for the three native SW decoders.

However, as soon as the stream starts, the implementation name would reveal this anyway. 

**In Chrome**
If we take a look outside WebRTC, there are at least two other ways of retrieving similar or even more detailed information. 1) Through the MediaCapabilities API which returns supported, smooth, and powerEfficient for any codec configuration that is requested. 2) Through WebGL which provides "Unmasked renderer" which is essentially the name of the graphics card chip (see https://webglreport.com/).

henbos:
Whether an implementation is HW or SW is incredibly valuable in order to ensure optimal user experience. Better APIs for codec selection are highly requested features by real world applications.

This information is already available in one way or another as per @drkron's comment, but obtaining it requires jumping through a couple of hoops. Having this available in the codec capability dictionary seems like it would be a good thing.

> However, as soon as the stream starts, the implementation name would reveal this anyway.

E.g. the implementation name is revealed in RTCPeerConnection.getStats()'s [encoderImplementation](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-encoderimplementation) and [decoderImplementation](https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-decoderimplementation) strings (which are exposed without special permissions).

snyderp:
> However, as soon as the stream starts, the implementation name would reveal this anyway.

> Whether an implementation is HW or SW is incredibly valuable in order to ensure optimal user experience. Better APIs for codec selection are highly requested features by real world applications.

I'm not doubting that this would be useful information for some parties to have at some point.  But learning it from a stream would require permission be given; it doesn't require a permission to learn in this route (if im following correctly).

Put differently, if it reveals something about the user's hardware beyond what could be learned by knowing the browser version and OS, then it needs to be gated some how, to keep it out of the hands of passive finger printers.

> the implementation name is revealed in RTCPeerConnection.getStats()

I believe that is being addressed here: https://github.com/w3c/webrtc-svc/issues/22

drkron:
Thanks for the comments!

As far as I know, there's no permission needed to start a stream. Permission is only needed to access the camera/microphone. A test stream could be started in the background without any permission. The client could then determine if the decoding is HW accelerated or not, either by looking at implementation name, or if that's removed, by looking at decode times.

In the current context with the existing MediaCapabilities API, WebGL, and WebRTC stats, I don't see how this adds any new information to be used for fingerprinting. It just packages it in a format that is more convenient for web creators using WebRTC.

Instead of gating each call to a function that could be used for fingerprinting, I think that the concept of a [privacy budget](https://github.com/bslassey/privacy-budget) is much more appealing.

snyderp:
> In the current context with the existing MediaCapabilities API, WebGL, and WebRTC stats, I don't see how this adds any new information to be used for fingerprinting. It just packages it in a format that is more convenient for web creators using WebRTC.

I don't understand the claim here.  Is this value available elsewhere?  If so, then you can expect at some point we will try to gate it, under the broader goal of isolating pages from hardware capabilities w/o specific opt-in / permission / similar

> Instead of gating each call to a function that could be used for fingerprinting, I think that the concept of a privacy budget is much more appealing.

1) Privacy Budget in general does not have sufficient detail to evaluate as a finger printing defense
2) I (and many others on PING) are extremely skeptical that the idea of privacy budgets are a useful, practical defense against fingerprinting (feel free to [read here for more details if you would like](https://brave.com/brave-fingerprinting-and-privacy-budgets/), where i've summarized.  The relevant part is the "Dynamic Privacy Budgets, And Where They Fail" section, not the Brave parts
3) even if there was a privacy budget situation, we would still need to be extremely cautious about what new budget-exhausting behavior was on the platform. "Privacy budget" doesn't mean we don't try to remove FP surface where it exists

drkron:
> > In the current context with the existing MediaCapabilities API, WebGL, and WebRTC stats, I don't see how this adds any new information to be used for fingerprinting. It just packages it in a format that is more convenient for web creators using WebRTC.
> 
> I don't understand the claim here. Is this value available elsewhere? If so, then you can expect at some point we will try to gate it, under the broader goal of isolating pages from hardware capabilities w/o specific opt-in / permission / similar
> 

This was written with this comment in mind:
https://github.com/w3c/webrtc-pc/pull/2356#issuecomment-552409803

**In summary:**
- Through the MediaCapabilities API it's possible to deduce if there's HW acceleration support for a particular codec.
- WebGL provides the name of the GPU chipset, which contains even more information.
- Once a WebRTC stream is started (which does not require any permission unless the camera/microphone is activated), it's possible to deduce if there's HW acceleration support by looking at the implementation name in getStats or by simply looking at decode times.

henbos:
Would it make more sense to have a promise-based `RTCRtpSender.getEncoderCapabilities()`/`RTCRtpReceiver.getDecoderCapabilities()` (or `RTCPeerConnection.getCodingCapabilities()`?), allowing a user agent to implement a user permissions prompt, should mitigating privacy concerns be needed and allowing more sensitive information in the future?

alvestrand:
Editors recommend moving the PR to the "small extensins" spec (presently https://github.com/henbos/webrtc-extensions), given the feature freeze.

drkron:
> Would it make more sense to have a promise-based `RTCRtpSender.getEncoderCapabilities()`/`RTCRtpReceiver.getDecoderCapabilities()` (or `RTCPeerConnection.getCodingCapabilities()`?), allowing a user agent to implement a user permissions prompt, should mitigating privacy concerns be needed and allowing more sensitive information in the future?

A question here before I continue the process at the extensions spec.

Would a promise-based function exist in parallel to the already existing functions? I guess it's too late to change the existing functions?

henbos:
I think it's too late to change the existing function, so there would have to be two methods.

snyderp:
I do not consider the privacy relevant aspects of this addressed.  Is there some other place to continue this?

> Through the MediaCapabilities API it's possible to deduce if there's HW acceleration support for a particular codec.
> WebGL provides the name of the GPU chipset, which contains even more information.
> Once a WebRTC stream is started (which does not require any permission unless the camera/microphone is activated), it's possible to deduce if there's HW acceleration support by looking at the implementation name in getStats or by simply looking at decode times.

These are all very much issues we are trying to address; the privacy concerning aspects of this proposal should be written expecting those issues will be addressed.  Simply moving this functionality to another spec does not address the problem

drkron:
> These are all very much issues we are trying to address; the privacy concerning aspects of this proposal should be written expecting those issues will be addressed. Simply moving this functionality to another spec does not address the problem

I think that what's exposed through MediaCapabilities API and WebGL is the biggest concern here, however they are outside of the WebRTC specification and cannot be expected to be solved within this forum.

Starting a stream just to determine if there's HW acceleration support for that particular codec is a bit far fetched, especially since the methods mentioned above are available.

I think that a good trade off right now is to keep the spec as is (unless there's a clear evidence that something is being misused) and be careful when adding new stuff.

henbos:
> I do not consider the privacy relevant aspects of this addressed. Is there some other place to continue this?

I filed https://github.com/henbos/webrtc-extensions/issues/21.

> Simply moving this functionality to another spec does not address the problem.

The primary reason for moving this is the Feature Freeze on webrtc-pc, let's continue the privacy discussions there.

Issue #2332, opened 2019-10-17
Explicit rollback steps (was: Firing the right events in "rollback": Which description to use?)


henbos:
In https://github.com/w3c/webrtc-pc/pull/2298/files the following was added:

> If connection.[[PendingLocalDescription]] is not null, then set description to connection. [[CurrentLocalDescription]], and set remote to false. Otherwise, set description to  connection. [[CurrentRemoteDescription]], and set remote to true.

Let's figure out if this is the right logic.

jan-ivar:
I think this is the right logic. This *description* value is never set. It's used in the remainder of the algorithm solely to undo events fired and tracks moved based on *remote* descriptions.

The most utilitarian way to undo these SRD actions (offer or answer, doesn't matter), is to re-apply the previous remote description if there is one—as it contains just the right the right remote-controlling SDP ingredients to do—or if there isn't, one mimic an empty description which we do as a special case just below. This keeps the algorithm from ballooning to twice the size.

henbos:
@jan-ivar Please take a look at this comment and let me know if I've discovered a bug or if I am confused again.

> I think this is the right logic. This description value is never set. It's used in the remainder of the algorithm solely to undo events fired and tracks moved based on remote descriptions.

We're on the same page here that this is the intent.

However, if you look at the steps, it modifies stuff like the [[Receptive]] slot and transports too. Which I suppose is only appropriate, e.g. early media use case. Mids also get assigned, but those are immutable between offer and answer, so that is not a concern, whichever SDP we get would be useful. But anything that could change between offer and answer could potentially get reverted to the "have-local/remote-offer" state rather than the previous stable state?

The [[Receptive]] one could be a big concern, if that value is wrong we could be in a state where a track that should unmute doesn't or a track that shouldn't unmute does based on packets on the wire.

SLD(offer) might have created bundle transports, which at SRD(answer) disappears.
In a follow up O/A that gets rolled back, setting _description_ to [[CurrentLocalDescription]] might cause transports to change not to the transports as they were at "stable" (e.g. SRD(answer)), but the transports as they were at "have-local-offer" (e.g. SRD(offer)).

The direction might be different... e.g. in stable we could be receiving and an offer to stop receiving (mute the track) comes in, then we roll that back: which direction do we use to make sure ontrack/onmute behaves correctly?

> The most utilitarian way to undo these SRD actions (offer or answer, doesn't matter), is to re-apply the previous remote description if there is one [...]

[...]

PR #2372, opened 2019-11-26
Revert transports on rollback

### LINKED ISSUE #2363: Transports on offer need to be rolled back
Follow-up issue to https://github.com/w3c/webrtc-pc/issues/2338.

When we roll back, we should say to null the internal slots for the transports if they were rolled back.

henbos:
Fixes #2363.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2372.html" title="Last updated on Nov 26, 2019, 3:01 PM UTC (fc31c20)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2372/00c4a67...henbos:fc31c20.html" title="Last updated on Nov 26, 2019, 3:01 PM UTC (fc31c20)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1269,6 +1269,23 @@
                     <code>true</code>, then abort these steps.</p>
                   </li>
                   <li>
+                    <p>If <var>description</var> is of type <code>"offer"</code>
+                    and the <a>signaling state</a> of <var>connection</var> is
+                    <code>"stable"</code> then for each <var>transceiver</var>
+                    in <var>connection</var>'s <a href="#transceivers-set">set
+                    of transceivers</a>, run the following steps:</p>
+                    <ol>
+                      <li>
+                        <p>Set <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\StableStateSenderTransport]]</a>
+                        to <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SenderTransport]]</a>.</p>
+                      </li>
+                      <li>
+                        <p>Set <var>transceiver</var>.<a>[[\Receiver]]</a>.<a>[[\StableStateReceiverTransport]]</a>
+                        to <var>transceiver</var>.<a>[[\Receiver]]</a>.<a>[[\ReceiverTransport]]</a>.</p>
+                      </li>
+                    </ol>
+                  </li>
+                  <li>
                     <p>If <var>remote</var> is <code>false</code>, then run one
                     of the following steps:</p>
                     <ol>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1853,6 +1870,27 @@ <h4>Set the RTCSessionDescription</h4>
                       </li>
                     </ol>
                   </li>
+                  <li>
+                    <p>If we are processing a rollback then for each
```
henbos: Note: I used the language "If we are processing a rollback" rather than 'If description is of type "rollback"' because until we fix #2332 by explicitly making rollback a separate part of steps, in this part of the algorithm the description variable has already been overwritten (e.g. it's type is not "rollback" even though we are processing a rollback). When all rollback steps are processed in a single place - I suggest here - then we can change this to 'If description is of type "rollback"'.

webrtc.html
```diff
@@ -1862,6 +1900,15 @@ <h4>Set the RTCSessionDescription</h4>
                         <a href="#transceivers-set">set of transceivers</a> run the
                         following steps:</p>
                         <ol>
+                          <li>
+                            <p>Set <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\StableStateSenderTransport]]</a>
+                            to <code>null</code>.
```
henbos: Resetting these internal slots is not strictly necessary: When we are in the stable state it doesn't matter what they are since they are only read from when we are performing rollback, at which point they have already been set (when applying the offer).

But conceptually these slots are not used unless we are performing an O/A exchange so I thought I would clean it up in all code paths where we return to stable state.

If you think this is unnecessary let me know and I'll remove these steps. (They don't have to be nulled above either.)

henbos:
Possible merge conflicts with #2384

webrtc.html
henbos: PR #2384 fixes this, when rebasing these steps should be part of the steps introduced in that PR.

webrtc.html
jan-ivar: Sorry for the delay in looking at these. I see this has already merged, but decided to answer here. Maybe we can clean this up in https://github.com/w3c/webrtc-pc/pull/2384.

I ran across this new language while reviewing https://github.com/w3c/webrtc-pc/pull/2384 and admit I got a bit confused by it. We have so many different states (current/pending), and at first I didn't get the narrow scope for these new ones, and found it hard to answer e.g. why we're still returning *SenderTransport* and not *StableStateSenderTransport* in getters, but then I realized these are *backup* slots, which helps understand their narrow scope.

I recommend the following:
1. Since these are backup slots, s/`StableStateSenderTransport`/`LastStableSenderTransport`/
2. Since the algorithm is getting long, skip setting them to `null`.

henbos: Thanks. To be addressed in #2384.

PR #2373, opened 2019-11-26
Revert [[ReceiveCodecs]] on rollback

### LINKED ISSUE #2368: Clarify what happens to codecs when rolling back
Spin-off from [here](https://github.com/w3c/webrtc-pc/issues/2324#issuecomment-558127202):

> What happens if you swap out codecs? Early media-reasoning would suggest that codecs are applied at SLD(offer) rather than SRD(answer). This implies that existing packets on the wire with an old codec would get dropped while have-local-offer. If offers were proposed to be changed and we roll back, then what we need to do on rollback is the revert codec settings.
> This assumes we don't do anything special for smooth transition when changing codecs.

henbos:
Fixes #2368.

This PR rests upon PR #2372.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2373.html" title="Last updated on Nov 26, 2019, 3:23 PM UTC (cd06757)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2373/00c4a67...henbos:cd06757.html" title="Last updated on Nov 26, 2019, 3:23 PM UTC (cd06757)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1283,6 +1283,10 @@
                         <p>Set <var>transceiver</var>.<a>[[\Receiver]]</a>.<a>[[\StableStateReceiverTransport]]</a>
                         to <var>transceiver</var>.<a>[[\Receiver]]</a>.<a>[[\ReceiverTransport]]</a>.</p>
                       </li>
+                      <li>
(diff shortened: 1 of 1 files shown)


henbos:
Possible merge conflicts with #2384

PR #2375, opened 2019-11-26
Construct RTCErrors with name 'OperationError' instead of 'RTCError'

### LINKED ISSUE #2330: Backwards compat concerns with RTCError
Since no browsers implement RTCError yet, concerns were raised about backwards 
compatibility with implementing it.

`RTCError` is both a *type* (with extra attributes), as well as the name of the error.

One idea to mitigate compat concerns would be to specify the error names browsers are using today. [E.g.](https://jsfiddle.net/jib1/r0vL78gw/):
```js
await pc.setRemoteDescription({type: "offer", sdp: "foo"}); // OperationError
```
On the upside it solves compat. On the downside, JS that cares about the attributes would need to check for them.

henbos:
Fixes #2330.

Before merging this PR, we need to check if this is in fact consistent with current implementations.
So far, I've noticed that there is one breaking change for Chrome's replaceTrack() which rejects with InvalidModificationError (though I doubt anybody is relying on that operation failing, it only fails on invalid input or when the PC is closed).

I/we need to take a closer look at Chrome and other browsers. (If different names are needed the PR can be updated to allow constructing with different names.)

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2375.html" title="Last updated on Nov 26, 2019, 5:44 PM UTC (73fd734)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2375/00c4a67...henbos:73fd734.html" title="Last updated on Nov 26, 2019, 5:44 PM UTC (73fd734)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -11585,7 +11585,7 @@ interface RTCError : DOMException {
                 <p>Invoke the <code><a data-cite="!WEBIDL#idl-DOMException">
                 DOMException</a></code> constructor of <var>e</var> with the
                 <code>message</code> argument set to <var>message</var> and the
-                <code>name</code> argument set to <code>"RTCError"</code>.</p>
+                <code>name</code> argument set to <code>"OperationError"</code>.</p>
                 <p class="note">This name does not have a mapping to a legacy
                 code so <var>e</var>'s <code>code</code> attribute will return
                 0.</p>


henbos:
If people can help and try to find out what error names can be present that would be helpful. Otherwise I think I'll create a codepen that makes all the operations that use RTCError fail that we can run on all browsers.

alvestrand:
@henbos will look at Chrome. @youennf will look at Safari.

PR #2385, opened 2019-11-28
Note that getSynchronizationSources works without sinks

### LINKED ISSUE #2240: Clarify that getSynchronizationSources() should return information even if the track has no sink (<video> tag)
When [Chrome tried to update its getSynchronizationSources() implementation](https://bugs.chromium.org/p/webrtc/issues/detail?id=10545#c14) to match [the spec](https://w3c.github.io/webrtc-pc/#dom-rtcrtpreceiver-getsynchronizationsources) about when the information is updated...

> When an audio or video frame from one or more RTP packets is delivered to the RTCRtpReceiver's MediaStreamTrack, the user agent MUST queue a task to update the relevant information for the RTCRtpContributingSource and RTCRtpSynchronizationSource dictionaries based on the content of those packets. 

... it started failing the [getSynchronizationSources() tests](https://cs.chromium.org/chromium/src/third_party/blink/web_tests/external/wpt/webrtc/RTCRtpReceiver-getSynchronizationSources.https.html) because without the MediaStreamTrack having a sink, Chrome was performing optimizations not to decode frames, and so it wasn't updating the SSRC information.

The question is: Is this spec compliant?

> As stated in the conformance section, requirements phrased as algorithms may be implemented in any manner so long as the end result is equivalent. So, an implementation does not need to literally queue a task for every frame, as long as the end result is that within a single event loop task execution, all returned RTCRtpSynchronizationSource and RTCRtpContributingSource dictionaries for a particular RTCRtpReceiver contain information from a single point in the RTP stream.

I would expect the MediaStreamTrack to, at least conceptually, receive frames as long as we are receiving decodable packets, whether or not the track forwards those frames to a sink such as a &lt;video&gt; tag.

On the other hand, why decode packets if we're not using them?

So...
1. Should we add a note to the spec saying "if the MediaStreamTrack does not have any sinks, this step MAY be optimized away, so that the SSRC information is not updated"?
2. Or should getSSRC be clarified to make the packet information mandatory even if there are no sinks and no decoding happening?
3. Or should decoding be mandatory, leaving no ambiguity here?

henbos:
Fixes #2240

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2385.html" title="Last updated on Nov 28, 2019, 2:50 PM UTC (dbf600e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2385/00c4a67...henbos:dbf600e.html" title="Last updated on Nov 28, 2019, 2:50 PM UTC (dbf600e)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7197,6 +7197,10 @@ interface RTCRtpReceiver {
       in order of ascending RTP timestamps. The user agent MUST keep information
       from RTP packets delivered to the <code><a>RTCRtpReceiver</a></code>'s
       <code><a>MediaStreamTrack</a></code> in the previous 10 seconds.</p>
+      <div class="note">Even if the <a>MediaStreamTrack</a> is not attached to
+      any sink for playout, <code>getSynchronizationSources</code> and
+      <code>getContributingSources</code> returns up-to-date information; sinks
+      are not a prerequisite for decoding RTP packets.</div>
       <div class="note">As stated in the <a href="#conformance">conformance
       section</a>, requirements phrased as algorithms may be implemented in
       any manner so long as the end result is equivalent. So, an
--- a/webrtc.html
+++ b/webrtc.html
@@ -7199,8 +7199,9 @@ interface RTCRtpReceiver {
       <code><a>MediaStreamTrack</a></code> in the previous 10 seconds.</p>
       <div class="note">Even if the <a>MediaStreamTrack</a> is not attached to
       any sink for playout, <code>getSynchronizationSources</code> and
-      <code>getContributingSources</code> returns up-to-date information; sinks
-      are not a prerequisite for decoding RTP packets.</div>
+      <code>getContributingSources</code> returns up-to-date information as long
+      as the track is not ended; sinks are not a prerequisite for decoding RTP
+      packets.</div>
       <div class="note">As stated in the <a href="#conformance">conformance
       section</a>, requirements phrased as algorithms may be implemented in
       any manner so long as the end result is equivalent. So, an


webrtc.html
```diff
@@ -7197,6 +7197,10 @@ <h2>Methods</h2>
       in order of ascending RTP timestamps. The user agent MUST keep information
       from RTP packets delivered to the <code><a>RTCRtpReceiver</a></code>'s
       <code><a>MediaStreamTrack</a></code> in the previous 10 seconds.</p>
+      <div class="note">Even if the <a>MediaStreamTrack</a> is not attached to
+      any sink for playout, <code>getSynchronizationSources</code> and
+      <code>getContributingSources</code> returns up-to-date information; sinks
+      are not a prerequisite for decoding RTP packets.</div>
```
henbos: Add "as long as the track is not ended"

webrtc.html
```diff
@@ -7197,6 +7197,10 @@ <h2>Methods</h2>
       in order of ascending RTP timestamps. The user agent MUST keep information
       from RTP packets delivered to the <code><a>RTCRtpReceiver</a></code>'s
       <code><a>MediaStreamTrack</a></code> in the previous 10 seconds.</p>
+      <div class="note">Even if the <a>MediaStreamTrack</a> is not attached to
+      any sink for playout, <code>getSynchronizationSources</code> and
+      <code>getContributingSources</code> returns up-to-date information; sinks
```
chxg: Would it make sense to change "returns" to "MUST return"?

henbos: This is a note, I think we're covered by the previous "MUST" above that this note is referring to :)

PR #2389, opened 2019-12-02
Fix a few references

dontcallmedom:
also minor markup improvements

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2389.html" title="Last updated on Dec 2, 2019, 4:41 PM UTC (90105b2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2389/ac24e7f...90105b2.html" title="Last updated on Dec 2, 2019, 4:41 PM UTC (90105b2)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -83,9 +83,9 @@
     <p>The terms <dfn>event</dfn>, <dfn data-cite="!HTML/webappapis.html#events-event-handlers">event
     handlers</dfn> and <dfn data-cite="!HTML/webappapis.html#event-handler-event-type">event
     handler event types</dfn> are defined in [[!HTML]].</p>
-    <p><code><dfn data-cite="!HIGHRES-TIME#dom-performance-timeorigin">performance.timeOrigin</dfn></code>
-    and <code><dfn data-cite="!HIGHRES-TIME#dom-performance-now">performance.now()</dfn></code>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -153,8 +153,8 @@ <h3>Introduction</h3>
       exchange of control messages (called a signaling protocol) over a
       signaling channel which is provided by unspecified means, but generally
       by a script in the page via the server, e.g. using
-      <code>XMLHttpRequest</code> [[XMLHttpRequest]] or Web Sockets
-      [[WEBSOCKETS-API]].</p>
+      <code>XMLHttpRequest</code> [[?XMLHttpRequest]] or Web Sockets
```
henbos: What's the difference between [[Foo]], [[\Foo]], [[!Foo]] and [[?Foo]]?

dontcallmedom: ` [[Foo]]` is a reference whose status (normative vs informative) inherits from the status of its context (normative in a normative section, informative in a non-normative one).
`[[!Foo]]` used to allow to distinguish normative from informative, but I think it no longer is needed (and would create a warning in an informative section).
`[[?Foo]]` allows to force informative references in a normative section.
`[[\Foo]]` is an internal slot named  `Foo`.

Issue #2370, opened 2019-11-26
Add a perfect negotiation example


jan-ivar:
The spec doesn't really discuss [perfect negotiation](https://blog.mozilla.org/webrtc/perfect-negotiation-in-webrtc/), as it's more of a recommended pattern for users than a concern for implementors. But an example would probably be great.

A need to reference this came up around *addIceCandidate* in https://github.com/w3c/webrtc-pc/issues/2367#issuecomment-558309247.

henbos:
If we cover how to use addIceCandidate safely inside the Perfect Negotiation example, I think we can close #2367.

henbos:
Another thing you may need to take into account: getParameters(), rollback, setParameters(). E.g. #2368

PR #2393, opened 2019-12-05
Add Perfect Negotiation example.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2370.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2393.html" title="Last updated on Dec 5, 2019, 3:02 PM UTC (a7af0ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2393/5cbd3ed...jan-ivar:a7af0ee.html" title="Last updated on Dec 5, 2019, 3:02 PM UTC (a7af0ee)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -11658,7 +11658,7 @@ signaling.onmessage = async ({data: {description, candidate}}) => {
       <p>The <code>offerIgnored</code> variable is needed, because
       the <code>RTCPeerConnection</code> object on the impolite side is never
       told about ignored offers. We must therefore suppress errors from
-      incoming candidates belonging to that offer.</p>
+      incoming candidates belonging to such offers.</p>
     </section>
     <!--
     <section>
--- a/webrtc.html
+++ b/webrtc.html
@@ -11594,10 +11594,15 @@ if (sender.dtmf.canInsertDTMF) {
     </section>
     <section>
       <h3>Perfect Negotiation Example</h3>
-      <p>Perfect negotiation is a recommended pattern to abstract away signaling
-      from the rest of an application. It lets applications operate on both
-      peer connection objects at once, on either end of the call, without having
-      to worry about signaling state, racing, or glare.</p>
+      <p>Perfect negotiation is a recommended pattern to manage negotiation
+      transparently, abstracting this asymmetric task away from the rest of an
+      application. This pattern has advantages over other patterns, like one
+      side always being the offerer, as it lets applications operate on both
+      peer connection objects simultaneously without risk of glare (an offer
+      coming in outside of <code>"stable"</code> state). The rest of the
+      application may use any and all modification methods and attributes,
+      without worrying about signaling state races.
+      </p>
       <p>It designates different roles to the two peers, with behavior to
       resolve signaling collisions between them:</p>
       <ol>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -11591,7 +11591,76 @@ <h3>DTMF Example</h3>
...
 }
       </pre>
-    </section><!--
+    </section>
+    <section>
+      <h3>Perfect Negotiation Example</h3>
+      <p>Perfect negotiation is a recommended pattern to abstract away signaling
+      from the rest of an application. It lets applications operate on both
+      peer connection objects at once, on either end of the call, without having
+      to worry about signaling state, racing, or glare.</p>
```
henbos: Can you add a definition (or just another sentence expanding upon) what "glare" means? This is a WebRTC-specific term that is never explained or mentioned in this document. You might also want to add a sentence clarifying what we may be concerned with to be "racing". You wouldn't need to worry about races if one side is always the offerer, for example. This helps clarifying what problems "perfect negotiation" solves.

webrtc.html
```diff
@@ -11591,7 +11591,76 @@ <h3>DTMF Example</h3>
...
+          <p>The <dfn>polite peer</dfn> uses rollback to avoid collision with an
+          incoming offer</p>
+        </li>
+        <li>
+          <p>The <dfn>impolite peer</dfn> ignores an incoming offer when this
+          would collide with its own</p>
+        </li>
+      </ol>
+      <p>Together, they manage signaling for the rest of the application in a
+      manner that doesn't deadlock.</p>
```
henbos: Suggest adding: "Which peer has which role needs to be determined before signaling starts, and is not covered in this example."

webrtc.html
```diff
@@ -11591,7 +11591,76 @@ <h3>DTMF Example</h3>
   console.log('DTMF function not available');
 }
       </pre>
-    </section><!--
+    </section>
+    <section>
+      <h3>Perfect Negotiation Example</h3>
+      <p>Perfect negotiation is a recommended pattern to abstract away signaling
+      from the rest of an application. It lets applications operate on both
```
henbos: I think it's clear to you and I what "abstract away signaling from the rest of the application" means, but while it sounds good to abstract things away, it might not be clear what "abstracting away" means or solves unless you're already familiar with this stuff. Do we need to clarify that this means that "If the application modifies the state of the peer connection in such a way that renegotiation is required, this will happen automatically due to event listeners. In other words, if addTrack() is performed on one endpoint, steps will automatically be taken to make sure ontrack fires at the other endpoint." ?

webrtc.html
```diff
@@ -11591,7 +11591,76 @@ <h3>DTMF Example</h3>
...
+      await pc.setRemoteDescription(description); // SRD rolls back as needed
+      if (description.type == "offer") {
+        await pc.setLocalDescription();
+        signaling.send({description: pc.localDescription});
+      }
+    } else if (candidate) {
+      try {
+        await pc.addIceCandidate(candidate);
+      } catch (e) {
+        if (!offerIgnored) throw e; // Suppress ignored offer's candidates
```
henbos: Are we guaranteed that if we ignore a candidate, the equivalent candidate will be re-generated at the other endpoint and sent again later so that we always end up getting remote candidates? Or is it possible that somehow candidates get lost and we never connect in some weird edge cases?

henbos:
I **strongly** suggest that we add WPTs that exercise Perfect Negotiation. Using these API this way is timing-sensitive and if browsers are not bug-free with regards to the timing of certain things, this pattern does not work.

webrtc.html
jan-ivar: @henbos Note we're not ignoring candidates, only errors from them bouncing.

PR #2401, opened 2019-12-09
Specify that SCTP transport goes to "closed" on DTLS close

### LINKED ISSUE #2360: Should SCTP transport enter "closed" state when DTLS transport does?
The spec says about SCTP transport state "closed":

A task is queued to update the [[SctpTransportState]] slot to "closed" when a SHUTDOWN or ABORT chunk is received or when the SCTP association has been closed intentionally, such as by closing the peer connection or applying a remote description that rejects data or changes the SCTP port.

It is not clear what should happen if the DTLS transport enters "closed" or "failed" state.
I suggest we update the spec to say "In addtion, the [[SctpTransportState]] slot is updated to "closed" when the DTLS transport enters "closed" or "failed" state".

I think queueing a task is the wrong thing here - if changes in DTLS state trigger changes to SCTP state, I think we'd want to have them happen "simultaneously" - the DTLS statechange handler should see the SCTP state as already changed.

alvestrand:
Fixes #2360

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2401.html" title="Last updated on Dec 12, 2019, 5:15 PM UTC (7ccc008)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2401/7fdd9b4...7ccc008.html" title="Last updated on Dec 12, 2019, 5:15 PM UTC (7ccc008)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -9272,11 +9272,22 @@ interface RTCSctpTransport : EventTarget {
                 "idl-def-RTCSctpTransportState.closed">closed</code></dfn></td>
                 <td>
                   <p>A task is queued to update the [[\SctpTransportState]]
-                  slot to <code>"closed"</code> when a SHUTDOWN or ABORT chunk
-                  is received or when the SCTP association has been closed
-                  intentionally, such as by closing the peer connection
-                  or applying a remote description that rejects data or
-                  changes the SCTP port.</p>
+                    slot to <code>"closed"</code> when:
+                    <ul>
+                      <li>a SHUTDOWN or ABORT chunk
+                        is received.</li>
+                      <li>the SCTP association has been closed
+                        intentionally, such as by closing the peer connection
+                        or applying a remote description that rejects data or
+                        changes the SCTP port.</li>
+                      <li>the underlying DTLS association has transitioned
+                        to "closed" state.</li>
+                    </ul>
+                    Note that the last transition is logical due to the
+                    fact that [[RFC8261]] section 6.1 specifies that
+                    SCTP over DTLS is single-homed, so no possibility
+                    of switching to an alternate transport exist.
+                  </p>
                 </td>
               </tr>
             </tbody>
--- a/webrtc.html
+++ b/webrtc.html
@@ -9273,16 +9273,18 @@ interface RTCSctpTransport : EventTarget {
                 <td>
                   <p>A task is queued to update the [[\SctpTransportState]]
                     slot to <code>"closed"</code> when:
-                    <ul>
-                      <li>a SHUTDOWN or ABORT chunk
-                        is received.</li>
-                      <li>the SCTP association has been closed
-                        intentionally, such as by closing the peer connection
-                        or applying a remote description that rejects data or
-                        changes the SCTP port.</li>
-                      <li>the underlying DTLS association has transitioned
-                        to "closed" state.</li>
-                    </ul>
+                    </p>
+                  <ul>
+                    <li>a SHUTDOWN or ABORT chunk
+                      is received.</li>
+                    <li>the SCTP association has been closed
+                      intentionally, such as by closing the peer connection
+                      or applying a remote description that rejects data or
+                      changes the SCTP port.</li>
+                    <li>the underlying DTLS association has transitioned
+                      to "closed" state.</li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -9272,11 +9272,22 @@ <h3><dfn>RTCSctpTransportState</dfn> Enum</h3>
                 "idl-def-RTCSctpTransportState.closed">closed</code></dfn></td>
                 <td>
                   <p>A task is queued to update the [[\SctpTransportState]]
-                  slot to <code>"closed"</code> when a SHUTDOWN or ABORT chunk
-                  is received or when the SCTP association has been closed
-                  intentionally, such as by closing the peer connection
-                  or applying a remote description that rejects data or
-                  changes the SCTP port.</p>
+                    slot to <code>"closed"</code> when:
```
dontcallmedom: add `</p>` (`<ul>` cannot be embedded in a `<p>`)

webrtc.html
```diff
@@ -9272,11 +9272,22 @@ <h3><dfn>RTCSctpTransportState</dfn> Enum</h3>
...
+                      <li>a SHUTDOWN or ABORT chunk
+                        is received.</li>
+                      <li>the SCTP association has been closed
+                        intentionally, such as by closing the peer connection
+                        or applying a remote description that rejects data or
+                        changes the SCTP port.</li>
+                      <li>the underlying DTLS association has transitioned
+                        to "closed" state.</li>
+                    </ul>
+                    Note that the last transition is logical due to the
```
dontcallmedom: add `<p>`

webrtc.html
```diff
@@ -9272,11 +9272,22 @@ <h3><dfn>RTCSctpTransportState</dfn> Enum</h3>
...
+                    fact that [[RFC8261]] section 6.1 specifies that
+                    SCTP over DTLS is single-homed, so no possibility
+                    of switching to an alternate transport exist.
```
lgrahl: Really a nit-pick: I think the authors of that document where more referring to SCTPs multi-homed capability which cannot be used in such a scenario. There's no technical reason why SCTP over DTLS cannot move to whatever transport is desired on the fly. /cc @tuexen

I agree that SCTP should be closed when DTLS closes but I don't entirely agree with the reason. There was another RFC you've quoted and I'd suggest using that instead. Quoting:

> The particular situation I'm staring at now involves incoming error alerts according to RFC 8446 (TLS 1.3) section 6.2 - the RFC says that "Upon transmission or receipt of a fatal alert message, both parties MUST immediately close the connection."
> 
> I think section 6.2.5 (quoted above) should be invoked immediately when there is a DTLS error alert. The peer endpoint has just become unreachable.

If that's stretching it too much, then we could consider to just say that it makes sense because of its tight integration into the peer connection instance. :slightly_smiling_face:

alvestrand: The reason there's a technical reason why SCTP over DTLS can't move is that WebRTC 1.0 specifies the use of RFC 8261, no extensions. And that one specifies that it's single homed.
We could let it move, but that would require writing an extension spec.

lgrahl: In SCTP over DTLS, SCTP is communicating over a virtual IP transport and doesn't have access to IP addresses. For me, that is the reason why they're saying it's *single-homed* (in SCTP terminology) in that RFC. Not because it can only live on top of that specific DTLS connection. But that is my interpretation of the word *single-homed*.
The RFC also doesn't explicitly say that SCTP assocs cannot move to another transport or what needs to happen when the DTLS connection is gone.

All I'm saying is: We might want to use our own words here to reason this behaviour instead of referring to the RFC alone. A to-be-smoothed suggestion:

> Note that the last transition is logical due to the fact that an SCTP association requires an established DTLS connection and that switching to an alternate transport has not been defined for this API.

rshpount: I agree with this. DTLS already runs on top of ICE, so the ICE transport layer should deal with actual address changes. SCTP is essentially twice removed from the actual transport and transport addresses.

alvestrand: I adopted the tweak, merged with the language I already had. It should be clear enough that the reader / implementor knows what to do.

PR #2402, opened 2019-12-09
Reject RID renegotiation

### LINKED ISSUE #2314: Can simulcast offers renegotiate rids?
We appear to allow remote simulcast offers to renegotiate rids and number of layers:
> *If the description is of type "offer" and contains a request to receive simulcast, use the order of the rid values specified in the simulcast attribute to create an RTCRtpEncodingParameters dictionary for each of the simulcast layers, populating the rid member according to the corresponding rid value, and let sendEncodings be the list containing the created dictionaries. Otherwise, let sendEncodings be an empty list.*

This differs from the local side, and conflicts with [5.4.1.](https://w3c.github.io/webrtc-pc/#simulcast-functionality):
> *The addTransceiver method establishes the simulcast envelope which includes the maximum number of simulcast streams that can be sent, as well as the ordering of the encodings* ... Once the envelope is determined, layers cannot be removed. 

If we allow the rids to be renegotiated (say, we start with 'low', 'med', and 'high', and then a renegotiation happens that changes to rids 'tall', 'grande', and 'venti'), then we have a referential integrity problem with payloadType. See https://github.com/w3c/webrtc-pc/issues/2313.

cc @docfaraday

alvestrand:
Fixes #2314 

Note: This is the first time we reject a SRD in the text section that comes after applying it successfully in JSEP. I'm not sure this is the right way to specify that, but doing it before JSEP gets weird too (repeated "look up this transceiver" steps). Suggestions welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2402.html" title="Last updated on Dec 12, 2019, 2:20 PM UTC (213fa33)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2402/7fdd9b4...213fa33.html" title="Last updated on Dec 12, 2019, 2:20 PM UTC (213fa33)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1593,6 +1593,15 @@
                             description</a>.</p>
                           </li>
                           <li>
+                            <p>If a suitable transceiver was found, and
+                              <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SendEncodings]]</a>
+                              is non-empty, and not equal to <var>sendEncodings</var>, we reject the
+                              <a>media description</a>. Reject <var>p</var> with a newly
+                              <a data-link-for="exception" data-lt="create">created</a>
+                              <code>InvalidAccessError</code> and abort these steps.
+                              This specification does not allow remotely initiated RID renegotiation.
+                            </p>
+                          <li>
                             <p>If a suitable transceiver was found (<var>transceiver</var>
                             is set) and <var>sendEncodings</var> is non-empty, set
                             <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SendEncodings]]</a>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1192,8 +1192,17 @@
           </li>
           <li>
             <p>In parallel, start the process to apply <var>description</var>
-            as described in <span data-jsep=
-            "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>.</p>
+              as described in <span data-jsep=
+                                  "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>,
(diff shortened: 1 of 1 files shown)


henbos:
> Note: This is the first time we reject a SRD in the text section that comes after applying it successfully in JSEP. I'm not sure this is the right way to specify that, but doing it before JSEP gets weird too (repeated "look up this transceiver" steps). Suggestions welcome.

Implementing this would yield out-of-sync behavior with JSEP. This may also have updated some prior transcievers before aborting.

I suggest moving this to here:

> In parallel, start the process to apply description as described in [JSEP] (section 5.5. and section 5.6.).

E.g. by saying

> ... start the process to apply description as described in [JSEP] (section 5.5. and section 5.6.), but with the following changes:
> 
> - If any of the simulcast bla bla bla

and make it clear that in this case we abort rather than apply the m= section.

alvestrand:
Moved as suggested. I think that looks clearer.

webrtc.html
```diff
@@ -1231,6 +1240,12 @@ <h4>Set the RTCSessionDescription</h4>
                     data-lt="create">created</a>
                     <code>InvalidAccessError</code> and abort these steps.</p>
                   </li>
+                  <li>
+                    <p>If the description attempted to renegotiate RIDs, as described
+                      above, then <a>reject</a> <var>p</var> with a newly
+                      <a data-link-for="exception" data-lt="create">created</a>
+                      <code>InvalidAccessError</code> and abort these steps.
```
henbos: This part is no longer needed

henbos: Nevermind

alvestrand: Yes, it is.

Issue #2396, opened 2019-12-05
Add "process remote tracks" steps to enable reusing steps in rollback


henbos:
Follow-up to https://github.com/w3c/webrtc-pc/pull/2384#discussion_r352723045

PR #2418, opened 2019-12-19
Add "process remote tracks" algorithm to fix missing rollback steps.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2396 and some bugs in the rollback algorithm missed in https://github.com/w3c/webrtc-pc/pull/2384.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2418.html" title="Last updated on Dec 19, 2019, 9:22 PM UTC (c0012dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2418/5690992...jan-ivar:c0012dc.html" title="Last updated on Dec 19, 2019, 9:22 PM UTC (c0012dc)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1175,7 +1175,7 @@
         <ol>
           <li data-tests="RTCPeerConnection-setLocalDescription-rollback.html,RTCPeerConnection-setRemoteDescription-rollback.html">
             <p>If <code><var>description</var>.type</code> is
-            <code>"rollback"</code> and <a>signaling state</a> is either
+            <code>"rollback"</code> and <var>connection</var>'s <a>signaling state</a> is either
             <code>"stable"</code>, <code>"have-local-pranswer"</code>, or
             <code>"have-remote-pranswer"</code>, then <a>reject</a> <var>p</var>
             with a newly <a data-link-for="exception" data-lt="create">created</a>
@@ -1296,7 +1296,7 @@
                         <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
                         to a new <code><a>RTCSessionDescription</a></code> object
                         constructed from <var>description</var>, set
-                        <a>signaling state</a> to <code>"have-local-offer"</code>,
+                        <var>connection</var>'s <a>signaling state</a> to <code>"have-local-offer"</code>,
                         and <a>release early candidates</a>.</p>
                       </li>
                       <!-- C) transition haveRemoteOffer or haveLocalProvAnswer to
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1837,61 +1808,32 @@ <h4>Set the RTCSessionDescription</h4>
...
-                              <li>
-                                <p>Remove <var>transceiver</var> from <var>connection</var>'s
-                                <a href="#transceivers-set">set of transceivers</a>.</p>
-                              </li>
-                            </ol>
-                          </li>
                           <li>
-                            <p>Otherwise, (if the <var>transceiver</var> was not just removed)
-                            run the following steps:</p>
+                            <p>If <var>remote</var> is <code>true</code>, run the following sub steps:</p>
```
henbos: Do we set "remote" correctly for "rollback"? setLocalDescription() with rollback and setRemoteDescription() with rollback  should behave exactly the same. I don't think we set "remote" to reflect which description is being rolled back?

In setRemoteDescription's implicit rollback we say "Let p be the result of setting the local RTCSessionDescription indicated by {type: "rollback"}." It makes sense that "remote" is false in this case, because only local descriptions will implicitly get rolled back.

In setLocalDescription, unless I'm reading this mistaken, if we explicitly perform "rollback" then we end up at "return the result of setting the local RTCSessionDescription indicated by {type, sdp}." So in this case, when we explicitly rollback, we set "remote" to false regardless if we're rolling back a local or remote offer.

Is this a bug? Should we do one of the following:
1. In all rollback cases, make sure "remote" reflects the offer being rolled back, not whether SLD or SRD is called.
2. Instead of saying "if remote is true" here, saying "if signalingState is have-remote-offer"?

webrtc.html
```diff
@@ -1837,61 +1808,32 @@ <h4>Set the RTCSessionDescription</h4>
...
-                                <p>If the <var>transceiver</var>'s <a>[[\FiredDirection]]</a>
-                                is either <code>"sendrecv"</code> or <code>"recvonly"</code>
-                                and <var>transceiver</var>'s <a>[[\CurrentDirection]]</a> is
-                                either <code>"sendonly"</code>, <code>"inactive"</code> or
-                                <code>null</code>, <a>process the removal of a remote track</a>
-                                with <var>transceiver</var> and <var>muteTracks</var> and
-                                set the <var>transceiver</var>'s <a>[[\Receptive]]</a> slot
-                                to <code>false</code>.</p>
+                                <p>Let <var>msids</var> be the a list of MSIDs that
+                                <var>transceiver</var>'s <a>[[\CurrentRemoteDescription]]</a> indicates
```
henbos: Please copy the [[LastStableFoo]] pattern above and add a [[LastStableAssociatedRemoteMediaStreams]] to the receiver. Parsing SDP could be burdensome if you're implementing this, and having a separate pattern for this one thing might beg the question what's special about it

webrtc.html
jan-ivar: Good catch! I picked 2 since it looks like we don't read _remote_ anywhere but here on rollback.

webrtc.html
jan-ivar: And we [already have it](https://w3c.github.io/webrtc-pc/#dfn-laststablestateassociatedremotemediastreams)! Will update to use it.

jan-ivar: @henbos And I'll remove the existing use of [[LastStableAssociatedRemoteMediaStreams]] which is incorrect (passes in MediaStream objects where MSIDs are expected). Fyi.

Issue #2424, opened 2019-12-29
Privacy: icecandidateerror use in port scanning


aboba:
https://medium.com/tenable-techblog/using-webrtc-ice-servers-for-port-scanning-in-chrome-ce17b19dd474

Suggestion from Justin: restrict ports used in RTCIceServer.url

PR #2425, opened 2019-12-29
Use of icecandidateerror for port scanning

aboba:
Fix for Issue https://github.com/w3c/webrtc-pc/issues/2424

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2425.html" title="Last updated on Dec 29, 2019, 4:38 AM UTC (11a8b66)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2425/e9065b3...11a8b66.html" title="Last updated on Dec 29, 2019, 4:38 AM UTC (11a8b66)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -11920,6 +11920,20 @@ interface RTCErrorEvent : Event {
       involvement.</p>
     </section>
     <section>
+      <h2>Port scanning</h2>
+      <p>Using {{RTCIceServer.url}} and {{onicecandidateerror}} an application
+      can probe the UDP and TCP ports of endpoints on local or remote networks.
+      This goes beyond what can be accomplished with other Web APIs such as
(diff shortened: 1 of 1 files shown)


lgrahl:
Is the solution suggesting that either `icecandidateerror` is unreliable or some ports are unusable? If yes, there are multiple issues with it:

* It's not deterministic,
* it's not testable,
* it's brittle because `icecandidateerror` may or may not fail, so relying on it is useless,
* it's brittle because it arrogates to know which ports are okay and which aren't,
* it still allows for fingerprinting.

I believe that's a deep rabbit hole. I'd suggest getting rid of `icecandidateerror` instead. We can live without it.

PR #2429, opened 2020-01-05
Touch up examples.

jan-ivar:
Fix up examples a bit.
- More readable, modern JS. Fix long lines/bad wrapping.
- Simple p2p example: add comment to not block negotiation on permission in production code.
- Warm-up example: fix bug where only one side would start; refactored.
- datachannel example: use simpler negotiated pattern; add brief app logic to show `send()`.
- perfect example: add brief app logic to highlight benefit.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2429.html" title="Last updated on Jan 7, 2020, 2:53 PM UTC (296b60d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2429/e9065b3...jan-ivar:296b60d.html" title="Last updated on Jan 7, 2020, 2:53 PM UTC (296b60d)">Diff</a>


webrtc.html
```diff
@@ -11344,9 +11320,33 @@ <h3>Perfect Negotiation Example</h3>
...
+    console.error(err);
+  }
+}
+
+pc.ontrack = ({streams}) =&gt; {
+  if (remoteView.srcObject) return;
+  remoteView.srcObject = streams[0];
+};
+
+// - The perfect negotiation logic, separated from the rest of the application ---
```
alvestrand: are the dashes serving some purpose, or are they a distraction?

jan-ivar: They form a separator, to illustrate separation of the perfect negotiation logic below (the remaining code) from the app logic above.

Unlike other comments, it does not apply only to the following lines/function. Do we have other ways to illustrate that?

PR #2444, opened 2020-01-22
Handle ICE state and candidate pair changes in the same task

### LINKED ISSUE #2283: Merge the steps for updating candidate pair and ice transport state
We currently do this in two separately queued tasks:

> When the ICE Agent indicates that the RTCIceTransportState for an RTCIceTransport has changed, the user agent MUST queue a task that runs the following steps:

> When the ICE Agent indicates that the selected candidate pair for an RTCIceTransport has changed, the user agent MUST queue a task that runs the following steps:

There is room for interpretation which task you queue first. In either case, since they are tasks queued separately and each task fires an event, it is observable which thing happened first.

I think the correct behavior would be to update both the ICE state and the candidate pair, and to fire both events (onstatechange, onselectedcandidatepairchanged)

henbos:
Fixes #2283.

If the selected candidate pair causes the ICE connection state to change, we surface all state changes first and fire all events later, in the following order:
- transport.selectedcandidatepairchange
- transport.statechange
- connection.iceconnectionstatechange
- connection.connectionstatechange

This gives a well-defined event firing order. It also makes it possible to inspect the state of the connection or transport and not find inconsistencies, e.g. in selectedcandidatepairchange we are guaranteed to see an up-to-date transport.state or vice versa.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2444.html" title="Last updated on Jan 23, 2020, 3:32 PM UTC (34fe4a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2444/7e02bc9...henbos:34fe4a2.html" title="Last updated on Jan 23, 2020, 3:32 PM UTC (34fe4a2)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<li lang="en">
+<html lang="en">
 <head>
   <meta charset="utf-8">
   <link href="webrtc.css" rel="stylesheet">
--- a/webrtc.html
+++ b/webrtc.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<li lang="en">
 <head>
   <meta charset="utf-8">
   <link href="webrtc.css" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


henbos:
Prior to this CL it would probably have been valid (up to interpretation) to update the ICE state first, fire an event, and then in a later task update the selected candidate pair and fire an event. Probably not intentional to be in a connected state while at the same time having a null candidate pair. This fixes that.

aboba:
One thing I noticed is that the check for a closed `RTCPeerConnection` has been removed.  Was this intentional?

webrtc.html
```diff
@@ -8132,10 +8132,14 @@ <h3><dfn>RTCIceTransport</dfn> Interface</h3>
...
-      {{RTCIceTransportState}} for an
-      {{RTCIceTransport}} has changed, the user agent MUST queue
-      a task that runs the following steps:</p>
+      <p>The {{RTCIceTransportState}} of an {{RTCIceTransport}} may change
+      because a candidate pair with a usable connection was found and selected
+      or it may change without the selected candidate pair changing - the
+      selected pair and {{RTCIceTransportState}} are related and are handled in
+      the same task:</p>
+      <p>When the [= ICE Agent =] indicates that an {{RTCIceTransport}} has
+      changed the selected candidate pair, the {{RTCIceTransportState}} or
```
jan-ivar: changed either

PR #2445, opened 2020-01-22
Add note explaining addTrack/addTransceiver-created transceivers

### LINKED ISSUE #2419: Explain the difference between addTrack and addTransciever-created transceivers
addTrack-transcievers can attach to remote offers, addTransciever-transceivers can't.

henbos:
Fixes #2419.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2445.html" title="Last updated on Jan 27, 2020, 9:49 AM UTC (e87f59d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2445/ffeb311...henbos:e87f59d.html" title="Last updated on Jan 27, 2020, 9:49 AM UTC (e87f59d)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5094,7 +5094,7 @@ interface RTCCertificate {
       When an offer is set as the remote description, any media descriptions
       in it not yet associated with a transceiver get associated with a new or
       existing transceiver. In this case, only unassociated transceivers that were
-      created via the {{RTCPeerConnection/addTrack()}} method get associated.
+      created via the {{RTCPeerConnection/addTrack()}} method may be associated.
       Unassociated transceivers created via the
       {{RTCPeerConnection/addTransceiver()}} method, however, won't get associated
       even if media descriptions are available in the remote offer. Instead,
--- a/webrtc.html
+++ b/webrtc.html
@@ -5083,9 +5083,9 @@ interface RTCCertificate {
       {{RTCPeerConnection/addTransceiver()}}-created transceivers to get
       associated, this endpoint must initiate an offer/answer exchange as well.
       This sets {{RTCPeerConnection/addTrack()}}-created and
-      {{RTCPeerConnection/addTransceiver()}}-created apart in one critical way
-      that is not observable from inspecting the properties of the
-      {{RTCRtpTransceiver}} object.
+      {{RTCPeerConnection/addTransceiver()}}-created transceivers apart in one
+      critical way that is in no way observable from inspecting the attributes
+      or methods of the {{RTCRtpTransceiver}} object.
     </p>
     <section>
       <h3>RTCPeerConnection Interface Extensions</h3>
--- a/webrtc.html
+++ b/webrtc.html
@@ -5065,6 +5065,28 @@ interface RTCCertificate {
       {{MediaStreamTrack}} and {{RTCRtpReceiver}} are
       surfaced to the application via the {{track}} event.
     </p>
+    <p>
+      In order for an {{RTCRtpTransceiver}} to send and/or receive media with
+      another endpoint this must be negotiated with SDP such that both endpoints
+      have an {{RTCRtpTransceiver}} that is [= associated =] with the same
+      [= media description =]. When creating an offer, the local side will
+      associate all unassociated transceivers with a media description in the
+      offer. When the remote side sets the offer, media descriptions that are
+      not associated with a transceiver yet will get associated with a
+      transceiver. In order for the remote side to associate a transceiver with
+      a media section in this step, it has to create a <i>new</i> transceiver,
+      except if it has an unassociated transceiver that was created via the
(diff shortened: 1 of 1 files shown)


aboba:
"In order for {{RTCPeerConnection/addTransceiver()}}-created transceivers to get
 associated, this endpoint must initiate an offer/answer exchange as well."

[BA] This makes it sound like we are recommending two O/A negotiations, one in each direction.

webrtc.html
```diff
@@ -5065,6 +5065,28 @@ <h2>RTP Media API</h2>
       {{MediaStreamTrack}} and {{RTCRtpReceiver}} are
       surfaced to the application via the {{track}} event.
     </p>
+    <p>
+      In order for an {{RTCRtpTransceiver}} to send and/or receive media with
+      another endpoint this must be negotiated with SDP such that both endpoints
+      have an {{RTCRtpTransceiver}} that is [= associated =] with the same
+      [= media description =]. When creating an offer, the local side will
+      associate all unassociated transceivers with a media description in the
```
jan-ivar: In SLD not createOffer

webrtc.html
```diff
@@ -5065,6 +5065,28 @@ <h2>RTP Media API</h2>
...
+      offer. When the remote side sets the offer, media descriptions that are
+      not associated with a transceiver yet will get associated with a
+      transceiver. In order for the remote side to associate a transceiver with
+      a media section in this step, it has to create a <i>new</i> transceiver,
```
jan-ivar: I think this explains the remedy before the problem is fully understood. I'd focus on describing what happens and its limitations first, and move explanations of remedies to the very tend

jan-ivar:
(I tried using GitHub's new suggestion feature, but it didn't work on multi-line comments, and got unwieldy, so pardon the rude edit). My explanations from that review:

> I like the word "enough". Makes me think of quantity, a helpful metaphor here. So I'll lean on it.
>
> I'd like to highlight the two sides and the subset of transceivers that get associated here (those on one side). This may seem obvious to us, but people might otherwise assume some magic that isn't there.

> I find the word "attach" as a stand-in for "associated" confusing terminology

> We didn't use "will" in almost the same sentence above.

> Toned down one part that sounded a bit critical perhaps (I don't disagree mind you) 😉

@henbos LMKWYT

jan-ivar:
I also tried to break up the long paragraph.

webrtc.html
```diff
@@ -5078,6 +5078,42 @@ <h2>RTP Media API</h2>
...
+      enough {{RTCPeerConnection/addTrack()}}-created transceivers. This sets
+      {{RTCPeerConnection/addTrack()}}-created and
+      {{RTCPeerConnection/addTransceiver()}}-created transceivers apart in a
+      critical way that is not observable from inspecting their attributes.
+    </p>
+    <p>
+      When creating an answer, only media media descriptions that were
+      present in the offer may be listed in the
+      answer. As a consequence, any transceivers that were not associated when
+      setting the remote offer remain unassociated when setting the local
```
jan-ivar:
```suggestion
      setting the remote offer remain unassociated after setting the local
```

webrtc.html
```diff
@@ -5078,6 +5078,42 @@ <h2>RTP Media API</h2>
...
+    <p>
+      In order for an {{RTCRtpTransceiver}} to send and/or receive media with
+      another endpoint this must be negotiated with SDP such that both endpoints
+      have an {{RTCRtpTransceiver}} object that is [= associated =] with the same
+      [= media description =].
+    </p>
+    <p>
+      When creating an offer, enough media descriptions will be generated to cover
+      all transceivers on that end. When this offer is set as the local description,
+      any unassociated transceivers get associated with media descriptions in the
```
jan-ivar: Sorry, one more thing I just caught. The word "unassociated" is new with this patch.

Prior terminology here [appears to be](https://w3c.github.io/webrtc-pc/#rtcrtptransceiver-interface) "disassociated". But then there are no uses of it as a noun (one use as a verb). Whichever word we pick we want to be consistent.

jan-ivar: And then there's https://english.stackexchange.com/questions/36416/dissociate-vs-disassociate

henbos: OK so the existing language is "disassociate" and "disassociated". I will use the same terminology here since this is a bit of a tangent to what this PR is attempting to do. If we want to revisit what language to use when referring to "disassociated transceivers" we can do it separately and for the entire document rather than just this section. We can also define the word, currently the explanation for disassociated does look like a proper definition of the word but we don't "&lt;dfn&gt;" it.

PR updated changing "unassociated" to "disassociated" for now.

webrtc.html
henbos: Done.

PR #2452, opened 2020-01-22
Clarify that removed codec need not be listed

### LINKED ISSUE #2371: setCodecPreferences() and removing a codec after initial negotiation
Spin-off from #2368.

Suppose I do the following:
```
setCodecPreferences(VP8, VP9);
O/A exchange
setCodecPreference(VP8);
O/A exchange
```

What should/does happen?
1. In the second offer, the VP9 payload type is not listed, and the a=video line does not reference it.
2. In the second offer, the VP9 payload type is still listed, but the a=video line does not reference it.
3. Invalid SDP is generated because JSEP demands payload types be set in stone and the a=video line to reference all PTs in some order.

If either 1) or 2) happens then setCodecPreference() has the effect that I though we desired. It was my assumption that 1) is what happens, but if it turns out PTs are more like m= sections and cannot be unlisted, then 2) makes more sense.

On the other hand, if 3) is the case, then we need to update webrtc-pc to say that the second setCodecPreference() throws an exception because we're changing the set of codecs after the initial negotiation. This is not something we currently guard against.

@alvestrand @jan-ivar

henbos:
Fixes #2371.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2452.html" title="Last updated on Jan 31, 2020, 9:13 AM UTC (03defe1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2452/7e02bc9...henbos:03defe1.html" title="Last updated on Jan 31, 2020, 9:13 AM UTC (03defe1)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7476,15 +7476,14 @@ interface RTCRtpTransceiver {
               include this {{RTCRtpTransceiver}} until this method is
               called again. Setting <var>codecs</var> to an empty sequence
               resets codec preferences to any default value.</p>
-              <p class="note">Preferred codecs have their payload types listed
+              <p class="note">Codecs have their payload types listed
               under each m= section in the SDP, defining the mapping between
               payload types and codecs. These payload types are referenced by
-              the a=video or a=audio lines in the order of preference. While a
-              previously negotiated codec's payload type must not be reused in
-              future offers or answers to map to a different codec, if a codec
-              that was previously negotiated is removed from the list, the
-              subsequent offers and answers do not need to continue to define
-              that payload type in the SDP.</p>
+              the m=video or m=audio lines in the order of preference. Since a
+              previously negotiated codec's payload type is not reused in
+              future offers or answers, if a codec that was previously
+              negotiated is removed from the list, the subsequent offers and
+              answers do not need to continue to define that payload type in the SDP.</p>
               <p>The <var>codecs</var> sequence passed into
               {{setCodecPreferences}} can only contain codecs that are
               returned by {{RTCRtpSender}}.{{RTCRtpSender/getCapabilities}}(<var>kind</var>) or
--- a/webrtc.html
+++ b/webrtc.html
@@ -7479,11 +7479,13 @@ interface RTCRtpTransceiver {
               <p class="note">Codecs have their payload types listed
               under each m= section in the SDP, defining the mapping between
               payload types and codecs. These payload types are referenced by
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -7476,6 +7476,14 @@ <h2>Methods</h2>
               include this {{RTCRtpTransceiver}} until this method is
               called again. Setting <var>codecs</var> to an empty sequence
               resets codec preferences to any default value.</p>
+              <p class="note">Codecs have their payload types listed
+              under each m= section in the SDP, defining the mapping between
+              payload types and codecs. These payload types are referenced by
+              the m=video or m=audio lines in the order of preference. Since a
```
henbos: Remove "Since"

webrtc.html
```diff
@@ -7476,6 +7476,14 @@ <h2>Methods</h2>
...
+              previously negotiated codec's payload type is not reused in
+              future offers or answers, if a codec that was previously
```
henbos: , / .

henbos:
I rephrased some things, please take a look.

webrtc.html
```diff
@@ -7476,6 +7476,16 @@ <h2>Methods</h2>
...
+              <p class="note">Codecs have their payload types listed
+              under each m= section in the SDP, defining the mapping between
+              payload types and codecs. These payload types are referenced by
+              the m=video or m=audio lines in the order of preference, and
+              codecs that are not negotiated do not apear in this list.
+              A previously negotiated codec that is subsequently removed
+              disappears from the m=video or m=audio line, and while its codec
+              payload type is not to be reused in future offers or answers,
+              its payload type may also be removed from the mapping of payload
+              types in the SDP.</p>
```
alvestrand: I think this is consistent with JSEP section 5.2.1:

```
   o  If codec preferences have been set for the associated transceiver,
      media formats MUST be generated in the corresponding order, and
      MUST exclude any codecs not present in the codec preferences.

   o  Unless excluded by the above restrictions, the media formats MUST
      include the mandatory audio/video codecs as specified in
      [RFC7874], Section 3, and [RFC7742], Section 5.
```
Should we have a reference to that section somewhere?

aboba: Yes, that seems like a good idea (though the reference doesn't cover the "not to be reused" aspect).

alvestrand: That's general SDP rules, I think.

Issue #2477, opened 2020-02-18
negotiationneeded may still fire when it shouldn't


jan-ivar:
In https://github.com/w3c/webrtc-pc/pull/2405 we added a "chain" step to the [update the negotiation-needed flag](https://w3c.github.io/webrtc-pc/#dfn-update-the-negotiation-needed-flag) algorithm, to avoid firing NN if negotiation had begun (but was still in `"stable"`):
> ...
> _6. [Chain](https://w3c.github.io/webrtc-pc/#dfn-chain) a step to queue a task that runs the following steps, to connection's operations chain:_
> 1. _If connection.[[IsClosed]] is true, abort these steps._
> 2. _If connection's signaling state is not "stable", abort these steps._
> 3. _If connection.[[NegotiationNeeded]] is false, abort these steps._
> 4. _Fire an event named negotiationneeded at connection._

Unfortunately, I believe this may fire twice now in an SRD(answer) edge-case:
```js
pc.setRemoteDesciption(answer); // note: no await
pc.addTransceiver("audio"); // will fire NN once "stable"; but so will SRD; = twice!
```
Also, because we chain-_then_-queue, by the time the queued task eventually runs, another already queued task may have begun negotiation: 🤦
```js
onmessage = ({data: {description}}) => pc.setRemoteDescription(description); // race me
await wait(10);
pc.addTransceiver("audio"); // queued NN task may race with queued onmessage
```

I see a couple ways to solve this, but let's recap **required behaviors**, so we don't break any:
1. Must defer any firing to after the current run of the event loop
2. Must not fire if closed
3. Must not fire if negotiation has begun (still in `"stable"`)
4. Must not fire if negotiation is in process (not in `"stable"`)
5. Must not fire twice
6. Must re-fire going back to `"stable"` if needed (including rollback)
7. Must not re-fire on negotiation failure

jan-ivar:
A basic **solution A** to the first problem may be to introduce yet another internal slot: **[[NegotiationBegun]]**, but setting and clearing it may get messy. And it doesn't address the 2nd.

A **solution B** might solve both: (Collapse the whole algorithm into one _queued-then-chained task_)

 _"Queue a task to [chain](https://w3c.github.io/webrtc-pc/#dfn-chain) the following steps to connection's operations chain:_
 1. _If connection.[[IsClosed]] is true, abort these steps._
 2. _If connection's signaling state is not "stable", abort these steps._
[...]

Issue #2476, opened 2020-02-15
JSEP reads & modifies transceivers in parallel, causing racy behavior.


jan-ivar:
[Webrtc-pc](https://w3c.github.io/webrtc-pc/#set-description) runs JSEP in the background (*"In parallel, start the process to apply description as described in [JSEP]"* and *"If ... applied successfully"* will *"queue a task"* to main thread with results).

But [section 5.9](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-25#section-5.9) ("Apply a Local Description") reads the transceiver list and read/writes the `mid` property ("[associated](https://w3c.github.io/webrtc-pc/#dfn-associated)"). Those are main-thread properties read/written from the wrong thread:
```
      *  If there is no RtpTransceiver associated with this m= section,
         find one and associate it with this m= section according to the
         following steps.  Note that this situation will only occur when
         applying an offer.

         +  Find the RtpTransceiver that corresponds to this m= section,
            using the mapping between transceivers and m= section
            indices established when creating the offer.

         +  Set the value of this RtpTransceiver's mid property to the
            MID of the m= section.
```
This read may be fine, since the only havoc JS can wreak here is stop a transceiver, which JSEP doesn't seem to care about. But it writes the `mid` which may race with JS-reads &  [ONN](https://w3c.github.io/webrtc-pc/#dfn-check-if-negotiation-is-needed). E.g.:
```js
const [transceiver] = pc.getTransceivers();

pc.setLocalDescription(offer); // note: no await
console.log(transceiver.mid); // null?
transceiver.sender.setStreams(streams); // will negotiationneeded fire?
```

[Section 5.10](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-25#section-5.10) ("Apply a Remote Description") also reads the transceiver list & read/writes the `mid`:

      *  If the m= section is not associated with any RtpTransceiver
         (possibly because it was dissociated in the previous step),
         either find an RtpTransceiver or create one according to the
         following steps:

         +  If the m= section is sendrecv or recvonly, and there are
            RtpTransceivers of the same type that were added to the
            PeerConnection by addTrack and are not associated with any
            m= section and are not stopped, find the first (according to
            the canonical order described in Section 5.2.1) such
            RtpTransceiver.

         +  If no RtpTransceiver was found in the previous step, create
            one with a recvonly direction.

         +  Associate the found or created RtpTransceiver with the m=
[...]

PR #2478, opened 2020-02-18
Queue then chain steps to update the negotiation-needed flag.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2477 and mitigates https://github.com/w3c/webrtc-pc/issues/2476.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2478.html" title="Last updated on Feb 18, 2020, 8:27 PM UTC (f7e1c09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2478/b9b0519...jan-ivar:f7e1c09.html" title="Last updated on Feb 18, 2020, 8:27 PM UTC (f7e1c09)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3850,8 +3850,9 @@ interface RTCSessionDescription {
         implementation that affect negotiation. If such changes occur, the user
         agent MUST queue a task to [= update the negotiation-needed
         flag =].</p>
-        <p>To <dfn>update the negotiation-needed flag</dfn> for
-        <var>connection</var>, run the following steps:</p>
+        <p data-tests="">To <dfn>update the negotiation-needed flag</dfn> for
+        <var>connection</var>, queue a task to [= chain =] the following steps
+        to <var>connection</var>'s [= operations chain =]:</p>
         <ol>
           <li class=untestable>
             <p>If <var>connection</var>.<a>[[\IsClosed]]</a> is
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3881,33 +3882,18 @@ <h4>Updating the Negotiation-Needed flag</h4>
...
-                <p>[= Fire an event =] named {{negotiationneeded}}
-                at <var>connection</var>.</p>
-              </li>
-            </ol>
-            <p class="note">This queueing prevents {{negotiationneeded}} from
-            firing prematurely, in the common situation where multiple
-            modifications to <var>connection</var> are being made at once.</p>
+          <li data-tests="RTCPeerConnection-onnegotiationneeded.html">
+            <p>[= Fire an event =] named {{negotiationneeded}}
+            at <var>connection</var>.</p>
```
alvestrand: The check against RTCSignalingState being not-"stable" is now in the chain only. Is that correct?

The check against NegotiationNeeded is now unnecessary, since NegotiationNeeded is set directly above in the same task, but it also means that NegotiationNeeded is potentially false for a "long" period while we're waiting for the chain to clear. Is that OK? (if multiple transitions false -> true are chained, only the first will fire a signal, so I suppose that's OK)

This is complicated......

Issue #2502, opened 2020-03-30
The threading model of webrtc-pc: When are effects of in-parallel stuff surfaced?


henbos:
JavaScript "pretending" to be single-threaded, things that happen either just happen (in the currently executing task), or they happen "in-parallel" (i.e. off the thread), and then the results are surfaced sometime in the future in a queued task.

According to my understanding of things, exposing the effects of something from an in-parallel operation before the queued task has executed would be a bug. I'm going to call this **_the state surfacing theorem_**. Feel free to challenge the validity of this theorem.

For example, `while (true) { console.log(sender.getSynchronizationSources()); }` should ALWAYS print the same result, because the thread is blocked and no queued task is able to execute to inform the JS thread that a new RTP packet has arrived.

Similarly, I would expect the following to print 0 forever and ever, because we are not letting the in-parallel SRD operation perform its queued task to create the JavaScript layer transceiver objects corresponding to the SDP m= sections:
```
pc.setRemoteDescription(offerThatCreatesATransceiver);
while (true) {
  console.log(pc.getTransceivers().length);
}
```

Why does this matter? Well JSEP has this to say about addTrack:

> If the PeerConnection is in the "have-remote-offer" state, the track will be attached to the first compatible transceiver that was created by the most recent call to setRemoteDescription() and does not have a local track.

But setRemoteDescription() executes in-parallel. And addTrack() executes immediately. So if I do `setRemoteDescription(); addTrack();`, has SRD happened when addTrack happens or not?

Based on recent discussions, I believe Chrome and Firefox has observably different implementations. I believe both of them violate _the state surfacing theorem_ but in different ways. Correct me if I'm wrong on Firefox's implementation.

- **Behavior A** (= what Chrome does): In-parallel operations previously queued happen first and synchronous operations happen second due to PostTasks ensuring a FIFO order. If you call SRD followed by addTrack, the JS thread is blocked until addTrack is performed, which only happens after the SRD operation has executed. The theorem is violated because what the result of the addTrack operation is could expose how SRD was executed.
- **Behavior B** (= probably what Firefox does?): In-parallel operations and JavaScript share the same set of transceiver objects, accessed by grabbing locks. The result of SRD + addTrack is undefined, because which operation modifies the set of transceiver objects first is racy. The theorem is violated in the cases when the race favors SRD happening first.
[...]

PR #2504, opened 2020-03-31
Restart process to apply description if addTrack called

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2502.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2504.html" title="Last updated on Apr 1, 2020, 9:55 PM UTC (2b3b2c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2504/544a656...jan-ivar:2b3b2c8.html" title="Last updated on Apr 1, 2020, 9:55 PM UTC (2b3b2c8)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1273,6 +1273,11 @@
                     <code>true</code>, then abort these steps.</p>
                   </li>
                   <li>
+                    <p>If the {{RTCPeerConnection/addTrack()}} method was called
+                    during the process to apply <var>description</var>, then
+                    abort these steps and start the process over.</p>
+                  </li>
+                  <li>
                     <p>If <var>description</var> is of type {{RTCSdpType/"offer"}}
                     and the [= signaling state =] of <var>connection</var> is
                     {{RTCSignalingState/"stable"}} then for each <var>transceiver</var>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1273,7 +1273,9 @@
                     <code>true</code>, then abort these steps.</p>
                   </li>
                   <li>
-                    <p>If any {{RTCPeerConnection/addTrack()}} methods succeeded
+                    <p>If <var>remote</var> is <code>true</code> and
+                    <var>description</var> is of type {{RTCSdpType/"offer"}},
+                    then if any {{RTCPeerConnection/addTrack()}} methods succeeded
                     during the process to apply <var>description</var>, abort
                     these steps and start the process over as if they had
                     succeeded prior, to include the extra transceiver(s) in the
--- a/webrtc.html
+++ b/webrtc.html
@@ -1273,9 +1273,11 @@
                     <code>true</code>, then abort these steps.</p>
                   </li>
                   <li>
-                    <p>If the {{RTCPeerConnection/addTrack()}} method was called
-                    during the process to apply <var>description</var>, then
-                    abort these steps and start the process over.</p>
+                    <p>If any {{RTCPeerConnection/addTrack()}} methods succeeded
+                    during the process to apply <var>description</var>, abort
+                    these steps and start the process over as if they had
+                    succeeded prior, to include the extra transceiver(s) in the
+                    process.</p>
                   </li>
                   <li>
                     <p>If <var>description</var> is of type {{RTCSdpType/"offer"}}


henbos:
Clever.

So if you do setRemoteDescription() followed by addTrack() the resulting behavior observed would always be _as if_ you invoked addTrack() first and setRemoteDescription() second?

Can you clarify, would this be any different from if you had awaited setRemoteDescription() first and performed addTrack() second?

jan-ivar:
> So if you do setRemoteDescription() followed by addTrack() the resulting behavior observed would always be as if you invoked addTrack() first and setRemoteDescription() second?

Yes.

> Can you clarify, would this be any different from if you had awaited setRemoteDescription() first and performed addTrack() second?

I think the net result is the same, but I'd appreciate a second pair of eyes. The assumption here is that the incoming SRD creates a transceiver that we want to associate with. In the latter case, we'd be in `"have-remote-offer"` so this transceiver's currentDirection would still be `null` and satisfy the absorption criteria in [addTrack](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtrack).

jan-ivar:
> I think the net result is the same

Verified in several browsers with https://jsfiddle.net/jib1/w672x5t9/

henbos:
If the net result is the same - fingers crossed - then Chrome acts almost like what this PR is proposing. Not by doing what the PR proposes, though.

Here's a codepen: https://codepen.io/henbos/pen/dyorzjV?editors=0010
If you run it in Chrome, I think you get the intended behavior except that the associated addTrack transceiver has its mid attribute exposed too early.

Chrome has a threading model that I have been arguing against for years, but while it blocks threads (undesired) it does have the benefit of always having a well-defined order.

Here's what happens:
JS calls SRD() and addTrack() without awaiting in-between. SRD() starts in-parallel work. The JS addTrack() blocks until JSEP has performed addTrack(), and then exposes the resulting transceiver. When SRD is done, a PostTask ensures transceivers get synchronized on the JS thread.

Our implementation does not perform addTrack() "synchronously" rather it fakes being synchronous by blocking the thread. On the JSEP thread however everything is single-threaded. On this thread, SRD() always gets executed first. This means that AddTrack() has a transceiver to recycle. When this is surfaced to JS, it exposes a new transceiver that is added to webrtc-pc's set of transceivers - even if this transceiver was created as the result of an in-parallel operation. Later, the task that was posted by SRD completing gets executed, and transceivers get synchronized to match the SRD callback.

henbos:
Chrome output of https://codepen.io/henbos/pen/dyorzjV?editors=0010:
```
#transceivers on offer: 2
pen.js:13 #transceivers after SRD (no await): 0
pen.js:15 #transceivers after addTrack (still no await): 1
pen.js:22 - addTrack()'s transceiver's mid: "0"
pen.js:24 #transceivers after await: 2
```
The output I would expect from intended/spec-compliant behavior:
```
#transceivers on offer: 2
pen.js:13 #transceivers after SRD (no await): 0
pen.js:15 #transceivers after addTrack (still no await): 1
pen.js:22 - addTrack()'s transceiver's mid: null
pen.js:24 #transceivers after await: 2
```

henbos:
The expected behavior is what I observe on Firefox 74.0!
I think Chrome can be fixed by simply making the mid null of addTrack-surfaced transceivers until the PostTask surfaces the mid a second time.

henbos:
v.2.0 also covering the mid value after awaiting the promise showing Firefox associates it:
https://codepen.io/henbos/pen/ZEGPXee?editors=0010

jan-ivar:
@henbos Thanks for the codepen! I've filed [bug 1626414](https://bugzil.la/1626414) on the 75 regression.

webrtc.html
```diff
@@ -1272,6 +1272,11 @@ <h4>Set the RTCSessionDescription</h4>
                     <p>If <var>connection</var>.<a>[[\IsClosed]]</a> is
                     <code>true</code>, then abort these steps.</p>
                   </li>
+                  <li>
+                    <p>If the {{RTCPeerConnection/addTrack()}} method was called
+                    during the process to apply <var>description</var>, then
+                    abort these steps and start the process over.</p>
```
henbos, line 1278: Would aborting the processes and starting over implicitly involve rolling back some JSEP layer actions, or do you imagine "starting over" is always "safe"?

For example, if the first attempt to SRD() didn't have a transceiver to associate with then a transceiver will be proposed to be created. But when starting over due to simultaneous addTrack(), it would discover that there is a transceiver to associate with. Ergo, the proposed transceiver that was never surfaced to JS is thrown away.

It may be implementation-specific how realized this proposed transceiver is: if an object is created or if there is just a struct saying that there should be a transceiver with such and such mids, but no such transceiver exists. It is clear that the intent is that there will be no transceiver SRD-created transceiver as far as JS is concerned. Just raising this point to make sure we don't miss some edge case complexity...
- Should we clarify that starting the process over means that any side-effects caused by the previous processes must be un-done?
- Are there any other side-effects to applying the process twice? For example, could transports get created in the first round and then re-created in the second?

I'm hoping the answer is everything's cool, just wanted a second pair of eyes on this. WDYT @jan-ivar ?

henbos:
> @henbos Thanks for the codepen! I've filed [bug 1626414](https://bugzil.la/1626414) on the 75 regression.

Add https://codepen.io/henbos/pen/ZEGPXee?editors=0010 to the bug and you can also verify that the association still happens after `await srdPromise;`

In Chrome I get...
```
#transceivers on offer: 2
#transceivers after SRD (no await): 0
#transceivers after addTrack (still no await): 1
- addTrack()'s transceiver's mid (before await): "0"
#transceivers after await: 2
- addTrack()'s transceiver's mid (after await): "0"
```
But the correct output should be...
```
#transceivers on offer: 2
#transceivers after SRD (no await): 0
#transceivers after addTrack (still no await): 1
- addTrack()'s transceiver's mid (before await): null
#transceivers after await: 2
- addTrack()'s transceiver's mid (after await): "0"
```
Which is exactly what I see in Firefox 74.0

webrtc.html
henbos: JSEP defines an all-or-nothing SetRemoteDescription and all-or-nothing rollback, but it does not define "starting over" during surfacing steps, so I imagine JSEP would have fully completed the SetRemoteDescription and then needs to change its mind somehow - but this may dig into our different views on webrtc-pc vs JSEP layers. I'd like both views to be valid interpretations for the sake of not having to re-implement things in either browser.

henbos: Do we dare to add a class="note" that says "After the processes of starting over is complete, the resulting set of transceivers and their associations should be the same as if addTrack() was performed first and setRemoteDescription() was performed second"?

jan-ivar: > Would aborting the processes and starting over implicitly involve rolling back some JSEP layer actions, or do you imagine "starting over" is always "safe"?

@henbos I don't think there are side effects, except maybe a skip in mid sequencing which seems tolerable? I'd like @docfaraday to weigh in to be sure. We're not "aborting" the JSEP process, only the subsequent queued success task, which doesn't contain any transitional conditionals (coining it) AFAICT, so I think it's safe since we haven't exposed anything to JS yet, and we will after re-processing. Repeated SRD's must already be allowed according to:
state diagram

> could transports get created in the first round and then re-created in the second?

JS-exposed transports haven't been created yet (they're created in the queued task).

I've updated the prose a bit to clarify intent. PTAL

webrtc.html
```diff
@@ -1272,6 +1272,13 @@ <h4>Set the RTCSessionDescription</h4>
                     <p>If <var>connection</var>.<a>[[\IsClosed]]</a> is
                     <code>true</code>, then abort these steps.</p>
                   </li>
+                  <li>
+                    <p>If any {{RTCPeerConnection/addTrack()}} methods succeeded
+                    during the process to apply <var>description</var>, abort
+                    these steps and start the process over as if they had
+                    succeeded prior, to include the extra transceiver(s) in the
+                    process.</p>
```
henbos, line 1280: OK so we don't roll anything back, we just apply the process again? The end result would be the same as if what we did from the start was "await SRD; addTrack; await SRD;"?

henbos: Makes sense to me

jan-ivar: > OK so we don't roll anything back, we just apply the process again?

Right, where _"process"_ = [section 5.6.](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-25#section-5.6) in JSEP.

> end result would be the same as if what we did from the start was "SRD; addTrack; await SRD;"?

Better. We're talking a subset of that, since that would run both JSEP section 5.6. **_and_** the SRD queued success steps in webrtc-pc twice, with side-effects like events firing twice. We wouldn't be firing events twice. But broadly speaking yes.

alvestrand:
@jan-ivar and @henbos to discuss.

PR #2505, opened 2020-03-31
Add [[Mid]] internal slot & separate JSEP state

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2476.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2505.html" title="Last updated on Apr 16, 2020, 6:34 PM UTC (73b113c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2505/544a656...jan-ivar:73b113c.html" title="Last updated on Apr 16, 2020, 6:34 PM UTC (73b113c)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1206,7 +1206,7 @@
             "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>,
             with these additional restrictions: Use <var>jsepSetOfTransceivers</var>
             as the source of truth with regard to what "RtpTransceivers" exist, and
-            their <a>[[\JsepMidProperty]]</a> internal slot as their "mid property".
+            their <a>[[\JsepMid]]</a> internal slot as their "mid property".
             If applying <var>description</var>
             leads to modifying a transceiver <var>transceiver</var>, and
             <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SendEncodings]]</a>
@@ -1463,7 +1463,7 @@
                                 [= media description =].</p>
                               </li>
                               <li data-tests="RTCPeerConnection-setDescription-transceiver.html,RTCPeerConnection-transceivers.https.html,RTCRtpTransceiver.https.html">
-                                <p>Set <var>transceiver</var>.<a>[[\Mid]]</a> to <var>transceiver</var>.<a>[[\JsepMidProperty]]</a>.</p>
+                                <p>Set <var>transceiver</var>.<a>[[\Mid]]</a> to <var>transceiver</var>.<a>[[\JsepMid]]</a>.</p>
                               </li>
                               <li data-tests="RTCRtpTransceiver-stop.html">
                                 <p>If <var>transceiver</var>.<a>[[\Stopped]]</a>
@@ -1674,7 +1674,7 @@
                           </li>
                           <li data-tests="RTCPeerConnection-setDescription-transceiver.html,RTCRtpTransceiver.https.html">
                             <p>Set <var>transceiver</var>.<a>[[\Mid]]</a> to
-                            <var>transceiver</var>.<a>[[\JsepMidProperty]]</a>.</p>
+                            <var>transceiver</var>.<a>[[\JsepMid]]</a>.</p>
                           </li>
                           <li>
                             <p>Let <var>direction</var> be an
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1196,19 +1206,23 @@ <h4>Set the RTCSessionDescription</h4>
...
-              <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SendEncodings]]</a>
-              is non-empty, and not equal to the encodings that would result from
-              processing <var>description</var>, the process of applying <var>description</var>
-              fails.
-              This specification does not allow remotely initiated RID renegotiation.
-            </p>
-
+            as described in <span data-jsep=
+            "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>,
+            with these additional restrictions: Use <var>jsepSetOfTransceivers</var>
```
henbos, line 1217: This may or may not do what we want, but I'd like more input from the working group on how to read the webrtc-pc versus JSEP specs. Because one way to resolve things is to say "copy things" and another is to add a note and say that the set of objects **are in fact different**, and clarify places that are needed when implications are not clear (like with addTransceiver and addTrack).

The reason I care is there may be edge-case differences, so we need to be careful here.

For example, if we are copying the states here, that means that we may be copying states of transceivers that have not yet been synchronized with state changes from the JSEP layer of a previous SLD or SRD operation. The JS state is NOT GUARANTEED TO BE UP-TO-DATE.

Example:
1. Starting point: our transceiver is not stopped
2. Apply SRD that stops the transceiver. This is successful and a task is posted to tell the JS layer of this.
3. Without awaiting (thus the PostTask has not happened yet): Perform another SRD. Because transceiver states are copied, we perform the second SRD as if the transceiver was not stopped. PostTask to tell JS layer of the result.
4. First SRD's PostTask synchronizes JS transceivers with result of SRD operation. Transceiver is stopped.
5. Second SRD's PostTask synchronizes JS transcievers with result of SRD operation. Transceiver is not stopped.

We just un-stopped a transceiver... what?

JS cannot be an authority on the true state of the underlying system. JS can only be a control knob that acts at a distance to inform JSEP what to do. But since JSEP can do things in parallel, JSEP is the authority and JS is the thing syncing up with JSEP, in my view.

henbos: The example works with swapping out mids too.

transciever.mid == null
SRD("mid:0")
SRD("mid:banana")

Is this now valid? When second SRD is applied mid is still null according to the copy, so assigning it "banana" should be fine, yes?

henbos: Or are we protected from this because PostTask must have happened before the second call, because without it the promise is not resolved, and without the promise resolving the operation chain protects us? :) I hope so

henbos: I think this is not an issue because of the operations chain, but keeping this thread unresolved for the sake of discussion later today

webrtc.html
```diff
@@ -964,6 +964,16 @@ <h4>Constructor</h4>
             <p>Let <var>connection</var> have an <dfn>[[\EarlyCandidates]]</dfn>
               internal slot, initialized to an empty list.</p>
           </li>
+          <li data-tests="RTCPeerConnection-transceivers.https.html, RTCRtpTransceiver.https.html">
+            <p>Let <var>connection</var> have a <dfn>[[\Mid]]</dfn>
```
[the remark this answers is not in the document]
jan-ivar: @henbos Thanks for catching that! Now fixed PTAL

webrtc.html
```diff
@@ -964,6 +964,16 @@ <h4>Constructor</h4>
...
+            This is the "mid" attribute as defined in <span data-jsep=
+            "initial-offers initial-answers">[[!JSEP]]</span>.</p>
+          </li>
+          <li data-tests="RTCPeerConnection-transceivers.https.html, RTCRtpTransceiver.https.html">
+            <p>Let <var>connection</var> have a <dfn>[[\CurrentMid]]</dfn>
+              internal slot, initialized to <code>null</code>.</p>
+          </li>
           <li data-tests="RTCPeerConnection-constructor.html">
             <p>Set <var>connection</var>'s [= signaling state =] to
             {{RTCSignalingState/"stable"}}.</p>
```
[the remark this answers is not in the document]
jan-ivar: That seems editorial and orthogonal to this PR. I suggest we open a new issue if there's a problem there.

webrtc.html
jan-ivar: I agree the operations chain protects us from this happening.

jan-ivar:
This should read better. @henbos note I kept a [[JsepMidProperty]] internal slot, because it got too unwieldy to use a temporary here, since it's per transceiver. Also ,since webrtc-pc never modifies it, we never need to copy it.

Issue #2512, opened 2020-04-14
Calling SRD twice isn't web compatible


jan-ivar:
Calling SRD twice (ditto SLD) may be well-defined in webrtc-pc, but it's not in JSEP.

![state diagram](https://w3c.github.io/webrtc-pc/images/peerstates.svg)

Firefox does something akin to a silent mini-rollback in-between, whereas Chrome does not, instead [failing](https://jsfiddle.net/jib1/xtk7np5m/) all offers different¹ from the first:
```js
InvalidAccessError: Failed to execute 'setRemoteDescription' on 'RTCPeerConnection':
Failed to set remote offer sdp: The order of m-lines in subsequent offer doesn't match
order from previous offer/answer.
```
In Firefox these succeed² instead. But its aggressive mini-rollbacks produce different transceiver objects even with identical offers, which doesn't seem great either. 

Who's right? Who won? You decide!

<sub>1) Turns out Chrome [allows m-lines to grow](https://jsfiddle.net/jib1/pk84hex9/), but not shrink. Also [media kinds cannot change](https://jsfiddle.net/jib1/fye9wcp1/).</sub>
<sub>2) We're fixing [some disappearing transceivers](https://bugzil.la/1630029) in Firefox atm, but that shouldn't distract from the different interpretations.</sub>

jan-ivar:
Last I heard of a real-world use case for offer trumping was the ["pushy SFU"](https://docs.google.com/presentation/d/1UrXARLbAfwmfK686rX_9W_-FGIc62ZDwVd079wIoX7I/edit?ts=5cdd966f#slide=id.g5b048ace1b_21_24) use case.

jan-ivar:
It was mentioned to me that a simple workaround for the pushy SFU case is as described in that slide: queue requests, basically visit `"stable"` in-between. So failure here may not matter much. OTOH it might inform how it should work.

fippo:
yeah, i've typically done a "if there is a cycle pending, put the operation into a queue, take stuff from the queue on next signalingstatechange to stable".
What this avoids is the following flow:
* SRD (1)
* SRD (2)
* createAnswer (1)
* createAnswer (2)
* SLD (1)
* SLD (2, :boom: can't call SLD twice)

jan-ivar:
[...]

PR #2514, opened 2020-04-20
Validate back-to-back offers as if answers were applied in between.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2512.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 23, 2020, 1:58 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fjan-ivar%2Fwebrtc-pc%2F9d8f02deac5e04fe2dfd0357ef1ef483377c9780%2Fwebrtc.html%3FisPreview%3Dtrue)

```
[33m📡 HTTP Error 520:[39m [36mhttps://rawcdn.githack.com/jan-ivar/webrtc-pc/9d8f02deac5e04fe2dfd0357ef1ef483377c9780/webrtc.html[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/webrtc-pc%232514.)._
</details>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1223,16 +1223,25 @@
             <p data-tests="">In parallel, start the process to apply <var>description</var>
             as described in <span data-jsep=
             "processing-a-local-desc processing-a-remote-desc">[[!JSEP]]</span>,
-            with these additional restrictions: Use <var>jsepSetOfTransceivers</var>
-            as the source of truth with regard to what "RtpTransceivers" exist, and
-            their <a>[[\JsepMid]]</a> internal slot as their "mid property".
-            If applying <var>description</var>
-            leads to modifying a transceiver <var>transceiver</var>, and
-            <var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SendEncodings]]</a>
-            is non-empty, and not equal to the encodings that would result from
-            processing <var>description</var>, the process of applying <var>description</var>
-            fails. This specification does not allow remotely initiated RID renegotiation.</p>
+            with these additional restrictions:<p>
             <ol>
+              <li>
+                <p>Use <var>jsepSetOfTransceivers</var> as the source of truth
+                with regard to what "RtpTransceivers" exist, and their
+                <a>[[\JsepMid]]</a> internal slot as their "mid property".</p>
+              </li>
+              <li>
+                <p>If <var>remote<var> is <code>true</code>, validate
+                back-to-back offers as if answers were applied in between.</p>
+              </li>
+              <li>
+                <p>If applying <var>description</var>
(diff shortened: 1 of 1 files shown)


jan-ivar:
[whitespace diff](https://github.com/w3c/webrtc-pc/pull/2514/files?diff=unified&w=1)

webrtc.html
```diff
@@ -1223,16 +1223,25 @@ <h4>Set the RTCSessionDescription</h4>
...
+            with these additional restrictions:<p>
             <ol>
+              <li>
+                <p>Use <var>jsepSetOfTransceivers</var> as the source of truth
+                with regard to what "RtpTransceivers" exist, and their
+                <a>[[\JsepMid]]</a> internal slot as their "mid property".</p>
+              </li>
+              <li>
+                <p>If <var>remote</var> is <code>true</code>, validate
+                back-to-back offers as if answers were applied in between.</p>
```
henbos, line 1235: I'm not sure what "as if answers were applied in between" means?
Notably, we are not saying to apply mock answers in between because that may lead to transmission, right? This only refers to validation?

Can this be clarified? Is it supposed to say "validate as if we are in the stable state" or...?

jan-ivar: Prose was based on https://github.com/w3c/webrtc-pc/issues/2512#issuecomment-614694127. How about adding _", by running the check for subsequent offers in stable state."_?

webrtc.html
```diff
@@ -1223,16 +1223,26 @@ <h4>Set the RTCSessionDescription</h4>
...
             <ol>
+              <li>
+                <p>Use <var>jsepSetOfTransceivers</var> as the source of truth
+                with regard to what "RtpTransceivers" exist, and their
+                <a>[[\JsepMid]]</a> internal slot as their "mid property".</p>
+              </li>
+              <li>
+                <p>If <var>remote</var> is <code>true</code>, validate
+                back-to-back offers as if answers were applied in between, by
+                running the check for subsequent offers in stable state.</p>
```
henbos, line 1236: nit: "in stable state" or "as if it was in stable state"?

PR #2540, opened 2020-06-18
Add language to block bad candidates.

### LINKED ISSUE #2531: Describe what happens with administratively prohibited candidates
If a TCP candidate is connecting to a port on the Do-not-connect list (from Fetch), something should happen to the candidate, and a connection attempt should not be made.

It seems OK to do what Chrome currently does, and just ignore the candidate - it does not show up in stats, in remote-description or in GetRemoteCandidates.

### LINKED ISSUE #2533: Do not allow DNS candidates in "relay" mode
This article:
https://www.digitaltrends.com/news/signal-vulnerability-hack-location/
described how to use DNS-valued candidates to get a rough geolocation capability.
This doesn't seem sensible to guard against when normal WebRTC procedures are in use, since the attacker can cause a call setup attempt to servers under his control too, but seems to be relevant when "relay" mode is in use, since that mode was explicitly introduced to avoid revealing the user's IP address-based location.

Suggested fix: Don't use this type of candidate when policy is "relay".

https://w3c.github.io/webrtc-pc/#dom-rtcicetransportpolicy-relay

Add:

"In this mode, remote candidates requiring resolution via some external mechanism, such as FQDN-valued candidates (require DNS lookup) and .local candidates (require MDN lookup), will be discarded."

### LINKED ISSUE #2534: Allow adminstrative rejection of candidates
In conjunction with the portscanning issue, it was stated (https://github.com/w3c/webrtc-pc/issues/2426#issuecomment-570528505) that "I think we should state that ICE-TCP ports should not connect to ports on the Fetch "bad ports" list - https://fetch.spec.whatwg.org/#port-blocking"

There might be other reasons for administratively rejecting a candidate - for instance overuse of resources, or local policy.

This seems to deserve its own issue.

alvestrand:
Fixes #2534
Fixes #2533 
Fixes #2531

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2540.html" title="Last updated on Jul 9, 2020, 2:50 PM UTC (7bf1051)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2540/30651fd...7bf1051.html" title="Last updated on Jul 9, 2020, 2:50 PM UTC (7bf1051)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3086,7 +3086,8 @@ interface RTCPeerConnection : EventTarget  {
                         <p>Let <var>p</var> be a new promise.</p>
                       </li>
                       <li>
-                        <p data-tests="">In parallel, add the ICE candidate
+                        <p data-tests="">In parallel, if the candidate is not
+                          [= administratively prohibited =], add the ICE candidate
                         <var>candidate</var> as described in <span data-jsep=
                         "addicecandidate">[[!JSEP]]</span>. Use
                         <var>candidate</var>.{{RTCIceCandidate/usernameFragment}} to identify the
@@ -3120,6 +3121,7 @@ interface RTCPeerConnection : EventTarget  {
                           </li>
                           <li data-tests="RTCPeerConnection-addIceCandidate.html">
                             <p>If <var>candidate</var> is applied successfully,
+                              or if the candidate was [= administratively prohibited =]
                             the user agent MUST queue a task that runs the
                             following steps:</p>
                             <ol>
@@ -3153,6 +3155,16 @@ interface RTCPeerConnection : EventTarget  {
                     </ol>
                   </li>
                 </ol>
+                <p>
+                  A candidate is <dfn>administratively prohibited</dfn> if the
+                  UA has decided not to make connections to this address. An
+                  UA MUST NOT make connections to addresses on the
+                  [[!Fetch]] [= block bad port =] list, and MAY choose
+                  to not make connections to other addresses. For privacy
+                  reasons, there is no indication to the user about whether
+                  or not an address/port is blocked; it behaves exactly as
+                  if there was no response from the address.
+                </p>
                 <p class="note">Due to WebIDL processing, {{RTCPeerConnection/addIceCandidate}}(<code>null</code>)
                 is interpreted as a call with the default dictionary present, which, in the
                 above algorithm, indicates end-of-candidates for all media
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3153,6 +3155,25 @@ <h2>Methods</h2>
                     </ol>
                   </li>
                 </ol>
+                <p>
+                  A candidate is <dfn>administratively prohibited</dfn> if the
+                  UA has decided not to allow connection attempts to this
+                  address.
```
alvestrand, line 3161: Sentence break here; the definition ends here.

webrtc.html
```diff
@@ -3153,6 +3155,25 @@ <h2>Methods</h2>
...
+                <p>
+                  If the {{RTCConfiguration/iceTransportPolicy}} member of the {{RTCConfiguration}} is {{RTCIceTransportPolicy/relay}}, candidates requiring external
+                  resolution, such as mDNS candidates and DNS candidates,
+                  MUST be prohibited.
+                </p>
+                <p>
+                  For privacy
+                  reasons, there is no indication to the user about whether
+                  or not an address/port is blocked; it behaves exactly as
+                  if there was no response from the address.
```
alvestrand, line 3175: Move this para to definition.

webrtc.html
```diff
@@ -3153,6 +3155,25 @@ <h2>Methods</h2>
...
+                  address.
+                  The UA MUST prohibit connections to addresses on the
+                  [[!Fetch]] [= block bad port =] list, and MAY choose
+                  to not make connections to other addresses.
+                </p>
+                <p>
+                  If the {{RTCConfiguration/iceTransportPolicy}} member of the {{RTCConfiguration}} is {{RTCIceTransportPolicy/relay}}, candidates requiring external
+                  resolution, such as mDNS candidates and DNS candidates,
+                  MUST be prohibited.
+                </p>
```
alvestrand, line 3170: Consider rejecting these candidates rather than silently dropping them.

webrtc.html
dontcallmedom: I wouldn't use "user" here (which would typically be used for the end-user, not the developer); Fetch uses "developer", so I guess we could re-use that language here.

fippo: the developer might still be able to figure this out based on lack of candidate pair stats (which is what bit us re tcp) or lack of remote stats?

alvestrand: It's OK to let the developer learn whether or not there was a TURN server that accepted his credentials, or a peer that accepted his ufrag/pw, at that address.
What's not OK is to let the developer know the difference between a blocked port and a port where nobody's home. I think that's what this is trying to say.

PR #2543, opened 2020-06-19
Use github actions to simplify spec maintenance tasks

dontcallmedom:


dontcallmedom:
the "validate-spec" action is failing because of a bug in the spec which #2541 would fix - which I guess demonstrates the action works as intended :)

Issue #2516, opened 2020-04-23
Does a PC depend on the page context?


alvestrand:
In a recent bug, we discovered that code in RTCPeerConnection depended on having a valid execution context.
This is no longer true if, for instance, the RTCPeerConnection's "window" is an iframe that has been destroyed - but references to the RTCPeerConnection may survive the iframe it was created in.
Apparently, some objects work fine without an execution context; some others don't. So it's up to the specification to say whether or not they are context dependent, and what happens if the context dies.

Alternatives I see (spec-wise):
- When the execution context dies, the RTCPeerConnection is closed, a "signalingStateChange" event is fired at it, and no further operation on it is possible.
- When the execution context dies, the RTCPeerConnection remains active and usable.

This may be related to the question in #2513 (another lifecycle issue).

youennf:
> * When the execution context dies, the RTCPeerConnection is closed, a "signalingStateChange" event is fired at it, and no further operation on it is possible.

When context dies, we probably do not want to fire an event.

> * When the execution context dies, the RTCPeerConnection remains active and usable.

There is obviously a time where the peer connection would need to go away so I am not sure we want that either.

It would be interesting to test the existing behavior in current implementations.
I would think that Safari is basically neutering the RTCPeerConnection, as if close() was called.

youennf:
WebKit has some limited tests in that area: https://github.com/WebKit/webkit/blob/master/LayoutTests/webrtc/pc-detached-document.html
It seems a great area for new WPT tests.

youennf:
For the above WebKit test, Chrome passes the two tests, not Firefox.
The second test seems to indicate that Chrome considers a detached RTCPeerConnection as if it was closed.

alvestrand:
It's important to specify whether we run the close algorithm or just close/discard the PC.

henbos:
See also #2490. Conclusion: run the close steps (don't fire any events) when page goes away

henbos:
Right before onpageunload I guess?

alvestrand:
Suggestion: Add to https://w3c.github.io/webrtc-pc/#garbage-collection
- Rename section to "Object lifecycle concerns"
- Add a paragraph:

When the <find reference>controlling object (or whatever is the right name for it)</> is destroyed, execute the PeerConnection's "close" algorithm.

Note: Closing fires a dozen events on a dozen objects. If we want to block them all, we'll have to do a real audit of the text.

henbos:
[...]

PR #2552, opened 2020-07-01
Close peer connections (modulo firing ended) on page unload.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2516.

cc @annevk in case I'm using the wrong hook here.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -867,6 +867,21 @@
         before the procedures fire any events or invoke any callbacks,
         so the modifications are made visible at a single point in time.</p>
 
+        <p>As one of the <a>unloading document cleanup steps</a>, run the
+        following steps:</a>
+        <ol>
+          <li class="untestable">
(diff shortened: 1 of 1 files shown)


annevk:
That looks reasonable, though I suspect we might want to eventually turn this into a thing where HTML calls an algorithm in WebRTC to ensure consistent ordering.

jan-ivar:
@annevk there was some question about how this works with bfcache. Are pages even unloaded then? Safari apparently queuest a task to fire an ended event in these steps, and were hoping to see that task run on revival, e.g. if the user hits the back button.

annevk:
They are, yes. As part of https://html.spec.whatwg.org/#unload-a-document. I'm not entirely convinced all of this is defined correctly though and interoperable across browsers.

jan-ivar:
@youennf please file a follow-up if needed to fire the event on bfcache revival if you think it is not covered by this. We couldn't figure it out definitively one way or the other.

PR #2554, opened 2020-07-08
Default scaleResolutionDownBy to 4:2:1 and drop last layers

### LINKED ISSUE #2535: Spec says scaleResolutionDownBy must default to 1, but browsers differ
The [spec](https://w3c.github.io/webrtc-pc/#dom-rtcrtpencodingparameters-scaleresolutiondownby) says: _"By default, the sender will not apply any scaling, (i.e., scaleResolutionDownBy will be 1.0)"_.

This is correct in Firefox (78 using a [fiddle](https://fiddle.jshell.net/jib1/yvg81otc/45/show) where `scaleResolutionDownBy` is commented out):
* Three videos of the same size

But incorrect in Chrome ([crbug 1088200](https://bugs.chromium.org/p/chromium/issues/detail?id=1088200)):
* One large, one medium, and one small video.

Unfortunately, the incorrect behavior is clearly a more _useful_ default, making it a web compat issue. I'm filing this spec issue for visibility and clarification of intent. Do we fix Chrome or change the spec? And if we change the spec, then to what?

cc @alvestrand @guidou

### LINKED ISSUE #2537: How to truncate simulcast layers?
In #2535 we seem to say that order of layers should be 4:2:1, but we have language saying...

> Let supportedEncodings be the maximum number of encodings that the implementation can support. If the length of sendEncodings is greater than supportedEncodings, truncate sendEncodings so that its length is supportedEncodings.

Are we truncating the end or the beginning of the list? It would make more sense to cut off the start if its 4:2:1, but it would make sense to cut off the end if the order was 1:2:4

henbos:
Fixes #2537.
Fixes #2535.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2554.html" title="Last updated on Jul 8, 2020, 2:58 PM UTC (fe13d85)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2554/bd9e586...henbos:fe13d85.html" title="Last updated on Jul 8, 2020, 2:58 PM UTC (fe13d85)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -5612,7 +5612,7 @@ interface RTCCertificate {
                      <code>2^(length of <var>sendEncodings</var> - encoding
                      index - 1)</code>. This results in smaller-to-larger
                      resolutions where the last encoding has no scaling applied
-                     to it, e.g. 4:2:1 if <var>maxN</var> is 3.
+                     to it, e.g. 4:2:1 if the length is 3.
                    </li>
                    <li>
                      <p>If the number of {{RTCRtpEncodingParameters}} now
--- a/webrtc.html
+++ b/webrtc.html
@@ -1645,6 +1645,13 @@
                             <var>supportedEncodings</var>.
                           </li>
                           <li>
+                            If <var>sendEncodings</var> is non-empty, set each
+                            encoding's
+                            {{RTCRtpEncodingParameters/scaleResolutionDownBy}}
+                            to <code>2^(<var>supportedEncodings</var> - encoding
+                            index - 1)</code>.
+                          </li>
+                          <li>
                             <p>As described by <span data-jsep=
                             "applying-a-remote-desc">[[!JSEP]]</span>, attempt to
                             find an existing {{RTCRtpTransceiver}}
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1644,6 +1644,13 @@ <h4>Set the RTCSessionDescription</h4>
                             truncate <var>sendEncodings</var> so that its length is
                             <var>supportedEncodings</var>.
                           </li>
+                          <li>
+                            If <var>sendEncodings</var> is non-empty, set each
+                            encoding's
+                            {{RTCRtpEncodingParameters/scaleResolutionDownBy}}
+                            to <code>2^(<var>supportedEncodings</var> - encoding
+                            index - 1)</code>.
```
Orphis, line 1652: We cannot use the maximum number of encodings supported by the client for the scaling factors. You need to use the number of encodings instead.
A client supporting 3 layers would have very different results from another supporting 6 layers for example.

henbos: You're right I was thinking they were the same but that's only if truncation happens. Updated PR.

henbos:
As noted by Florent offline, if we want to support "please give me the default" after-the-fact in setParameters(), the algorithm needs to be repeated there too

PR #2555, opened 2020-07-08
Stop transports that are no longer used and fire events.

### LINKED ISSUE #2486: What happens to stopped transports?
We [expose](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-transport) _transceiver._**[[Sender]]**.**[[SenderTransport]]** to JS, which can then hold on to it:

```js
pc.onnegotiationneeded = () => { /* negotiation code */ };

const transport1 = pc.getTransceivers()[0].sender.transport;
const transport2 = pc.getTransceivers()[1].sender.transport;
```

There are a couple of ways to shut down a transport:
 1. `pc.getTransceivers()[0].stop()`
 2. `pc.setConfiguration({bundlePolicy: "max-bundle"})` if it were previously `"max-compat"`.

What, if anything should happen to `transport.state` and `transport2.state`?

Also, will events still fire on these objects? Will `getRemoteCertificates()` work or throw something?

I wasn't able to repro any of these circumstances in Chrome because of the lack of `transceiver.stop()` and https://bugs.chromium.org/p/chromium/issues/detail?id=1058826.

henbos:
Fixes #2486.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2555.html" title="Last updated on Jul 9, 2020, 2:21 PM UTC (58ecc88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2555/bd9e586...henbos:58ecc88.html" title="Last updated on Jul 9, 2020, 2:21 PM UTC (58ecc88)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1908,8 +1908,26 @@
                   </li>
                   <li data-tests="RTCPeerConnection-onnegotiationneeded.html">
                     <p>If <var>connection</var>'s [= signaling state =] is now
-                    {{RTCSignalingState/"stable"}}, [= clear the negotiation-needed
-                    flag =] and [= update the negotiation-needed flag =].
+                    {{RTCSignalingState/"stable"}}, run the following steps:</p>
+                    <ol>
+                      <li>
+                        <p>For any <var>transceiver</var> that was removed from
+                        the [= set of transceivers =] in a previous step, if any
+                        of its transports
+                        (<var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SenderTransport]]</a>
+                        or <var>transceiver</var>.<a>[[\Receiver]]</a>.<a>[[\ReceiverTransport]]</a>)
+                        are no longer referenced by a non-stopped transceiver,
+                        close the {{RTCDtlsTransport}}s and their associated
+                        {{RTCIceTransport}}s. This results in events firing on
+                        these objects in a queued task.</p>
+                        </li>
+                      </p>
+                      </li>
+                      <li data-tests="RTCPeerConnection-onnegotiationneeded.html">
+                        <p>[= Clear the negotiation-needed flag =] and
+                        [= update the negotiation-needed flag =].</p>
+                      </li>
+                    </ol>
                   </li>
                   <li>
                     <p>If <var>connection</var>'s [= signaling state =]
--- a/webrtc.html
+++ b/webrtc.html
@@ -1916,8 +1916,9 @@
                         of its transports
                         (<var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SenderTransport]]</a>
                         or <var>transceiver</var>.<a>[[\Receiver]]</a>.<a>[[\ReceiverTransport]]</a>)
-                        are no longer referenced by a non-stopped transceiver,
-                        close the {{RTCDtlsTransport}}s and their associated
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -1908,8 +1908,25 @@ <h4>Set the RTCSessionDescription</h4>
...
-                    {{RTCSignalingState/"stable"}}, [= clear the negotiation-needed
-                    flag =] and [= update the negotiation-needed flag =].
+                    {{RTCSignalingState/"stable"}}, run the following steps:</p>
+                    <ol>
+                      <li>
+                        <p>For any <var>transceiver</var> that was removed from
+                        the [= set of transceivers =] in a previous step, if any
+                        of its transports
+                        (<var>transceiver</var>.<a>[[\Sender]]</a>.<a>[[\SenderTransport]]</a>
+                        or <var>transceiver</var>.<a>[[\Receiver]]</a>.<a>[[\ReceiverTransport]]</a>)
```
jan-ivar, line 1918: Unsure about "for any". I worry it's not clear how many times we close the same transport, or what order events fire. Maybe I'm being pedantic but how about:

- Let *transports* be an empty set.
- For each _transceiver_ that was removed from the _set of transceivers_ in a previous step, add  _transceiver_.[[Sender]].[[SenderTransport]] and _transceiver_.[[Receiver]].[[ReceiverTransport]] to _transports_.
- For each _transport_ in _transports_, **close the RTCDtlsTransport** _transport_.

Then add a **close the RTCDtlsTransport** algorithm somewhere near [here](https://w3c.github.io/webrtc-pc/#dfn-dtlstransportstate) that takes care of firing `statechange` and then queues a task to close its `transport.iceTransport` (or maybe in the same task like https://github.com/w3c/webrtc-pc/pull/2444?) 

Then to avoid confusion, amend this text [near here](https://w3c.github.io/webrtc-pc/#dfn-dtlstransportstate) (addition in **bold**):

"When the underlying DTLS transport needs to update the state of the corresponding RTCDtlsTransport object for any other reason, **and the new state is not `"closed"`,**"

henbos: Due to a lack of time I did a lazy version of making sure its not closed twice: I added the caveat "are still not closed". But I can do the pedantic version of this later if you think it is needed.

webrtc.html
```diff
@@ -1908,8 +1908,26 @@ <h4>Set the RTCSessionDescription</h4>
                   </li>
                   <li data-tests="RTCPeerConnection-onnegotiationneeded.html">
                     <p>If <var>connection</var>'s [= signaling state =] is now
-                    {{RTCSignalingState/"stable"}}, [= clear the negotiation-needed
-                    flag =] and [= update the negotiation-needed flag =].
+                    {{RTCSignalingState/"stable"}}, run the following steps:</p>
+                    <ol>
+                      <li>
```
alvestrand, line 1913: add needs-test

webrtc.html
```diff
@@ -1908,8 +1908,26 @@ <h4>Set the RTCSessionDescription</h4>
...
+                        <p>For any <var>transceiver</var> that was removed from
```
henbos, line 1914:
```suggestion
                        <p class="needs-test">For any <var>transceiver</var> that was removed from
```

PR #2597, opened 2020-11-05
Allow piggybacking getCapabilities on most recent offer or answer

### LINKED ISSUE #2460: `getCapabilities` seems to leak hardware capabilities w/o a permission
Moved from the WebRTC-SVC repo: https://github.com/w3c/webrtc-svc/issues/22

Opened by snyderp

Apologies if I'm misreading the spec, but if I'm reading it correctly it looks like a site can learn about the visitors underlying hardware capabilities w/o a permission prompt or some other positive, affirmative action by the visitor.

Is my reading of the spec correct then, there is a FP vector exposed by the current text that would need to be mitigated (e.g. sites couldn't access it by default).

Otherwise, if this is addressed elsewhere, could you kindly point me to where, so I dont make the same mistake twice? :) Thanks!

henbos:
(Issues have been moved around, add a Fixes <issue link> here before landing.)

While the fingerprint note previously already said the capabilities returned could be a subset of the real capabilities in privacy-sensitive contexts, this PR makes it clearer that this can be even more true prior to calling createOffer or createAnswer.

This also resolves any synchronicity issues by allowing the result to be incomplete if the user agent doesn't know some capabilities yet (e.g. may require accessing hardware).

This PR also makes the requirements of this API a little bit clearer by saying that after creating an offer or answer, the capabilities returned MUST match the SDP.

This attempts to clarify two things:
- Allow getCapabilities not to leak anything that we are not already leaking in createOffer.
- By "piggybacking" on createOffer, we get around the limitations of this synchronous API by encouraging you to have already awaited an asynchronous API.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2597.html" title="Last updated on Dec 2, 2020, 11:24 AM UTC (47e9097)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2597/60f2135...henbos:47e9097.html" title="Last updated on Dec 2, 2020, 11:24 AM UTC (47e9097)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -10133,7 +10133,7 @@ interface RTCRtpReceiver {
                   {{RTCPeerConnection/createAnswer}} there MAY be capabilities
                   that are either unknown when the synchronous
                   {{getCapabilities()}} call is made, or that the user agent
-                  choses not to disclose for other reasons (e.g. privacy). If
+                  chooses not to disclose for other reasons (e.g. privacy). If
                   {{getCapabilities()}} is called after
                   {{RTCPeerConnection/createOffer}} or
                   {{RTCPeerConnection/createAnswer}} has resolved, the
--- a/webrtc.html
+++ b/webrtc.html
@@ -8774,6 +8774,18 @@ interface RTCRtpSender {
                   privacy-sensitive contexts, browsers can consider mitigations
                   such as reporting only a common subset of the capabilities.
                 </p>
+                <p class="no-test-needed">
+                  Prior to calling {{RTCPeerConnection/createOffer}} or
+                  {{RTCPeerConnection/createAnswer}} there MAY be capabilities
+                  that are either unknown when the synchronous
+                  {{getCapabilities()}} call is made, or that the user agent
+                  choses not to disclose for other reasons (e.g. privacy). If
+                  {{getCapabilities()}} is called after
+                  {{RTCPeerConnection/createOffer}} or
+                  {{RTCPeerConnection/createAnswer}} has resolved, the
+                  capabilities returned by this method MUST match that of the
+                  most recently created offer or answer.
+                </p>
               </dd>
               <dt data-tests=
               "RTCRtpParameters-codecs.html,RTCRtpSender-setParameters.html">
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -8774,6 +8774,18 @@ <h2>
                   privacy-sensitive contexts, browsers can consider mitigations
                   such as reporting only a common subset of the capabilities.
                 </p>
+                <p class="no-test-needed">
+                  Prior to calling {{RTCPeerConnection/createOffer}} or
+                  {{RTCPeerConnection/createAnswer}} there MAY be capabilities
+                  that are either unknown when the synchronous
+                  {{getCapabilities()}} call is made, or that the user agent
+                  choses not to disclose for other reasons (e.g. privacy). If
```
jan-ivar, line 8782:
```suggestion
                  chooses not to disclose for other reasons (e.g. privacy). If
```

webrtc.html
```diff
@@ -10116,6 +10128,18 @@ <h2>
...
```
jan-ivar, line 10136:
```suggestion
                  chooses not to disclose for other reasons (e.g. privacy). If
```

webrtc.html
```diff
@@ -8774,6 +8774,18 @@ <h2>
...
+                  Prior to calling {{RTCPeerConnection/createOffer}} or
+                  {{RTCPeerConnection/createAnswer}} there MAY be capabilities
+                  that are either unknown when the synchronous
+                  {{getCapabilities()}} call is made, or that the user agent
+                  chooses not to disclose for other reasons (e.g. privacy). If
+                  {{getCapabilities()}} is called after
+                  {{RTCPeerConnection/createOffer}} or
+                  {{RTCPeerConnection/createAnswer}} has resolved, the
+                  capabilities returned by this method MUST match that of the
+                  most recently created offer or answer.
```
henbos, line 8787: Note that before this PR it is spec-compliant to implement getCapabilities() as "return {};". After this PR, that is not spec-compliant.

To address the concern about getCapabilities() being static, but offers and answers are on a per-peer connection basis, how about instead of saying "capabilities returned by this method MUST match that of the most recently created offer or answer" it can say "capabilities returned by this method MUST match _or be a superset_ of _any previously_ created offer or answer _by any peer connections in the document_".

1. This is consistent with the method's description that this is an "optimistic view" of the capabilities.
2. Calling getCapabilities() after calling createOffer() could theoretically add to the list of the capabilities previously returned, but it could _never subtract from it_.
3. Because you cannot have capabilities exposed in SDP that are not also included in getCapabilities(), this PR means you no longer* have to parse SDP to get a full set of capabilities that your peer connection exposes.
4. In practise, no browser has to update their implementation.

Today's fingerprinting note **already** makes getCapabilities() a foot-gun, but this PR makes it slightly less of a foot-gun by adding some requirement to what it MUST return.

_* In practise of course you don't have to do this, but a browser could return a subset of capabilties from what is exposed in SDP and still be compliant._

Tagging @jan-ivar @youennf for thoughts but this can wait until the interim, no rush

henbos: With regards to 2), I guess the "optimistic view" of capabilities would allow you to actually be over-optimistic before createOffer(), then do createOffer() and find out you don't have a HW encoder, and then remove that from the set of capabilities. But my point is you should never have something in the SDP that is not also in the a later getCapabilities() call.

henbos:
To be revisited after interim discussions, this PR might not reflect the final proposal yet:
https://docs.google.com/presentation/d/17Z_vR4pOXn-9wthiqbM9GC7JE7vfbdNffEedBPqj1Mw/edit#slide=id.gad716a442e_1_0

henbos:
I'll abandon this one and make a new PR

Issue #1477, opened 2017-07-13
Minor inconsistencies around RTCSessionDescription vs. RTCSessionDescriptionInit


taylor-b:
I noticed some minor issues related to the use of `RTCSessionDescription` (the interface) and `RTCSessionDescriptionInit` (the dictionary):

- There's a section called "Set the RTCSessionSessionDescription". That's one too many "Session"s.
- This section talks about setting an `RTCSessionDescription`, but it's actually an `RTCSessionDescriptionInit` that goes into `setLocalDescription` and `setRemoteDescription`. Is there a missing "construct `RTCSessionDescription` from `RTCSessionDescriptionInit` step"?
- If we do go that route, then the constructor of `RTCSessionDescription` is no longer just for legacy purposes.

It may also be nice to explain somewhere in human-readable language _why_ we have two things. Is it just for legacy purposes, to support the following?

`pc.setLocalDescription(new RTCSessionDescription({sdp: sdpString, type: "offer"}));`

But does this work? `setLocalDescription` and `setRemoteDescription` no longer accept `RTCSessionDescription` as arguments. Or do they, because an `RTCSessionDescription` is capable of being converted to an `RTCSessionDescriptionInit`?

jan-ivar:
The reason we have two things is that dictionaries cannot be attributes:
```js
readonly attribute RTCSessionDescription? localDescription;
```
...yet at the same time we want to write:
```js
pc.setLocalDescription({sdp: sdpString, type: "offer"});
```
> an RTCSessionDescription is capable of being converted to an RTCSessionDescriptionInit?

👍
The former browser object is valid input to a method expecting the latter dictionary, but not vise versa.

Conceptually they're interchangeable. Unfortunately, the better name is taken by the legacy constructor. We have prose that mixes concept and type in places, like:

*"The initial value of `mid` is null. Setting a new `RTCSessionDescription` may change it to a non-null value, as defined in [JSEP] (section 5.5. and section 5.6.)."*

In this case, [section 5.5](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-20#section-5.5) actually uses "SessionDescription" (though [section 4.1.8](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-20#section-4.1.8) suggests this means `RTCSessionDescription`). In either case, it sounds odd to say:

[...]

PR #2614, opened 2020-12-10
Disambiguate all uses of RTCSessionDescription.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/1477.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2614.html" title="Last updated on Dec 10, 2020, 2:04 PM UTC (a3ade6c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2614/610c8da...jan-ivar:a3ade6c.html" title="Last updated on Dec 10, 2020, 2:04 PM UTC (a3ade6c)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2698,14 +2698,15 @@
                           {{RTCSdpType/"rollback"}}) run the following steps:
                         </p>
                         <ol>
-                          <p>
-                            Let <var>pendingDescription</var> be either
-                            <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
-                            or
-                            <var>connection</var>.<a>[[\PendingRemoteDescription]]</a>,
-                            whichever one is not <code>null</code>.
-                          </p>
-
+                          <li>
+                            <p>
+                              Let <var>pendingDescription</var> be either
+                              <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
+                              or
+                              <var>connection</var>.<a>[[\PendingRemoteDescription]]</a>,
+                              whichever one is not <code>null</code>.
+                            </p>
+                          </li>
                           <li>
                             <p>
                               For each <var>transceiver</var> in the
@@ -3215,7 +3216,7 @@ interface RTCPeerConnection : EventTarget  {
   constructor(optional RTCConfiguration configuration = {});
   Promise&lt;RTCSessionDescriptionInit&gt; createOffer(optional RTCOfferOptions options = {});
   Promise&lt;RTCSessionDescriptionInit&gt; createAnswer(optional RTCAnswerOptions options = {});
-  Promise&lt;undefined&gt; setLocalDescription(optional OptionalRTCSessionDescriptionInit description = {});
+  Promise&lt;undefined&gt; setLocalDescription(optional RTCLocalSessionDescriptionInit description = {});
   readonly attribute RTCSessionDescription? localDescription;
   readonly attribute RTCSessionDescription? currentLocalDescription;
   readonly attribute RTCSessionDescription? pendingLocalDescription;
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2696,6 +2698,14 @@ <h4>
                           {{RTCSdpType/"rollback"}}) run the following steps:
                         </p>
                         <ol>
+                          <p>
```
dontcallmedom, line 2701:
```suggestion
                          <li><p>
```

webrtc.html
```diff
@@ -2696,6 +2698,14 @@ <h4>
...
+                            Let <var>pendingDescription</var> be either
+                            <var>connection</var>.<a>[[\PendingLocalDescription]]</a>
+                            or
+                            <var>connection</var>.<a>[[\PendingRemoteDescription]]</a>,
+                            whichever one is not <code>null</code>.
+                          </p>
```
dontcallmedom, line 2707:
```suggestion
                          </p></li>
```

webrtc.html
```diff
@@ -3207,7 +3215,7 @@ <h3>
   constructor(optional RTCConfiguration configuration = {});
   Promise&lt;RTCSessionDescriptionInit&gt; createOffer(optional RTCOfferOptions options = {});
   Promise&lt;RTCSessionDescriptionInit&gt; createAnswer(optional RTCAnswerOptions options = {});
-  Promise&lt;undefined&gt; setLocalDescription(optional RTCLocalSessionDescriptionInit description = {});
+  Promise&lt;undefined&gt; setLocalDescription(optional OptionalRTCSessionDescriptionInit description = {});
```
dontcallmedom, line 3218: dictionary names are not exposed to JS developers, but they do take a slot in the IDL space - I'd be reluctant to make such a change at this stage.

If we we were to change it, I think it would need to be `RTCOptionalSessionDescriptionInit` (since pretty much all of our IDL names are prefixed with RTC)

dontcallmedom (changes requested):
looks good directionally, but with some changes needed

webrtc.html
henbos: I don't have a strong preference here, but since the name doesn't matter much and it's only used for the SLD call, I suggest we keep the old name and merge this editorial PR.

henbos (approved):
There are some comments but happy with "editors can integrate"

dontcallmedom (approved):

henbos (approved):

henbos:
The chairs/etc will have to decide what to do with editorial issues this point in time.

jan-ivar:
@dontcallmedom Should we merge these editorial PRs to get them in, or hold off?

henbos:
We thought it looked good but the concern came up of "what if we accidentally introduce a bug before the cut" so we need some stance on what to do with minor editorial changes this late in the process

dontcallmedom:
from a process perspective merging them in is OK; but the concerns about introducing late bugs in what looks editorial is also reasonable. My guess is that if enough eyeballs have taken a hard look at the PR, then merging should be OK.

jan-ivar:
> My guess is that if enough eyeballs have taken a hard look at the PR, then merging should be OK.

Thanks @dontcallmedom and @henbos. I've looked over the PR again and I found no mistakes.
 * I did a full pass of the [Files changed](https://github.com/w3c/webrtc-pc/pull/2614/files) diff.
 * I followed all _"session description"_ links in the [preview](https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2614.html).
 * I also did the same in the [preview diff](https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2614/610c8da...jan-ivar:a3ade6c.html) though I found it unreliable (it trips over webidl that is unchanged by this patch)
 * I checked the new _"pendingDescription"_ variable added instead of referring to the description _"being rolled back"_.
 * I checked the three apply-rewordings to avoid the phrases _"prior to applying"_ and _"is applied"_ wrt the setting algorithms (replaced by _"prior to_ pendingDescription _being set"_ and _"is set successfully"_ to denote more accurately what is being referred to: the former refers to pending descriptions, the latter refers to being set as current description, which frankly remains vague but is thankfully non-normative).

I think this is an improvement., so I'd like to get this in before the Dec 15th deadline. @alvestrand and @aboba can I get an `Editors can integrate`?

jan-ivar:
> which frankly remains vague but is thankfully non-normative

Actually, it's not vague because an `"answer"` can never be pending (that would be `"pranswer"`), so this is good.

henbos:
I am happy with editors can integrate but I have already approved it. So can @alvestrand or @aboba give their thumbs up or down?

alvestrand (approved):

jan-ivar:
Thanks @alvestrand I take that as a sign to go ahead.

PR #2617, opened 2020-12-14
Add note to getCapabilities to avoid InvalidModificationError

### LINKED ISSUE #2616: Add clarifying note to getCapabilities() to prevent capabilities being removed
See slides https://docs.google.com/presentation/d/1CeNeno5XuDhm1mpnVyE9eT14YKZgZUtgQsJfC8uqEpA/edit#slide=id.gad716a442e_1_0 for more information

henbos:
Fixes #2616

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/2617.html" title="Last updated on Dec 15, 2020, 2:46 AM UTC (384df80)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2617/9915912...henbos:384df80.html" title="Last updated on Dec 15, 2020, 2:46 AM UTC (384df80)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8780,7 +8780,8 @@ interface RTCRtpSender {
                     {{RTCRtpTransceiver/setCodecPreferences()}} algorithm and
                     what inputs it throws {{InvalidModificationError}} on,
                     and should also be consistent with information revealed by
-                    {{createOffer}} and {{createAnswer}} to ensure any
+                    {{createOffer}} and {{createAnswer}} about codecs
+                    negotiated for sending, to ensure any
                     privacy mitigations are effective.
                   </p>
                 </div>
--- a/webrtc.html
+++ b/webrtc.html
@@ -10132,7 +10132,8 @@ interface RTCRtpReceiver {
                     {{RTCRtpTransceiver/setCodecPreferences()}} algorithm and
                     what inputs it throws {{InvalidModificationError}} on,
                     and should also be consistent with information revealed by
-                    {{createOffer}} and {{createAnswer}} to ensure any
+                    {{createOffer}} and {{createAnswer}} about codecs
+                    negotiated for reception, to ensure any
                     privacy mitigations are effective.
                   </p>
                 </div>
--- a/webrtc.html
+++ b/webrtc.html
@@ -8778,7 +8778,7 @@ interface RTCRtpSender {
                   <p class="no-test-needed">
                     The capabilities returned need to be consistent between
                     invocations from the same document to ensures that codecs
-                    returned can be used in
+                    returned can be used as arguments of
                     {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
                     InvalidModificationError.
                   </p>
@@ -10128,7 +10128,7 @@ interface RTCRtpReceiver {
                   <p class="no-test-needed">
                     The capabilities returned need to be consistent between
                     invocations from the same document to ensures that codecs
-                    returned can be used in
+                    returned can be used as arguments of
                     {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
                     InvalidModificationError.
                   </p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -10127,11 +10127,12 @@ interface RTCRtpReceiver {
                 </p>
                 <div class="note">
                   <p>
-                    The capabilities returned need to be consistent between
-                    invocations from the same document to ensures that codecs
-                    returned can be used as arguments to
-                    {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
-                    {{InvalidModificationError}}.
+                    The codec capabilities returned affect the
+                    {{RTCRtpTransceiver/setCodecPreferences()}} algorithm and
+                    what inputs it throws {{InvalidModificationError}} on,
+                    and should also be consistent with information revealed by
+                    {{createOffer}} and {{createAnswer}} to ensure any
+                    privacy mitigations are effective.
                   </p>
                 </div>
               </dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -8775,12 +8775,12 @@ interface RTCRtpSender {
                   such as reporting only a common subset of the capabilities.
                 </p>
                 <div class="note">
-                  <p class="no-test-needed">
+                  <p>
                     The capabilities returned need to be consistent between
                     invocations from the same document to ensures that codecs
                     returned can be used as arguments to
                     {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
-                    InvalidModificationError.
+                    {{InvalidModificationError}}.
                   </p>
                 </div>
               </dd>
@@ -10125,12 +10125,12 @@ interface RTCRtpReceiver {
                   such as reporting only a common subset of the capabilities.
                 </p>
                 <div class="note">
-                  <p class="no-test-needed">
+                  <p>
                     The capabilities returned need to be consistent between
                     invocations from the same document to ensures that codecs
                     returned can be used as arguments to
                     {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
-                    InvalidModificationError.
+                    {{InvalidModificationError}}.
                   </p>
                 </div>
               </dd>
--- a/webrtc.html
+++ b/webrtc.html
@@ -8774,6 +8774,15 @@ interface RTCRtpSender {
                   privacy-sensitive contexts, browsers can consider mitigations
                   such as reporting only a common subset of the capabilities.
                 </p>
+                <div class="note">
+                  <p class="no-test-needed">
+                    The capabilities returned need to be persistent between
+                    invocations from the same document to ensures that codecs
+                    returned can be used in
+                    {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
+                    InvalidModificationError.
+                  </p>
+                </div>
               </dd>
               <dt data-tests=
               "RTCRtpParameters-codecs.html,RTCRtpSender-setParameters.html">
@@ -10115,6 +10124,15 @@ interface RTCRtpReceiver {
                   privacy-sensitive contexts, browsers can consider mitigations
                   such as reporting only a common subset of the capabilities.
                 </p>
+                <div class="note">
+                  <p class="no-test-needed">
+                    The capabilities returned need to be persistent between
+                    invocations from the same document to ensures that codecs
+                    returned can be used in
+                    {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
+                    InvalidModificationError.
+                  </p>
+                </div>
               </dd>
               <dt data-tests="RTCRtpReceiver-getParameters.html">
                 <dfn data-idl="">getParameters</dfn>
--- a/webrtc.html
+++ b/webrtc.html
@@ -8778,7 +8778,7 @@ interface RTCRtpSender {
                   <p class="no-test-needed">
                     The capabilities returned need to be consistent between
                     invocations from the same document to ensures that codecs
-                    returned can be used as arguments of
+                    returned can be used as arguments to
                     {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
                     InvalidModificationError.
                   </p>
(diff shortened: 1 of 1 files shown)


henbos (commented):
Please take a look @dontcallmedom @alvestrand @jan-ivar @aboba and feel free to edit-and-merge if necessary

dontcallmedom (approved):

alvestrand (approved):
A minimal addition, which I think is the Right Thing to do.

aboba:
It's a bit too minimal. The slides were more clear. Not sure what "consistent" means in this context.

henbos:
> It's a bit too minimal. The slides were more clear. Readers could misconstrue "consistent" to mean "the same".

I was a bit vague on purpose since we're not adding any "MUSTs" this late in the game.

Alternatively we could say that you are not allowed to remove any capability from the set of capabilities after exposing them and still allow adding to the set of capabilities when finding out about more hardware. So if you were overly-optimistic in the past then you have to be overly optimistic in future calls as well, even after finding out a codec doesn't exist. This would prevent IllegalModificationError while still allowing more flexibility by the implementation.

However if we give an implementation the flexibility to add to the set of capabilities over time, then:
1. You're better off parsing SDP than calling getCapabilities().
2. The usefulness of getCapabilities() is implementation-specific and not interoperable.
3. getCapabilities() is supposed to be "the most optimistic view of the capabilities" but if it under-reports initially that does not sound very optimistic.

So I'm not sure interpreting "consistent" as "the same" is a bad idea, but I'm also not sure we want to explicitly say that "the same" is the only valid option without revisiting this in webrtc-extensions.

So... this is hairy. Feel free to edit this PR.

jan-ivar (commented):
@henbos I don't see anything in the existing spec suggesting "the same", so I don't think a note can say that.

In fact, I think that suggests it can change, which seems more useful. E.g. we can polyfill a later fix:
```js
RTCRtpSender.getCapabilitiesAsync = async function() {
  await new RTCPeerConnection().createOffer();
  RTCSender.getCapabilities(arguments);
}
```
Since `getCapabilities` is synchronous and so trivial to call, why respect results that predate the current run of the event loop that calls the synchronous [setCodecPreferences](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setcodecpreferences)? The algorithm itself doesn't seem to:

* "If the intersection between codecs and RTCRtpSender.getCapabilities(kind).codecs or the intersection between codecs and RTCRtpReceiver.getCapabilities(kind).codecs only contains RTX, RED or FEC codecs or is an empty set, throw InvalidModificationError."

From my reading, this isn't referring to the result of some earlier invocation of that method, but to the result of an immediate (actual or hypothetical) invocation of that method right now.

> getCapabilities() is supposed to be "the most optimistic view of the capabilities"

The word "optimistic" to me means callers may always get `InvalidModificationError` and have to handle it anyway, so no invariant there.

> Alternatively we could say that you are not allowed to remove any capability from the set of capabilities after exposing them ... This would prevent IllegalModificationError 

But if the user agent had a good reason to remove it, i.e. it's no longer available, then don't we prefer `IllegalModificationError` to spurious errors later?

What exactly are we concerned about here? User agents removing perfectly fine codecs for privacy reasons? I doubt such mitigations would remove things already exposed to JS, as that defeats the purpose. Also, such privacy mitigations would have to apply to `createOffer` and `createAnswer` as well to be effective, or someone will just shim a better getCapabilities using createOffer and an SDP library.

I think a note merely needs to remind implementers to keep `getCapabilities` and `createOffer`/`createAnswer` consistent.

webrtc.html
```diff
@@ -8774,6 +8774,15 @@ <h2>
                   privacy-sensitive contexts, browsers can consider mitigations
                   such as reporting only a common subset of the capabilities.
                 </p>
+                <div class="note">
+                  <p>
+                    The capabilities returned need to be consistent between
+                    invocations from the same document to ensures that codecs
+                    returned can be used as arguments to
+                    {{RTCRtpTransceiver/setCodecPreferences()}} without risk of
+                    {{InvalidModificationError}}.
```
jan-ivar, line 8783:
```suggestion
                    The codec capabilities returned affect the
                    {{RTCRtpTransceiver/setCodecPreferences()}} algorithm and
                    what inputs it throws {{InvalidModificationError}} on,
                    and should also be consistent with information revealed by
                    {{createOffer}} and {{createAnswer}} to ensure any
                    privacy mitigations are effective.
```

webrtc.html
```diff
@@ -10115,6 +10124,15 @@ <h2>
...
```
jan-ivar, line 10133:
```suggestion
                    The codec capabilities returned affect the
                    {{RTCRtpTransceiver/setCodecPreferences()}} algorithm and
                    what inputs it throws {{InvalidModificationError}} on,
                    and should also be consistent with information revealed by
                    {{createOffer}} and {{createAnswer}} to ensure any
                    privacy mitigations are effective.
```

jan-ivar (commented):
Here's my suggestion for a note, based on my comments above.

webrtc.html
aboba: This makes more sense to me.

webrtc.html
aboba: I like it!

aboba:
In general, I like Jan-Ivar's suggested direction.  My question is whether it is perhaps ambiguous to ask for consistency between `getCapabilities` and the SDP in **both** `createOffer`/`createAnswer`.  The SDP from these methods can be different, right? Is the **last** of these method called definitive?

For example, I interpret "consistency" to mean that `RTCRtpSender.getCapabilities()` includes the codecs included on the SDP m-lines specified as `send/recv` or `sendonly`, and `RTCRtpReceiver.getCapabilities()` includes the codecs included on the SDP m-lines specified as `send/recv` or `recvonly`.  So directionality matters.

jan-ivar:
@aboba That sounds right. I kept it vague using "consistent with information revealed by"... But since we have two notes, perhaps we can differentiate direction somehow.

webrtc.html
```diff
@@ -8774,6 +8774,16 @@ <h2>
                   privacy-sensitive contexts, browsers can consider mitigations
                   such as reporting only a common subset of the capabilities.
                 </p>
+                <div class="note">
+                  <p>
+                    The codec capabilities returned affect the
+                    {{RTCRtpTransceiver/setCodecPreferences()}} algorithm and
+                    what inputs it throws {{InvalidModificationError}} on,
+                    and should also be consistent with information revealed by
+                    {{createOffer}} and {{createAnswer}} to ensure any
```
jan-ivar, line 8783:
```suggestion
                    {{createOffer}} and {{createAnswer}} about codecs
                    negotiated for sending, to ensure any
```

webrtc.html
```diff
@@ -10115,6 +10125,16 @@ <h2>
...
```
jan-ivar, line 10134:
```suggestion
                    {{createOffer}} and {{createAnswer}} about codecs
                    negotiated for reception, to ensure any
```

webrtc.html
```diff
@@ -8774,6 +8774,17 @@ <h2>
                   privacy-sensitive contexts, browsers can consider mitigations
                   such as reporting only a common subset of the capabilities.
                 </p>
+                <div class="note">
+                  <p>
+                    The codec capabilities returned affect the
+                    {{RTCRtpTransceiver/setCodecPreferences()}} algorithm and
+                    what inputs it throws {{InvalidModificationError}} on,
+                    and should also be consistent with information revealed by
+                    {{createOffer}} and {{createAnswer}} about codecs
```
jan-ivar, line 8783:
```suggestion
                    {{RTCPeerConnection/createOffer()}} and {{RTCPeerConnection/createAnswer()}} about codecs
```

webrtc.html
```diff
@@ -10115,6 +10126,17 @@ <h2>
...
```
jan-ivar, line 10135:
```suggestion
                    {{RTCPeerConnection/createOffer()}} and {{RTCPeerConnection/createAnswer()}} about codecs
```

jan-ivar:
@aboba does that work? 

@alvestrand @henbos feel free to merge when you wake up if you agree, so we can get this in in time.

cc @dontcallmedom @koalie

henbos:
> @henbos I don't see anything in the existing spec suggesting "the same", so I don't think a note can say that.

The sameness is not called out in the spec, but is inferred from the intended use cases of the APIs. For example, we have recommended the Perfect Negotiation pattern. But if the user of this API has to do a dance of calling negotiation methods before getCapabilities() works, then we can no longer "abstract away negotiation" from the application, and methods like setCodecPreferences() become "special" methods.

First of all, you are entirely correct that this isn't called out in the spec. But let me remind everybody that there is also nothing suggesting that I have to call createOffer() in order for getCapabilities() to work. So why would you? For example, it would be reasonable to expect the following code snippet to work:

```
// Setup code.
pc.onnegotiationneeded = (perfect negotiation pattern);
// Application code.
const transceiver = pc.addTransceiver("video");
let codecs = getCapabilities().codecs;
codecs = getCodecsWithH264First(codecs);
transceiver.setCodecPreferences(codecs);
```

Or, here's another example, that doesn't even involve Perfect Negotiation, but where SDP wouldn't have been created by your peer connection before getCapabilities():

```
await pc.setRemoteDescription(remoteOffer);
// Transceiver created by offer.
const transceiver = pc.getTransceivers()[0];
let codecs = getCapabilities().codecs;
codecs = getCodecsWithH264First(codecs);
transceiver.setCodecPreferences(codecs);
const answer = await pc.createAnswer();
...
```

All of this being said though, I agree with the sentiments:

> Since getCapabilities is synchronous and so trivial to call, why respect results that predate the current run of the event loop that calls the synchronous setCodecPreferences?

This is good. I can't think of any use case where you would have to remember "old" codecs. And if it is only to be used within the same event loop cycle, this would be consistent with get/setParameters() semantics.

However, we cannot simply assume that createOffer() or createAnswer() has already been called. It might have been setRemoteDescription() that was the first call.

henbos (commented):
Without introducing normative steps, I think this is about as good as it gets. Because this doesn't explicitly say WHEN createAnswer() is called, I think this note could even imply a working use case of setRemoteDescription()+getCapabilities()+createAnswer().

However, to really nail this, we need webrtc-extensions. But for now, let's merge?

henbos:
Alright then, I'm merging, we can follow up in webrtc-extensions. For now it's "just a note" and I think it ended up being a pretty good one without having to touch any normative steps.

jan-ivar:
> But if the user of this API has to do a dance of calling negotiation methods before getCapabilities() works, then we can no longer "abstract away negotiation" from the application

`createOffer`, unlike `setLocalDescription`, doesn't change signaling state, so it shouldn't interfere with negotiation, should it? Using my polyfill [above](https://github.com/w3c/webrtc-pc/pull/2617#pullrequestreview-551961464) to illustrate a workaround:
```
// Setup code.
await RTCRtpSender.getCapabilitiesAsync();
pc.onnegotiationneeded = (perfect negotiation pattern);
// Application code.
const transceiver = pc.addTransceiver("video");
transceiver.setCodecPreferences(getCodecsWithH264First(RTCRtpSender.getCapabilities().codecs));
```

> there is also nothing suggesting that I have to call createOffer() in order for getCapabilities() to work

True, though I don't think this impacts interpretation, since specs are written for implementers, not users. And following it leads us here. We could add a second note saying we messed up, and here's a workaround.

henbos:
> True, though I don't think this impacts interpretation, since specs are written for implementers, not users. And following it leads us here. We could add a second note saying we messed up, and here's a workaround.

Following it leads us here, but only after several editors and implementers discussing ways to avoid what I think is a less desirable but more straight-forward interpretation, which would be to have getCapabilities() block JS and always expose the same capabilities (HW or SW, known after block). It would even be more useful, but also taboo. Even with our note this is still a valid interpretation and having to do said workaround is not clear even to implementers. But I say let's follow-up in https://github.com/w3c/webrtc-extensions/issues/57, for now the note serves as a "heads up, there be dragons here".

PR #2636, opened 2021-03-22
simulcast example: remove signalling offer handling

fippo:
which does not apply to this example

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-pc/pull/2636.html" title="Last updated on Mar 22, 2021, 11:02 AM UTC (7a1bdc5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2636/5568f10...fippo:7a1bdc5.html" title="Last updated on Mar 22, 2021, 11:02 AM UTC (7a1bdc5)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -16278,11 +16278,6 @@ signaling.onmessage = async ({data: {description, candidate}}) =&gt; {
   try {
     if (description) {
       await pc.setRemoteDescription(description);
-      // if we got an offer, we need to reply with an answer
-      if (description.type == 'offer') {
-        await pc.setLocalDescription();
-        signaling.send({description: pc.localDescription});
-      }
     } else if (candidate) {
       await pc.addIceCandidate(candidate);
     }


webrtc.html
```diff
@@ -16278,11 +16278,6 @@ <h3>
   try {
     if (description) {
       await pc.setRemoteDescription(description);
-      // if we got an offer, we need to reply with an answer
-      if (description.type == 'offer') {
-        await pc.setLocalDescription();
-        signaling.send({description: pc.localDescription});
-      }
```
jan-ivar, line 16285: The conferencing server may send an offer to receive simulcast. So applications may need to handle this.

Also, to handle renegotiation for stuff I don't think we should customize the negotiation handling code.

I vote against merging this.

jan-ivar (changes requested):

webrtc.html
jan-ivar: We could add something to the comment:
```
// if we got an offer, we need to reply with an answer. This might
// happen e.g. if a server later wants to send media back to the client.
```

PR #2637, opened 2021-03-22
A merge guide for updating webrtc-pc

alvestrand:
Proposal. Sketchy.

### DIFF
--- a/merge-guide.md
+++ b/merge-guide.md
@@ -21,6 +21,10 @@ different people interested), extension specs are preferred. For extensions that
 "small" (add a few methods, extend an argument list or enum, the function description
 is explanation enough), PRs against webrtc-extensions is preferred.
 
+In order to explore whether such an extension is a viable idea, file an issue on the
+webrtc-extensions repo. (This needs to be done when issues are found with the proposed
+extensions already present, of course.)
+
 ## Landing in webrtc-extensions
 
 We will land extensions in webrtc-extensions on the following criteria:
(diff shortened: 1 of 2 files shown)


merge-guide.md
```diff
@@ -0,0 +1,47 @@
...
+
+In order to have a new feature considered for inclusion, write it up as an extension spec
+or as a pull request against webrtc-extensions.
+
+## Landing in webrtc-extensions
+
+We will land extensions in webrtc-extensions on the following criteria:
+
+* There is rough consensus in the WG that this API makes sense to consider
+* There is commitment from at least one browser vendor that they intend to implement
```
dontcallmedom, line 24: I think this should also say what goes into webrtc-extensions vs an extension spec - my sense is anything that can be defined cleanly as an extension spec should be defined that way (e.g. doesn't require monkey patching algorithms).

dontcallmedom (commented):
this lgtm, with possibly some clarification needed on webrtc-extensions vs extension specs

merge-guide.md
alvestrand: I am thinking that "large features" (webrtc-encoded-transform) should be their own spec. "Small features" (priority would have been one if -extensions existed at the time, header control is one) should go into the extensions spec.

Monkey-patching will be equally hard in extensions and in standalone specs.

alvestrand:
@dontcallmedom is there a preferred format / template for this sort of document?

dontcallmedom:
I realize there is an aspect this document doesn't cover: where issues should be filed between -pc and -extensions.

In terms of format, I don't think there is any particular template to follow; maybe we could put this into the README.md instead of a separate file (and if not, the README should prominently link to it).

alvestrand:
I added text to the README.md pointing here, and text to this document talking about where to file issues. Does it look ready now?

dontcallmedom (approved):
LGTM

PR #2679, opened 2021-10-01
limit number of ice servers to 64

fippo:
limits the number of ICE server entries to 64.
This value is arbitrarily chosen, more than a few servers
do not make sense in practice. As candidates from each
server need a unique priority, the number space is limited.

See
  https://bugs.chromium.org/p/webrtc/issues/detail?id=13195
and
  https://mailarchive.ietf.org/arch/msg/mmusic/isyEwdluHuTl5mMnVUDrmawBnIU/
for background information.

We need to decide whether to throw an error (and if yes, which one) or silently ignore as it is done for certificates

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-pc/pull/2679.html" title="Last updated on Nov 15, 2021, 11:22 AM UTC (012738b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2679/5bff3e5...fippo:012738b.html" title="Last updated on Nov 15, 2021, 11:22 AM UTC (012738b)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -286,8 +286,11 @@
                 <dd>
                   <p>
                     An array of objects describing servers available to be used
-                    by ICE, such as STUN and TURN servers. At most 64 elements
-                    are allowed.
+                    by ICE, such as STUN and TURN servers.
+                  </p>
+                  <p class="note">
+                    Existing implementations only utilize the first 32 entries;
+                    the others are ignored.
                   </p>
                 </dd>
                 <dt data-tests="RTCConfiguration-iceTransportPolicy.html">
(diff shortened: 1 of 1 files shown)


alvestrand:
Firefox starts emitting console warnings at 5. Should we go with a lower value?
Should we ignore or throw?

fippo:
I think we can get histograms of the current usage from Chrome but that will take several release cycles.

jan-ivar:
32 and ignoring seems more web compatible

webrtc.html
```diff
@@ -288,6 +288,10 @@ <h2>
                     An array of objects describing servers available to be used
                     by ICE, such as STUN and TURN servers.
                   </p>
+                  <p class="note">
+                    Existing implementations only utilize the first 32 entries;
+                    the others are ignored.
+                  </p>
                 </dd>
```
alvestrand, line 295: I'd leave the language as normative in the processing steps, but formulate as

If the number of ICE servers exceeds an implementation-defined limit, ignore the ICE servers above the threshold.
This implementation defined limit MUST be at least 32.

jan-ivar:
Updated title to match limit in the PR which is32.

Issue #2698, opened 2021-11-20
Connection establishment should be "failed" even without candidates


jan-ivar:
I tried to detect connection establishment failure by withholding all candidates, while still signaling end-of-candidates. The spec says I should see ["failed"](https://w3c.github.io/webrtc-pc/#dom-rtcicetransportstate-failed) fired: _"The RTCIceTransport has finished gathering, received an indication that there are no more remote candidates, finished checking all candidate pairs, and all pairs have either failed connectivity checks or have lost consent."_

There are zero candidate pairs in this case, which is a 🚩 perhaps, but still passes a pedantic (robot) test: "I have have finished all zero of them", or "there were zero successful pairs with consent", if you will. Zero shoes that fit.

 But none of the browsers appear to fire anything indicating failure. Things just stall. This seems wrong since the UA knows nothing more can happen, but fails to communicate this.

Here's a diff of a [fiddle](https://jsfiddle.net/jib1/uq6pk40r/)'s output in Chrome (which comes closest) compared to what I'd expect according to the spec:
```diff
  ice1.state = new
  ice1.gatheringState = new
  ice1.getSelectedCandidatePair() = null
+ ice1.state = gathering
  pc1.iceGatheringState = gathering
  ice2.state = new
  ice2.gatheringState = new
  ice2.getSelectedCandidatePair() = null
+ ice2.state = gathering
  pc2.iceGatheringState = gathering
  ice1.gatheringState = complete
  pc1.iceGatheringState = complete
+ ice1.state = failed
+ pc1.connectionState = failed
+ pc1.iceConnectionState = failed
  ice2.gatheringState = complete
  pc2.iceGatheringState = complete
+ ice2.state = failed
+ pc2.connectionState = failed
+ pc2.iceConnectionState = failed
```
Importantly, the connection state would have to go directly from "new" to "failed" here, bypassing ["checking"](https://w3c.github.io/webrtc-pc/#dom-rtcicetransportstate-checking) which says: _"The RTCIceTransport has received at least one remote candidate ..."_

But when I look at the [state transition diagram](https://w3c.github.io/webrtc-pc/#fig-non-normative-ice-transport-state-transition-diagram), and there's no arrow from "new" to "failed":

<img width="626" alt="image" src="https://user-images.githubusercontent.com/3136226/142705007-29cb8f6e-098a-4e0e-a811-8cc77f2f6a7d.png">

I think we should draw an arrow, and add some tests.

[...]

PR #2704, opened 2021-12-09
new->failed transition & wait for PAC timer if > 0 local candidates

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2698.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2704.html" title="Last updated on Dec 9, 2021, 1:45 PM UTC (59583ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2704/9cb71e8...jan-ivar:59583ae.html" title="Last updated on Dec 9, 2021, 1:45 PM UTC (59583ae)">Diff</a>


webrtc.html
```diff
@@ -12409,7 +12409,9 @@ <h3>
                     The {{RTCIceTransport}} has finished gathering, received an
                     indication that there are no more remote candidates,
                     finished checking all candidate pairs, and all pairs have
-                    either failed connectivity checks or have lost consent.
+                    either failed connectivity checks or lost consent, and
+                    either zero local candidates were gathered or the PAC timer
+                    has expired [[RFC8863]].
```
youennf, line 12414: Sounds good, we need to figure out whether revealing PAC timer value is a fingerprinting issue or not.
For instance, can it reveal network type? If a fixed value, seems ok

jan-ivar:
I just noticed Preview doesn't show any of the diagrams, so the modified diagram is:

webrtc.html
jan-ivar: [RFC8864 section 4](https://www.rfc-editor.org/rfc/rfc8863.html#section-4) says _"the ICE agent will start its timer once it believes ICE connectivity checks are starting,"_ and _"The RECOMMENDED duration for the PAC timer is equal to the agent's connectivity check transaction timeout, including all retransmissions"_, so this seems network agnostic to me.

juberti: This looks right to me. The PAC timer is meant to be independent of path type, so I agree there shouldn't be a fingerprinting concern.

PR #2754, opened 2022-07-20
remove interaction between encoding.active and simulcast ~rid

### LINKED ISSUE #2735: webrtc-pc specifies using '~' in a=simulcast, but does not support RFC 7728 (RTP pause), which is a prerequisite
webrtc-pc has support for pausing simulcast encodings, which RFC 8853 says requires support for RFC 7728, at least in an answer:

"An answerer that receives an offer without RTP stream pause/resume
   capability MUST NOT mark any simulcast streams as initially paused in
   the answer."

However, RFC 7728 allows partial implementation; there are 8 different "profiles" (that are negotiated with a=rtcp-fb:ccm pause config=\<number\>):

https://datatracker.ietf.org/doc/html/rfc7728#section-9

webrtc-pc probably ought to specify the minimum level of support for RFC 7728.

fippo:
since this would imply setParameters can trigger changes to the
SDP that require negotiation.

Applications that rely on this behavior can still inspect the SDP
(or rely on other means of signaling) and manually change the
encodings active property to not send a particular encoding.

Fixes #2735

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-pc/pull/2754.html" title="Last updated on Jul 20, 2022, 12:27 PM UTC (89fdf17)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2754/fbf4a93...fippo:89fdf17.html" title="Last updated on Jul 20, 2022, 12:27 PM UTC (89fdf17)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2521,19 +2521,6 @@
                                       <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}.
                                     </p>
                                   </li>
-                                  <li>
-                                    <p>
-                                      Update the paused status as indicated by
-                                      [[RFC8853]] of each simulcast
-                                      layer by setting the
-                                      {{RTCRtpEncodingParameters/active}}
-                                      member on the corresponding dictionaries
-                                      in
-                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
-                                      to <code>true</code> for unpaused or to
-                                      <code>false</code> for paused.
-                                    </p>
-                                  </li>
                                 </ol>
                               </li>
                               <li data-tests=


fippo:
@docfaraday can you take a look, also at the test in https://github.com/web-platform-tests/wpt/pull/34912

webrtc.html
```diff
@@ -2521,19 +2521,6 @@ <h4>
...
-                                      Update the paused status as indicated by
-                                      [[RFC8853]] of each simulcast
-                                      layer by setting the
-                                      {{RTCRtpEncodingParameters/active}}
-                                      member on the corresponding dictionaries
-                                      in
-                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
-                                      to <code>true</code> for unpaused or to
-                                      <code>false</code> for paused.
-                                    </p>
```
jan-ivar, line 2535: This removal is good. Though as mentioned by @Orphis in the meeting I wonder if it is sufficient.

Do we perhaps also need to say somewhere something like _"For offers and answers [[RFC7728]] is not supported."_ and make it a non-normative(?) reference. Perhaps in 5.4.1 after _"Alternatives and restrictions (described in [RFC8853]) are not supported."_? 

The reason I wonder this is that the algorithm steps here happen on main thread *after* the description has been generated in the background, and is mostly preoccupied with updating [[SendEncodings]] which (I think??) is mostly used to hold the JS observable states (on main thread) to match what is mostly described in IETF specs, and not actually used anywhere to control what actually gets paused or not in the back end AFAICT (though I could be wrong about this).

jan-ivar: I volunteer to add this in a follow-up to unblock this.

jan-ivar (approved):

Issue #2743, opened 2022-06-15
SLD/SRD(answer) trips over itself when narrowing simulcast envelope


jan-ivar:
https://github.com/w3c/webrtc-pc/pull/2081 added simulcast rejection in answer (by truncating layers): _"If ... simulcast is not supported or desired, ... description rejects any of the offered layers, ... [OR] update the paused status ... [THEN MODIFY] transceiver.[[[Sender]]](https://w3c.github.io/webrtc-pc/#dfn-sender).[[[SendEncodings]]](https://w3c.github.io/webrtc-pc/#dfn-sendencodings)"_

Unfortunately, this runs afoul of language added in https://github.com/w3c/webrtc-pc/issues/2314: _"If applying description leads to modifying a transceiver transceiver, and transceiver.[[[Sender]]](https://w3c.github.io/webrtc-pc/#dfn-sender).[[[SendEncodings]]](https://w3c.github.io/webrtc-pc/#dfn-sendencodings) is non-empty, and not equal to the encodings that would result from processing description, the process of applying description fails."_

When read together, this says SRD(answer) must fail the following if the answer doesn't support simulcast or rejects a layer:
```js
pc.addTransceiver("video", {sendEncodings: [{rid: "a"}, {rid: "b"}]});
```
This seems accidental, because:
 1. failing defeats the detailed modifications that #2081 made.
 2. #2314 say its purpose was to _"not allow remotely initiated RID renegotiation"_.

Remotely initiated = remote offer, not remote answer.

The solution would be to narrow the #2314 language so it doesn't catch #2081.

aboba:
Feedback from June 19, 2022 WEBRTC WG Virtual Interim: "Ready for PR"

PR #2757, opened 2022-07-27
Fix "does not allow remotely initiated RID renegotiation" step

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2743.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2757.html" title="Last updated on Jul 27, 2022, 9:03 PM UTC (46c60bb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2757/d6ed794...jan-ivar:46c60bb.html" title="Last updated on Jul 27, 2022, 9:03 PM UTC (46c60bb)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1834,7 +1834,10 @@
                   </li>
                   <li>
                     <p>
-                      If applying <var>description</var> leads to modifying a
+                      If <var>remote</var> is <code>true</code>, and
+                      <var>description</var> is of type {{RTCSdpType/"offer"}}
+                      and contains a request to receive simulcast, and applying
+                      <var>description</var> leads to modifying a
                       transceiver <var>transceiver</var>, and
                       <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
                       is non-empty, and not equal to the encodings that would


aboba (approved):

jan-ivar:
Editor's agree the original language was probably overly strict compare to initial intent of it (as judged by the summary in the next sentence after the modified language). But the restrictive language may still match current implementations. So we should write some tests to confirm whether this is true or not before merging this.

alvestrand (approved):
This effectively says "allow answers to change RIDs".
This may be too open-ended, but I think that will be fixed by #2758 once it's finished.

Approving.

jan-ivar:
I've written some tentative WPT tests in https://jsfiddle.net/jib1/opsv1ugf/ with console.log statements instead of asserts in places, and the results are in. Chrome, Safari and Edge pretty much yield the same results. Firefox only passes the first (existing) test since we're about to implement this. In Chrome:

(Note the last two PASSes illustrate current behavior, not spec compliance)
```
FAIL createAnswer() attaches to an existing transceiver with a remote simulcast offer: Expected exactly one transceiver - Expected 1, got 2 failed
PASS SRD(simulcastOffer) creates a simulcast transceiver with rids
FAIL SRD(simulcastOffer) creates a simulcast transceiver with scaleResolutionDownBys: Expected scaleResolutionDownBys - Expected 4,2,1, got ,, failed
PASS SRD(simulcastOffer) creates a simulcast transceiver with actives
foo,bar,baz
foo,bar
foo
foo
foo
PASS Remote reoffer can narrow but not (re-)expand simulcast envelope. Never fails.
0,1,2
0,1
0
0
0
PASS Remote reanswer can narrow but not (re-)expand simulcast envelope. Never fails.
4/6
PASS 
```
To summarize:
1. SRD never fails in response, ignoring changes if needed instead
2. remote offers can narrow the simulcast envelope
3. remote answers can narrow the simulcast envelope
3. neither remote offers nor answers can expand the simulcast envelope, but doesn't fail
4. Once narrowed, the envelope stays narrowed and cannot be re-expanded
5. Once present, rid ids are never removed

1, 2 and 5 violate the spec I think. We should perhaps assess how web compatible we think it would be to enforce the spec at this point.

jan-ivar:
4 also violates the spec since it doesn't fail, I think.

aboba:
These results seem reasonable, in general.  So as a goal, it would seem like a good idea to align the spec to the implementations.

jan-ivar:
If we agree on this, then I think this means the paragraph this PR is modifying instead goes away entirely, because of 1.

jan-ivar:
@Orphis does this way forward we suggest make sense to you?

alvestrand:
I am worried about 7.

In particular, if the offer is 1,2,3 and the answer is 1,foo,2, exactly what is being removed?

If we want to be lenient, I would treat any answer that has changes as erroneous layers and ignore them. Thus:

1,2,3 + 1,2 -> 1,2 (narrowing allowed)
1,2,3 + 1,3,2 -> 1
1,2,3 + 1,foo,2 -> 1
1,2,3 + foo, 1, 2 -> one layer with no RID

And CreateOffer() after 1,2,3 + 1,2 should offer 1,2 (no widening allowed from either side).

jan-ivar:
The conversation moved to https://github.com/w3c/webrtc-pc/issues/2762#issuecomment-1205366956 where the current suggestion seems to be to fail over 7.

Issue #2722, opened 2022-04-22
sRD(offer) completely overwrites pre-existing transceiver.[[Sender]].[[SendEncodings]]


docfaraday:
Presently, the language that describes how to handle simulcast in a remote offer says that [[SendEncodings]] is completely replaced based on the rids in the simulcast attribute. While this works fine for transceivers that are not yet associated, for already associated transceivers (which have already populated [[SendEncodings]]), this is not appropriate.

We need to specify what happens on sRD(offer) when there is already an associated transceiver. Since we (rightly) allow sRD(answer) to remove pre-existing rids, we probably need to allow sRD(offer) to remove pre-existing rids as well (since the base simulcast spec requires the answerer to handle this situation). We also need to ensure that the language around createAnswer does the right thing if the offer tries to add a rid (ie; the answer will not contain that new rid).

It probably does not make sense to allow sRD(offer) to add or reorder rids on an already associated transceiver, but I suppose if there's a use-case we want to support we could try something like that. We also do not want to modify any preexisting element in [[SendEncodings]]; we should not try to automatically update scaleResolutionDownBy, for example. So sRD(offer) would only be able to remove elements from [[SendEncodings]], and no other modification would be possible, pretty much the same as the existing rules for sRD(answer).

fippo:
isn't that related to chrome's continued lack of compliance in its so-called "spec-compliant simulcast"?

docfaraday:
Possibly? I'm not sure exactly what you're referring to wrt bugs in Chrome simulcast.

fippo:
see #2155

docfaraday:
So the "lack of compliance" you're referring to is the lack of support for the change in https://github.com/w3c/webrtc-pc/pull/2155?

Whatever the case, https://github.com/w3c/webrtc-pc/pull/2155 is closely related to this issue.

jan-ivar:
Yes we're failing to handle the "already associated" case correctly here, which was what I missed in https://github.com/w3c/webrtc-pc/pull/2155#discussion_r272185898 (because the "else" of creating a new set of transceivers each time, is clearly not what anyone meant).

What @docfaraday proposed SGTM.

aboba:
Feedback from June 19, 2022 WEBRTC WG Virtual Interim: "Ready for PR"

Issue #2737, opened 2022-05-16
Modifications to [[SendEncodings]] from setParameters and sRD can be racy


docfaraday:
setParameters has the following flow in brief:

1. Synchronously executed state and validity checking
2. In parallel, apply the new parameters to the RTP stream being transmitted (might involve queuing, might not)
3. A queued task (that is enqueued during step 2) that updates [[SendEncodings]], and then resolves the promise

sRD(answer) can reject an offered simulcast, which also results in a modification to [[SendEncodings]] (by truncating to size 1). It seems like it would be fairly easy for a message handler to call sRD(answer) so that the truncation happens between steps 1 and 2, or steps 2 and 3.

jan-ivar:
@docfaraday mentioned offline a similar issue with addTrack and SRD(simulcastOffer).

jan-ivar:
My sense is there are two aspects to this problem:
1. The setParameters algorithm's approach of updating entire dictionaries instead of individual members overwrites unrelated data at times
2. The inherent race of two sides competing to modify the `scaleResolutionDownBy` property

1 seems more problematic to me than 2.

IOW, if setParameters were a three-way merge (soving 1) then the only remaining problem would seem to be 2 which seems fine since it's an inherent race between parties (a few milliseconds earlier and you'd get one outcome, vs. a few milliseconds later and you'd get the other).

docfaraday:
> @docfaraday mentioned offline a similar issue with addTrack and SRD(simulcastOffer).

After more offline discussion, it seems like this particular question (what to do when setParameters races against sRD(simulcast offer)) depends on what the expected behavior is in the non-racy case. So if JS calls addTrack, then performs a get/setParameters cycle (which will have one encoding), what should be done with a subsequent sRD(simulcast offer)? Do we interpret the unicast get/setParameters cycle as establishing unicast, and reject the simulcast in our answer? Do we have the sRD(simulcast offer) stomp the encodings entirely? Or do we try to merge the two parameter sets somehow?

docfaraday:
Also, we need to deal with setParameters racing against rollback.

docfaraday:
There are a handful of types of race here, on further thought.

addTrack, then a race between sRD(simulcast recv offer) and setParameters(unicast)
- There are multiple questions here. First we need to decide what addTrack followed by unicast setParameters means. Does it mean that we are definitely not doing simulcast, and we should reject additional encodings in the sRD? Or does it still allow simulcast.

[...]

PR #2758, opened 2022-07-27
Don't let offers to receive simulcast overwrite existing [[SendEncodings]]

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2722. cc @docfaraday

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2758.html" title="Last updated on Oct 13, 2022, 12:49 PM UTC (561937b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2758/aac8167...jan-ivar:561937b.html" title="Last updated on Oct 13, 2022, 12:49 PM UTC (561937b)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2471,8 +2471,7 @@
                                       the previously negototiated layers, then
                                       remove the dictionaries that correspond to
                                       the missing layers from
-                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}, but do not reorder
-                                      them.
+                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}.
                                     </p>
                                   </li>
 				  <li id="rm-simulcast-pause" class="diff-rm"><!-- kept for candidate amendment management --></li>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1758,21 +1758,30 @@
                   <li>
                     <p id="remote-rid-reneg">
                       If <var>remote</var> is <code>true</code>, and
-                      <var>description</var> is of type {{RTCSdpType/"offer"}}
-                      and contains a request to receive simulcast, and applying
-                      <var>description</var> leads to modifying a
-                      transceiver <var>transceiver</var>, and
-                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
-                      is non-empty, and not equal to the encodings that would
-                      result from processing <var>description</var>, the
-                      process of applying <var>description</var> fails. This
-                      specification does not allow remotely initiated RID
-                      renegotiation.
+                      <var>description</var> is of type
+                      {{RTCSdpType/"offer"}}, then for each
+                      [= media description =] requesting to receive simulcast
+                      that already has an existing {{RTCRtpTransceiver}} object,
+                      <var>transceiver</var>, associated with it, as described in
+                      <span data-jsep="applying-a-remote-desc">[[!RFC8829]]</span>,
+                      if the rid values specified in the simulcast attribute (up
+                      to the maximum number the implementation supports) differ
+                      in name, order or number from the ones previously used to
+                      populate
+                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}},
+                      then [= description fails | fail =] the process of
+                      applying <var>description</var>.
+                      <div class="note">
+                        This specification does not allow remotely initiated RID
+                        renegotiation.
+                      </div>
                     </p>
                   </li>
                   <li>
                     <p>
-                      If the process to apply <var>description</var> fails for
+                      If the <dfn data-lt="description fails"
+                      id="set-pc-configuration">process to apply
+                      <var>description</var> fails</dfn> for
                       any reason, then the user agent MUST queue a task that
                       runs the following steps:
                     </p>
@@ -2322,7 +2331,8 @@
                               <li>
                                 <p>
                                   If the <var>description</var> is of type
-                                  {{RTCSdpType/"offer"}} and contains a request
+                                  {{RTCSdpType/"offer"}} and the
+                                  [= media description =] contains a request
                                   to receive simulcast, use the order of the
                                   rid values specified in the simulcast
                                   attribute to create an
@@ -2363,42 +2373,16 @@
                               <li>
                                 <p>
                                   If a suitable transceiver was found
-                                  (<var>transceiver</var> is set) and
+                                  (<var>transceiver</var> is set),
                                   <var>proposedSendEncodings</var> is non-empty,
-                                  run the following steps:
+                                  and
+                                  <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                                  is empty, then set
+                                  <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                                  to <var>proposedSendEncodings</var>, and set
+                                  <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[LastReturnedParameters]]}}
+                                  to <code>null</code>.
                                 </p>
-                                <ol>
-                                  <li class="needs-test">
-                                    <p>
-                                      If
-                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
-                                      is non-empty, then for each encoding,
-                                      <var>encoding</var>, in
-                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}},
-                                      copy all of <var>encoding</var>'s members
-                                      and their values to the encoding with the
-                                      same {{RTCRtpCodingParameters/rid}} in
-                                      <var>proposedSendEncodings</var>, if there
-                                      is one.
-                                    </p>
-                                    <p>
-                                      Set
-                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
-                                      to <var>proposedSendEncodings</var>, and set
-                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[LastReturnedParameters]]}}
-                                      to <code>null</code>.
-                                      <div class="note">
-                                        If this step produces a net change in an
-                                        existing non-empty
-                                        <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}},
-                                        then the process of applying
-                                        <var>description</var> fails. This
-                                        specification does not allow remotely
-                                        initiated RID renegotiation.
-                                      </div>
-                                    </p>
-                                  </li>
-                                </ol>
                               </li>
                               <li>
                                 <p>
--- a/webrtc.html
+++ b/webrtc.html
@@ -1764,14 +1764,18 @@
                       that already has an existing {{RTCRtpTransceiver}} object,
                       <var>transceiver</var>, associated with it, as described in
                       <span data-jsep="applying-a-remote-desc">[[!RFC8829]]</span>,
-                      if the rid values specified in the simulcast attribute (up
-                      to the maximum number the implementation supports) differ
-                      in name, order or number from the ones previously used to
-                      populate
-                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}},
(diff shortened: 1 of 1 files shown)


jan-ivar:
I recommend [ignoring whitespace](https://github.com/w3c/webrtc-pc/pull/2758/files?diff=unified&w=1) when reviewing.

jan-ivar:
Feedback is it might be simpler to spell the failure criteria earlier in 4.1 explicitly instead of relying on it referencing whether this code that's being changed results in a change or not. That failure test would then (for the case where existing SendEncodings exist) disallow changes in SendEncodings.length and changes in ordering, or changes in rid names

webrtc.html
```diff
@@ -2438,12 +2438,40 @@ <h4>
...
+                                      <var>receiveEncodings</var>, if there is one.
+                                    </p>
+                                    <p>
+                                      Set
+                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                                      to <var>receiveEncodings</var>, and set
+                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[LastReturnedParameters]]}}
+                                      to <code>null</code>.
+                                      <div class="note">
+                                        If this step produces a net change in an
```
alvestrand, line 2464: What's a "net change"?

webrtc.html
```diff
@@ -2438,12 +2438,40 @@ <h4>
...
+                                        existing non-empty
+                                        <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}},
+                                        then the process of applying
+                                        <var>description</var> fails. This
+                                        specification does not allow remotely
+                                        initiated RID renegotiation.
```
alvestrand, line 2470: I don't think you can have a normative failure in a "note". If this is to be a note, it has to point to a "raise error and abort these steps" in a normative section.

webrtc.html
```diff
@@ -2438,12 +2438,40 @@ <h4>
...
+                                    <p>
+                                      If
+                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                                      is non-empty, then for each encoding,
+                                      <var>encoding</var>, in
+                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}},
+                                      copy all of <var>encoding</var>'s members
+                                      and their values to the encoding with the
+                                      same {{RTCRtpCodingParameters/rid}} in
+                                      <var>receiveEncodings</var>, if there is one.
```
alvestrand, line 2455: This process allows the order of RIDs applied to layers to change - old {rid: a, rid:b} can be changed to new {rid: b, rid: a}. Since these are lists, not sets, that's a change.

alvestrand (commented):
Think this still needs some work, in particular about being explicit about where inappropriate changes (adding RIDs, reordering RIDs) are being applied.

I think the only modification operation we really want to allow is to delete trailing RIDs (possibly dropping down to singlecast by deleting all RIDs?)

webrtc.html
```diff
@@ -2404,24 +2404,24 @@ <h4>
                                   each of the simulcast layers, populating the
                                   {{RTCRtpCodingParameters/rid}} member
                                   according to the corresponding rid value, and
-                                  let <var>sendEncodings</var> be the list
+                                  let <var>receiveEncodings</var> be the list
```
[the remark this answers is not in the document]
alvestrand: On careful review, it seems to be what's being proposed by the remote end that is the basis for modifying sender.sendEncoding. I suggest we call it proposedSendEncodings instead, to avoid confusion with a possible later extension for receive encodings.

(this was intended to be part of the first review comment, but somehow never landed in this comment)

webrtc.html
docfaraday: Agreed. I think we need to think of this as "remote offers can remove rids from [[SendEncodings]], but cannot make other changes". An attempt to add a rid probably should result in an error, while an attempt to reorder rids might be appropriate to ignore (although throwing an error in that case might be fine too).

webrtc.html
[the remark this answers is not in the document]
docfaraday: Yeah, I agree with the naming change to proposedSendEncodings.

webrtc.html
jan-ivar: The note refers to existing normative prose higher up, since it's too late to fail here. Search for "This specification does not allow remotely initiated RID renegotiation." — I'll try to add a link.

webrtc.html
jan-ivar: The attempt to change is needed to trigger failure elsewhere.

The way things are written right now, the prose that fails is higher up in 4.5: _"If applying description leads to modifying a transceiver transceiver, and transceiver.[[[Sender]]](https://w3c.github.io/webrtc-pc/#dfn-sender).[[[SendEncodings]]](https://w3c.github.io/webrtc-pc/#dfn-sendencodings) is non-empty, and not equal to the encodings that would result from processing description, the process of applying description fails. This specification does not allow remotely initiated RID renegotiation."_

This means, oddly, that to actually trigger that failure condition requires us to (attempt to) make a change here.

Yes this is terrible action-at-a-distance spec writing that I got some good [feedback](https://github.com/w3c/webrtc-pc/pull/2758#issuecomment-1198249698) on already on how to rewrite.

webrtc.html
```diff
@@ -1834,19 +1834,30 @@ <h4>
...
+                      that already has an existing {{RTCRtpTransceiver}} object,
+                      <var>transceiver</var>, associated with it, as described in
+                      <span data-jsep="applying-a-remote-desc">[[!RFC8829]]</span>,
+                      if the rid values specified in the simulcast attribute (up
+                      to the maximum number the implementation supports) differ
+                      in name, order or number from the ones previously used to
+                      populate
+                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}},
+                      then [= description fails | fail =] the process of
+                      applying <var>description</var>.
```
alvestrand, line 1849: This is clear. Thank you!

alvestrand (approved):

webrtc.html
```diff
@@ -2437,10 +2449,13 @@ <h4>
                               <li>
                                 <p>
                                   If a suitable transceiver was found
-                                  (<var>transceiver</var> is set) and
-                                  <var>sendEncodings</var> is non-empty, set
+                                  (<var>transceiver</var> is set),
+                                  <var>proposedSendEncodings</var> is non-empty,
+                                  and
+                                  <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                                  is empty, then set
```
jan-ivar, line 2456: This line is bogus since [[[SendEncodings](https://w3c.github.io/webrtc-pc/#dfn-sendencodings)]] is never empty.

jan-ivar:
I've updated this PR based on https://github.com/w3c/webrtc-pc/issues/2762#issuecomment-1258792891, with one difference: it only fails over rid name mismatch on the first encoding in re-offers, and treats mismatch past that as absence.

IOW I saw no reason to treat known out-of-order rids any different from unknowns rids. This seems to match implementations (both result in layer reduction). @alvestrand @docfaraday  PTAL.

jan-ivar:
The conflict is trivial, and I volunteer to resolve it on merge. This seems preferable to a rebase + force push that would lose patch history.

webrtc.html
```diff
@@ -1840,14 +1840,18 @@ <h4>
...
+                      if the first encoding in
+                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                      [=map/contains=] a {{RTCRtpCodingParameters/rid}} member, and its
+                      value does not match the first rid in the simulcast attribute,
                       then [= description fails | fail =] the process of
                       applying <var>description</var>.
                       <div class="note">
+                        A change in the number of rid values is tolerated in
+                        remote offers as long as the first rid matches the rid
+                        of the first encoding that was previously negotiated.
```
alvestrand, line 1801: , or there are no rid values. (? - turning off simulcast via omitting the attribute could theoretically happen in subsequent offer/answer in a handoff situation)

webrtc.html
```diff
@@ -2449,16 +2453,69 @@ <h4>
...
+                                      is `1`, and the lone encoding
+                                      [=map/contains=] no
+                                      {{RTCRtpCodingParameters/rid}} member, set
+                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                                      to <var>proposedSendEncodings</var>.
+                                    </p>
+                                  </li>
+                                  <li class="needs-test">
+                                    <p>
+                                      Assert: The first encoding of
```
alvestrand, line 2423: We haven't used "Assert:" language before; I'd suggest saying "At this point, " instead.

webrtc.html
```diff
@@ -2449,16 +2453,69 @@ <h4>
...
+                                <ol>
+                                  <li class="needs-test">
+                                    <p>
+                                      If the length of
+                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                                      is `1`, and the lone encoding
+                                      [=map/contains=] no
+                                      {{RTCRtpCodingParameters/rid}} member, set
+                                      <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}
+                                      to <var>proposedSendEncodings</var>.
```
alvestrand, line 2418: and stop (I'm not clear if it should say "go to end of section", "stop processing RIDs", or "go on to the next transceiver").

alvestrand (commented):
This is close to being right, I think. But I worry about the removal (rather than disabling) of non-sent layers.

I think this means that the sequence:

offer: rid: 1,2,3
answer: rid: 1,2

means that subsequent offer is

offer: rid:1,2
and that
answer: rid:1,2,3
will lead to rejection.

Does this have an interaction with pr-answer? (ex: offer: 1,2,3; pr-answer: 1,2; answer: 1,2,3)?

jan-ivar:
> This is close to being right, I think. But I worry about the removal (rather than disabling) of non-sent layers.

Having SRD toggle the `active` attributes would create more races like we're trying to remove in https://github.com/w3c/webrtc-pc/issues/2737 I fear.

> offer: rid:1,2
> and that
> answer: rid:1,2,3
> will lead to rejection.

This is already invalid in JSEP O/A. An answer must stay within the envelope of an offer AFAIK.

> Does this have an interaction with pr-answer? (ex: offer: 1,2,3; pr-answer: 1,2; answer: 1,2,3)?

I don't think so. This PR isn't modifying any answer behavior.

webrtc.html
jan-ivar: I'll test implementations and add that if it works.

webrtc.html
dontcallmedom: I wasn't aware this convention existed, but I confirm it's one that is specifically [supported by ReSpec](https://github.com/w3c/respec/issues/1762), supported in other specs and defined in [Infra](https://infra.spec.whatwg.org/#assertions)

webrtc.html
jan-ivar: I originally had "and skip the remaining sub-steps" but since we're in nested sub-steps of sub-steps I worried it wouldn't be clear which sub-steps to skip. Fortunately, the remaining steps we care to skip are all no-ops in this case (all aspects being compared for are equal already), which is why I suggest leaving it this way.

This has the added advantage of running the assert.

dontcallmedom:
@jan-ivar can you rebase on main so that I add the necessary metadata on Rec changes, following #2713 ?

jan-ivar:
This PR needs more work to incorporate additonal findings:
1. Its restriction that `encodings[0].rid` must match the first rid in a=simulcast seems too strict, when implementations [allow rid pruning not just from the tail](https://github.com/w3c/webrtc-pc/issues/2762#issuecomment-1264036923)
2. While letting remote offers prune rids is required by JSEP, browsers' immediate exposure of this effect on `encodings` in`"have-remote-offer"` seems poorly executed as [it cannot be rolled back, and effects stack over back-to-back SRDs](https://github.com/w3c/webrtc-pc/issues/2762#issuecomment-1271875081).

Instead, waiting until local answer to prune `encodings` avoids both problems (no `encodings` to rollback and no stacking).

Such waiting would maintain invariants the spec offers today, which seems more conservative as a first PR here, while moving us closer to how browsers work.

I think that's the right place to stop, as whether we should go any further I think requires further WG discussion.

jan-ivar:
PR updated. This simplifies quite a bit, which seems promising. PTAL.

aboba (approved):

jan-ivar:
@dontcallmedom this one seems to be missing amendments as well, yet was able to merge. Did you change this requirement?

dontcallmedom:
merges aren't prevented by CI failure in our current setup - the merge was done here despite such [a failure](https://github.com/w3c/webrtc-pc/actions/runs/3242493394/jobs/5315829728) :)

Could you prepare a separate PR to update amendments.json?

jan-ivar:
To qualify the title of this PR, remote offers to receive simulcast _still_ trounce unicast [[SendEncodings]] if it has a lone encoding without a `rid`.

PR #2773, opened 2022-09-22
add url attribute to RTCIceCandidate

fippo:
aligning with how this is defined in webrtc-stats
      https://w3c.github.io/webrtc-stats/#dom-rtcicecandidatestats-url
and add deprecation note to the url attribute on the RTCIceEvent

@youennf since you mentioned this was implemented in Safari, can you please check if it actually shows the correct value? Given the trouble I had reconstructing the url in the lowest layers I would not be surprised if this was the empty strings.

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-pc/pull/2773.html" title="Last updated on Oct 26, 2022, 6:50 PM UTC (73a0a9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2773/67341a6...fippo:73a0a9c.html" title="Last updated on Oct 26, 2022, 6:50 PM UTC (73a0a9c)">Diff</a>

### DIFF
--- a/amendments.json
+++ b/amendments.json
@@ -196,6 +196,15 @@
       "id": 16
     }
   ],
+  "rtcpeerconnectioniceevent-attributes": [
+    {
+      "description": "Add deprecation notice to RTCPeerConnectionIceEvent.url",
+      "pr": 2773,
+      "type": "correction",
+      "status": "candidate",
+      "id": 23
+    }
+  ],
   "rtcicecandidate-attributes": [
     {
       "description": "Add RTCIceCandidate.relayProtocol",
@@ -204,6 +213,13 @@
       "type": "addition",
       "status": "candidate",
       "id": 16
+    },
+    {
+      "description": "Add RTCIceCandidate.url",
+      "pr": 2773,
+      "type": "addition",
+      "status": "candidate",
+      "id": 23
     }
   ],
   "rtcicecredentialtype-enum": [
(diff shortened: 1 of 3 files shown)


fippo:
looks like Safari is actually setting a value but returning IP addresses when asked for stun servers with a hostname. That doesn't look like anyone is actually using this so this should still be safe to remove.

youennf:
Since Safari is implementing it, I would prefer we keep the current API and add it to candidate. Plus add a note that we plan to deprecate the event url.
This will leave some time for Safari to try doing the migration, at which point, if it succeeds, we can remove it altogether.

alvestrand:
Note - I checked the [test for RTCPeerConnectionIceEvent-constructor](https://github.com/web-platform-tests/wpt/blob/master/webrtc/RTCPeerConnectionIceEvent-constructor.html), and it does check that you can construct an RTCPeerConnectionIceEvent with an URL argument, but never checks that the resulting event has an URL in it.

Labelling "test needed" ....

fippo:
Safari still implementing something that was removed isn't a big problem, no (unless we add a negative test)? Nobody cares about chrome still implementing addStream or RTCIceServer.url.
I can help adding a polyfill to adapter if needed

alvestrand:
Seems worth doing a 2-step approach.

jan-ivar:
@fippo can you update this PR to do a two-stage approach?

youennf:
Discussed at today's editor meeting.
@fippo, are you ok with the two step approach?
If not, we probably need more discussion at a next WebRTC interim.

fippo:
done. I *think* I got the amendment right (as a separate commit)

alvestrand:
apparently the generation is failing because the framesDropped error hasn't been fixed yet.

PR #2779, opened 2022-10-06
Allow offers to turn off or prune simulcast just like answers

### LINKED ISSUE #2762: Simulcast: Implementations do not fail (and that seems good)
In [PR2757](https://github.com/w3c/webrtc-pc/pull/2757), Jan-Ivar said: 

"I've written some tentative WPT tests in https://jsfiddle.net/jib1/opsv1ugf/ with console.log statements instead of asserts in places, and the results are in. Chrome, Safari and Edge pretty much yield the same results. Firefox only passes the first (existing) test since we're about to update set/getParameters to spec. In Chrome:

(Note the last 3 PASSes illustrate implemented behaviors, not spec compliance)
```
FAIL createAnswer() attaches to an existing transceiver with a remote simulcast offer: Expected exactly one transceiver - Expected 1, got 2 failed
PASS SRD(simulcastOffer) creates a simulcast transceiver with rids
FAIL SRD(simulcastOffer) creates a simulcast transceiver with scaleResolutionDownBys: Expected scaleResolutionDownBys - Expected 4,2,1, got ,, failed
PASS SRD(simulcastOffer) creates a simulcast transceiver with actives
foo,bar,baz
foo,bar
foo
foo
foo
PASS Remote reoffer can narrow but not (re-)expand simulcast envelope. Never fails.
0,1,2
0,1
0
0
0
PASS Remote reanswer can narrow but not (re-)expand simulcast envelope. Never fails.
0,1
0
0
PASS Remote reanswer altering rids is treated as removal. Never fails.
5/7
PASS 
```
To summarize:
1. SRD never fails in response, ignoring changes if needed instead
2. remote offers can narrow the simulcast envelope
3. remote answers can narrow the simulcast envelope
3. neither remote offers nor answers can expand the simulcast envelope, but don't fail
4. Once narrowed, the envelope stays narrowed and cannot be re-expanded
5. Once present, rid ids are never removed from an encoding
6. renaming rids in remote answers is treated as layer removal, but doesn't fail 

1, 2, 6 and 7 violate the spec I think. We should perhaps assess how web compatible we think it would be to enforce the spec at this point."

jan-ivar:
Fixes #2762 (specifically https://github.com/w3c/webrtc-pc/issues/2762#issuecomment-1264036923, the other observations in that issue are being addressed by https://github.com/w3c/webrtc-pc/pull/2758 instead).

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2779.html" title="Last updated on Oct 6, 2022, 1:56 PM UTC (22c1b39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2779/0349b39...jan-ivar:22c1b39.html" title="Last updated on Oct 6, 2022, 1:56 PM UTC (22c1b39)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2420,6 +2420,7 @@
                               <li>
                                 <p>
                                   If <var>description</var> is of type
+                                  {{RTCSdpType/"offer"}},
                                   {{RTCSdpType/"answer"}} or
                                   {{RTCSdpType/"pranswer"}}, and
                                   <var>transceiver</var>.
@@ -2440,9 +2441,9 @@
                                   </li>
                                   <li>
                                     <p>
-                                      If <var>description</var> rejects any of
+                                      If <var>description</var> is missing any of
                                       the offered layers, then remove the
-                                      dictionaries that correspond to rejected
+                                      dictionaries that correspond to missing
                                       layers from
                                       <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendEncodings]]}}.
                                     </p>


alvestrand:
Make sure that rollback will restore the removed layers.

Issue #2724, opened 2022-04-22
The language around setting a description appears to prohibit renegotiation of RIDs


docfaraday:
"If the description attempted to renegotiate RIDs, as described above, then [reject](https://w3c.github.io/webrtc-pc/#dfn-reject) p with a newly [created](https://webidl.spec.whatwg.org/#dfn-create-exception) [InvalidAccessError](https://webidl.spec.whatwg.org/#invalidaccesserror) and abort these steps."

While it is fine to prohibit a local reoffer from adding or removing RIDs, or changing the simulcast envelope, it is a violation of RFC8853 for an offerer to refuse to honor a remote answer that rejects a previously negotiated RID, and it is also a violation of RFC8853 for an answerer to refuse to honor a remote offer because it removed a previously negotiated RID.

jan-ivar:
The _"described above"_ language says: _"This specification does not allow **remotely initiated** RID renegotiation."_. Remotely initiated = remote offer.

We explicitly prohibited remote offers from renegotiating RIDs in https://github.com/w3c/webrtc-pc/issues/2314. Do we have new information to revisit that decision?

> it is a violation of RFC8853 for an offerer to refuse to honor a remote answer that rejects a previously negotiated RID

Side-note: if you remove "previously negotiated" from that sentence then I agree, and I've filed https://github.com/w3c/webrtc-pc/issues/2743 on that.

But reading [RFC8853](https://www.rfc-editor.org/rfc/rfc8853.html) I find very little mention of previously negotiated state. To me it doesn't appear to make any distinction between initial O/A and renegotiation. I therefore find it hard to infer the requirements of these simulcast attributes on subsequent O/A.

I know JSEP has such language, but it's not clear to me how that maps here.

For instance: RFC8853's example mentions an offer to send 3 layers, with an answer to receive 2. Is a subsequent identical O/A acceptable because the net result is the same? If so, what if the answer doesn't reject the 3rd layer the second time? Now we've added a layer, is that acceptable?

> it is also a violation of RFC8853 for an answerer to refuse to honor a remote offer because it removed a previously negotiated RID.

Same here, if you could quote where it says that it would help.

jan-ivar:
So with https://github.com/w3c/webrtc-pc/pull/2758/ merged, the current language here is: _"if none of the encodings in transceiver.[[Sender]].[[SendEncodings]] contain a rid member whose value matches any of the rids in the simulcast attribute, then fail the process of applying description."_

[...]

PR #2794, opened 2022-10-26
Don't fail sRD(offer) over rid mismatch, just answer with unicast.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2724.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2794.html" title="Last updated on Nov 8, 2022, 4:43 PM UTC (a9399f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2794/67341a6...jan-ivar:a9399f5.html" title="Last updated on Nov 8, 2022, 4:43 PM UTC (a9399f5)">Diff</a>

### DIFF
--- a/amendments.json
+++ b/amendments.json
@@ -316,5 +316,15 @@
       "status": "candidate",
       "id": 22
     }
-  ]
+  ],
+  "answer-missing-encodings": [
+    {
+      "description": "Don't fail sRD(offer) over rid mismatch, just answer with unicast.",
+      "pr": 2794,
+      "difftype": "modify",
+      "type": "correction",
+      "status": "candidate",
+      "id": 23
+    }
+  ],
 }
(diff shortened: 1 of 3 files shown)


jan-ivar:
@dontcallmedom This PR removes an earlier amendment, but I seem to be tripping over something. Is there a correct way to do this?

dontcallmedom:
the narrow technical issue reported in the console is
> The container with id process-to-apply-description marked as amended cannot embed the other container of amendment apply-description-restrictions

But if my reading is correct, the real source of the problem is that this pull request reverts #2757 and the part of #2758 that was under the id `remote-rid-reneg` - if so, I think the correct way to describe this in amendments.json would be simply to  remove altogether the current entry under id  `remote-rid-reneg` (since it's no longer a change to the text of the current Rec).

alvestrand:
Scheduled for November interim.

aboba (approved):

PR #2808, opened 2022-12-14
Allow encoder resolution alignment in scaleResolutionDownBy.

### LINKED ISSUE #2802: scaleResolutionDownBy's odd-resolution requirement trips up HW encoders
Some hardware encoders require specific alignment values that need to be taken into consideration when requesting specific resolutions.

https://w3c.github.io/webrtc-pc/#dfn-rtp-media-api  says:
"When video is rescaled, for example for certain combinations of width or height and [scaleResolutionDownBy](https://w3c.github.io/webrtc-pc/#dom-rtcrtpencodingparameters-scaleresolutiondownby) values, situations when the resulting width or height is not an integer may occur. In such situations the user agent MUST use [the integer part of the result](https://tc39.github.io/ecma262/#eqn-floor)."

I would like to propose that this text be amended to support selecting a resolution supported by the encoder greater than the calculated scaled resolution. It appears the [Chrome implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/video/config/encoder_stream_factory.cc;l=452;drc=2df668b7cbf6c1d0766b6ee0ae8147adc8830f2e) may already be doing this.

jan-ivar:
Fixed #2802.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -7484,13 +7484,15 @@ interface RTCCertificate {
         possible designs have been discussed in <a href=
         "https://github.com/w3c/webrtc-pc/issues/1283">GitHub issue 1283</a>.
       </p>
-      <p>
+      <p id="rtp-video-rescaled">
         When video is rescaled, for example for certain combinations of width
         or height and {{RTCRtpEncodingParameters/ scaleResolutionDownBy}}
         values, situations when the resulting width or height is not an integer
         may occur. In such situations the user agent MUST use <a href=
         "https://tc39.github.io/ecma262/#eqn-floor">the integer part of the
-        result</a>. What to transmit if the integer part of the scaled width or
+        result</a>. The user agent MAY increase the calculated scaled resolution
+        to the nearest alignment supported by the encoder.
+        What to transmit if the integer part of the scaled width or
         height is zero is implementation-specific.
       </p>
       <p>
--- a/amendments.json
+++ b/amendments.json
@@ -385,5 +385,15 @@
       "status": "candidate",
       "id": 26
     }
+  ],
+  "rtp-video-rescaled": [
+    {
+      "description": "Allow encoder resolution alignment in scaleResolutionDownBy.",
+      "pr": 2808,
+      "difftype": "modify",
+      "type": "correction",
+      "status": "candidate",
+      "id": 27
+    }
   ]
 }
--- a/base-rec.html
+++ b/base-rec.html
@@ -7718,7 +7718,7 @@ interface <a class="internalDFN idlID" data-link-type="interface" href="#dom-rtc
         and timeline for a more complex handling of this situation. Some
         possible designs have been discussed in <a href="https://github.com/w3c/webrtc-pc/issues/1283">GitHub issue 1283</a>.
       </p></div>
-      <p class="needs-tests">
+      <p class="needs-tests" id="rtp-video-rescaled">
         When video is rescaled, for example for certain combinations of width
         or height and <a data-link-type="idl" href="#dom-rtcrtpencodingparameters-scaleresolutiondownby" class="internalDFN" id="ref-for-dom-rtcrtpencodingparameters-scaleresolutiondownby-2"><code><code>scaleResolutionDownBy</code></code></a>
         values, situations when the resulting width or height is not an integer


webrtc.html
```diff
@@ -7484,14 +7485,15 @@ <h2>
         possible designs have been discussed in <a href=
         "https://github.com/w3c/webrtc-pc/issues/1283">GitHub issue 1283</a>.
       </p>
-      <p>
+      <p id="rtp-video-rescaled">
         When video is rescaled, for example for certain combinations of width
         or height and {{RTCRtpEncodingParameters/ scaleResolutionDownBy}}
         values, situations when the resulting width or height is not an integer
```
jan-ivar, line 7491: Hmm, should we limit this sentence to `scaleResolutionDownBy`, to not step on RFC8829?

This seems to match better the extent of this spec's concerns, which I think are that `scaleResolutionDownBy` behaves deterministically.

webrtc.html
```diff
@@ -7472,7 +7472,8 @@ <h2>
       </p>
       <p>
         Following the rules in <span data-jsep="imageattr">[[!RFC8829]]</span>,
-        the video MAY be downscaled in order to fit the SDP constraints. The
+        the video MAY be downscaled in order to fit the SDP constraints and
+        alignment restrictions of the encoder. The
```
jan-ivar, line 7476:
```suggestion
        alignment limitations of the encoder. The
```

webrtc.html
```diff
@@ -7484,14 +7485,15 @@ <h2>
...
-      <p>
-        When video is rescaled, for example for certain combinations of width
-        or height and {{RTCRtpEncodingParameters/ scaleResolutionDownBy}}
-        values, situations when the resulting width or height is not an integer
-        may occur. In such situations the user agent MUST use <a href=
-        "https://tc39.github.io/ecma262/#eqn-floor">the integer part of the
-        result</a>. What to transmit if the integer part of the scaled width or
-        height is zero is implementation-specific.
+      <p id="rtp-video-rescaled">
+        When video is rescaled using {{RTCRtpEncodingParameters/scaleResolutionDownBy}},
```
jan-ivar, line 7489: Rephrase to clarify rescaling requirement applies over time.
```suggestion
        Whenever video is rescaled as a result of {{RTCRtpEncodingParameters/scaleResolutionDownBy}},
```

webrtc.html
```diff
@@ -7472,7 +7472,8 @@ <h2>
       </p>
       <p>
         Following the rules in <span data-jsep="imageattr">[[!RFC8829]]</span>,
-        the video MAY be downscaled in order to fit the SDP constraints. The
+        the video MAY be downscaled in order to fit the SDP constraints and
+        alignment limitations of the encoder. The
```
Pehrsons, line 7476: What about downscaling due to cpu overuse detection or BWE?

webrtc.html
```diff
@@ -7484,14 +7485,15 @@ <h2>
...
-        height is zero is implementation-specific.
+      <p id="rtp-video-rescaled">
+        Whenever video is rescaled as a result of {{RTCRtpEncodingParameters/scaleResolutionDownBy}},
+        situations when the resulting width or height is not an integer
+        may occur. The user agent MUST NOT transmit video larger than
+        <a href="https://tc39.github.io/ecma262/#eqn-floor">the integer part</a>
+        of the scaled width and height from
+        {{RTCRtpEncodingParameters/scaleResolutionDownBy}}. What to transmit if
+        the integer part of the scaled width or height is zero is
+        [=implementation-defined=].
```
Pehrsons, line 7496: This needs to be a bit broader -- in case of resolution alignment it is not just the integer part of the scaled dimensions being zero that is implementation-defined.

Also [some codecs impose a minimum resolution](https://developer.android.com/reference/android/media/MediaCodecInfo.VideoCapabilities#getSupportedWidths()) that could impact this.

webrtc.html
jan-ivar: I'll move mention of these to the paragraph above, and restore this paragraph to not disturb it, as it appears to have a lot of history.

webrtc.html
jan-ivar: This particular paragraph now only discusses scaleResolutionDownBy, so hopefully it is clear that this sentence only speaks to that: basically a clarification of the sentence that precedes it, which otherwise would less unclear.

I don't think it is claiming this to be the _only_ implementation-defined behavior.

Regarding minimum codec resolution, that is a good point. I've added that. PTAL.

jan-ivar:
I find looking at the [preview](https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2808.html#dfn-rtp-media-api) helpful here as there are multiple separate paragraphs meant to be read in sequence.

PR #2812, opened 2022-12-28
add test annotations for msid generation

fippo:
and move test annotiation for setStreams which is currently not tested in that test suite.

@dontcallmedom did I do this right?

(note that currently msid-generate does not test setStreams, i'll fix that soon)


fippo:
(more tests as of https://github.com/web-platform-tests/wpt/commit/8f1829e65a578fc45357023a3a1622354a4b0467)

PR #2818, opened 2023-02-01
Use markup convnetions for algorithms

dontcallmedom:
helps with cross-linking and automatic extractions


fippo:
nit: s/convnetions/conventions/ in the commit message 💅

dontcallmedom:
making a draft since merging this will require changes to respec (in particular resolution of https://github.com/w3c/respec/issues/4376 )

alvestrand:
The change here seems like it adds "class=" attributes to &lt;dfn&gt; tags. Are these well documented, or are they too a WIP?

dontcallmedom:
using the `class` attribute on the `<dfn>` element is standard feature of ReSpec to give a type to a definition.

The list of types is shared with Bikeshed where it is [reasonably well-documented](https://speced.github.io/bikeshed/#dfn-types) (although there is discussion to move that documentation [elsewhere and improve it](https://github.com/speced/spec-dfn-contract)).

The fact that the [`class` attribute can be used](https://github.com/w3c/respec/blob/994d63b031b691475a5da7edcb625bf01faf6c27/src/core/dfn.js#L92)  (or even as far as I understand is the preferred the way) to associate a type to a definition is not well-documented on the other hand.

dontcallmedom:
with https://github.com/w3c/respec/issues/4376 fixed, this should be ready to merge

PR #2826, opened 2023-02-13
Add internal slots for signaling state, connection state, ice gatheri…

jan-ivar:
…ng state, and ice connection state, and fix ongatheringstatechange to use the right one.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2826.html" title="Last updated on Mar 10, 2023, 11:17 AM UTC (ef49a06)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2826/c0cd7f2...jan-ivar:ef49a06.html" title="Last updated on Mar 10, 2023, 11:17 AM UTC (ef49a06)">Diff</a>

### DIFF
--- a/amendments.js
+++ b/amendments.js
@@ -127,8 +127,8 @@ function showAmendments() {
     const target = document.getElementById(section);
     let wrapper = document.createElement("div");
     if (amendments[section][0].difftype !== "append") {
-      if (["LI"].includes(target?.tagName)) {
-	wrapper = document.createElement("li");
+      if (["LI", "DD"].includes(target?.tagName)) {
+	wrapper = document.createElement(target.tagName);
 	wrapper.className = "skip";
       }
     } else {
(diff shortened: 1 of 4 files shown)


alvestrand:
Please don't merge until the REC process has completed.

dontcallmedom:
What triggers the usage of the amendment process (and for our repo, the use of the "editorial" label or not) is based on the [class of change as described in the W3C Process](https://www.w3.org/2021/Process-20211102/#correction-classes) - essentially, anything that might affect the conformance of an existing implementation.

I had read the fix on the `gatheringstatechange` event as crossing that line (hence the removal of the `Editorial` label), but now your description of this being mostly an alignment with other normative text and the text not making sense as is makes a convincing argument that it is in fact a class 2 change, and so the "Editorial" label should be reinstated, and the pull request should not touch `amendments.json` at all - I'll get that done.

PR #2853, opened 2023-04-03
Use the url spec to parse ice server urls (Take 2).

### LINKED ISSUE #2660: Unclear how URLs are processed
Chrome throws an exception for
```js
new RTCPeerConnection({ iceServers: [{ urls:["stun:turn.example.org/test"] }] });
```
whereas Firefox and Safari do not.

Where does https://w3c.github.io/webrtc-pc/#dom-peerconnection process these?

Some processing appears to happen in https://w3c.github.io/webrtc-pc/#set-the-configuration but that does not seem to be invoked from the constructor. (That also suggests using a URL parser that browsers do not have, which seems somewhat dubious.)

jan-ivar:
Fixes #2660. @annevk PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2853.html" title="Last updated on Apr 4, 2023, 2:11 PM UTC (1a5b430)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2853/fb2be18...jan-ivar:1a5b430.html" title="Last updated on Apr 4, 2023, 2:11 PM UTC (1a5b430)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3117,7 +3117,7 @@
                       <li>
                         <p>
                           If <var>parsedURL</var>'s' [=url/scheme=] is <code>"turn"</code> or
-                          <code>"turns"</code>, and either
+                          <code>"turns"</code>, and either of
                           <var>server</var>.{{RTCIceServer/username}} or
                           <var>server</var>.{{RTCIceServer/credential}} do
                           [=map/exist|not exist=], then [= exception/throw =] an
--- a/webrtc.html
+++ b/webrtc.html
@@ -3063,7 +3063,7 @@
                   <li>
                     <p>
                       If <var>urls</var> is empty, [= exception/throw =] a
-                      {{SyntaxError}}.
+                      "{{SyntaxError}}" {{DOMException}}.
                     </p>
                   </li>
                   <li>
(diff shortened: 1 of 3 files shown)


webrtc.html
```diff
@@ -3073,44 +3073,54 @@ <h4>
...
-                          "scheme">turn</code> or <code class=
-                          "scheme">turns</code>, and
-                          <var>server</var>.{{RTCIceServer/credential}} is not
-                          a <span class="idlMemberType">DOMString</span>, then
-                          [= exception/throw =] an {{InvalidAccessError}}.
+                          If <var>parsedURL</var>'s' [=url/scheme=] is <code>"turn"</code> or
+                          <code>"turns"</code>, and neither
+                          <var>server</var>.{{RTCIceServer/username}} nor
+                          <var>server</var>.{{RTCIceServer/credential}}
+                          [=map/exist=], then [= exception/throw =] an {{InvalidAccessError}}.
```
alvestrand, line 3123: I don't think it helps to have credential without username. The previous text insisted that both need to exist.

webrtc.html
```diff
@@ -3073,44 +3073,54 @@ <h4>
...
+                        </ul>
+                      </li>
+                      <li>
+                        <p>If <var>parsedURL</var>'s [=url/scheme=] is not implemented by the
+                        user agent, then [=exception/throw=] a {{NotSupportedError}}.</p>
+                      </li>
+                      <li>
+                        <p>Let <var>hostAndPortURL</var> be result of
+                        <a data-cite="!url#concept-url-parser">parsing</a> the concatenation of
+                        `"https://"` and <var>parsedURL</var>'s [=url/path=].</p>
```
alvestrand, line 3100: This is a decidedldy odd way of parsing the arguments to the URL.
It assumes that 7064 and 7065 are completely consistent with the relevant parts of the URL spec. I'd prefer to have the URL spec learn to parse TURN URLs - @annevk may have opinions on the proper way to do this.

alvestrand (changes requested):
The username/password thing needs updating. As to the appropriateness of running the URL algorithm twice, once upon a synthetic URL, I'll let @annevk have an opinion on that.

webrtc.html
jan-ivar: Whoops, I inverted the logic incorrectly. Thanks for spotting that.

alvestrand (approved):

PR #2854, opened 2023-04-05
Export links to addTransceiver/setParameters steps + unnumber some steps

jan-ivar:
This should make it simpler for https://github.com/w3c/webrtc-extensions/pull/147 and https://w3c.github.io/webrtc-svc/#setparameters to reference this spec.

Also, the latter extends the list of conditions, which is currently a numbered list, which seems confusing, since there's no sequential order implied AFAIK (these are all conditions that result in the same error thrown, making order unobservable).

This PR makes that list unnumbered, to avoid confusion. WebRTC-SVC should update to match.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2854.html" title="Last updated on Apr 5, 2023, 10:06 PM UTC (43a6330)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2854/fb2be18...jan-ivar:43a6330.html" title="Last updated on Apr 5, 2023, 10:06 PM UTC (43a6330)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -8969,7 +8969,9 @@ interface RTCRtpSender {
                     newly [= exception/created =] {{InvalidStateError}}.
                   </li>
                   <li id="setparameters-validation">
-                    Validate <var>parameters</var> by running the following steps:
+                    Validate <var>parameters</var> by running the following
+                    <dfn class="export" data-dfn-for="RTCRtpSender">
+                    setParameters validation steps</dfn>:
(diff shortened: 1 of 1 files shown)


jan-ivar:
Did the same for addTransceiver's validation steps.

Orphis (approved):

Issue #2865, opened 2023-04-27
The steps in "Update the connection state" say to fire connectionstatechange when [[IsClosed]] is set to true


docfaraday:
From 4.4.1.3 Update the connection state:

"An [RTCPeerConnection](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection) object has an aggregated [[[ConnectionState]]](https://w3c.github.io/webrtc-pc/#dfn-connectionstate). Whenever the state of an [RTCDtlsTransport](https://w3c.github.io/webrtc-pc/#dom-rtcdtlstransport) changes or when the [[[IsClosed]]](https://w3c.github.io/webrtc-pc/#dfn-isclosed) slot turns true, the user agent MUST update the connection state by queueing a task that runs the following steps:

1. Let connection be this [RTCPeerConnection](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection) object.

2. Let newState be the value of deriving a new state value as described by the [RTCPeerConnectionState](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectionstate) enum.

3. If connection.[[[ConnectionState]]](https://w3c.github.io/webrtc-pc/#dfn-connectionstate) is equal to newState, abort these steps.

4. Set connection.[[[ConnectionState]]](https://w3c.github.io/webrtc-pc/#dfn-connectionstate) to newState.

5. [Fire an event](https://dom.spec.whatwg.org/#concept-event-fire) named [connectionstatechange](https://w3c.github.io/webrtc-pc/#event-connectionstatechange) at connection."

This directly contradicts 4.4.4 Garbage collection:

"An [RTCPeerConnection](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection) object MUST not be garbage collected as long as any event can cause an event handler to be triggered on the object. When the object's [[[IsClosed]]](https://w3c.github.io/webrtc-pc/#dfn-isclosed) internal slot is true, no such event handler can be triggered and it is therefore safe to garbage collect the object."

In addition, most places suppress event firing when [[IsClosed]] is true. There are some exceptions, though, and those are likely bugs too. Just grep through for "Fire an event" and find the places that don't do an [[IsClosed]] check.

jan-ivar:
[...]

PR #2876, opened 2023-06-02
Don't fire connectionstatechange on pc.close().

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2865.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2876.html" title="Last updated on Jul 13, 2023, 1:32 PM UTC (2b3d707)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2876/bc552d5...jan-ivar:2b3d707.html" title="Last updated on Jul 13, 2023, 1:32 PM UTC (2b3d707)">Diff</a>


jan-ivar:
I'm getting validation errors, but I don't see anything wrong. @dontcallmedom am I using the append amendments type wrong?
> "file:/home/runner/work/webrtc-pc/webrtc-pc.w3c/webrtc.html":2675.15-2675.146: info warning: Possible misuse of “aria-label”. (If you disagree with this warning, file an issue report or send e-mail to www-validator@w3.org.)
    "file:/home/runner/work/webrtc-pc/webrtc-pc.w3c/webrtc.html":2705.13-2705.58: error: Element “div” not allowed as child of element “ol” in this context. (Suppressing further errors from this subtree.)

youennf (approved):
I don't think this is a change of behaviour.
Close is probably setting the connection stage to closed synchronously.
Then the algorithm will try to update and be a no-op.
Worth clarifying though

alvestrand (approved):
I don't understand the markup, but agree with the behavior change (which I think is probably the current behavior of browsers; test needed), so marking myself as approving.

Issue #2889, opened 2023-08-15
Editorial mistakes in iceServers limit


jan-ivar:
While we added an iceServers limit in https://github.com/w3c/webrtc-pc/pull/2679, some editorial mistakes makes it not work:
- A later [merge mistake](https://github.com/w3c/webrtc-pc/pull/2691/files#diff-6eb57d2440cec3e5472ee0c3031da1f0408af0f3539f38a371517f70e01d5210L3077) in #2691 (my bad) erased the truncation in setConfiguration
- A typo in the [constructor](https://w3c.github.io/webrtc-pc/#dom-peerconnection) fails to truncate anything: _"If configuration.[iceServer](https://w3c.github.io/webrtc-pc/#dom-rtcconfiguration-iceservers) is an array truncate it to the maximum number of supported elements."_ It should be `s/iceServer/iceServers/`

PR #2890, opened 2023-08-15
Move iceServers limit to setConfiguration steps.

jan-ivar:
Fixes https://github.com/w3c/webrtc-pc/issues/2889.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1277,13 +1277,6 @@
                   checks are unnecessary.
                 </p>
               </li>
-              <li data-tests="RTCConfiguration-iceServers.html">
-                <p>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -3045,10 +3038,21 @@ <h4>
                   </li>
                 </ol>
               </li>
+              <li>
```
alvestrand, line 3041: doesn't the data-tests tag follow along?

jan-ivar: Yes. I'm landing that test in https://phabricator.services.mozilla.com/D186256 today, so I might as well add it now. Thanks!

PR #2902, opened 2023-10-06
Prevent GC of non-closed RTCDataChannels

### LINKED ISSUE #2624: Should garbage collecting RTCDataChannels be observable?
A useful feature of data channels is they can detect a remote peer closing. From my [stackoverflow answer](https://stackoverflow.com/a/66309925/918910):
> The best way to detect the other peer closing the connection is to always negotiate SCTP data channels with your connection and listen for a data channel's [closed](https://w3c.github.io/webrtc-pc/#idl-def-RTCSctpTransportState.closed) event:

This works in all major browsers even when data channels aren't used for anything else, modulo a Firefox GC bug:
```js
pc2.ondatachannel = e => window.keepalive = e.channel; // Firefox GC workaround
```
Without this workaround or other code actually using the data channel on `pc2`, Firefox will [automatically fire a `closed` event on `pc1` after a few of seconds](https://jan-ivar.github.io/dummy/dcclose_bug.html) due to GC/CC. We'd like to fix Firefox, except the spec isn't super clear Firefox is in the wrong.

[§ 6.4 Garbage Collection](https://w3c.github.io/webrtc-pc/#garbage-collection-0) in full:
>  An RTCDataChannel object MUST not be garbage collected if its
>
> *    [[ReadyState]] slot is "connecting" and at least one event listener is registered for open events, message events, error events, closing events, or close events.
> *    [[ReadyState]] slot is "open" and at least one event listener is registered for message events, error events, closing events, or close events.
> *    [[ReadyState]] slot is "closing" and at least one event listener is registered for error events, or close events.
> *    [underlying data transport](https://w3c.github.io/webrtc-pc/#dfn-data-transport) is established and data is queued to be transmitted.

**Interpretation A**: It's only talking about the surfaced JS channel object, not its [underlying data transport](https://w3c.github.io/webrtc-pc/#dfn-data-transport).

The problem with this interpretation is that the spec does not appear to accommodate an underlying data transport outliving its associated JS channel object. Instead it appears to assume the opposite in several places ([1](https://w3c.github.io/webrtc-pc/#closing-procedure), [2](https://w3c.github.io/webrtc-pc/#announcing-a-data-channel-as-closed), [3](https://w3c.github.io/webrtc-pc/#dfn-receive-an-rtcdatachannel-message)).

**Interpretation B**: It's talking about the surfaced JS channel object _and_ its underlying data transport.

The problem with this interpretation is that it makes garbage collection observable, and would require "fixing" the other browsers to have the same behavior Firefox has (what we called a "bug" above).

Neither of these interpretations seem appealing.

**Proposal:** Have an RTCPeerConnection hold strong references to all its RTCDataChannels that have not been closed.

jan-ivar:
Fixes #2624.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2902.html" title="Last updated on Nov 2, 2023, 3:00 PM UTC (0c99003)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2902/682049e...jan-ivar:0c99003.html" title="Last updated on Nov 2, 2023, 3:00 PM UTC (0c99003)">Diff</a>

### DIFF
--- a/amendments.json
+++ b/amendments.json
@@ -520,5 +520,14 @@
       "status": "candidate",
       "id": 36
     }
+  ],
+  "closing-procedure": [
+    {
+      "description": "Prevent GC of non-closed RTCDataChannels",
+      "pr": 2902,
+      "type": "addition",
+      "status": "candidate",
+      "id": 38
+    }
   ]
 }
--- a/amendments.json
+++ b/amendments.json
@@ -525,7 +525,7 @@
     {
       "description": "Prevent GC of non-closed RTCDataChannels",
       "pr": 2902,
-      "type": "addition",
+      "type": "correction",
       "status": "candidate",
       "id": 38
     }
--- a/webrtc.html
+++ b/webrtc.html
@@ -13972,7 +13972,7 @@ interface RTCSctpTransport : EventTarget {
                 {{RTCDataChannelState/"closing"}} and [= fire an event =] named
                 {{RTCDataChannel/closing}} at <var>channel</var>.
               </p>
-             </li>
+            </li>
             <li>
               <p>
                 Run the following steps in parallel:
--- a/webrtc.html
+++ b/webrtc.html
@@ -13953,6 +13953,12 @@ interface RTCSctpTransport : EventTarget {
                 [= underlying data transport =] was closed.
               </p>
             </li>
+            <li class="no-test-needed">
+              <p>
+                Let <var>connection</var> be the {{RTCPeerConnection}} object
+                associated with <var>channel</var>.
+              </p>
+            </li>
             <li>
               <p>
                 Remove <var>channel</var> from <var>connection</var>.{{RTCPeerConnection/[[DataChannels]]}}.
--- a/webrtc.html
+++ b/webrtc.html
@@ -1313,6 +1313,13 @@
               </li>
               <li>
                 <p>
+                  Let <var>connection</var> have a
+                  <dfn data-dfn-for="RTCPeerConnection">[[\DataChannels]]</dfn> internal slot, initialized to
+                  an empty [=ordered set=].
+                </p>
+              </li>
+               <li>
+                <p>
                   Let <var>connection</var> have an <dfn data-dfn-for="RTCPeerConnection">[[\Operations]]</dfn>
                   internal slot, representing an [= operations chain =],
                   initialized to an empty list.
@@ -13328,6 +13335,12 @@ interface RTCTrackEvent : Event {
                   </li>
                   <li>
                     <p>
+                      [=list/Append=] <var>channel</var> to
+                      <var>connection</var>.{{RTCPeerConnection/[[DataChannels]]}}.
+                    </p>
+                  </li>
+                  <li>
+                    <p>
                       Return <var>channel</var> and continue the following
                       steps in parallel.
                     </p>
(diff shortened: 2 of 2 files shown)


jan-ivar:
Test forthcoming in https://phabricator.services.mozilla.com/D190744.

webrtc.html
```diff
@@ -13326,6 +13333,12 @@ <h2>
                       negotiation-needed flag =] for <var>connection</var>.
                     </p>
                   </li>
+                  <li>
+                    <p>
+                      [=list/Append=] <var>channel</var> to
+                      <var>connection</var>.{{RTCPeerConnection/[[DataChannels]]}}.
```
youennf, line 13339: It seems this should be done for both createDataChannel code and ondatachannel code paths.
We could add it to https://w3c.github.io/webrtc-pc/#announcing-a-data-channel-instance for instance.

amendments.json
```diff
@@ -520,5 +520,14 @@
       "status": "candidate",
       "id": 36
     }
+  ],
+  "closing-procedure": [
+    {
+      "description": "Prevent GC of non-closed RTCDataChannels",
```
youennf, line 526: non-closing/non-closed

webrtc.html
```diff
@@ -1312,6 +1312,13 @@ <h4>
                 </p>
               </li>
               <li>
+                <p>
+                  Let <var>connection</var> have a
+                  <dfn data-dfn-for="RTCPeerConnection">[[\DataChannels]]</dfn> internal slot, initialized to
+                  an empty [=ordered set=].
```
alvestrand, line 1318: Why "ordered set" in particular?

webrtc.html
alvestrand, line 13339: Similarly, why "append" and not "add"?

alvestrand (approved):
My only nit here is that "Ordered set" sounds prescriptive of implementation while seeming to have no JS-visible effect. I like minimally-constraining descriptions.

webrtc.html
jan-ivar: This appears to be recommended, and there's no infra for unordered sets. From https://infra.spec.whatwg.org/#sets:

image

webrtc.html
jan-ivar: See above.

webrtc.html
```diff
@@ -13940,6 +13953,17 @@ <h4>
...
             </li>
+            <li class="no-test-needed">
+              <p>
+                Let <var>connection</var> be the {{RTCPeerConnection}} object
+                associated with <var>channel</var>.
+              </p>
+            </li>
+            <li>
+              <p>
+                Remove <var>channel</var> from <var>connection</var>.{{RTCPeerConnection/[[DataChannels]]}}.
```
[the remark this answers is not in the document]
jan-ivar: Hmm, that "procedure" is also an algorithm called from https://w3c.github.io/webrtc-pc/#data-transport-closing-procedure

There's also this note suggesting the [closing procedure](https://w3c.github.io/webrtc-pc/#data-transport-closing-procedure) is called from somewhere _other_ than `close()` but it's not.
image

amendments.json
jan-ivar: Given where we're removing the channel I'm not sure we're preventing GC of non-closing channels?

Orphis (approved):

henbos (approved):

jan-ivar:
Test is https://wpt.fyi/results/webrtc/RTCDataChannel-GC.html

Issue #2888, opened 2023-08-11
setCodecPreferences vs unidirectional codecs


fippo:
(for after TPAC, this is a minor detail and we have more important things to discuss)

https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setcodecpreferences
Step 6 says that
> If the intersection between codecs and RTCRtpSender.getCapabilities(kind).codecs or the intersection between codecs and 
> RTCRtpReceiver.getCapabilities(kind).codecs only contains RTX, RED or FEC codecs or is an empty set, throw InvalidModificationError. 
> This ensures that we always have something to offer, regardless of transceiver.direction.

This means that you must negotiate at least one codec that you can send and receive.

While this seems harmless, it causes trouble with asymmetric codec support and that is a fairly common thing. E.g. in Chrome on Windows we have  14 send codecs and 21 receive codecs, with the asymmetric ones ranging from H264 over VP9 to AV1.
The worst example we found was Android where the H264 send and receive profiles are not an exact string match and we have support for sending 42e032, 4d0032 and 640032 while being able to receive 42001f, 4d001f and 64001f. This lead to errors in a fairly innocent attempt to restrict the codecs to H264 variants:
```
const codecs = RTCRtpSender.getCapabilities('codecs')
  .filter(c => c.mimeType.toLowerCase() === 'video/H264');
someTransceiver.setCodecPreferences(codecs);
```

The sentence "this ensures that we always have something to offer" is there to avoid the subsequent problem, not being able to negotiate a codec when the transceiver direction changes like this:
```
t.direction = 'recvonly'
t.setCodecPreferences({a-codec-you-can-only-encode})
```
If this was taken to createOffer and setLocalDescription it might lead to an m= line that is rejected which would cause the transceiver to be  stopped which would be quite surprising.

setCodecPreferences might check the direction in that case but that just leads to the following scenario:
```
t.setCodecPreferences({a-codec-you-can-only-encode})
t.direction = 'recvonly'
```
One approach to solving this is that an attempt to set the direction to 'recvonly' might throw. This also requires a compability check in setCodecPreferences in case the direction was set before setCodecPreferences.

The other alternative would be to let createOffer (which knows about send+recv codes as well as the direction) throw an error when it detects this mismatch. It gets a bit finicky to figure out which transceiver was causing this, the mid *might* not be available.

Overall I think this is an edge case but the first example is a fairly common flow that is currently broken (depending on the codec support)

[...]

PR #2910, opened 2023-11-15
Document test updates associated with amendments

dontcallmedom:
This is a first pass at associating amendments to the relevant test cases and updates made to WPT as part of the amendment documenting process.

I have not documented associated test updates for the following PRs: #2758 #2759 #2794 #2797 #2801 #2847 #2853 #2861 - most likely because no updates were brought, but it could be that I missed them too.

It would be useful if someone could review that the annotations made in `amendments.json` are correct. Namely that:
* [x] the changes in the following PRs aren't usefully testable (either because they refer to implementation dependent decisions, or because they don't impact what's exposed to JS): #2679 #2686 #2700 #2704  #2753 #2767 #2808 #2814 #2834 #2841 #2885
* [x] the changes in the following PRs were already tested (in general, because the change in the spec is made to match implementations): #2680 #2742 #2829 #2875 #2876
* [x] the association between the other PRs and their WPT PRs is correct and covers sufficiently the said change:
   * [x] #2687 → web-platform-tests/wpt#43171
   * [x] #2689 → web-platform-tests/wpt#43167
   * [x] #2691 → web-platform-tests/wpt#43166
   * [x] #2708 -> web-platform-tests/wpt#21025
   * [x] #2744 #2748 #2832 → web-platform-tests/wpt#35703 web-platform-tests/wpt#43172
   * [x] #2754 → web-platform-tests/wpt#34912
   * [x] #2763 → web-platform-tests/wpt#36157
   * [x] #2772 → web-platform-tests/wpt#37477
   * [x] #2773 → web-platform-tests/wpt#36572
   * [x] #2774 → web-platform-tests/wpt#37477
   * [x] #2775 → web-platform-tests/wpt#37477
   * [x] #2785 → web-platform-tests/wpt#43173
   * [x] #2797 → web-platform-tests/wpt#37477
   * [x] #2799 → web-platform-tests/wpt#37477
   * [x] #2800 → web-platform-tests/wpt#37477
   * [x] #2813 → web-platform-tests/wpt#36155 web-platform-tests/wpt#37477
   * [x] #2813 → web-platform-tests/wpt#37477
   * [x] #2902 → https://phabricator.services.mozilla.com/D190744
   * [x] #2909 → web-platform-tests/wpt#41663
   * [x] #2913 → web-platform-tests/wpt#43601

Merging this PR will also require that any new amendment documents its test status (either as "already-tested", "not-testable", or as an array of pointers to WPT pull requests).

***
[...]

PR #2926, opened 2024-01-16
make setCodecPreferences only look at receive codecs

fippo:
aligning with [JSEP](https://www.rfc-editor.org/rfc/rfc8829.html#name-setcodecpreferences): 
> setCodecPreferences does not directly affect which codec the implementation decides to send. 
> It only affects which codecs the implementation indicates that it prefers to receive

partial fix for #2888

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -11183,7 +11183,7 @@ interface RTCRtpTransceiver {
               <dd>
                 <p>
                   The {{setCodecPreferences}} method overrides the default
-                  codec preferences used by the <a>user agent</a>. When
+                  receive codec preferences used by the <a>user agent</a>. When
                   generating a session description using either
                   {{RTCPeerConnection/createOffer}} or
                   {{RTCPeerConnection/createAnswer}}, the <a>user agent</a>
@@ -11221,8 +11221,6 @@ interface RTCRtpTransceiver {
                 <p>
                   {{setCodecPreferences}} will reject attempts to set <var>codecs</var>
                   [= codec match | not matching =] codecs found in
-                  {{RTCRtpSender}}.{{RTCRtpSender/getCapabilities}}(<var>kind</var>)
-                  or
                   {{RTCRtpReceiver}}.{{RTCRtpReceiver/getCapabilities}}(<var>kind</var>),
                   where <var>kind</var> is the kind of the
                   {{RTCRtpTransceiver}} on which the method is called.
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

alvestrand (approved):
Discussed in the January 16 WEBRTC WG meeting. I believe this PR is consistent with the outcome of that discussion.

henbos (approved):

fippo:
Test added in https://github.com/web-platform-tests/wpt/pull/44318

fippo:
Amendment worked but validation now throws
`"file:/home/runner/work/webrtc-pc/webrtc-pc.w3c/webrtc.html":13397.15-13397.67: error: Element “div” not allowed as child of element “dl” in this context. (Suppressing further errors from this subtree.)`
but I am not touching a `dl`, only a `dt`? @dontcallmedom you are my only hope!

dontcallmedom:
#2910 had the same bogus error, which I didn't quite get to debug; will raise again in my todo list :)

webrtc.html
```diff
@@ -11181,9 +11181,10 @@ <h2>
                 <dfn data-idl="">setCodecPreferences</dfn>
               </dt>
               <dd>
+		<div id="setcodecpreferences-receive">
```
henbos, line 11184: I think I found one more place we forgot to upate. Should we also delete this sentence?

https://w3c.github.io/webrtc-pc/#dfn-final-steps-to-create-an-answer says
> If transceiver.[direction](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-direction) is "[sendonly](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiverdirection-sendonly)" or "[sendrecv](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiverdirection-sendrecv)", exclude any codecs not included in the [list of implemented send codecs](https://w3c.github.io/webrtc-pc/#dfn-list-of-implemented-send-codecs) for kind.

Since it should be OK for pc1 to receive X and pc2 to receive Y, we shouldn't filter the preference on send codecs because that would exclude X from pc1's offer if X was a recvonly codec.

fippo: That sentence needs to stay. Until we are ready to more officially deal with answer codecs that are not in the offer which JSEP allows.

webrtc.html
```diff
@@ -11221,8 +11222,6 @@ <h2>
                 <p>
                   {{setCodecPreferences}} will reject attempts to set <var>codecs</var>
                   [= codec match | not matching =] codecs found in
-                  {{RTCRtpSender}}.{{RTCRtpSender/getCapabilities}}(<var>kind</var>)
-                  or
```
henbos, line 11225: Shouldn't the transceiver direction filtering when generating the SDP already take care of this? https://github.com/w3c/webrtc-pc/issues/2932#issuecomment-1933628242

If we don't allow preferring sendonly codecs for the sendonly transceiver usecase, then it's impossible to do an O/A exchange that results in a sendonly codec WITHOUT having to modify SDP

henbos (commented):
There be dragons

henbos (changes requested):
Removing my approval because I think I've identified that this PR breaks sendonly use cases without modifying SDP and considering we already have a transciever.direction filtering in place, I don't see any problem with what the spec says today. See https://github.com/w3c/webrtc-pc/issues/2932#issuecomment-1933634998

fippo:
Can you give an example of a sendonly codec?

webrtc.html
henbos: Nevermind

henbos (approved):
Suspecting what I was reacting to was an implementation bug, not a spec issue. Approving again

alvestrand:
Editors can integrate once tests have landed (and @dontcallmedom has ensured that amendment thingy is right.

jan-ivar (approved):

jan-ivar (approved):

PR #2948, opened 2024-03-19
Clarify when RTCIceCandidate's relayProtocol and url members are null.

### LINKED ISSUE #2947: RTCIceCandidate's relayProtocol and url members can be null but not absent
The nullable readonly attribute [relayProtocol](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-relayprotocol) says: _"This is only present for local relay candidates."_ which is wrong since readonly attributes are always [present](https://infra.spec.whatwg.org/#map-exists), instead we should clarify when it is `null`.

Similarly, [url](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-url) says: _"For local candidates of type "[srflx](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidatetype-srflx)" or type "[relay](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidatetype-relay)" this is the URL of the ICE server from which the candidate was obtained."_ — add _"Otherwise, it is null"_.

This seems minor and obvious, so marking it editorial.

jan-ivar:
Fixes #2947.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/2948.html" title="Last updated on Mar 19, 2024, 10:03 PM UTC (8ea7c8b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2948/a913714...jan-ivar:8ea7c8b.html" title="Last updated on Mar 19, 2024, 10:03 PM UTC (8ea7c8b)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -6366,24 +6366,21 @@ interface RTCIceCandidate {
                 </dd>
                 <dt class="add-to-usernameFragment-desc">
                   <dfn>relayProtocol</dfn> of type <span class=
-                  "idlMemberType">RTCIceServerTransportProtocol</span>
+                  "idlMemberType">RTCIceServerTransportProtocol</span>, readonly, nullable
                 </dt>
                 <dd class="add-to-usernameFragment-desc">
-                  <p>
-                    The protocol used by the endpoint to communicate with the TURN server. This
-                    is only present for local relay candidates.
-                  </p>
+                  For local relay candidates this is the protocol used by the endpoint to
+                  communicate with the TURN server. For all other candidates it is `null`.
                 </dd>
                 <dt class="add-to-usernameFragment-desc">
                   <dfn>url</dfn> of type <span class=
-                  "idlMemberType">DOMString</span>
+                  "idlMemberType">DOMString</span>, readonly, nullable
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -6366,24 +6366,21 @@ <h2>
...
                   <dfn>relayProtocol</dfn> of type <span class=
-                  "idlMemberType">RTCIceServerTransportProtocol</span>
+                  "idlMemberType">RTCIceServerTransportProtocol</span>, readonly, nullable
                 </dt>
                 <dd class="add-to-usernameFragment-desc">
-                  <p>
-                    The protocol used by the endpoint to communicate with the TURN server. This
-                    is only present for local relay candidates.
-                  </p>
+                  For local relay candidates this is the protocol used by the endpoint to
```
alvestrand, line 6372: for consistency of language, should this be "for local candidates of type {{relay}}"?
someone could interpret the difference in markup between the two sections as meaning that there are other candidates of type relay.

youennf (approved):

PR #2958, opened 2024-04-17
Rename to "codec dictionary match" algorithm and remove index consiency claim

### LINKED ISSUE #2955: Confusion over duplicates in setCodecPreferences? Why allow them?
I spotted a dubious claim in the [setCodecPreferences](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setcodecpreferences) algorithm: 
- _"4. Remove any duplicate values in codecs. Start at the back of the list such that the order of the codecs is maintained; the index of the first occurrence of a codec within the list is the same before and after this step."_

The claim about indexes remaining the same seems wrong. E.g. with codecs `[A, A₁, B]` removing `A₁` results in `[A, B]` necessarily altering the index B.*

We should also link "duplicate" to the [codec match](https://w3c.github.io/webrtc-pc/#dfn-codec-match) algorithm.

But why allow duplicates in the first place? This seems to only invite trouble.

---
<sub>*) This is implementation language, e.g. c++, operating on infra or WebIDL types, so JS concepts like sparse arrays shouldn't apply.

jan-ivar:
Fixes #2955.

Once merged, links in the following places would need to be updated:
1. https://w3c.github.io/webrtc-extensions/#addtransceiver

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -11299,7 +11299,7 @@ interface RTCRtpTransceiver {
                 </p>
                 <p>
                   {{setCodecPreferences}} will reject attempts to set <var>codecs</var>
-                  [= codec match | not matching =] codecs found in
+                  [= codec dictionary match | not matching =] codecs found in
                   {{RTCRtpReceiver}}.{{RTCRtpReceiver/getCapabilities}}(<var>kind</var>),
                   where <var>kind</var> is the kind of the
                   {{RTCRtpTransceiver}} on which the method is called.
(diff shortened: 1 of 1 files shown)


alvestrand:
Why rename?
I can't see anything else that needs to use the name "codec match", and as pointed out in the CL, renaming makes extra work in multiple other places. I fail to see the benefit of that.

The other changes in the CL look fine to me.

alvestrand (approved):

jan-ivar:
@jan-ivar to merge a PR for https://w3c.github.io/webrtc-extensions/#addtransceiver right after merging this.

Issue #2930, opened 2024-01-18
Consider making RTCIceCandidatePair an interface


jan-ivar:
Currently, [RTCIceCandidatePair](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidatepair) is a dictionary containing two interfaces:
```webidl
dictionary RTCIceCandidatePair {
  RTCIceCandidate local;
  RTCIceCandidate remote;
};
```
This seems like a mistake, allowing JS to compose invalid pairs and pass them as inputs.

Valid pairs are [formed](https://w3c.github.io/webrtc-extensions/#dfn-form) exclusively by the ICE Agent, a reality better expressed by a more restrictive contract:
```webidl
[Exposed=Window]
interface RTCIceCandidatePair {
  readonly attribute RTCIceCandidate local;
  readonly attribute RTCIceCandidate remote;
};
```
This would benefit web developers by enforcing an invariant, and spec writers by improved WebIDL type checking, simpler algorithms, and even simpler APIs in the future.

I believe this should be mostly backwards compatible since all APIs that take pairs expect valid ones. cc @sam-vi

I propose we make this change.

jan-ivar:
@sam-vi thoughts on this? Should we discuss it at the next meeting perhaps?

foolip:
This dictionary is also used as a method argument in webrtc-extensions. How would that be affected?

jan-ivar:
I think the goal would be for [selectCandidatePair](https://w3c.github.io/webrtc-extensions/#dom-rtcicetransport-selectcandidatepair) and [removeCandidatePair](https://w3c.github.io/webrtc-extensions/#dom-rtcicetransport-removecandidatepair) to take the interface object (instead of a dictionary with two member interface objects).

The two methods throw `NotFoundError` on invalid pairs already. Pairs serve as handles to resources effectively.

I didn't find any WPT tests for these methods, so they're hopefully not implemented yet @sam-vi is that right?

jan-ivar:
> Pairs serve as handles to resources effectively.

Not in an owning way, I should say. E.g. we could have used IDs for this instead,  but interface objects seem better.

sam-vi:
> @sam-vi thoughts on this? Should we discuss it at the next meeting perhaps?

I agree with this proposal. I will work on some next steps.

> This seems like a mistake, allowing JS to compose invalid pairs and pass them as inputs.

[...]

PR #2961, opened 2024-04-18
Convert RTCIceCandidatePair dictionary to an interface

sam-vi:
Addresses: #2930.

Change the `RTCIceCandidatePair` dictionary to an interface and change the required dictionary members to readonly attributes backed by internal slots.

No constructor is described in the IDL to prevent an application from constructing arbitrary pairs. Only the user agent can construct valid `RTCIceCandidatePair` in response to the ICE agent forming candidate pairs, and to represent the selected candidate pair.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -12670,35 +12670,56 @@ interface RTCIceTransport : EventTarget {
         </section>
         <section id="rtcicecandidatepair">
           <h3>
-            <dfn>RTCIceCandidatePair</dfn> Dictionary
+            <dfn>RTCIceCandidatePair</dfn> Interface
           </h3>
+          <p>
+            This interface represents an ICE candidate pair, described in Section 4 in [[RFC8445]]. An {{RTCIceCandidatePair}} is a pairing of a local and a remote {{RTCIceCandidate}}.
+          </p>
+          <p>
+            To <dfn class="abstract-op">create an RTCIceCandidatePair</dfn> with {{RTCIceCandidate}} objects, |local:RTCIceCandidate| and |remote:RTCIceCandidate|, run the following steps:
+          </p>
+          <ol class="algorithm">
+            <li>
+              Let |candidatePair:RTCIceCandidatePair| be a newly created {{RTCIceCandidatePair}} object.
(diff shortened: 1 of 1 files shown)


alvestrand:
Since it's an observable change, it needs an amendments.json update.
@dontcallmedom  may need to help.

dontcallmedom:
I have added the basic amendment description, but the CI will fail (as designed) until a test update can be referenced.

It may be that for this particular change, it would be enough to test it via the idl_harness change, which itself will be done automatically once this pull request gets merged (since WPT will automatically get the updated IDL extracted from the spec). If that is so, it would be appropriate to merge as is, and file an issue to update the amendment once the WPT update lands (which given the various intermediary steps involved would probably take a few days)

sam-vi:
So if I understand correctly, if the test via idl_harness change is deemed sufficient, then there is no further action right now, correct?

Once the IDL is updated in WPT, the new `rtcicecandidatepair` amendment should be updated with 
```
"tests": [
  "webrtc/idlharness.https.window.html"
],
"testUpdates": [
  "web-platform-tests/wpt#___"
],
```

In that case, could the PR be considered at the next editors' call?

alvestrand:
I'm not sure how to navigate the update of the WPT idlharness test to get a number to put into this PR.
May need a followup.

alvestrand:
@dontcallmedom help?

dontcallmedom:
this is the kind of PR that updates the IDL fragments on which the idl harness relies: https://github.com/web-platform-tests/wpt/pull/46056

again, the WPT PR would only get generated some days after this WebRTC PR gets merged, so let's not block on it (but instead, let's file an issue when this PR gets merged)

saschanaz:
Have you filed implementation bugs for this?

saschanaz:
cc @jan-ivar, I think W3C in general needs WHATWG-like PR template to make sure bugs are filed for all browser engines.

jan-ivar:
Firefox does not implement getSelectedCandidatePair() yet so this is already covered by [bug 1307994](https://bugzil.la/1307994).

PR #2972, opened 2024-05-08
Separate Send/ReceiveCodecs and NegotiatedCodecs

### LINKED ISSUE #2956: receiver.getParameters().codecs seems under-specified.
[receiver.getParameters](https://w3c.github.io/webrtc-pc/#dom-rtcrtpreceiver-getparameters) has this non-normative note:

image

It suggests that post sRD(answer), `receiver.getParameters().codecs` is updated to only reflect what's been negotiated.

But I can't find any normative steps to support this note:
- [[ReceiveCodecs]] is [initialized]() to _"the [list of implemented receive codecs](https://w3c.github.io/webrtc-pc/#dfn-list-of-implemented-receive-codecs) for kind"_
- [[ReceiveCodecs]] is only ever assigned to [[LastStableStateReceiveCodecs]] (on rollback)
- [[ReceiveCodecs]] does not appear to ever be modified
- The "enabled" flag is only ever set to true

Is the note wrong or are we missing some normative steps? What was the intent?

alvestrand:
This makes it unambiguous what to return in RTCRtpSender/Receiver's getParameters() function (null before negotiation, filled in after, which corresponds to implementations).

Fixes #2956

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1864,9 +1864,9 @@
                           <li>
                             <p>
                               Set
-                              <var>transceiver</var>.{{RTCRtpTransceiver/[[Receiver]]}}.{{RTCRtpReceiver/[[LastStableStateReceiveCodecs]]}}
+                              <var>transceiver</var>.{{RTCRtpTransceiver/[[Receiver]]}}.{{RTCRtpReceiver/[[LastStableStateNegotiatedCodecs]]}}
                               to
-                              <var>transceiver</var>.{{RTCRtpTransceiver/[[Receiver]]}}.{{RTCRtpReceiver/[[ReceiveCodecs]]}}.
+                              <var>transceiver</var>.{{RTCRtpTransceiver/[[Receiver]]}}.{{RTCRtpReceiver/[[NegotiatedCodecs]]}}.
                             </p>
                           </li>
                         </ol>
@@ -2189,11 +2189,15 @@
                               </li>
                               <li>
                                 <p>
-                                  For each of the codecs that <var>description</var> negotiates for receiving, execute the following steps:
+                                  Clear the <var>transceiver</var>.{{RTCRtpTransceiver/[[Receiver]]}}.{{RTCRtpReceiver/[[NegotiatedCodecs]]}} slot.
+                                </p>
+                                <p>
+                                  Then, For each of the codecs that <var>description</var> negotiates for receiving, execute the following steps:
                                       <ol>
                                         <li>Locate the matching codec description in <var>transceiver</var>.{{RTCRtpTransceiver/[[Receiver]]}}.{{RTCRtpReceiver/[[ReceiveCodecs]]}}.</li>
                                         <li>If the matching codec description is not found, abort these steps.</li>
                                         <li>Set the "enabled" flag in the matching codec description to "true".</li>
+                                        <li>Add the codec description to the {{RTCRtpReceiver/[[NegotiatedCodecs]]}} internal slot.</li>
                                       </ol>
                                   
                                 </p>
@@ -2244,11 +2248,15 @@
                                   </li>
                                   <li>
                                     <p>
-                                      For each of the codecs that <var>description</var> negotiates for sending, execute the following steps:
+                                      Clear the <var>transceiver</var>.{{RTCRtpTransceiver/[[Receiver]]}}.{{RTCRtpReceiver/[[NegotiatedCodecs]]}} slot.
+                                    </p>
+                                    <p>
+                                      Then, for each of the codecs that <var>description</var> negotiates for sending, execute the following steps:
                                       <ol>
                                         <li>Locate the matching codec description in <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendCodecs]]}}.</li>
                                         <li>If the matching codec description is not found, abort these steps.</li>
                                         <li>Set the "enabled" flag in the matching codec description to "true".</li>
+                                        <li>Add the codec description to the {{RTCRtpReceiver/[[NegotiatedCodecs]]}} internal slot.</li>
                                       </ol>
                                   </li>
                                   <li>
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -2244,13 +2226,37 @@ <h4>
...
                                     <p>
-                                      For each of the codecs that <var>description</var> negotiates for sending, execute the following steps:
+                                      Clear the <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[NegotiatedCodecs]]}} slot.
+                                      <br>
+                                      Then, for each of the codecs that <var>description</var> negotiates for sending, execute the following steps:
                                       <ol>
                                         <li>Locate the matching codec description in <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendCodecs]]}}.</li>
                                         <li>If the matching codec description is not found, abort these steps.</li>
                                         <li>Set the "enabled" flag in the matching codec description to "true".</li>
+                                        <li>Add the codec description to the sender's {{RTCRtpSender/[[NegotiatedCodecs]]}} internal slot.</li>
```
jan-ivar, line 2236: This puts tuples in [[NegotiatedCodecs]], which while expedient, seems a bit confusing since they'll always have "enabled" true. Is this on purpose?

Also since these are internal objects kept in lists in internal slots, these entries are added by reference, so if we ever introduce code that sets an enabled flag to false, it would affect both [[SendCodecs]] and [[NegotiatedCodecs]] potentially causing the codec entry to disappear from getParameters(). Is this desirable?

Even if desirable, this behavior seems a bit hard to grasp from this prose.

The other thing I noticed is that `payloadType` isn't filled in.

While [[[SendCodecs]]](https://w3c.github.io/webrtc-pc/#dfn-sendcodecs) is _"a list of [tuple](https://infra.spec.whatwg.org/#tuple)s, each containing an [RTCRtpCodecParameters](https://w3c.github.io/webrtc-pc/#dom-rtcrtpcodecparameters) dictionary and an "enabled" boolean"_ it's actually initialized to _"an [implementation-defined](https://infra.spec.whatwg.org/#implementation-defined) list of [RTCRtpCodecCapability](https://w3c.github.io/webrtc-pc/#dom-rtcrtpcodeccapability) dictionaries"_

Spot the different dictionary types above. RTCRtpCodecParameters != RTCRtpCodecCapability

The former contains payloadType, the latter doesn't. So it needs to be filled in here I think.

It might be good to use language shown in https://infra.spec.whatwg.org/#example-tuple to be more precise. This includes declaring the tuple where it's created with member names like `codec` and `enabled`, and "Let _entry_ be ..." etc.

This would let us be more precise and be clear whether we want to put the tuples in [[NegotiatedCodecs]] or simply the _entry_.codec.

jan-ivar (changes requested):

webrtc.html
```diff
@@ -8800,6 +8806,9 @@ <h3>
               [=RTCRtpSender/list of implemented send codecs=], with the "enabled" flag
               set in an implementation defined manner.
             </p>
+            <p>
+              Let <var>sender</var> have a <dfn data-dfn-for="RTCRtpSender">[[\NegotiatedCodecs]]</dfn> internal slot, consisting of {{RTCRtpCodecParameters}}, and initialized to an empty list.
```
jan-ivar, line 8810: Ah, this says [[NegotiatedCodecs]] contains RTCRtpCodecParameters, not tuples. Good, but means the code above needs to be fixed to not add tuples.

webrtc.html
alvestrand: The intent of "negotiatedCodecs" is to reflect the currently negotiated codecs. The "enabled" flag has no purpose there, so it's not carried along.
Good point about using RTCRtpCodecCapability and filling in the PT field - this accentuates the need for it being a copy not a reference.

webrtc.html
alvestrand: Or rather ... construct an RTCRtpCodecCapability from the RTCRtpCodec and add the PT from the description.

webrtc.html
alvestrand: Actually RTCRtpCodecCapability is a dictionary with no extra members over RTCRtpCodec. We shouldn't be using it.

amendments.json
```diff
@@ -604,6 +604,14 @@
       "type": "correction",
       "status": "candidate",
       "id": 26
+    },
+    {
+      "description": "Separate Send/ReceiveCodecs and NegotiatedCodecs.",
+      "pr": 2972,
+      "difftype": "modify",
```
alvestrand, line 611:
```suggestion
      "difftype": "modify",
      "tests": [
        "webrtc/RTCRtpSender-getParameters.html",
        "webrtc/RTCRtpReceiver-getParameters.html",
        ],
        "testUpdates": [
           "web-platform-tests/wpt#46840"
         ],
```

amendments.json
```diff
@@ -604,6 +604,21 @@
...
+      "tests": [
+        "webrtc/RTCRtpSender-getParameters.html",
+        "webrtc/RTCRtpReceiver-getParameters.html",
```
dontcallmedom, line 614:
```suggestion
        "webrtc/RTCRtpReceiver-getParameters.html"
```

webrtc.html
```diff
@@ -2244,13 +2226,37 @@ <h4>
                                   </li>
                                   <li>
                                     <p>
-                                      For each of the codecs that <var>description</var> negotiates for sending, execute the following steps:
+                                      Clear the <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[NegotiatedCodecs]]}} slot.
+                                      <br>
+                                      Then, for each of the codecs that <var>description</var> negotiates for sending, execute the following steps:
                                       <ol>
                                         <li>Locate the matching codec description in <var>transceiver</var>.{{RTCRtpTransceiver/[[Sender]]}}.{{RTCRtpSender/[[SendCodecs]]}}.</li>
                                         <li>If the matching codec description is not found, abort these steps.</li>
```
jan-ivar, line 2234: Which steps are aborted here? One of the nested for-loops or sLD/sRD without error? Either seems wrong.

Not failing assumes negotiated codecs is a subset of this client's [[SendCodecs]]. The [1.0 spec](https://github.com/w3c/webrtc-pc/issues/2956#issuecomment-2072500942) didn't have this:

image

...which, AFAIK meant `sender.getParameters().codecs` might include _remote_ codecs not found locally, but found in the remote answer, which [JSEP](https://rtcweb-wg.github.io/jsep/#rfc.section.5.3.1) seems to allow: _"Any currently available media formats that are not present in the current remote description MUST be added after all existing formats."_

Why is this new limitation needed?

It's also unclear what constitutes a "matching codec description".

jan-ivar (changes requested):

PR #2983, opened 2024-07-12
Update localDescription etc. attributes rather than just their sdp.

### LINKED ISSUE #2977: PC.local_description and friends - snapshot views or dynamic views?
In the current specification, current_local_description is described as an attribute that returns the content of the [[CurrentLocalDescription]] slot.
Returning an object in Javascript is basically passing a reference, not cloning the value.

So - in the current model, if we have:

x = pc.current_local_description
... time passes
a new candidate is added
... time passes
what is the value of x.sdp? does it include the new candidate or not?

This came up while hunting for a thread access issue inside WebRTC; it made it clear to me that life would be much simpler if those accessors returned a copy of the current state of [[LocalDescription]], and did not change after returning. But that is not the way attributes usually work.

What should we intend here?
(all the attributes of an RTCSessionDescription are readonly, so the Javascript side can't change it. The question is whether the platform can change it.)

jan-ivar:
Fixes #2977.

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -4586,9 +4586,10 @@ interface RTCPeerConnection : EventTarget  {
                                     <var>connection</var>.{{RTCPeerConnection/[[PendingRemoteDescription]]}}
                                     is not <code>null</code>, and represents
                                     the ICE [= generation =] for which
-                                    <var>candidate</var> was processed, add
-                                    <var>candidate</var> to
-                                    <var>connection</var>.{{RTCPeerConnection/[[PendingRemoteDescription]]}}.sdp.
+                                    <var>candidate</var> was processed, set
+                                    <var>connection</var>.{{RTCPeerConnection/[[PendingRemoteDescription]]}}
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -12274,6 +12278,17 @@ <h3>
           changing. The selected pair and {{RTCIceTransportState}} are related
           and are handled in the same task.
         </p>
+        <p>
+          To <dfn class="abstract-op" data-lt="clone a description">clone a description</dfn>
+          with members <var>type</var> and <var>sdp</var>, return a new {{RTCSessionDescription}}
+          object constructed with {{RTCSessionDescription/type}} set to <var>type</var>,
+          and {{RTCSessionDescription/sdp}} set to <var>sdp</var>.
+        </p>
+        <ol class=algorithm>
```
dontcallmedom, line 12287: does this `<ol>` serve a purpose?

alvestrand (approved):
Looks good to me.

Issue #2950, opened 2024-03-25
Missing tests for candidate amendments


dontcallmedom:
The following pull requests have brought candidate amendments for which no tests are documented at the moment:
* [  ] #2758
* [  ]  #2759
* [  ]  #2760
* [  ]  #2794
* [  ]  #2801
* [  ]  #2853
* [  ]  #2861
* [  ]  #2935

(this may be because there is no test, or because I couldn't find the relevant test)

This is a follow up to #2910

PR #2996, opened 2024-09-03
Complete ice server url validation

### LINKED ISSUE #2912: Fix ICE server URL querystring validation
follow up to #2660: #2853 left a TODO: `validate ?transport=udp|tcp`

dontcallmedom:
close #2912

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/pull/2996.html" title="Last updated on Oct 8, 2024, 7:16 AM UTC (22f85c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/2996/4888a48...22f85c4.html" title="Last updated on Oct 8, 2024, 7:16 AM UTC (22f85c4)">Diff</a>

### DIFF
--- a/amendments.json
+++ b/amendments.json
@@ -57,8 +57,14 @@
   "validate-ice-server-algo": [
     {
       "description": "Use the url spec to parse ice server urls",
-      "pr": 2853,
+      "pr": [2853, 2996],
       "type": "correction",
+      "tests": [
+	"webrtc/RTCConfiguration-iceServers.html"
+      ],
+      "testUpdates": [
+	"web-platform-tests/wpt#47959"
+      ],
       "status": "candidate",
       "id": 33
     }
--- a/webrtc.html
+++ b/webrtc.html
@@ -3143,7 +3143,7 @@
                   For "turn" and "turns" schemes, this and the steps below validate
                   [[!RFC7065]] section 3.1.</p>
               </li>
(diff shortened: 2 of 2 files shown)


dontcallmedom:
(the CI error is because we don't have tests for this; as noted in #2950 this is only a continuation of the lack of test for this change; see also #2997)

jan-ivar:
> ... reusing the application/x-www-form-urlencoded parser means making ?&transport=udp parsable

Less custom URL parsing seems in line with being _"consistent with how we treat URLs elsewhere in the platform."_. https://github.com/w3c/webrtc-pc/issues/2660#issuecomment-1494140703

> an alternative would be to instead do the split on `=` ourselves (rather than invoking the URL algorithm), limiting the set of acceptable query strings to `?transport=udp` and `?transport=tcp`

OTOH, if this simple string compare lets us keep this in REC, why not do that for now, and add application/x-www-form-urlencoded back in later?

That seems a reasonable way to increment here to me.

dontcallmedom:
I've switched to using a simple string comparison; I'm not sure using `application/x-www-form-urlencoded` parsing buys us much when it is not clear the RFC means it that way.

jan-ivar:
Yeah that algorithm supports multiple query parameters whereas our RFC does not, so it's arguably odd to support `&` at all

jan-ivar (approved):

PR #3000, opened 2024-09-09
Clarify that stopping of sending and receiving happens in parallel

jan-ivar:
Transmission of media happens off main thread, so stopping it necessarily needs to as well, to not block main thread on it.


henbos (approved):

alvestrand:
Seems to be unobservable, so editorial.

PR #3005, opened 2024-09-19
Fix timestamp definition to align with webrtc-stats

### LINKED ISSUE #3004: RTCStats::timestamp discrepancy between webrtc-stats and webrtc-pc
In webrtc-stats it [says](https://w3c.github.io/webrtc-stats/#dom-rtcstats-timestamp):

> Timestamps are expressed with [DOMHighResTimeStamp](https://www.w3.org/TR/hr-time-3/#dom-domhighrestimestamp) [[HIGHRES-TIME](https://w3c.github.io/webrtc-stats/#bib-highres-time)], and are defined as [Performance](https://www.w3.org/TR/hr-time-3/#dom-performance).[timeOrigin](https://www.w3.org/TR/hr-time-3/#dom-performance-timeorigin) + [Performance](https://www.w3.org/TR/hr-time-3/#dom-performance).[now](https://www.w3.org/TR/hr-time-3/#dom-performance-now)() at the time the information is collected.

Which is consistent with most timestamps that we've defined for the last several years and ensures monotonically incrementing etc, so it's always possible to do "delta fooStat / delta timestamp".

But the timestamp definition in webrtc-pc is really old and it [says](https://w3c.github.io/webrtc-pc/#dom-rtcstats-timestamp):

> The [timestamp](https://github.com/w3c/webrtc-pc/issues/new#dom-rtcstats-timestamp), of type [DOMHighResTimeStamp](https://www.w3.org/TR/hr-time-3/#dom-domhighrestimestamp), associated with this object. The time is relative to the UNIX epoch (Jan 1, 1970, UTC).

It doesn't really say but this sounds like a wall clock that is not necessarily always monotonically incrementing. I think this is a mistake and it should say Performance here as well to align with all other timestamps, otherwise there will be edge cases where things break.

We noticed a bug relating to wall clock vs performance where it would diverge if the laptop went to sleep and then the page resumed...

henbos:
Fixes #3004

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/3005.html" title="Last updated on Sep 30, 2024, 2:23 PM UTC (09f394b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/3005/0c2ee4b...henbos:09f394b.html" title="Last updated on Sep 30, 2024, 2:23 PM UTC (09f394b)">Diff</a>

### DIFF
--- a/amendments.json
+++ b/amendments.json
@@ -1091,6 +1091,9 @@
       "pr": 3005,
       "tests": [
   "webrtc/RTCPeerConnection-getStats-timestamp.https.html"
+      ],
+      "testUpdates": [
+  "web-platform-tests/wpt#48361"
       ]
     }
   ]


jan-ivar:
Not editorial IMHO, so you'll need to do an amendment

henbos:
WPTs in progress: https://chromium-review.googlesource.com/c/chromium/src/+/5887191

jan-ivar (approved):

PR #3018, opened 2024-11-07
Filter codecs preferences on transceiver direction

### LINKED ISSUE #3006: Align spec /w codec direction decision
How to deal with codecs and directionality was discussed at TPAC 2024:
- [Minutes](https://www.w3.org/2024/09/24-webrtc-minutes.html#4c6d)
- [Slides](https://docs.google.com/presentation/d/1dFntwgSd9MrLnd1Sf58eBHstkodUd30WSKGIaK5K2CM/edit#slide=id.g302dfc81162_4_26)
- [Recording](https://www.youtube.com/watch?v=aDKpOfGhbXk&t=44m39s)

There was a strong preference for Proposal A, i.e. to "keep it simple": sendonly transceivers can do all send codecs, recvonly transcievers can do all receive codecs, and sendrecv transceivers can only do codecs and profiles that you can BOTH send AND receive with. So if you want to do fancy unidirectional codecs or profiles you use multiple m= sections.

Let this issue track making sure the spec aligns with Proposal A and allow us to merge other codec related issues as fixed by this issue

henbos:
Fixes #3006.

As decided by the WG, this PR implements [Proposal A at TPAC](https://docs.google.com/presentation/d/1dFntwgSd9MrLnd1Sf58eBHstkodUd30WSKGIaK5K2CM/edit#slide=id.g302dfc81162_4_26), i.e:
- A sendonly transceiver can prefer sendonly codecs.
- A recvonly transceiver can prefer recvonly codecs.
- A sendrecv transceiver can only prefer codecs that we can both send and receive with. If you want to use both sendonly codecs and recvonly codecs you need to have two transceivers.

This is supported via filtering based on RTCRtpTransceiver.direction. In addition, the case where filtering results in an empty codec preferences this is treated as "no preference" as per [Proposal A at Virtual Interim](https://docs.google.com/presentation/d/1tfFiR2xqyEybbmnGUALqdWHpDx2W47YnlO65gI0HvBA/edit#slide=id.g30790f41697_14_26).

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/3018.html" title="Last updated on Nov 28, 2024, 1:44 PM UTC (50e9841)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/3018/9db156a...henbos:50e9841.html" title="Last updated on Nov 28, 2024, 1:44 PM UTC (50e9841)">Diff</a>

### DIFF
--- a/amendments.json
+++ b/amendments.json
@@ -894,14 +894,14 @@
   "setcodecpreferences-filter-direction": [
     {
       "description": "setCodecPreferences takes both send and receive codecs which get filtered at negotiation based on transceiver direction",
-      "pr": 1337,
-      "type": "correction",
+      "pr": 3018,
+      "type": "addition",
       "status": "candidate",
       "tests": [
 	"webrtc/RTCRtpTransceiver-setCodecPreferences-direction.html"
       ],
       "testUpdates": [
-        "web-platform-tests/wpt#1337"
+        "web-platform-tests/wpt#1234"
       ],
       "id": 50
     }
--- a/amendments.json
+++ b/amendments.json
@@ -876,24 +876,9 @@
       "id": 40
     }
   ],
-  "setcodecpreferences-receive": [
+  "setcodecpreferences-send-and-receive": [
     {
-      "description": "setCodecPreferences only takes into account receive codecs",
-      "pr": 2926,
-      "type": "correction",
-      "status": "candidate",
-      "tests": [
-	"webrtc/RTCRtpTransceiver-setCodecPreferences.html"
-      ],
-      "testUpdates": [
-        "web-platform-tests/wpt#44318"
-      ],
-      "id": 42
-    }
-  ],
-  "setcodecpreferences-filter-direction": [
-    {
-      "description": "setCodecPreferences takes both send and receive codecs which get filtered at negotiation based on transceiver direction",
+      "description": "setCodecPreferences supports both send and receive codecs (filtered by direction)",
       "pr": 3018,
       "type": "addition",
       "status": "candidate",
--- a/amendments.json
+++ b/amendments.json
@@ -891,6 +891,21 @@
       "id": 42
     }
   ],
+  "setcodecpreferences-filter-direction": [
+    {
+      "description": "setCodecPreferences takes both send and receive codecs which get filtered at negotiation based on transceiver direction",
+      "pr": 1337,
+      "type": "correction",
+      "status": "candidate",
+      "tests": [
+	"webrtc/RTCRtpTransceiver-setCodecPreferences-direction.html"
+      ],
+      "testUpdates": [
+        "web-platform-tests/wpt#1337"
+      ],
+      "id": 50
+    }
+  ],
   "internalslot-jitterbuffertarget": [
     {
       "description": "Add control for the receiver's jitter buffer",
--- a/amendments.json
+++ b/amendments.json
@@ -1029,12 +1029,12 @@
       "type": "addition",
       "status": "candidate",
       "tests": [
-	"webrtc/RTCRtpTransceiver-setCodecPreferences-direction.html"
+  "webrtc/protocol/h264-unidirectional-codec-offer.https.html"
       ],
       "testUpdates": [
-        "web-platform-tests/wpt#1234"
+        "web-platform-tests/wpt#49421"
       ],
-      "id": 50
+      "id": 51
     }
   ],
   "internalslot-jitterbuffertarget": [
--- a/webrtc.html
+++ b/webrtc.html
@@ -3729,14 +3729,11 @@ interface RTCPeerConnection : EventTarget  {
                         </li>
                         <li>
                           <p>
-                            The <i>codec preferences</i> of a [= media
-                            description =]'s [= associated =] transceiver,
-                            <var>transceiver</var>, is said to be the value of
-                            <var>transceiver</var>.{{RTCRtpTransceiver/[[PreferredCodecs]]}}
-                            with the following filtering applied (or said not
-                            to be set if
-                            <var>transceiver</var>.{{RTCRtpTransceiver/[[PreferredCodecs]]}}
-                            is empty):
+                            Let <var>filteredCodecs</var> be the result of
+                            applying the following filter on
+                            <var>transceiver</var>.{{RTCRtpTransceiver/[[PreferredCodecs]]}}.
+                            The filtering MUST NOT change the order of the codec
+                            preferences:
                           </p>
                           <ol>
                             <li>
@@ -3769,8 +3766,11 @@ interface RTCPeerConnection : EventTarget  {
                             </li>
                           </ol>
                           <p>
-                            The filtering MUST NOT change the order of the
-                            codec preferences.
+                            The <i>codec preferences</i> of a [= media
+                            description =]'s [= associated =] transceiver,
+                            <var>transceiver</var>, is said to be the value of
+                            <var>filteredCodecs</var> if non-empty and said to
+                            be unset otherwise.
                           </p>
                         </li>
                         <li>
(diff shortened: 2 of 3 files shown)


amendments.json
```diff
@@ -891,6 +891,21 @@
       "id": 42
     }
   ],
+  "setcodecpreferences-filter-direction": [
```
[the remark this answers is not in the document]
dontcallmedom: if this is reverting the changes in `setcodecpreferences-receive`, the relevant block should be removed from amendments.json

amendments.json
```diff
@@ -891,6 +891,21 @@
...
+    {
+      "description": "setCodecPreferences takes both send and receive codecs which get filtered at negotiation based on transceiver direction",
+      "pr": 3018,
+      "type": "addition",
```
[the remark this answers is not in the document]
dontcallmedom: I think it's an addition (and maybe the description should better reflect this, e.g. as "support for receive-only and send-only codecs"?)

amendments.json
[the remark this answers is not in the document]
henbos: OK I removed "setcodecpreferences-receive" from both JSON and the HTML files, replacing them by the new addition. The bots are still happy.

webrtc.html
```diff
@@ -11325,31 +11325,39 @@ <h2>
                 <dfn data-idl="">setCodecPreferences</dfn>
               </dt>
               <dd>
-		<div id="setcodecpreferences-receive">
+    <div id="setcodecpreferences-send-and-receive">
                 <p>
                   The {{setCodecPreferences}} method overrides the default
-                  receive codec preferences used by the <a>user agent</a>. When
+                  codec preferences used by the <a>user agent</a>. When
```
jan-ivar, line 11331: Can we add a note somewhere to explain how setCodecPreferences works differently in the offerer vs answerer role? 

For example search for "Looking at setCodecPreferences as a sender-control leads to surprises" in https://blog.mozilla.org/webrtc/cross-browser-support-for-choosing-webrtc-codecs/

webrtc.html
```diff
@@ -11325,31 +11325,39 @@ <h2>
...
                   specific codecs (including RTX/RED/FEC). It also allows an
-                  application to cause a remote peer to prefer the codec that
-                  appears first in the list for sending.
+                  application to affect the remote peer's default codecs;
+                  influencing both what they should send by default and the
+                  default codec order of subsequently generated remote SDP. But
+                  just like the local peer can override its defaults, so can the
+                  remote peer. The local peer MUST be prepared to receive any
+                  codec that has been negotiated, even ones not first in the
+                  list.
```
jan-ivar, line 11350: We should avoid competing normative language with JSEP. One way might be to put explanations in non-normative notes.

webrtc.html
henbos: Notes clarified, including that the answer defaults to offer order

webrtc.html
henbos: I changed these to notes and tried to be more clear

henbos (commented):
PTAL @jan-ivar I believe I addressed your comments

webrtc.html
```diff
@@ -3769,8 +3766,11 @@ <h2>
                             </li>
```
henbos, line 3766: Self-review: The current filtering says to exclude codecs not in the list of send or receive capabilities. It's rather binary, either the codec was included or it was excluded, which doesn't work well with profile level IDs... for example:
- I can send H265 profile-level-id=156
- I can receive H265 profile-level-id=180

The current filter would allow a sendonly transceiver to offer 156 and a recvonly transceiver to offer 180, but a sendrecv transceiver would say that 156 is not in the list of receive codecs and that 180 is not in the list of send codecs... in reality we can sendrecv 156 because 156 is a subset of 180.

But I think this should be addressed in a separate PR

henbos: Technically the spec doesn't say anything about profiles, so a spec-compliant implementation could list every possible profile-level-id from 1..180 as separate codec entries and this would work, but that is absurd :) we only list the highest supported level per direction

henbos: I filed #3020

henbos:
@jan-ivar Have I addressed your feedback? If you could take a look before the editor's meeting

webrtc.html
```diff
@@ -11325,31 +11325,48 @@ <h2>
...
+                  remote endpoint which codec the local endpoint prefers to
+                  receive. The answerer defaults to using the order in the SDP
+                  offer, but overrides this with their own codec preferences if
+                  they have any.
+                </p>
+                <p class="note">
+                  An RTCRtpSender defaults to sending what the remote endpoint
+                  indicated that it prefers to receive, but the application can
+                  change which codec to send amongst negotiated codecs in
+                  setParameters() at any time. An RTCRtpReceiver is prepared to
```
henbos, line 11358: Link to "codec" specifically

webrtc.html
```diff
@@ -11325,31 +11325,48 @@ <h2>
...
+                <p class="note">
+                  If the m= section is used for receiving, the order of the
+                  codecs in the SDP (both the offer and the answer) tells the
+                  remote endpoint which codec the local endpoint prefers to
+                  receive. The answerer defaults to using the order in the SDP
+                  offer, but overrides this with their own codec preferences if
+                  they have any.
+                </p>
+                <p class="note">
+                  An RTCRtpSender defaults to sending what the remote endpoint
```
henbos, line 11355: Use curly bracket links here and elsewhere

webrtc.html
```diff
@@ -11325,31 +11325,48 @@ <h2>
                 <dfn data-idl="">setCodecPreferences</dfn>
               </dt>
               <dd>
-		<div id="setcodecpreferences-receive">
+    <div id="setcodecpreferences-send-and-receive">
                 <p>
                   The {{setCodecPreferences}} method overrides the default
-                  receive codec preferences used by the <a>user agent</a>. When
+                  codec preferences used by the <a>user agent</a>. When
```
henbos, line 11331: user agent "...as input to negotiation"

webrtc.html
```diff
@@ -11325,31 +11325,48 @@ <h2>
...
-                  application to cause a remote peer to prefer the codec that
-                  appears first in the list for sending.
+                  specific codecs (including RTX/RED/FEC) by listing all codecs
+                  except for the ones to disable.
+                </p>
+                <p class="note">
+                  If the m= section is used for receiving, the order of the
+                  codecs in the SDP (both the offer and the answer) tells the
+                  remote endpoint which codec the local endpoint prefers to
+                  receive. The answerer defaults to using the order in the SDP
```
henbos, line 11350: Even if the m= section is not used for receiving,

jan-ivar (approved):
LGTM. I might add another note to cover the surprising case if the answerer receives a recvonly offer: in this case it's too late for the answerer to use setCodecPreferences to set what to send, and they need to use setParameters instead.

dontcallmedom (changes requested):
quoting @henbos
> Add real test before landing!
(marking this as requesting changes to make sure this doesn't get overlooked)

amendments.json
```diff
@@ -876,19 +876,19 @@
...
       "status": "candidate",
       "tests": [
-	"webrtc/RTCRtpTransceiver-setCodecPreferences.html"
+	"webrtc/RTCRtpTransceiver-setCodecPreferences-direction.html"
       ],
       "testUpdates": [
-        "web-platform-tests/wpt#44318"
+        "web-platform-tests/wpt#1234"
```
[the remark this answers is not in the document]
henbos: Chromium issue for adding WPTs: https://crbug.com/379551577

webrtc.html
```diff
@@ -3767,8 +3764,11 @@ <h2>
                             </li>
                           </ol>
                           <p>
-                            The filtering MUST NOT change the order of the
-                            codec preferences.
+                            The <i>codec preferences</i> of a [= media
+                            description =]'s [= associated =] transceiver,
+                            <var>transceiver</var>, is said to be the value of
+                            <var>filteredCodecs</var> if non-empty and said to
+                            be unset otherwise.
```
henbos, line 3771: Note to self there is also a reference to preferred codecs internal slots in setParameters that should be filtered on sending

amendments.json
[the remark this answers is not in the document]
henbos: Done

henbos:
@dontcallmedom PTAL can you please approve (remove "Changes requested") now that test coverage has been added? See https://github.com/web-platform-tests/wpt/pull/49421/files

dontcallmedom (approved):

PR #3023, opened 2024-11-21
Allow level-id to be different in codec match

### LINKED ISSUE #3020: Codec negotiation and level IDs
If you are capable of level-id=X then you are also capable of level-id=Y where Y<X, assuming mimeType, profile-id, tier-flag and tx-mode is the same. I.e.
- If I can send H265 level-id=180 and receive level-id=156 that means I can sendrecv 156.
- If I can receive H265 level-id=156 and send level-id=180 that means I can sendrecv 156.

Today's filter seems to assume that every unique combination of codec, profile and level-id is included in the list:
> If transceiver.[direction](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-direction) is "[sendonly](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiverdirection-sendonly)" or "[sendrecv](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiverdirection-sendrecv)", exclude any codecs not included in the [list of implemented send codecs](https://w3c.github.io/webrtc-pc/#dfn-list-of-implemented-send-codecs) for kind.
>
> If transceiver.[direction](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-direction) is "[recvonly](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiverdirection-recvonly)" or "[sendrecv](https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiverdirection-sendrecv)", exclude any codecs not included in the [list of implemented receive codecs](https://w3c.github.io/webrtc-pc/#dfn-list-of-implemented-receive-codecs) for kind.

But in reality, user agents only list the highest level-id they support. Such as 186, rather than all possible values between 0 and 186 as separate codec values. This means you can offer one level-ID from your getCapabilities but this gets downgraded to a level-ID that is not in your getCapabilities with interesting consequences:
- What got negotiated (`getParameters().codecs`) is different from what was preferred (`setCodecPreferences`) or what was in `getCapabilities`.
- If sender capability is 186, but 180 was negotiated, can I still `setParameters` with 186 or do I have to change to 180?

henbos:
Fixes #3020

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/3023.html" title="Last updated on Jan 9, 2025, 11:07 AM UTC (fa63f7e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/3023/70955a8...henbos:fa63f7e.html" title="Last updated on Jan 9, 2025, 11:07 AM UTC (fa63f7e)">Diff</a>


henbos:
@aboba Does this wording make sense to you?

webrtc.html
taste1981, line 1: Comment on the text before the change here:

In RFCs we typically have inferred values for FMTP line if not specified. 

[RFC 7798](https://datatracker.ietf.org/doc/html/rfc7798)For H.265 profile-id inferred to 1, tier-flag inferred to 0, level-id inferred to 93, tx-mode inferred to SRST.
[AV1 RTP spec](https://aomediacodec.github.io/av1-rtp-spec/#72-sdp-parameters): profile inferred to 0, level-idx inferred to 5, tier inferred to 0.

I feel the comparison of sdpFmtpLine should be based on the actual inferred or explicitly signaled parameters. 
If one has FMTP line and another not, we consider them match if they are actually the same parameters though one does not explicitly signal them?

henbos: A bit of a separate issue but yes I agree, if "foo=0" can be inferred from a line that doesn't contain it, then it should also be equal to a line that does explicitly contain it. Likewise, I would imagine that the order of parameters does not matter here? E.g. "foo=0;bar=1" is the same as "bar=1;foo=0"?

taste1981: > A bit of a separate issue but yes I agree, if "foo=0" can be inferred from a line that doesn't contain it, then it should also be equal to a line that does explicitly contain it. Likewise, I would imagine that the order of parameters does not matter here? E.g. "foo=0;bar=1" is the same as "bar=1;foo=0"?

Correct. Order does not matter.

henbos: OK I added a sentence to clarify this while we're at it

webrtc.html
```diff
@@ -11467,21 +11468,26 @@ <h2>
...
                     <p>
                       If either (but not both) of <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}}
                       and <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}} are [= map/exist | missing =],
                       or if they both [=map/exist=] and <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}}
                       is different from <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}}, return
-                      <code>false</code>.
+                      <code>false</code>. Two FMTP lines are the same as long as they convey the same
+                      information, including omitting parameters with well defined defaults or using
+                      different order. In this step, the user agent MUST consider two FMTP lines
+                      identical if the "media format" are the same but the highest complying bitstream
```
alvestrand, line 11480:
```suggestion
                      identical if the "media format" is the same but the highest complying bitstream
```

fippo:
if you want to go this route you will also need to take care of [RFC6184](https://www.rfc-editor.org/rfc/rfc6184#section-8.1)s `level-asymmetry-allowed`, no? All of this is reasonably well-defined in IETF specs thougdh (well, not quite for HEVC) already

webrtc.html
henbos: Done + reworded

henbos:
> if you want to go this route you will also need to take care of [RFC6184](https://www.rfc-editor.org/rfc/rfc6184#section-8.1)s level-asymmetry-allowed, no?

No this is just about the codec matching algorithm used to sanity check input to setParameters/setCodecPreferences, what is actually negotiated based on this is defined somewhere else.

As for sending the selected codec, the description is kind of vague, [RTCRtpEncodingParmaters.codec](https://w3c.github.io/webrtc-pc/#dom-rtcrtpencodingparameters-codec) just says:
> Optional value selecting which codec is used for this encoding's RTP stream. If absent, the user agent can chose to use any negotiated codec.

And [setParameters](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-setparameters) just says:
> [In parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel), configure the media stack to use parameters to transmit sender.[[[SenderTrack]]](https://w3c.github.io/webrtc-pc/#dfn-sendertrack).

So it's just "codec used for selecting what to send", we might want to create an issue about clarifying not to send greater than the level-id that was negotiated, but that seems more like an editorial note.

henbos:
> not to send greater than the level-id

We might want to update getParameters().encodings[i].codec to match the level-id that you find in getParamters().codecs though? Separate issue though because that is the same as what happens with setCodecPreferences() today

webrtc.html
aboba: The text is ambiguous because not all fmtp parameters are included in the "media format" for matching purposes. As an example, the statement applies to H.265 but not H.264 where `profile-level-id` combines `profile` with `level-id`.

webrtc.html
```diff
@@ -11467,21 +11468,26 @@ <h2>
...
                   <li>
+                    <div id="codec-match-without-level-id">
                     <p>
                       If either (but not both) of <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}}
                       and <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}} are [= map/exist | missing =],
                       or if they both [=map/exist=] and <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}}
                       is different from <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}}, return
-                      <code>false</code>.
+                      <code>false</code>. Two FMTP lines are the same as long as they convey the same
+                      information, including omitting parameters with well defined defaults or using
```
henbos, line 11478: Change "information" to "media format"

amendments.json
```diff
@@ -1242,5 +1242,16 @@
...
+  ],
+  "codec-match-without-level-id": [
+    {
+      "description": "Two codecs are considered the same even if level-id is not",
+      "type": "correction",
+      "status": "candidate",
+      "difftype": "modify",
+      "id": 51,
+      "pr": 3023,
+      "testUpdates": "not-testable"
```
henbos, line 1254: Remote SDP is testable.

henbos:
Codec wizard @aboba, can you do another review?

henbos:
Btw I had said during the meeting that I thought all instances of "codec matching" could ignore level, but upon a second look I realized that we don't need to be that "loose" in most cases. For example if you addTransceiver, the only valid input is from getCapabilities since there is no chance that anything else has been negotiated for that transceiver yet. Ultimately we only need to "ignoreLevels=true" at a couple of places:
- When negotiation completes, the check to see if the codec was negotiated away and for maybe clearing setParameters's codec value, this should not remove the codec because of a level mismatch.
- When setParameters is called, this should not throw due to level mismatch (= "Proposal A").
- In addition I clarified that the "RTCRtpEncodingParameters.codec" picks codec based on "ignoreLevels=true" for the same reason that setParameters is supposed to work here, but this was never called out before.

In all other cases we can "ignoreLevels=false", which the PR reflects now.

webrtc.html
```diff
@@ -11470,18 +11482,57 @@ <h2>
                     <p>
                       If either (but not both) of <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}}
                       and <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}} are [= map/exist | missing =],
-                      or if they both [=map/exist=] and <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}}
-                      is different from <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}}, return
-                      <code>false</code>.
+                      return <code>false</code>.
```
aboba, line 11485: RTP payload specs can specify default values of fmtp parameters. So it is possible for codecs to match even if a peer's sdpFmtpLine is missing. Example: In H.265, default level-id is 3.1, default profile is Main.  Are default values included in the algorthym?

webrtc.html
```diff
@@ -11470,18 +11482,57 @@ <h2>
...
+                      <li>
+                        <div id="codec-match-without-level-id">
+                        <p>
+                          If <var>ignoreLevels</var> is <code>false</code> and the highest complying bitstream
+                          levels inferred from <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}} and
+                          <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}} are different, return <code>false</code>.
+                        </p>
+                        <p class="note">
+                          Even if <var>ignoreLevels</var> is <code>true</code>, some codecs (such as H.264) have
+                          bistream levels that are not distinct from their media formats and as such cannot be
```
aboba, line 11523: Maybe "have bitstream levels that are not distinct from their media formats and as such" -> "include levels in the media format, so that the level"

aboba (changes requested):

webrtc.html
fippo: FWIW: some implementations backfill these values and simplify the life of the JS layer that way

henbos: I have a note that says "In some cases a parameter can be omitted and still be inferred", that should cover default values

webrtc.html
```diff
@@ -11497,18 +11509,56 @@ <h2>
...
-                      or if they both [=map/exist=] and <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}}
-                      is different from <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}}, return
-                      <code>false</code>.
+                      return <code>false</code>.
                     </p>
                   </li>
+                  <li>
+                    <p>
+                      If both <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}} and
+                      <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}} [=map/exist=], run the following
```
alvestrand, line 11518:
```suggestion
                      <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}} [=map/exist=], and are in key/value format, run the following
```

alvestrand (approved):

webrtc.html
```diff
@@ -11466,10 +11475,13 @@ <h2>
                 </ol>
 		</div>
                 <div id="codecmatch-algorithm">
-                <p>  The <dfn class="export">codec dictionary match</dfn> algorithm
-                  given two {{RTCRtpCodec}} dictionaries
-                  <var>first</var> and <var>second</var> is as follows:
+                <p>
+                  The <dfn class="export">codec dictionary match</dfn> algorithm given two
+                  {{RTCRtpCodec}} dictionaries <var>first</var> and <var>second</var>, and
+                  an <var>ignoreLevels</var> boolean defaulting to <code>false</code> if not
```
henbos, line 11481: Self-review: Also need to ignoreLevel=true in the createOffer/createAnswer filter or else a "sendrecv" transceiver will filter out both level IDs when sender and receiver caps have different level IDs!

aboba (approved):

webrtc.html
```diff
@@ -11497,18 +11517,56 @@ <h2>
...
+                          Even if <var>ignoreLevels</var> is <code>true</code>, some codecs (such as H.264) include
+                          levels in the media format, so that the level cannot be ignored in this algorithm.
+                        </p>
+                        </div>
+                      </li>
+                    </ol>
+                  </li>
                   <li>
                     <p>
                       Return <code>true</code>.
```
jan-ivar, line 11566: If one of the inputs is garbage the old algorithm returned false. Doesn't this risk returning true if the line is not parseable?

jan-ivar (approved):

webrtc.html
```diff
@@ -11497,18 +11517,56 @@ <h2>
...
+                      <li>
+                        <div id="codec-match-without-level-id">
+                        <p>
+                          If <var>ignoreLevels</var> is <code>false</code> and the highest complying bitstream
+                          levels inferred from <var>first</var>.{{RTCRtpCodec/sdpFmtpLine}} and
+                          <var>second</var>.{{RTCRtpCodec/sdpFmtpLine}} are different, return <code>false</code>.
+                        </p>
+                        <p class="note">
+                          Even if <var>ignoreLevels</var> is <code>true</code>, some codecs (such as H.264) include
+                          levels in the media format, so that the level cannot be ignored in this algorithm.
```
henbos, line 11558: More accurately "..., so that ignoring the level requires codec-specific parsing" to reflect level ID in H.264 being something that can be parsed out in some cases

henbos:
Merging based on Editors can integrate

henbos:
Removing needs test label since tests now exist

Issue #3010, opened 2024-10-22
Opus mono/stereo and remote track channelCount


henbos:
Opus is a codec that is capable of both mono and stereo.

In SDP, you can specify "stereo=X" to say what your receiver preference is, but importantly:
- This is just a preference, regardless what the value of X is and regardless if "stereo=" is even present in the SDP, the decoder has to be prepared for both mono and stereo input on the wire. It's completely optional in every way.

Today in Chrome, you get mono (track.getSettings().channelCount == 1) by default, but if "stereo=1" is used you get stereo (channelCount == 2). Importantly, the decoder will downmix or upmix to match the desired output (e.g. if stereo is used you will get a stereo track even if the input on the wire is mono).

We would like to support stereo without SDP munging. The question is what is the expected behavior?
1. Opus is a stereo codec, so always output stereo regardless of signal on the wire.
2. Adjust the output of the decoder to match what is on the wire (mono = mono, stereo = stereo). This could change on the fly.

While I originally thought 2) was more intuitive, I'm told that Firefox does 1). It also turns out that the audio team prefers to do 1) to avoid some complexity and to reduce the risk of audio glitches due to reconfiguring the decoder in the event that the signal switches back and forth on the fly (use case: virtual audio ssrc and speaker switching between mono and stereo signals).

**Proposal A**: Mandate "always stereo" for stereo codecs.
**Proposal B**: Mandate dynamically switching between mono and stereo.
**Proposal C**: Mandate nothing; it is up to the user agent.

henbos:
@jan-ivar @youennf @alessiob

fippo:
see also https://github.com/w3c/webrtc-extensions/issues/63

henbos:
Ah that old issue... it does seem to be in favor of defaulting to stereo=1 but allowing mono via channelCount:1 constraints.

However it does not capture the nuance between Proposal A or Proposal B here. For example, even if we _encode_ mono or stereo depending on the local track's channelCount, there's a separate question regarding if when we _decode_ this signal if the remote track's channelCount would be 1 or 2? According to the audio experts, there's "no harm" in having channelCount be 2 on the remote track regardless if the signal being decoded was mono or stereo

henbos:
Action items:
- @jan-ivar to check if Firefox has mono or stereo remote track in the event that the signal on the wire is mono.
- Overall: stereo should work without SDP munging :)
- @henbos to file an issue about exposing channelCount

henbos:
[...]

PR #3025, opened 2024-11-26
Add remote track channelCount

### LINKED ISSUE #3011: Remote track.getSettings().channelCount
If this is 2 for stereo and 1 for mono codecs then we can test support for opus stereo.
- remoteTrack.getSettings().channelCount is currently implemented in Chrome, this was implemented before we added [this table](https://w3c.github.io/webrtc-pc/#mediatracksupportedconstraints-mediatrackcapabilities-mediatrackconstraints-and-mediatracksettings).

Related to #3010 where we also discuss if a stereo decoder would need to dynamically switch channelCount on the fly based on signal on the wire, or if it is OK for a stereo codec to always output a stereo track.

henbos:
Fixes #3011

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-pc/pull/3025.html" title="Last updated on Mar 28, 2025, 1:40 PM UTC (953516d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/3025/fe6fcc3...henbos:953516d.html" title="Last updated on Mar 28, 2025, 1:40 PM UTC (953516d)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -16743,8 +16743,8 @@ async function gatherStats(pc) {
                   As a setting, this is the number of {{RTCRtpCodec/channels}}
                   that the decoder produced when the last audio frame was
                   decoded. The number of channels that the decoder produces MUST
-                  reflect the number of channels configured in response to SDP
-                  negotiation. Before any frames have been decoded, this setting
+                  reflect the number of channels configured when the local SDP
+                  was set. Before any frames have been decoded, this setting
                   MUST NOT [=map/exist=].
                   <p class="note">
                     For example, the opus audio codec has a "stereo" flag. If


amendments.json
```diff
@@ -1242,5 +1242,21 @@
   "web-platform-tests/wpt#48361"
       ]
     }
+  ],
+  "remote-track-settings-channel-count": [
+    {
+      "description": "Use Performance.timeOrigin + Performance.now() for stats timestamps",
```
alvestrand, line 1248: This description seems a bit off?

webrtc.html
```diff
@@ -16583,6 +16583,21 @@ <h4>
                   double rounded to the tenth decimal place.
                 </td>
               </tr>
+              <div id="remote-track-settings-channel-count">
+              <tr id="def-constraint-channelCount">
+                <th data-tests="">
```
alvestrand, line 16588: should data-tests refer to the same test as the amendments update?

amendments.json
henbos: Oops

alvestrand (approved):

henbos:
(Jan-Ivar's fiddle from meeting: https://jsfiddle.net/jib1/3Lh9zaxq/)

webrtc.html
```diff
@@ -16610,6 +16610,21 @@ <h4>
...
+                <th data-tests="RTCRtpEncodingParameters-codec-opus-stereo.https.html">
+                  channelCount
+                </th>
+                <td>
+                  {{ConstrainULong}}
+                </td>
+                <td>
+                  As a setting, this is the number of channels of the latest
+                  audio frame received, reflecting the number of
+                  {{RTCRtpCodec/channels}} that the decoder outputted.
```
henbos, line 16624: Comment from Youenn: make this more clear that you MUST do what the decoder is capable of

jan-ivar (approved):

base-rec.html
```diff
@@ -15949,6 +15949,7 @@ <h4 id="x9-3-1-mediatracksupportedconstraints-mediatrackcapabilities-mediatrackc
                   double rounded to the tenth decimal place.
                 </td>
               </tr>
+              <div id="remote-track-settings-channel-count"><!-- ammendment --></div>
```
dontcallmedom, line 15952:
```suggestion
 ```
not needed since it's an `append` amendment

File: webrtc.html [hunk 5]
```diff
@@ -16677,6 +16677,21 @@ <h4>
                   double rounded to the tenth decimal place.
                 </td>
               </tr>
+              <div id="remote-track-settings-channel-count">
+              <tr id="def-constraint-channelCount">
```
dontcallmedom, line 16681:
```suggestion
              <tr id="def-constraint-channelCount">
```

webrtc.html
```diff
@@ -16677,6 +16677,21 @@ <h4>
...
+                <td>
+                  {{ConstrainULong}}
+                </td>
+                <td>
+                  As a setting, this is the number of channels of the latest
+                  audio frame received, reflecting the number of
+                  {{RTCRtpCodec/channels}} that the decoder outputted.
+                </td>
+              </tr>
+              </div>
```
dontcallmedom, line 16694:
```suggestion
              </div>
```

amendments.json
```diff
@@ -1243,6 +1243,21 @@
       ]
     }
   ],
+  "remote-track-settings-channel-count": [
```
dontcallmedom, line 1246:
```suggestion
  "def-constraint-channelCount": [
```

dontcallmedom (commented):
fix and simplify amendment description

amendments.json
```diff
@@ -1243,6 +1243,22 @@
...
+      "id": 53,
+      "pr": 3025,
+      "tests": [
+  "webrtc/RTCRtpEncodingParameters-codec-opus-stereo.https.html"
+      ],
+      "testUpdates": [
+  "web-platform-tests/wpt#48599",
+  "web-platform-tests/wpt#49392"
+    ]
+  },
```
dontcallmedom, line 1261:
```suggestion
    }
  ],
```

dontcallmedom:
the CI error is something that needs to be fixed in the amendment script AFAICT, will take a look

webrtc.html
henbos: Also to [refresh our memories](https://github.com/w3c/webrtc-pc/issues/3010#issuecomment-2487877298) about the conclusion: we have go-ahead to return 2 here and as per discussion, saying MUST here was preferred.

I updated this sentence to say "A stereo capable decoder MUST produce stereo and return the value 2.

henbos: [The WPT](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/webrtc/RTCRtpEncodingParameters-codec-opus-stereo.https.html) also reflects the "always do stereo on the receiving track" decision.

webrtc.html
```diff
@@ -16690,6 +16690,20 @@ <h4>
...
+                  channelCount
+                </th>
+                <td>
+                  {{ConstrainULong}}
+                </td>
+                <td>
+                  As a setting, this is the number of channels of the latest
+                  audio frame received, reflecting the number of
+                  {{RTCRtpCodec/channels}} that the decoder outputted. A stereo
+                  capable decoder MUST produce stereo and return the value 2.
```
Pehrsons, line 16704: Define "stereo capable" ?
For opus I interpret the intention to be what [RFC 6716 2.1.2](https://datatracker.ietf.org/doc/html/rfc6716#section-2.1.2) refers to as "stereo decoder". If so this seems fine, since it reflects the data represented by the track. I'm not sure that's obvious from the proposed text though. The bit with "that the decoder outputted" seems clearer, but OTOH seems to contradict "latest audio frame received" in case it doesn't match the decoder configuration.

henbos: The intent is that is the browser has implemented opus stereo, then the decoder will output stereo - regardless if the signal on the wire is mono or stereo. You could in theory flip-flop between mono and stereo here, but our audio folks did not want to do this because they didn't see the use case for it and they were concerned that re-initializing the opus decoder in the middle of the stream (the use case is CSRC where the stream can switch between mono and stereo at any time and multiple times throughout the meeting).

Is this OK with you? I can clarify the text to make it more clear

henbos: I.e. I would only expose change from 2 to 1 or vice versa if you switch which codec you use, but as long as you use opus you should always get 2

Pehrsons: From what I can tell of [the libwebrtc opus decoder](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/api/audio_codecs/opus/audio_decoder_opus.cc;l=60,95) (referenced above in https://github.com/w3c/webrtc-pc/pull/3025#discussion_r2016844668) it outputs what was signaled. Not sure what you mean by "implements" above but it doesn't sound like you mean "signaled" or "configured"? Are you proposing to return 2 even when negotiating [`stereo=0`](https://datatracker.ietf.org/doc/html/rfc7587#section-6.1)?

henbos: I think we should say something along the lines that the channelCount reflects the number of channels that the codec used to decode the last audio frame was configured to use according to what was negotiated, with a note to clarify what stereo=0 and stereo=1 means for opus as an example

Pehrsons: Looking a bit more at [RFC7587](https://datatracker.ietf.org/doc/html/rfc7587#section-6.1), `stereo` is a receive-only param so not something remote peers can influence. It seems fair to note though, that a client could choose to use a mono decoder when the remote peer signals `sprop-stereo=0`.

Not sure if signaling should be mentioned. There seems to be a number of decisions left to the implementation here. Something about what was output by the decoder seems fine to me.

henbos: > Looking a bit more at [RFC7587](https://datatracker.ietf.org/doc/html/rfc7587#section-6.1), stereo is a receive-only param so not something remote peers can influence.

Yes, opus can handle both mono and stereo bitstreams, so the "stereo" flag is nothing more than a hint, kind of like saying "if you send me stereo you're wasting some bits because I'm going to downsample it to mono anyway". So in a way it's an optional parameter, but our implementation - on the decode side - appears to respect it. The sender could still do something different though...

> Not sure if signaling should be mentioned. There seems to be a number of decisions left to the implementation here. Something about what was output by the decoder seems fine to me.

Ah too late, can you take a look anyway @Pehrsons ? I just updated the PR to say that it is the number of channels output by the decoder, but then I went one step further and added that this MUST reflect what was configured in response to SDP negotiation. Plus a note about opus.

I believe and hope that this description matches all implementations, and if it doesn't we could revisit it in the future

henbos: An earlier comment by @youennf was that we should add MUSTs if possible. We could downgrade to SHOULD but that is not testable

Pehrsons: I'm not sure about "negotiated". Since the stereo param is receive-only it makes sense to me if the setting matches the stereo param in the local description. Other people know signaling better than me, to say whether those are equivalent for this case.

Otherwise LGTM

henbos: OK - changed from "negotiated" to "when the local SDP was set", which is more precise.

@jan-ivar @Pehrsons I think this fully captures what we want now and includes that "MUST" language. Can I have the "Editors can integrate" back? :)

jan-ivar: After disscussions with @padenot, we've found value in returning channelCount 1 when a stereo capable decoder is receiving mono packets. It's useful because down stream websites can cut processing in half.

I'll comment on the issue in a bit (after updating a fiddle to show this)

henbos: I understand if you want to get it right on the first PR, but considering I'll be OOO for months and this PR currently reflects how [libwebrtc-based opus decoders already work today](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/api/audio_codecs/opus/audio_decoder_opus.cc;l=60,95), meaning all browsers that change the default to stereo=1 and exposed channelCount would pass [these WPTs](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/webrtc/RTCRtpEncodingParameters-codec-opus-stereo.https.html), can we leave "channelCount:1 on a stereo=1 decoder" as a separate issue and pull request?

We can already claim consensus on everyone wanting stereo opus by default so it would be good to make progress even if we don't have consensus on "mono on stereo opus" which is a new feature request. @jakobivarsson

webrtc.html
```diff
@@ -16732,6 +16732,29 @@ <h4>
...
+                <td>
+                  {{ConstrainULong}}
+                </td>
+                <td>
+                  As a setting, this is the number of {{RTCRtpCodec/channels}}
+                  that the decoder produced when the last audio frame was
+                  decoded. The number of channels that the decoder produces MUST
+                  reflect the number of channels configured when the local SDP
+                  was set. Before any frames have been decoded, this setting
+                  MUST NOT [=map/exist=].
```
jan-ivar, line 16748: I don't believe the constrainable pattern allows absence. Is there a precedent in the spec for a setting to not exist and come into being later? I don't think we should allow that. I've filed https://github.com/w3c/webrtc-pc/issues/3045

```suggestion
                  was set.
```

jan-ivar (changes requested):
Unfortunately, I'm recording disagreement on the proposed behavior here, so I think we need to resolve this before we can merge this PR and tests.

Sorry about your timeframe constraints, but I'm confident we can make progress on this in a much shorter timeframe.

I'm happy to commandeer the PR to help assist with that, or feel free to nominate someone else of your choosing.

webrtc.html
henbos: > Is there a precedent in the spec for a setting to not exist and come into being later?

Yes this is how remote track settings already work in the video frames (width, height and frameRate) so I thought we should be consistent with audio frames.

henbos:
”Mono on stereo=1 decoder” is behavior I proposed to my colleagues already. My arguments in favor of this were basically:
1. It makes sense.
2. It makes end-to-end testing easier.

You brought up a point a about down stream processing being cut in half that I believe was not discussed at the time. Still, the current state of the PR represents what I was able to get agreement on, which was to exclude this particular use case and focus on what was already implemented (stereo=1 being always stereo) for now.

That’s why I thought it might be easier to do it in two steps since the first step (specifically talking about ”we want stereo” and ”we want channelCount) has agreement and is already implemented in libwebrtc, while the second one (mono over stereo=1) isn’t implemented by anyone yet (if I understand correctly) and may need further discussions with my colleagues.

But if your preference is to do both steps in the same PR I am happy to appoint it to you.

henbos:
The way to decouple step 1 from step 2 would be to delete the sentence "The number of channels that the decoder produces MUST reflect the number of channels configured when the local SDP was set." and to delete the NOTE. Then it would be up to the implementation. But we can also wait until we know what the plan is.

PR #3050, opened 2025-05-16
Validate TURN username and credentials

### LINKED ISSUE #3049: Validate an ICE server url is missing length check for username
The "validate an ICE server URL" steps in
https://w3c.github.io/webrtc-pc/#set-the-configuration
do not specify a length check on the username for TURN urls.

RFC 8489 limits the username length to 509 bytes (down from 512 bytes in RFC 5389).
This has led to diverging behavior which is [trivally observable](https://jsfiddle.net/fippo/7xot4avw/7/)
* Firefox throws an InvalidAccessError but never updated for RFC 8489 so throws at 512 characters (and has a typo in the error message).
* Safari throws at 509 characters but it throws a TypeError.
* Chromium silently ignores usernames longer than 509 characters since this caused the ["NAT SlipStream" attack in 2020](https://datatracker.ietf.org/meeting/110/materials/slides-110-rtcweb-nat-slipstream-00) (thanks @alvestrand!) and never sends the STUN packets.

There is [no WPT covering this](https://wpt.fyi/results/webrtc/RTCConfiguration-iceServers.html).

Technically Chrome is following the spec and might have frustrated attackers.
But it seems easier to align on helping developers and add the following bullet point:
* If parsedURL's' scheme is "turn" or "turns", and the UTF-8 representation of server.username exceeds the 509 byte limit defined by RFC 8489 section, then throw an InvalidAccessError.

It is unclear to me why Safari attempts to restrict the credential in the same way [here](https://github.com/WebKit/WebKit/blob/e72e89897081efd608c057fe8c33f468a10bda4b/Source/WebCore/Modules/mediastream/RTCPeerConnection.cpp#L551). The credential is never transmitted over the network in clear but only input to the TURN authentication hash.

Semi-related, why is "do not exist" used in step 7 and therefore forbids rejecting an empty username?
Apparently 3/4 implementations [do not agree either](https://wpt.fyi/results/webrtc/RTCConfiguration-iceServers.html?label=experimental&label=master&aligned). What is the TURN client supposed to put into the username field? Firefox actually puts an empty string there which is unlikely to be useful and fails to proceed with an empty password.

jan-ivar:
Fixes #3049 (for discussion)

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -3154,8 +3154,10 @@
                   If <var>parsedURL</var>'s' [=url/scheme=] is <code>"turn"</code> or
                   <code>"turns"</code>, and either of
                   <var>server</var>.{{RTCIceServer/username}} or
-                  <var>server</var>.{{RTCIceServer/credential}} do
-                  [=map/exist|not exist=], then [= exception/throw =] an
+                  <var>server</var>.{{RTCIceServer/credential}} are
+                  [=map/exist|missing=] or their UTF-8 representations
+                  fail to conform to [[RFC8489]] section 14.3 and [[RFC8265]]
+                  section 4.1 respectively, then [= exception/throw =] an
                   {{InvalidAccessError}}.
                 </p>
               </li>


youennf (approved):

ss2J2 (approved):

jan-ivar:
@dontcallmedom note https://github.com/w3c/webrtc-pc/pull/3050/changes/385df716bd25f0a1c0048b2105cc3e9ad8f90491 was needed for amendments parser to recognize there's no append/modify conflict.

jan-ivar:
Test forthcoming in https://phabricator.services.mozilla.com/D287053

PR #3121, opened 2026-06-30
Expand definition of "administratively prohibited."

VergeA:
I am working on the Connection Allowlists specification being incubated in WICG: https://github.com/WICG/connection-allowlists

This feature will allow sites to restrict network access to particular allowlisted endpoints. WebRTC is handled by providing a boolean flag to either fully enable or fully disable RTCPeerConnection usage. The corresponding spec PR monkey-patches the RTCPeerConnection constructor to throw an error when disabling: https://github.com/WICG/connection-allowlists/pull/13

This mirrors the concept of "administratively prohibited", but applies to the entire PC interface instead of individual ICE candidates. This PR defines a new subsection containing the definition of "administratively prohibited", and has the RTCPeerConnection constructor and addIceCandidate refer to this common definition.

***
<a href="https://pr-preview.s3.amazonaws.com/VergeA/webrtc-pc/pull/3121.html" title="Last updated on Jun 30, 2026, 5:31 PM UTC (7f90593)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/3121/42d503e...VergeA:7f90593.html" title="Last updated on Jun 30, 2026, 5:31 PM UTC (7f90593)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -1199,6 +1199,16 @@
               </p>
             </li>
           </ol>
+		  <section>
+		  	<h4>
+			  User Agent Restrictions
+		    </h4>
+			<p>
+			  The user agent may disallow some or all behavior of {{RTCPeerConnection}},
+			  including disabling use of the interface entirely. Such behavior is defined
+			  as <dfn>administratively prohibited</dfn>.
+		    </p>
+		  </section>
           <section>
             <h4>
               Constructor
@@ -1215,6 +1225,12 @@
                   specified here, [= exception/throw =] an {{UnknownError}}
                   with the {{DOMException/message}} attribute set to an
                   appropriate description.
+                </p>
+              </li>
+			  <li class="no-test-needed">
+                <p>
+                  If the use of {{RTCPeerConnection}} is [= administratively
+				  prohibited =], throw a {{NotAllowedError}}.
                 </p>
               </li>
               <li class="no-test-needed">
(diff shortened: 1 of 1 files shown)


webrtc.html
```diff
@@ -4691,7 +4707,7 @@ <h2>
                     </li>
                   </ol>
                   <p>
-                    A candidate is <dfn>administratively prohibited</dfn> if
+                    A candidate is [= administratively prohibited =] if
```
domfarolino, line 4710: Given the next of this paragraph, it sounds like this term is being defined twice, which is unclear.

VergeA: I can restructure to clarify we're referencing an existing definition. Maybe something like this?
 
"The user agent can decide not to allow connection attempts to a candidate by declaring that candidate to be [= administratively prohibited =]."

webrtc.html
```diff
@@ -1216,6 +1226,12 @@ <h4>
                   with the {{DOMException/message}} attribute set to an
                   appropriate description.
                 </p>
+              </li>
+			  <li class="no-test-needed">
+                <p>
+                  If the use of {{RTCPeerConnection}} is [= administratively
+				  prohibited =], throw a {{NotAllowedError}}.
```
jan-ivar, line 1233: I don't think we should add this. See my comment in https://github.com/WICG/connection-allowlists/pull/13#discussion_r3514681741.

jan-ivar (changes requested):

webrtc.html
VergeA: Ack, thanks! Will abandon this PR and structure the Connection Allowlist integration to work more like CSP.

PR #3135, opened 2026-08-26
Clarify when events are fired on SCTP & DTLS transports

### LINKED ISSUE #3126: RTCSctpTransport-events.html expects RTCSctpTransport.onstatechange when the PC is closed
```
  const pc1ClosedWaiter = waitForState(pc1.sctp, 'closed');
  const pc2ClosedWaiter = waitForState(pc2.sctp, 'closed');
  pc1.close();
  await pc1ClosedWaiter;
```

Spec says nothing about firing an event for this state transition during the closing procedure for RTCPeerConnection, and the general pattern is not to fire events due to close():

https://w3c.github.io/webrtc-pc/#ref-for-dfn-sctptransportstate-1

This test-case is in Interop 2026.

jan-ivar:
Replaces normative steps inside enum state descriptions about when events are expected to be fired, with explicit algorithms that fire events called from the appropriate place.

Fixes #3126. 

This is meant to be behavior neutral. If it is not, don't merge.

@docfaraday PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/3135.html" title="Last updated on Aug 26, 2026, 6:55 PM UTC (950cab0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/3135/42d503e...jan-ivar:950cab0.html" title="Last updated on Aug 26, 2026, 6:55 PM UTC (950cab0)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2052,15 +2052,21 @@
                           {{RTCSctpTransportState/"connecting"}}.
                         </p>
                       </li>
-                      <li>
+                      <li data-tests="">
                         <p>
                           If <var>description</var> is of type
                           {{RTCSdpType/"answer"}}, and it initiates the closure
                           of an existing SCTP association, as defined in
-                          [[RFC8841]], Sections 10.3 and 10.4, set
+                          [[RFC8841]], Sections 10.3 and 10.4,
+                          [= announce the rtcsctptransport as closed | announce =]
+                          <var>connection</var>.{{RTCPeerConnection/[[SctpTransport]]}}
+                          as closed, and set
                           <var>connection</var>.{{RTCPeerConnection/[[SctpTransport]]}} to
                           <code>null</code>.
                         </p>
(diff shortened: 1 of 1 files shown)


docfaraday:
This looks good to me.

youennf:
We probably need to check what implementations are doing, in particular:
- whether RTCDTLSTransport.statechange event is fired in case of pc.close. According https://wpt.fyi/results/webrtc/RTCDtlsTransport-state.html?label=master&label=experimental&aligned, it seems that Chrome is firing, Safari and Firefox are not.
- whether RTCSctpTransport.statechange event is fired in case of pc.close. I did not see a WPT test maybe there is.
- If those two events are firing, what order is being used (hopefully DTLS then SCTP as seems to indicate the spec and the PR).

PR #3136, opened 2026-08-27
Specify RTCIceCandidate object creation for getLocal/RemoteCandidates()

### LINKED ISSUE #3120: Do we need to specify a comparison algorithm for RTCIceCandidate?
If we want to avoid copies of remote candidates piling up in the return of RTCIceTransport.getRemoteCandidates(), we're going to need to detect duplicates. Even if we say "If you pass dupes to addIceCandidate, you get dupes", that doesn't save us because we also need to incorporate candidates from remote descriptions, which can be set repeatedly.

There are lots of ways we could do this.

* Do we just string compare the candidate-attribute + ufrag and call it a day? That won't work with Chrome's candidates since the string form is different in SDP.
* If we compare only specific fields, which fields are compared? Surely at least ufrag, address, port, protocol, and tcpType. But should we pay attention to others? (type? component? priority?)
* How sophisticated should address comparison be? Do we normalize IPv6? Do we normalize weird IPv4 representations? IDNA?

jan-ivar:
Adds algorithms and internal slots to specify RTCIceCandidate object management.

Fixes #3120. Amendments and test updates to follow.

@docfaraday PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webrtc-pc/pull/3136.html" title="Last updated on Aug 27, 2026, 1:59 PM UTC (7d97e43)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-pc/3136/fcacb9b...jan-ivar:7d97e43.html" title="Last updated on Aug 27, 2026, 1:59 PM UTC (7d97e43)">Diff</a>

### DIFF
--- a/webrtc.html
+++ b/webrtc.html
@@ -2849,6 +2849,14 @@
                               firing on these objects in a queued task.
                             </p>
                           </li>
+                          <li>
+                            <p>
+                              For each {{RTCIceTransport}} <var>transport</var>
+                              associated with <var>connection</var>, [= prune stale
+                              candidates =] of <var>transport</var> of
+                              <var>connection</var>.
+                            </p>
+                          </li>
                           <li id="init-rtcrtpencodingparameters-codec">
                             <p>For each <var>transceiver</var> in <var>connection</var>'s [= set of transceivers =]:</p>
                             <ol>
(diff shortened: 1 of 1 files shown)


docfaraday:
This is a start.

* The steps for "selected candidate pair was changed" won't work for prflx; those never appear in SDP and should never appear in the getters either. They probably shouldn't be in the slots either.
* We don't want to process every m-section that uses a given transport when populating the candidate slots. Each transport should be grabbing candidates from just one m-section, whichever one is authoritative for that transport (either the bundle-tag or a standalone). This leads me to...
* Do we really want to perform deduplication in "collect ICE candidates" (step 3.2.2)? That would mean the resulting array could differ from what's in the SDP, if the SDP contains dupe candidates (which is possible, but admittedly silly).
* The "collect ICE candidates" step will need to update sdpMid/sdpMLineIndex in pre-existing candidates; bundle renegotiation allows transports to move from one bundle-tag to another.
* We seem to be collecting lazily in the getters. This will result in different observations depending on whether we've called the getter previously; if we kick off an ICE restart (we're in have-X-offer) and call the corresponding getter for the first time, we'll get the new candidates only, whereas if we have called the getter before starting that negotiation we'll get the old candidates too. That said...
* Do we really want the getter to give us both generations of candidates if we're in have-X-offer on an ICE restart? If we do that, how do we handle non-ICE-restart cases that change the candidates (eg; first negotiation has RTCP candidates, reoffer does not, or cases where a transport moves from one m-section to another, or even cases where the other end just removes a candidate attr for whatever reason)?
 
I'd advocate having two slots for each candidate type; one containing candidates from the current description, and the other containing candidates from the pending description. Let's call these [[CurrentRemoteCandidates]], [[PendingRemoteCandidates]], [[CurrentLocalCandidates]], and [[PendingLocalCandidates]]. Whenever the SDP (pending or current) changes, we build a list of candidate init dicts for all of the candidates in it, and update the corresponding (pending or current) slot based on that list, reusing objects if they were present in either the pending/current slots. This should be robust against all kinds of stuff, unlike trying to use ufrag to distinguish old from new. The "surface a local candidate" and addIceCandidate cases can be optimized in the spec algorithm, or we could leave that up to the implementation; it isn't really necessary to build the whole init dict list again for those cases.

PR #3140, opened 2026-09-17
Mark RTCIceTransport's component attribute as feature at risk

### LINKED ISSUE #3137: RTCIceTransport.component should be removed from the spec
Since there's only one RTCIceTransport per [RTCDtlsTransport](https://w3c.github.io/webrtc-pc/#dom-rtcdtlstransport-icetransport), and only one of *those* per [sender](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-transport)/[receiver](https://w3c.github.io/webrtc-pc/#dom-rtcrtpreceiver-transport)/[sctp](https://w3c.github.io/webrtc-pc/#dom-rtcsctptransport-transport), there's no way to have a separate RTCIceTransport for rtcp. That means `component` would always be "rtp". Nobody implements `component` right now, so we should probably just remove it.

jan-ivar:
Fixes #3137


youennf:
Marking as editorial as we are not changing the spec, but are planning to do it, probably in the coming months, before end of the year.

youennf (approved):

